PR #223, opened 2017-01-31
Ensure all email has a from address.

rushtong:
Hot fix for missing from address in new configuration.


davidangb (approved):
code here looks good!

PR #241, opened 2017-06-28
Include MongoDB, Elasticsearch, GCS in health check [GAWB-1962]

aednichols:
- `deadlocks` is built into Dropwizard, it's detecting deadlocked Java threads. Not sure how useful this is for us, but might as well keep it around
- The MySQL check is built in to JDBI, under the covers it's doing a `SELECT 1` via `DBIFactory` and `DBIHealthCheck`
- From what I've read, `{ ping: 1 }` is the MongoDB equivalent of `SELECT 1`
- Elasticsearch check is the same class as in Ontology, it just wasn't registered to run
- Google storage - attempt to read ~the directory listing at the root of the bucket~ the bucket's metadata

Example output (I hacked the MongoDB command so it fails, and broke the bucket name):

```
{
    "deadlocks": {
        "healthy": true
    },
    "elastic-search": {
        "healthy": true,
        "message": "ClusterHealth is GREEN"
    },
    "google-cloud-storage" : {
        "healthy" : false,
        "message" : "GCS bucket unreachable: 404 Not Found\n{\n  \"code\" : 404,\n  \"errors\" : [ {\n    \"domain\" : \"global\",\n    \"message\" : \"Not Found\",\n    \"reason\" : \"notFound\"\n  } ],\n  \"message\" : \"Not Found\"\n}"
    },
    "mongodb": {
        "healthy": false,
        "message": "Exception querying Mongo instance: Command failed with error 59: 'no such command: pudding' on server mongodb-dev-1:27017. The full response is { \"ok\" : 0.0, \"errmsg\" : \"no such command: pudding\", \"code\" : 59, \"bad cmd\" : { \"pudding\" : \"1\" } }"
    },
    "mysql": {
        "healthy": true
    }
}
```

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/consent-ws/src/main/java/org/broadinstitute/consent/http/MongoHealthCheck.java
+++ b/consent-ws/src/main/java/org/broadinstitute/consent/http/MongoHealthCheck.java
@@ -1,20 +1,33 @@
 package org.broadinstitute.consent.http;
 
 import com.codahale.metrics.health.HealthCheck;
-import com.mongodb.Mongo;
+import com.mongodb.MongoClient;
+import com.mongodb.BasicDBObject;
+import com.mongodb.MongoException;
 
 public class MongoHealthCheck extends HealthCheck {
  
-    private final Mongo mongo;
+    private final MongoClient mongo;
+    private final String databaseName;
  
-    protected MongoHealthCheck(Mongo mongo) {
+    protected MongoHealthCheck(MongoClient mongo, String databaseName) {
         super();
         this.mongo = mongo;
+        this.databaseName = databaseName;
     }
  
     @Override
     protected Result check() throws Exception {
-        mongo.getDatabaseNames();
+
+        // https://stackoverflow.com/a/16823314/818054
+        BasicDBObject ping = new BasicDBObject("ping", "1");
+
+        try {
+            mongo.getDatabase(this.databaseName).runCommand(ping);
+        } catch (MongoException e) {
+            return Result.unhealthy("Exception querying Mongo instance: " + e.getMessage());
+        }
+
         return Result.healthy();
     }
  
--- a/consent-ws/src/main/java/org/broadinstitute/consent/http/cloudstore/GCSStore.java
+++ b/consent-ws/src/main/java/org/broadinstitute/consent/http/cloudstore/GCSStore.java
@@ -5,6 +5,7 @@ import com.google.api.client.http.*;
 import com.google.api.client.http.javanet.NetHttpTransport;
 import com.google.api.services.storage.Storage;
 import com.google.api.services.storage.StorageScopes;
+import com.google.api.services.storage.model.Bucket;
 import com.google.api.services.storage.model.ObjectAccessControl;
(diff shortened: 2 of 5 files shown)


consent-ws/src/main/java/org/broadinstitute/consent/http/MongoHealthCheck.java
```diff
@@ -1,20 +1,33 @@
...
-    protected MongoHealthCheck(Mongo mongo) {
+    protected MongoHealthCheck(MongoClient mongo, String databaseName) {
         super();
         this.mongo = mongo;
+        this.databaseName = databaseName;
     }
  
     @Override
     protected Result check() throws Exception {
-        mongo.getDatabaseNames();
```
aednichols, line 17: `getDatabaseNames` is deprecated, so I replaced it with a [command](https://docs.mongodb.com/manual/reference/command/ping/) recommended by the MongoDB documentation (also referenced in the SO link)

davidangb (approved):
👍 did not test.

I'm a bit more concerned about the gcs bucket health check here than I am in ontology - as usage ramps up for consent (usage is extremely low now), the bucket will start to fill up and the health check will become heavyweight.

consent-ws/src/main/java/org/broadinstitute/consent/http/MongoHealthCheck.java
rushtong: 👍

consent-ws/src/main/java/org/broadinstitute/consent/http/cloudstore/GCSStore.java
```diff
@@ -156,6 +157,13 @@ public String putStorageDocument(InputStream stream, String type, String fileNam
         return url.toString();
     }
 
+    public Bucket getBucketMetadata() throws IOException, GeneralSecurityException {
```
rushtong, line 160: Can we get a positive/negative test for this method? I started looking in consent for this but couldn't find it, shame on us. There **are** examples in Ontology where we do this (StoreOntologyServiceTest) that uses a mocked CloudStore object which should serve as a model.

rushtong (changes requested):
Love the revised approach of checking bucket metadata. I would like to see a test for this though.

consent-ws/src/main/java/org/broadinstitute/consent/http/cloudstore/GCSStore.java
aednichols: This method is mostly a wrapper for a Google API; I feel like it would be more helpful to do something like Ontology's `ElasticSearchHealthCheckTest`, which verifies the business logic of the health check itself. Thoughts?

rushtong: Sorry I'm late responding to this. I agree, that makes more sense.

PR #242, opened 2017-07-26
ES Upgrade: GAWB-1487 (consent)

rushtong:
## Addresses: 
Elastic Search Upgrade to 5.5.0: https://broadinstitute.atlassian.net/browse/GAWB-1487

## Requires:
* Ontology PR: https://github.com/broadinstitute/consent-ontology/pull/80

Config PRs:
* Local: https://github.com/broadinstitute/firecloud-develop/pull/568
* Dev: https://github.com/broadinstitute/firecloud-develop/pull/569
* FiaB: https://github.com/broadinstitute/firecloud-develop/pull/570

## Changes:
* Upgrade elastic search
* Use REST client for all ES transactions instead of transport client
* Use mock server for ES testing instead of embedded ES server
* Use smaller ontology file for tests for better performance
* Bug fix for deleting ontology endpoint. Follow-up story to check that the UI is making the correct request format is here: https://broadinstitute.atlassian.net/browse/GAWB-2378
* Change how we delete ontologies. Instead of removing them completely from ES, instead we mark them as `usable=false`.
* Add `/status` swagger description

## Testing
See `DEVNOTES.md` for instructions on spinning up a local (unpopulated) ES instance. You can then populate the indexes with content from example data-use.owl and diseases.owl files. These files also exist in dev GCS so can be retrieved from there and used. Alternatively, you can point your local configs to the dev ES instance (which should be populated by the time this PR is picked up).

The APIs that are affected are:
* GET /status
* POST /api/ontology
* PUT /api/ontology

After a POST/PUT, you can query the es instance directly to compare what was added/removed. Some example queries you can use:
```
curl -v -XPOST 'http://elasticsearch5a1.dsde-dev.broadinstitute.org:9200/ontology-local/_search' -H 'Content-Type: application/json' -d '
{
  "query": {
    "term" : { "ontology": "organization" } 
  }
}' | jq '.'
curl -v -XPOST 'http://elasticsearch5a1.dsde-dev.broadinstitute.org:9200/ontology-local/_search' -H 'Content-Type: application/json' -d '
{
  "query": {
    "term" : { "ontology": "disease" } 
  }
}' | jq '.'
curl -v -XGET 'http://elasticsearch5a1.dsde-dev.broadinstitute.org:9200/ontology-local/_search?size=10' -H 'Content-Type: application/json' -d '
{
  "query": {
    "term" : { "usable": true } 
  }
}' | jq '.'
curl -v -XGET 'http://elasticsearch5a1.dsde-dev.broadinstitute.org:9200/ontology-local/_search?size=10' -H 'Content-Type: application/json' -d '
{
  "query": {
    "term" : { "usable": false } 
  }
}' | jq '.'
```

---

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off
- [x] **Product Owner** sign off
- [x] **Submitter**: Delete branch after merge
- [x] **Submitter**: **Test this change works on dev environment after deployment**. YOU own getting it fixed if dev isn't working for ANY reason!
- [x] **Submitter**: Mark JIRA issue as resolved once this checklist is completed

### DIFF
--- a/consent-ws/src/main/java/org/broadinstitute/consent/http/service/ontologyIndexer/ElasticSearchSupport.java
+++ b/consent-ws/src/main/java/org/broadinstitute/consent/http/service/ontologyIndexer/ElasticSearchSupport.java
@@ -4,9 +4,12 @@ import org.apache.http.HttpHost;
 import org.broadinstitute.consent.http.configurations.ElasticSearchConfiguration;
 import org.elasticsearch.client.RestClient;
 
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
 import java.util.stream.Collectors;
 
-public class ElasticSearchRestClient {
+@SuppressWarnings("WeakerAccess")
+public class ElasticSearchSupport {
 
     public static RestClient getRestClient(ElasticSearchConfiguration configuration) {
         HttpHost[] hosts = configuration.
@@ -15,6 +18,18 @@ public class ElasticSearchRestClient {
             map(server -> new HttpHost(server, 9200, "http")).
             collect(Collectors.toList()).toArray(new HttpHost[configuration.getServers().size()]);
         return RestClient.builder(hosts).build();
+    }
+
+    public static String getIndexPath(String index) {
+        return "/" + index;
+    }
+
+    public static String getTermIdPath(String index, String termId) throws UnsupportedEncodingException {
+        return "/" + index + "/ontology_term/" + URLEncoder.encode(termId, "UTF-8");
+    }
+
+    public static String getClusterHealthPath(String index) {
+        return "/_cluster/health/" + index;
     }
 
 }
--- a/consent-ws/src/main/java/org/broadinstitute/consent/http/service/ontologyIndexer/ElasticSearchSupport.java
+++ b/consent-ws/src/main/java/org/broadinstitute/consent/http/service/ontologyIndexer/ElasticSearchSupport.java
@@ -13,7 +13,7 @@ import java.util.stream.Collectors;
 @SuppressWarnings("WeakerAccess")
 public class ElasticSearchSupport {
 
-    public static RestClient getRestClient(ElasticSearchConfiguration configuration) {
+    public static RestClient createRestClient(ElasticSearchConfiguration configuration) {
         HttpHost[] hosts = configuration.
             getServers().
             stream().
--- a/consent-ws/src/main/java/org/broadinstitute/consent/http/service/ontologyIndexer/ElasticSearchSupport.java
+++ b/consent-ws/src/main/java/org/broadinstitute/consent/http/service/ontologyIndexer/ElasticSearchSupport.java
@@ -1,6 +1,8 @@
 package org.broadinstitute.consent.http.service.ontologyIndexer;
 
+import org.apache.http.Header;
 import org.apache.http.HttpHost;
+import org.apache.http.message.BasicHeader;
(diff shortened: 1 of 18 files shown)


davidangb (commented):
same comments as on your consent-ontology PR:
* I reviewed at a high level for architecture; did not test or look closely at syntax
* might not want to reinitialize the rest client for each request
* keep an eye out for mocks becoming very complex

consent-ws/src/main/java/org/broadinstitute/consent/http/service/ontologyIndexer/ElasticSearchSupport.java
```diff
@@ -0,0 +1,39 @@
...
+
+@SuppressWarnings("WeakerAccess")
+public class ElasticSearchSupport {
+
+    public static RestClient getRestClient(ElasticSearchConfiguration configuration) {
+        HttpHost[] hosts = configuration.
+            getServers().
+            stream().
+            map(server -> new HttpHost(server, 9200, "http")).
+            collect(Collectors.toList()).toArray(new HttpHost[configuration.getServers().size()]);
```
ahaessly, line 21: not sure if this applies, but should we be using https? would that change this code?

consent-ws/src/main/java/org/broadinstitute/consent/http/service/ontologyIndexer/IndexOntologyService.java
```diff
@@ -47,12 +51,12 @@ public String getIndexName() {
...
-
-        utils.validateIndexExists(client, indexName);
-
-        try {
+        try(RestClient client = getRestClient()) {
             for (StreamRec streamRec : streamRecList) {
 
+                // Deprecate everything that might already exist for this ontology file
+                utils.bulkDeprecateTerms(client, indexName, streamRec.getOntologyType());
+
```
ahaessly, line 66: should we not deprecate until after we know the new ontology file is not problematic?

ahaessly (approved):
see above suggestions/questions

consent-ws/src/main/java/org/broadinstitute/consent/http/service/ontologyIndexer/ElasticSearchSupport.java
rushtong: Good question - I don't know the right answer. All of the example code we have and use is `http`, but I'll do some testing with `https` and see if that still works.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/ontologyIndexer/IndexOntologyService.java
rushtong: At present, we don't have a way to distinguish between the two cases (existing terms and new terms) after posting them to ES. The previous code would leave in place old terms that were not in the new file which was definitely a bug. Also, since we're just deprecating terms (by setting a field to `false`), we can always fix things up after the fact if there is a problem with the file.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/ontologyIndexer/ElasticSearchSupport.java
rushtong: `https` doesn't work - we don't have that protocol set up on the ES servers.

davidangb (approved):
👍

ahaessly (approved):
looks good. 👍

PR #244, opened 2017-09-21
Accommodate DUO ontology import

rushtong:
## Addresses:
https://broadinstitute.atlassian.net/browse/GAWB-2523
Requires: https://github.com/broadinstitute/consent-data-use/pull/6
Requires: https://github.com/broadinstitute/consent-ontology/pull/83

## Changes:
* Migrate away from using terms that are no longer in use in DUOS
* Include the reference to direct ontology imports (i.e., only one level of import depth)
* Upgrade owl-api library to for imported ontology functions.
* Added an ES index check that creates the desired index if it doesn't exist.

## Testing:
* I suggest running this against a local ES instance that points to a locally mounted data directory so that you can re-use the populated index when testing consent. It should also be set up to point to a new bucket in broad-dsde-dev GCS for local testing, or you can use the one I have set up for this (`ontology-2523`). I don't have a set-up that spins up both consent and ontology at once, but that would be better for testing.
```
es:
  image: docker.elastic.co/elasticsearch/elasticsearch:5.5.0
  ports:
    - "9200:9200"
    - "9300:9300"
  volumes:
    - ../../data:/usr/share/elasticsearch/data
  environment:
    transport.host: 127.0.0.1
    xpack.security.enabled: "false"
    http.host: 0.0.0.0
```
* Test upload of the `data-use.owl` file. Terms in DUO with labels, descriptions, and parents should be visible in ontology.

---

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off
- [x] **Product Owner** sign off
- [x] **Submitter**: Delete branch after merge
- [x] **Submitter**: **Test this change works on dev environment after deployment**. YOU own getting it fixed if dev isn't working for ANY reason!
- [x] **Submitter**: Mark JIRA issue as resolved once this checklist is completed


consent-ws/src/main/java/org/broadinstitute/consent/http/service/ontology/IndexerUtils.java
```diff
@@ -1,35 +1,107 @@
...
  */
 public class IndexerUtils {
 
-    private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(IndexerUtils.class);
+    private static final String FIELD_DEFINITION_PROPERTY = "IAO_0000115";
+    private static final String FIELD_HAS_EXACT_SYNONYM_PROPERTY = "hasExactSynonym";
+    private static final String FIELD_LABEL_PROPERTY = "label";
+    private static final String FIELD_DEPRECATED_PROPERTY = "deprecated";
+    private static final Logger logger = LoggerFactory.getLogger(IndexerUtils.class);
+    private static final ArrayList<String> IRI_FILTERS = new ArrayList<>(Arrays.asList("DOID", "DUOS", "DUO"));
```
rushtong, line 42: **Note for Reviewer** This approach is less than ideal. Without filtering, we get a range of undesired terms from DUO that only add confusion to the index. Even if we move this to a configuration, that still requires a code update. I think a better solution would be to accept a list of IRI filter terms from the API ... looking for reviewer feedback on this question and if we should add a story for updating the API to accept an optional filter list.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/ontology/IndexerUtils.java
```diff
@@ -1,35 +1,107 @@
...
+
+    /**
+     * Generate terms from an ontology file, along with any direct ontology imports.
+     *
+     * @param streamRec The StreamRec that refers to an an ontology file.
+     * @return Collection of Terms generated from the ontology file
+     * @throws IOException The IOException
+     * @throws OWLOntologyCreationException The OWLOntologyCreationException
+     */
+    public Collection<Term> generateTerms(StreamRec streamRec) throws IOException, OWLOntologyCreationException {
```
rushtong, line 81: **Note for Reviewer:** This method is an extraction from where it was previously and is mostly unchanged, except to handle ontology imports.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/ontology/IndexerUtils.java
davidangb: I agree feels not-ideal ... but practically, we don't change the "whitelist" of ontologies very often, so I don't mind this as-is.

consent-ws/src/main/resources/changesets/changelog-consent-35.0.xml
```diff
@@ -0,0 +1,49 @@
...
+        <!-- Non Profit Migration -->
+        <sql dbms="hsql, mysql">
+            update election set useRestriction = REPLACE(useRestriction, 'http://www.broadinstitute.org/ontologies/DUOS/Non_profit', 'http://purl.obolibrary.org/obo/DUO_0000018')
+        </sql>
+        <sql dbms="hsql, mysql">
+            update consents set useRestriction = REPLACE(useRestriction, 'http://www.broadinstitute.org/ontologies/DUOS/Non_profit', 'http://purl.obolibrary.org/obo/DUO_0000018')
+        </sql>
+
+
+        <!-- For Profit Migration to Not: Non Profit -->
```
aednichols, line 39: This is my favorite part of this migration - for-profit is otherwise unrepresentable in DUO?

aednichols (approved):
Ran test plan, inspected

consent-ws/src/main/resources/changesets/changelog-consent-35.0.xml
rushtong: The goal is to reduce, as much as possible, the terms we use in DUOS so converting `For-Profit` to `not: Non-Profit` is functionally equivalent. Hope that makes sense.

davidangb (approved):
:+1:

PR #248, opened 2017-11-22
GAWB-2733: upgrade dependencies

davidangb:
Upgrading some dependencies due to proactive warnings from SourceClear.


rushtong (approved):
👍 Looks good. We should watch for mongo issues - we've had trouble with the java library and mongo host names. I tested locally connecting to dev and the current configuration works as expected, so the workarounds we've implemented still work.

PR #250, opened 2018-01-22
GAWB-3001: Accept, require, save, and return Data Use primitive in all consent interactions

rushtong:
## Addresses 
* https://broadinstitute.atlassian.net/browse/GAWB-3001

This PR does not address back-population - that will be handled manually by Moran.

This PR also does not enforce any kind of logical equivalence between the structured use restriction and structured data use primitive. There are future stories to address the removal of the structured use restriction and make consents solely reliant on having a data use primitive. 

## See Also 
* ORSP PR: https://github.com/broadinstitute/orsp/pull/331
* Consent UI PR: https://github.com/DataBiosphere/consent-ui/pull/165
* FC-Develop PR: https://github.com/broadinstitute/firecloud-develop/pull/940

## Testing
* Test create/update consent from both Consent UI and from ORSP
* Test retrieval in all of the get calls (can be done from swagger)

I have tested this on a dev fiab. The fiab liquibase jobs are not (at this time) working for custom branches, so I manually made the db changes and was able to successfully create/update/retrieve consents (via the UI) with the new data use primitive.

---

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/consent-ws/src/main/java/org/broadinstitute/consent/http/db/ConsentMapper.java
+++ b/consent-ws/src/main/java/org/broadinstitute/consent/http/db/ConsentMapper.java
@@ -1,7 +1,5 @@
 package org.broadinstitute.consent.http.db;
 
-import org.apache.commons.collections.Closure;
-import org.apache.log4j.Logger;
 import org.broadinstitute.consent.http.models.Consent;
 import org.broadinstitute.consent.http.models.DataUseDTO;
 import org.broadinstitute.consent.http.models.grammar.UseRestriction;
@@ -14,8 +12,6 @@ import java.sql.SQLException;
 
 public class ConsentMapper implements ResultSetMapper<Consent> {
 
-    private Logger logger = Logger.getLogger("ConsentMapper");
-
     public Consent map(int index, ResultSet r, StatementContext ctx) throws SQLException {
         Consent consent = new Consent();
         consent.setConsentId(r.getString("consentId"));
@@ -24,11 +20,10 @@ public class ConsentMapper implements ResultSetMapper<Consent> {
         consent.setDulName(r.getString("dulName"));
         try {
             consent.setUseRestriction(UseRestriction.parse(r.getString("useRestriction")));
-            consent.setDataUse(DataUseDTO.parseDataUse(r.getString("dataUse")));
         } catch (IOException e) {
-            logger.error("Unable to parse object from database: " + e.getMessage());
             throw new SQLException(e);
         }
+        consent.setDataUse(DataUseDTO.parseDataUse(r.getString("dataUse")).orElse(null));
         consent.setName(r.getString("name"));
         consent.setCreateDate(r.getTimestamp("createDate"));
         consent.setSortDate(r.getTimestamp("sortDate"));
--- a/consent-ws/src/main/java/org/broadinstitute/consent/http/db/ConsentMapper.java
+++ b/consent-ws/src/main/java/org/broadinstitute/consent/http/db/ConsentMapper.java
@@ -1,6 +1,7 @@
 package org.broadinstitute.consent.http.db;
 
 import org.broadinstitute.consent.http.models.Consent;
+import org.broadinstitute.consent.http.models.DataUseDTO;
 import org.broadinstitute.consent.http.models.grammar.UseRestriction;
 import org.skife.jdbi.v2.StatementContext;
 import org.skife.jdbi.v2.tweak.ResultSetMapper;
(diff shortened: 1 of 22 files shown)


consent-ws/src/main/java/org/broadinstitute/consent/http/resources/Resource.java
```diff
@@ -25,7 +26,13 @@ protected Logger logger() {
...
     protected Response createExceptionResponse(Exception e) {
         try {
-            return dispatch.get(e.getClass()).handle(e);
+            logger().warn("Returning error response to client: " + e.getMessage());
+            ExceptionHandler handler = dispatch.get(e.getClass());
+            if (handler != null) {
+                return handler.handle(e);
+            } else {
+                return Response.serverError().entity(new Error(e.getMessage(), Response.Status.INTERNAL_SERVER_ERROR.getStatusCode())).build();
+            }
```
rushtong, line 37: Added some logging here. During testing, some of the logged errors were pretty useless.

consent-ws/src/test/java/org/broadinstitute/consent/http/AbstractTest.java
```diff
@@ -180,4 +180,33 @@ public void mockValidateTokenResponse() throws IOException {
...
+                setUseRestriction(useRestriction).
+                setDataUse(dataUse).
+                setName(UUID.randomUUID().toString()).
+                setCreateDate(createDate).
+                setLastUpdate(createDate).
+                setSortDate(createDate).
+                build();
+    }
+
+    protected MongoClient setUpMongoClient() throws IOException {
```
rushtong, line 197: Three test classes were doing the same thing in copy/pasta mode. Abstracted here to clean up the duplicate code and to prep for upgrading flapdoodle (see https://broadinstitute.atlassian.net/browse/GAWB-2440).

consent-ws/src/main/resources/changesets/changelog-consent-36.0.xml
```diff
@@ -0,0 +1,15 @@
...
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet author="grushton" id="114">
+        <addColumn tableName="consents">
+            <column name="dataUse" type="text"/>
+        </addColumn>
+        <sql dbms="hsql, mysql">UPDATE consents SET dataUse = ''</sql>
+        <sql dbms="hsql, mysql">ALTER TABLE consents MODIFY dataUse text NOT NULL</sql>
```
rushtong, line 12: mysql 5 doesn't let us create a `text` column with a default value in an `add column` statement, so manually doing the same thing with the extra sql statements.

consent-ws/src/main/java/org/broadinstitute/consent/http/db/ConsentMapper.java
```diff
@@ -13,20 +14,25 @@
...
+        } catch (IOException e) {
+            throw new SQLException(e);
+        }
+        consent.setName(r.getString("name"));
         consent.setCreateDate(r.getTimestamp("createDate"));
         consent.setSortDate(r.getTimestamp("sortDate"));
-        consent.setLastUpdate((r.getTimestamp("lastUpdate") == null) ? null : r.getTimestamp("lastUpdate"));
-        consent.setTranslatedUseRestriction((r.getString("translatedUseRestriction") == null) ? null : r.getString("translatedUseRestriction"));
+        consent.setLastUpdate(r.getTimestamp("lastUpdate"));
+        consent.setTranslatedUseRestriction(r.getString("translatedUseRestriction"));
```
davidangb: are the null checks no longer needed?

consent-ws/src/main/resources/changesets/changelog-consent-36.0.xml
davidangb: You could use `VARCHAR`, which does accept a default. I see you're using `TEXT` for all the other columns, so I think consistency is better and text is ok.

davidangb (approved):
code looks good, did not test. lots of test-class cleanup in here!

consent-ws/src/main/java/org/broadinstitute/consent/http/db/ConsentMapper.java
```diff
@@ -13,20 +14,25 @@
...
+            consent.setUseRestriction(UseRestriction.parse(r.getString("useRestriction")));
         } catch (IOException e) {
             throw new SQLException(e);
         }
-        consent.name = r.getString("name");
+        try {
+            consent.setDataUse(DataUseDTO.parseDataUse(r.getString("dataUse")));
+        } catch (IOException e) {
+            throw new SQLException(e);
+        }
```
kyuksel: Couldn't the two `try-catch` blocks above be merged into one since they're checking for and throwing the same exceptions? It also made me wonder if there is a more specific `Exception` subclass than `IOException` that could provide more clarity.

consent-ws/src/main/java/org/broadinstitute/consent/http/models/DataUseDTO.java
```diff
@@ -272,5 +281,20 @@ public void setNonBiomedical(Boolean nonBiomedical) {
     public String toString() {
         return new GsonBuilder().create().toJson(this);
     }
-    
+
+    public static DataUseDTO parseDataUse(String str) throws IOException {
+        if (str == null || str.isEmpty()) {
```
kyuksel: Would using `java.util.Optional` alleviate the `null` pain?

consent-ws/src/main/java/org/broadinstitute/consent/http/db/ConsentMapper.java
rushtong: Technically, no, but I'm going to revert this to keep the changes down. Intellij was simplifying things for me. Reverting to reduce the risk of regression.

kyuksel (approved):
👍 ing with minor comments. Unfortunately, I'm too unfamiliar with the functionality to give more meaningful feedback. Didn't test.

consent-ws/src/main/java/org/broadinstitute/consent/http/db/ConsentMapper.java
rushtong: Great point - they certainly could. As written, we'd lose which statement threw the exception, but I can revise that.

consent-ws/src/main/java/org/broadinstitute/consent/http/models/DataUseDTO.java
rushtong: I would really like for it to be helpful, but I think it just adds extra code.
```
        Optional<String> opt = Optional.ofNullable(str);
        if (!opt.isPresent() || opt.get().isEmpty()) {
            return null;
            ...
```

kyuksel: The way to use `Optional` would be to include it in the type signature 

```java
public static Optional<DataUseDTO> parseDataUse(String str) { ... }
```

like we do in Scala so `parseDataUse` wouldn't throw and any errors would be handled upstream of this method.

I can't assess if this is worthwhile, though.

rushtong: Actually, returning an optional does help us with the mapper since we can handle exceptions much better!

rushtong: Yes, actually, the Optional approach is much cleaner, thanks!

consent-ws/src/main/resources/changesets/changelog-consent-36.0.xml
rushtong: Yes, I chose `text` to be consistent with how we're storing `useRestriction` currently.

davidangb (approved):
still :+1:, nice cleanup

kyuksel:
👍

PR #257, opened 2018-03-08
Set replicas equal to server count - 1 [WIP]

aednichols:
This change specifies that ES shards be replicated to every other server in the cluster, if any. This overrides the ES default of replicating to exactly 1 other server, regardless of how many are in the cluster.

In production-like environments, we will go from 1 -> 2 expected replicas.

On FiaBs, we will go from 1 -> 0 expected replicas, so they will no longer show as yellow.

This is all predicated on my untested assumption that Consent creates its own index every time; if this is not true, some accommodation will have to be made to apply this setting change to existing indexes.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/consent-ws/src/main/java/org/broadinstitute/consent/http/service/ontology/IndexOntologyService.java
+++ b/consent-ws/src/main/java/org/broadinstitute/consent/http/service/ontology/IndexOntologyService.java
@@ -17,6 +17,7 @@ public class IndexOntologyService implements Managed {
     private final String indexName;
     private IndexerUtils utils = new IndexerUtils();
     private RestClient client;
+    private int targetReplicaCount;
 
     @Override
     public void start() throws Exception { }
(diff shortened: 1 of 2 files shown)


rushtong:
Consent does not initially rely on creating its own index, it assumes that one exists already, and if not, only then does it try to create one. I'm hesitant about adding something to consent that will side-affect a separate service, especially since that service is now shared with Orchestration. Ontology not only uses the same service, it also speaks to the same index (as does Orch), so this change would also directly affect ontology. 

In theory, I'm leery of adding this level of infrastructure configuration into the code. In practice, consent and ontology are already strongly coupled (i.e. they both read and/or write to the same GCS bucket, both read and/or write to the same ES index), so I can perhaps live with it. Let's talk.

PR #259, opened 2018-03-19
GAWB-3340: Check for invalid dul values

rushtong:
## Addresses:
Fixes a bug discovered while working on https://broadinstitute.atlassian.net/browse/DSDECOM-130

Issue is tracked here for the GAWB project: https://broadinstitute.atlassian.net/browse/GAWB-3340

## Changes:
* Preemptively prevent invalid DUL values from being posted to a consent.
* Allow for non-fatal GCS file deletion in the case where invalid data exists. Log to sentry for manual follow-up.
* Multiple test fixes for previously improper usage of `assertThat`s in the code:
  * ConsentAcceptanceTest.java
  * ConsentsServiceTest.java
  * ElectionReviewTest.java
  * HelpReportTest.java

## Notes:
Discovered an unrelated bug (See https://broadinstitute.atlassian.net/browse/GAWB-3313) where querying for consents by association type returns a cartesian product. The test class, `ConsentsServiceTest.java`, is a mess and needs significant revision to get working correctly. 

---

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentResource.java
+++ b/consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentResource.java
@@ -1,5 +1,6 @@
 package org.broadinstitute.consent.http.resources;
 
+import com.google.api.client.http.GenericUrl;
 import com.google.gson.Gson;
 import io.dropwizard.auth.Auth;
 import org.broadinstitute.consent.http.enumeration.Actions;
@@ -78,6 +79,9 @@ public class ConsentResource extends Resource {
             if (rec.getDataUse() == null) {
                 throw new IllegalArgumentException("Data Use Object is required.");
             }
+            if (rec.getDataUseLetter() != null) {
+                checkValidDUL(rec);
+            }
             Consent consent = api.create(rec);
             auditServiceAPI.saveConsentAudit(consent.getConsentId(), AuditTable.CONSENT.getValue(), Actions.CREATE.getValue(), dacUser.getEmail());
             URI uri = info.getRequestUriBuilder().path("{id}").build(consent.consentId);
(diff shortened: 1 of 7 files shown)


consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentResource.java
```diff
@@ -78,6 +79,9 @@ public Response createConsent(@Context UriInfo info, Consent rec, @Auth User use
             if (rec.getDataUse() == null) {
                 throw new IllegalArgumentException("Data Use Object is required.");
             }
+            if (rec.getDataUseLetter() != null) {
```
kyuksel, line 82: `checkValidDUL()` checks `if (rec.getDataUseLetter() != null)` so no need for duplicating it here? Similarly for line#111.

Alternatively, `if (rec.getDataUseLetter() != null)` could be removed from `checkValidDUL()` if that makes more sense.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentResource.java
```diff
@@ -180,4 +187,15 @@ public ConsentResource() {
         this.electionAPI = AbstractElectionAPI.getInstance();
     }
 
+    private void checkValidDUL(Consent rec) {
```
kyuksel, line 190: It'd be nice to make methods like this less side-effecty and more strongly typed like we do in our Scala codebase. Not a blocker...just something to think about in case new Java standard/non-standard libraries make that more practical now.

kyuksel (approved):
👍 with minor comments inline. Didn't test.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentResource.java
rushtong: I think what we could do to make this more strongly typed is to convert the `dataUseLetter` field to be a URL type, but that's a non-trivial refactor. Agreed that the model object could be smarter.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentResource.java
rushtong: Good point - I think this double-check a result of some refactoring. Will definitely clean this up some more.

davidangb (approved):
:+1: lgtm, did not test

PR #260, opened 2018-03-29
GAWB-3291: Remove timeouts

rushtong:
## Addresses:
Needed for, but does not complete https://broadinstitute.atlassian.net/browse/GAWB-3291

Also addresses reported production sentry errors (search Sentry for `DatabaseMatching` or `SocketTimeoutException`).

## See Also:
Consent UI PR: https://github.com/DataBiosphere/consent-ui/pull/174. The two can proceed separately - they are related but not dependent. 

## Changes
* Primary change is to remove the timeout for ontology calls which was set to 10 seconds. That's much too low for match checking which is fairly resource intensive.
* Minor refactoring to clean up the code a bit.
* Added swagger doc for the undocumented API.
* Errors are being correctly reported to Sentry, so no need for logging changes.

## Testing
Existing tests covered the timeout case already so no need for more. 

---

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseMatchingServiceAPI.java
+++ b/consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseMatchingServiceAPI.java
@@ -12,6 +12,7 @@ import org.broadinstitute.consent.http.models.matching.RequestMatchingObject;
 import org.broadinstitute.consent.http.models.matching.ResponseMatchingObject;
 import org.broadinstitute.consent.http.util.DarConstants;
 import org.bson.Document;
+import org.glassfish.jersey.client.ClientProperties;
 
 import javax.ws.rs.client.Client;
 import javax.ws.rs.client.Entity;
@@ -42,7 +43,10 @@ public class DatabaseMatchingServiceAPI extends AbstractMatchingServiceAPI {
         this.dataAccessAPI = darAPI;
         this.consentAPI = consentAPI;
         this.dsAPI = dsAPI;
-        this.matchServiceTarget = client.target(config.getMatchURL());
+        Integer timeout = 1000 * 60 * 3; // 3 minute timeout so ontology can properly do matching.
+        client.property(ClientProperties.CONNECT_TIMEOUT, timeout);
+        client.property(ClientProperties.READ_TIMEOUT, timeout);
+        matchServiceTarget = client.target(config.getMatchURL());
     }
 
     @Override
--- a/consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseMatchingServiceAPI.java
+++ b/consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseMatchingServiceAPI.java
@@ -113,20 +113,17 @@ public class DatabaseMatchingServiceAPI extends AbstractMatchingServiceAPI {
             logger.error("Error finding single matchData Access Request: "+ dar.getString(DarConstants.DAR_CODE) + " does not have a proper Use Restriction.");
             throw new Exception("Data Access Request: " + dar.getString(DarConstants.DAR_CODE) + " cannot be matched. Missing Use Restriction field.");
         }
-        if (consent != null) {
-            Match match = createMatch(consent.getConsentId(), dar.get(DarConstants.ID).toString(), false, false);
-            RequestMatchingObject requestObject = createRequestObject(consent, dar);
-            String json = new Gson().toJson(requestObject);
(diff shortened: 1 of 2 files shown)


consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseMatchingServiceAPI.java
```diff
@@ -105,15 +101,23 @@ public Match findMatchForPurpose(String purposeId){
...
+            logger.error("Data Access Request is null");
+            throw new IllegalArgumentException("Data Access Request cannot be null");
+        }
         if(!dar.containsKey(DarConstants.RESTRICTION)){
-            logger().error("Error finding single matchData Access Request: "+ dar.getString(DarConstants.DAR_CODE) + " does not have a proper Use Restriction.");
+            logger.error("Error finding single matchData Access Request: "+ dar.getString(DarConstants.DAR_CODE) + " does not have a proper Use Restriction.");
             throw new Exception("Data Access Request: " + dar.getString(DarConstants.DAR_CODE) + " cannot be matched. Missing Use Restriction field.");
         }
-        if(consent != null && dar != null){
+        if (consent != null) {
```
ahaessly: We already checked if consent was null above. you can remove this check and the return null below

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseMatchingServiceAPI.java
```diff
@@ -27,26 +27,22 @@
...
+    private WebTarget matchServiceTarget;
+    private GenericType<ResponseMatchingObject> rmo = new GenericType<ResponseMatchingObject>(){};
+    private Logger logger = Logger.getLogger("DatabaseMatchingServiceAPI");
 
     public static void initInstance(Client client, ServicesConfiguration config) {
         MatchAPIHolder.setInstance(new DatabaseMatchingServiceAPI(client, config, AbstractConsentAPI.getInstance(), AbstractDataAccessRequestAPI.getInstance(), AbstractDataSetAPI.getInstance()));
     }
 
-    protected DatabaseMatchingServiceAPI(Client client, ServicesConfiguration config, ConsentAPI consentAPI, DataAccessRequestAPI darAPI, DataSetAPI dsAPI){
+    DatabaseMatchingServiceAPI(Client client, ServicesConfiguration config, ConsentAPI consentAPI, DataAccessRequestAPI darAPI, DataSetAPI dsAPI){
```
ahaessly, line 42: should this actually be private?

ahaessly (changes requested):
a few comments

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseMatchingServiceAPI.java
rushtong: It's used by a test which is in the same package space, so the default package-protected works here. Making it private would break the test.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseMatchingServiceAPI.java
rushtong: Right ... I added the null check above to fail fast and neglected to clean up the if. Will fix!

ahaessly (approved):
👍

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseMatchingServiceAPI.java
```diff
@@ -105,24 +101,29 @@ public Match findMatchForPurpose(String purposeId){
     }
 
     private Match singleEntitiesMatch(Consent consent, Document dar) throws Exception {
+        if (consent == null) {
```
aednichols, line 108: I like the pattern of checking preconditions at the top of a method, I think this is a nice improvement

aednichols (commented):
We removed the custom timeout - do we know if there is now a default timeout being enforced, or no timeout?

No timeout is probably bad, a network or server issue could easily put all of the application's threads into limbo.

rushtong:
@aednichols - good question - and it might be infinity. Double checking the version we have.

rushtong:
:-( It is infinity. Will bump to something more reasonable than 10 seconds.

aednichols (approved):
👍 LGTM, thanks for checking the timeout!

PR #261, opened 2018-03-29
Fix NPE by ignoring null votes [GAWB-3302]

aednichols:
This changeset addresses this part of the story:

>- To be safe, we should protect against a null vote by ignoring it. In other words, a null vote should not be reported in the match summary.

I still need to do
>- Examine the code to determine if null is a reasonable case (my first instinct is that null is a valid case).
>- Further debugging to determine how this happened if null is not a reasonable case.

---

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseSummaryAPI.java
+++ b/consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseSummaryAPI.java
@@ -110,8 +110,8 @@ public class DatabaseSummaryAPI extends AbstractSummaryAPI {
             List<Integer> electionIds = reviewedElections.stream().map(e -> e.getElectionId()).collect(Collectors.toList());
             List<Vote> votes = voteDAO.findVotesByElectionIds(electionIds);
 
-            List<Vote> agreementYesVotes = filterVotes(votes, Boolean.TRUE);
-            List<Vote> agreementNoVotes = filterVotes(votes, Boolean.FALSE);
+            List<Vote> agreementYesVotes = filterAgreementVotes(votes, Boolean.TRUE);
+            List<Vote> agreementNoVotes = filterAgreementVotes(votes, Boolean.FALSE);
 
             if (CollectionUtils.isNotEmpty(agreementYesVotes) || CollectionUtils.isNotEmpty(agreementNoVotes)) {
                 summaryList.add(createSummary(0, agreementYesVotes.size(), agreementNoVotes.size()));
@@ -122,7 +122,7 @@ public class DatabaseSummaryAPI extends AbstractSummaryAPI {
         return summaryList;
     }
 
-    private List<Vote> filterVotes(List<Vote> votes, Boolean desiredValue) {
+    private List<Vote> filterAgreementVotes(List<Vote> votes, Boolean desiredValue) {
         return votes.stream().filter(
                 v -> v.getType().equals(VoteType.AGREEMENT.getValue()) && v.getVote() != null && v.getVote().equals(desiredValue)
         ).collect(Collectors.toList());
--- a/consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseSummaryAPI.java
+++ b/consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseSummaryAPI.java
@@ -122,9 +122,9 @@ public class DatabaseSummaryAPI extends AbstractSummaryAPI {
         return summaryList;
     }
 
-    private List<Vote> filterVotes(List<Vote> votes, Boolean voteValue) {
+    private List<Vote> filterVotes(List<Vote> votes, Boolean desiredValue) {
         return votes.stream().filter(
-                v -> v.getType().equals(VoteType.AGREEMENT.getValue()) && v.getVote() != null && v.getVote().equals(voteValue)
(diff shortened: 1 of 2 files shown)


rushtong (approved):
👍 Looks good!

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseSummaryAPI.java
```diff
@@ -108,19 +109,25 @@ public Summary describeDataRequestSummaryCases(String electionType) {
...
             }
         }else{
             summaryList.add(createSummary(0,0,0));
         }
         return summaryList;
     }
 
+    private List<Vote> filterAgreementVotes(List<Vote> votes, Boolean desiredValue) {
+        return votes.stream().filter(
+                v -> v.getType().equals(VoteType.AGREEMENT.getValue()) && v.getVote() != null && v.getVote().equals(desiredValue)
```
davidangb, line 127: not important: you could collapse
```
v.getVote() != null && v.getVote().equals(desiredValue)
```
to
```
v.getVote() == desiredValue
```
( side effect: if you really really wanted to retrieve the null votes, this method would still work )

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseSummaryAPI.java
```diff
@@ -108,19 +109,25 @@ public Summary describeDataRequestSummaryCases(String electionType) {
...
-                Map<Boolean, List<Vote>> partition =
-                        agreementVotes.stream()
-                                .collect(Collectors.partitioningBy(v -> v.getVote()));
-                summaryList.add(createSummary(0, partition.get(Boolean.TRUE).size(), partition.get(Boolean.FALSE).size()));
+
+            List<Vote> agreementYesVotes = filterAgreementVotes(votes, Boolean.TRUE);
+            List<Vote> agreementNoVotes = filterAgreementVotes(votes, Boolean.FALSE);
+
+            if (CollectionUtils.isNotEmpty(agreementYesVotes) || CollectionUtils.isNotEmpty(agreementNoVotes)) {
+                summaryList.add(createSummary(0, agreementYesVotes.size(), agreementNoVotes.size()));
```
davidangb, line 117: this is a business-logic question, not a syntax one: does the app really want this `if` clause? If there truly are no votes, should it return `createSummary(0, 0, 0)`, or should it skip adding a summary to the list?

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseSummaryAPI.java
aednichols: I thought in Java `==` did not work on `Boolean`, only `boolean`

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseSummaryAPI.java
aednichols: Was wondering that too.

rushtong: I believe it should not skip the summary in the case of no votes. The data is passed to a google charts API and each element needs to know its values, even if they are all `0`.

aednichols: Skipping the summary in the case of no votes is the existing behavior, before this PR.

Do you want me to remove skipping in this PR?

rushtong: Oops, sorry, I missed that detail. Let me see if I can test this a little better.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseSummaryAPI.java
davidangb: good call on identity vs. equivalence. I was counting on everything using e.g. `Boolean.TRUE` instead of `new Boolean(true)`. How about `desiredValue.equals(v.getVote())` then? I guess that wouldn't support the case where `desiredValue` is null.

davidangb: ( this also isn't really important, don't hold the PR on this at all )

aednichols: It actually should support that case OK, because you can pass anything into a `.equals()`; I wanted to make the `null` check explicit since we (evidently) have not paid enough attention to `null` here in the past

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseSummaryAPI.java
rushtong: Tested this locally and the UI handles skipped cases just fine - which is as it should be. I don't suggest any changes to the approach in this PR.

davidangb (approved):
:+1:

PR #268, opened 2018-05-29
BTRX - 226 - Show election related DUL and sDUL in every page

LeoForconesi:
### Description

_In order to obtain the correct DUL, some changes were needed._
_- Add endpoint logic to download DUL from consent if it has not any election, or use election's DUL url_

_- Use last election version if there is no id for it given through query param, otherwise bring the specified election's info._

_- Map relevant info from election's query result in order to return it to the Ui when needed_

**Jira Story**
[BTRX-226-Show election related DUL and sDUL in every page](https://broadinstitute.atlassian.net/browse/BTRX-226)

**Note:** Ui branch [lf-BTRX-226-electionDulandSdul](https://github.com/DataBiosphere/consent-ui/tree/lf-BTRX-226-electionDulandSdul) is required to see the expected behavior.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataUseLetterResource.java
+++ b/consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataUseLetterResource.java
@@ -12,7 +12,6 @@ import org.broadinstitute.consent.http.enumeration.Actions;
 import org.broadinstitute.consent.http.enumeration.AuditTable;
 import org.broadinstitute.consent.http.models.Consent;
 import org.broadinstitute.consent.http.models.DACUser;
-import org.broadinstitute.consent.http.models.Election;
 import org.broadinstitute.consent.http.models.User;
 import org.broadinstitute.consent.http.service.*;
 import org.broadinstitute.consent.http.service.users.AbstractDACUserAPI;
@@ -34,14 +33,12 @@ import java.util.UUID;
 public class DataUseLetterResource extends Resource {
 
     private final ConsentAPI api;
-    private final ElectionAPI electionApi;
     private final GCSStore store;
     private final DACUserAPI dacUserAPI;
     private final AuditServiceAPI auditServiceAPI;
 
     public DataUseLetterResource(GCSStore store) {
         this.api = AbstractConsentAPI.getInstance();
-        this.electionApi = AbstractElectionAPI.getInstance();
         this.store = store;
         this.dacUserAPI = AbstractDACUserAPI.getInstance();
         this.auditServiceAPI = AbstractAuditServiceAPI.getInstance();
--- a/consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataUseLetterResource.java
+++ b/consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataUseLetterResource.java
@@ -145,7 +145,7 @@ public class DataUseLetterResource extends Resource {
                     .header("Content-Disposition", "attachment; filename=" + targetFile.getName())
                     .build();
         } catch (UnknownIdentifierException e) {
-            throw new NotFoundException(String.format("Could not find consent with id %s", consentId));
+            throw new NotFoundException(e);
         } catch (IOException e) {
             logger().error("Error when trying to read/write the file " + e.getMessage());
         } catch (GeneralSecurityException e) {
--- a/consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataUseLetterResource.java
+++ b/consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataUseLetterResource.java
@@ -131,13 +131,17 @@ public class DataUseLetterResource extends Resource {
     public Response getDUL(@PathParam("id") String consentId) {
         String msg = String.format("GETing Data Use Letter for consent with id '%s'", consentId);
         logger().debug(msg);
+        String fileUrl, fileName;
         try {
-            Election election = electionApi.describeConsentElection(consentId);
             Consent consent = api.retrieve(consentId);
-
-            String fileUrl  = election == null ? consent.getDataUseLetter() : election.getDataUseLetter();
-            String fileName = election == null ? consent.getDulName() : election.getDulName();
-
+            if (consent.getLastElectionStatus() != null) {
+                Election election = electionApi.describeConsentElection(consentId);
+                fileUrl = election.getDataUseLetter();
+                fileName = election.getDulName();
+            } else {
+                fileUrl = consent.getDataUseLetter();
+                fileName = consent.getDulName();
+            }
             HttpResponse r = store.getStorageDocument(fileUrl);
             File targetFile = new File(fileName);
             FileUtils.copyInputStreamToFile(r.getContent(), targetFile);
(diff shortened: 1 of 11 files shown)


consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataUseLetterResource.java
```diff
@@ -125,12 +125,15 @@ public Consent updateDUL(
     @GET
     @Produces(MediaType.APPLICATION_OCTET_STREAM)
     @RolesAllowed({"ADMIN","CHAIRPERSON","MEMBER","DATAOWNER"})
-    public Response getDUL(@PathParam("id") String consentId) {
+    public Response getDUL(@PathParam("id") String consentId, @QueryParam("electionId") Integer electionId) {
         String msg = String.format("GETing Data Use Letter for consent with id '%s'", consentId);
```
ahaessly: you might want to add in the election id to the debug statement

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseConsentAPI.java
```diff
@@ -473,4 +473,17 @@ public Consent getByName(String name) throws UnknownIdentifierException {
...
 
+    @Override
+    public Consent retrieveElectionDul(Integer electionId, Consent consent) {
+        Election election;
+        if (electionId != null) {
+            election = electionDAO.findElectionWithFinalVoteById(electionId);
+        } else {
+            election = electionDAO.getElectionWithFinalVoteByReferenceIdAndType(consent.consentId, ElectionType.TRANSLATE_DUL.getValue());
+        }
+        consent.setDataUseLetter(election.getDataUseLetter());
```
ahaessly: Do you really want to be updating the consent? What if you just return the election object at this point?
My concern about updating the consent is what happens if the current election is canceled/archived and a new election is started? Do you remove the values that have been set for the DataUseLetter and name when this happens? (Meaning that now you have to keep these attributes in sync with the "correct" election). 

If you really do want to make this change on the consent, I would rename the method to updateConsentWithElectionDul and return void since you are modifying the input paraml. Also, I would probably want to set this when the election is created instead of the retrieving the Dul.

ahaessly (commented):
see suggestion

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataUseLetterResource.java
LeoForconesi: Adding election id to debug statement.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseConsentAPI.java
LeoForconesi: You are right Andrea, i didn't thought that. Please review again if now it looks better.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseConsentAPI.java
```diff
@@ -473,4 +473,18 @@ public Consent getByName(String name) throws UnknownIdentifierException {
         return consent;
     }
 
+    @Override
+    public Election retrieveElectionDul(Integer electionId, String consentId) throws UnknownIdentifierException{
```
ahaessly: This should now be named retrieveElection

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataUseLetterResource.java
```diff
@@ -125,13 +126,17 @@ public Consent updateDUL(
...
-        String msg = String.format("GETing Data Use Letter for consent with id '%s'", consentId);
+    public Response getDUL(@PathParam("id") String consentId, @QueryParam("electionId") Integer electionId) {
+        String msg = String.format("GETing Data Use Letter for consent with id '%s' and Election Id '%s", consentId, electionId);
         logger().debug(msg);
+        Election election = null;
         try {
             Consent consent = api.retrieve(consentId);
-            String fileUrl  = consent.getDataUseLetter();
-            String fileName = consent.getDulName();
+            if (StringUtils.isNotEmpty(consent.getLastElectionStatus())) {
```
ahaessly, line 135: You could check if electionId is null and if so retrieve the consent and get the the url and name from the consent. Otherwise call retrieveElectionDul and get the url and name from the election. In this second case if the election is null, you could throw an UnknownIdentifierException for the election id.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseConsentAPI.java
```diff
@@ -473,4 +473,18 @@ public Consent getByName(String name) throws UnknownIdentifierException {
...
+        Election election;
+        if (electionId != null) {
+            election = electionDAO.findElectionWithFinalVoteById(electionId);
+        } else {
+            election = electionDAO.getElectionWithFinalVoteByReferenceIdAndType(consentId, ElectionType.TRANSLATE_DUL.getValue());
+        }
+        if (election.getDataUseLetter() == null) {
+            throw new UnknownIdentifierException(String.format("Could not found DUL for elecionId: '%s", election.getElectionId()));
```
ahaessly: should say "Could not find DUL..."

ahaessly (approved):
a few suggestions

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataUseLetterResource.java
LeoForconesi: I see what you mean, it sounds good, but that way we would lose the ability to bring
a dul from the last election version without passing an electionId, if such exists (which is used from the ui).

The reason why is because:

first we bring consent, the method used here brings some fields from its most recent election if this exists. If it doesnt the value in LastElectionStatus is null.

So, if consent.lastElectionStatus is null then it means that there's no election associated to
the consent for which we use consent's dul and dulurl.

On the other hand, if this value is not null,  it proceeds to check if an electionId is passed, resulting to bring the last election when the value is null, or the corresponding to the passed id.

You can see this last behavior described above, in retrieveElection, where we also throw the exception you mentioned.

Please if you agree to make this change anyway let me know.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseConsentAPI.java
```diff
@@ -473,4 +473,18 @@ public Consent getByName(String name) throws UnknownIdentifierException {
...
 
+    @Override
+    public Election retrieveElectionDul(Integer electionId, String consentId) throws UnknownIdentifierException{
+        Election election;
+        if (electionId != null) {
+            election = electionDAO.findElectionWithFinalVoteById(electionId);
+        } else {
+            election = electionDAO.getElectionWithFinalVoteByReferenceIdAndType(consentId, ElectionType.TRANSLATE_DUL.getValue());
+        }
+        if (election.getDataUseLetter() == null) {
```
ahaessly: actually, i don't think you need this check at all now that we are returning the election. and the code that calls it checks whether the election is null

ahaessly (approved):
approve with change to method name and not throwing exception from getElection

PR #290, opened 2018-07-30
BTRX 278 Group Names Already Voted

tadeork:
Jira issue [BTRX-271](https://broadinstitute.atlassian.net/browse/BTRX-278)
**Notes to use this new end-point:**
This end-point works only for Admin users. It should be used via a REST client, such as Postman.
_What it's needed?_
1. Admin user token
2. point to ``/api/consents/group-names/``
3. Headers: ``Accept: application/json`` ``Authorization: Bearer token_admin`` ``Content-Type: multipart/form-data``
4. Body of the request: ``Key -> data; Value -> file``

The info of the group names comes from the ORSP db. Run this query:
```
select  dur.vault_consent_id as consentId, i.summary as groupName
from issue i 
inner join data_use_restriction dur on dur.consent_group_key = i.project_key
where i.type = 'Consent Group'
and dur.vault_export_date is not null;
```

Export to _csv_ file. This file should be exported to JSON with the following format:
```
[ 
    {
        "consentId": "testId",
        "groupName": "lorem ipsum / 123"
    },
    {
        "consentId": "testId2",
        "groupName": "lorem ipsum / 124"
    },
    {
       "consentId": "testId3",
       "groupName": "lorem ipsum / 125"
    }
]
```

The result of this last export is the file that the end-point can process.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
```diff
@@ -85,4 +98,24 @@ protected Logger logger() {
         return Logger.getLogger("ConsentsResource");
     }
 
+
+    @POST
+    @Path("group-names")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.MULTIPART_FORM_DATA)
+    @RolesAllowed("ADMIN")
+    public Response updateGroupName(@Context UriInfo info, @Auth User user,
```
rushtong: Nit: method should be similarly pluralized.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
```diff
@@ -85,4 +98,24 @@ protected Logger logger() {
         return Logger.getLogger("ConsentsResource");
     }
 
+
+    @POST
+    @Path("group-names")
```
rushtong: Good choice on hyphenated path name

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
```diff
@@ -85,4 +98,24 @@ protected Logger logger() {
...
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.MULTIPART_FORM_DATA)
+    @RolesAllowed("ADMIN")
+    public Response updateGroupName(@Context UriInfo info, @Auth User user,
+                                    @FormDataParam("data") InputStream uploadedDataSet,
+                                    @FormDataParam("data") FormDataBodyPart data) {
+        ObjectMapper objectMapper = new ObjectMapper();
+        try {
+            List<Consent> groupNames = objectMapper.readValue(uploadedDataSet, new TypeReference<List<Consent>>(){});
```
rushtong: The variable name here is confusing. The json is parsed into "consents", not a list of pairs.

I would not expect coercing this to a consent. It makes more sense to me to parse two string pairs as a DTO instead of a full `Consent` object. The API should then be very specific about the values it takes for update. I suggest something like a `List<ImmutablePair<String, String>>` as the transfer object instead of a full consent. This API risks overstepping its scope if it thinks it is working with a real consent.

There should be some validation somewhere on the values to check for emptiness, duplicated names, etc.

rushtong (changes requested):
Please also review with @ahaessly, but see comments inline.

Could use some tests for both success and various failure cases at the dao and api layers. For example, a test on failed json parsing and invalid data in the json.

Lastly, I think this needs some documentation built into the code. I see that we're not documenting this via swagger - that's fine. However, there should be something somewhere that covers the steps described in this PR.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
tadeork: I don't quite understand what 'duplicated names' means, it's about the consentId or the groupName?

rushtong: Group name, in this model, can be duplicated, but in ORSP, they are definitely unique. I don't know what the requirement is here though. 

Another thought came to me about validation. If I have the same consent id three times, but with three different names - is that a valid input json? I think the validation should probably kick that out.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
```diff
@@ -85,4 +99,28 @@ protected Logger logger() {
...
     }
 
+
+    @POST
+    @Path("group-names")
+    @Produces(value = {MediaType.APPLICATION_JSON, MediaType.MULTIPART_FORM_DATA})
+    @Consumes(value = {MediaType.MULTIPART_FORM_DATA, MediaType.APPLICATION_JSON})
+    @RolesAllowed("ADMIN")
+    public Response updateGroupName(@Context UriInfo info, @Auth User user,
+                                    @FormDataParam("data") InputStream uploadedDataSet,
```
ahaessly: not sure what the annotations actually do, but is it correct for both this and the one below to be data?

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseConsentAPI.java
```diff
@@ -485,4 +479,28 @@ public Election retrieveElection(Integer electionId, String consentId) {
         }
         return election;
     }
+
+    @Override
+    public List<ConsentGroupNameDTO> verifyConsentsGroupNames(List<ConsentGroupNameDTO> consentGroupNames) {
```
ahaessly: i would change the name of this method to indicate it will actually do the update if the names are valid. something like verifyAndUpdateConsentGroupNames or updateConsentGroupNamesWhenValid

consent-ws/src/test/java/org/broadinstitute/consent/http/ConsentResourceTest.java
```diff
@@ -143,4 +151,59 @@ private Election updateElection(String consentId, Integer electionId) throws IOE
...
+
+        WebTarget webTarget = client.target(path2Url("/consents/group-names"));
+        MultiPart multipart = createFormData("consent-group-name-duplicated", "json");
+
+        mockValidateTokenResponse();
+        Response response = webTarget.request(MediaType.APPLICATION_JSON_TYPE)
+                .header("Authorization", "Bearer access-token")
+                .post(Entity.entity(multipart, multipart.getMediaType()));
+
+        assertEquals(BAD_REQUEST, response.getStatus());
```
ahaessly: check the response for the expected invalid group name

ahaessly (commented):
see comments

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
```diff
@@ -85,4 +99,59 @@ protected Logger logger() {
         return Logger.getLogger("ConsentsResource");
     }
 
+
+    /**
+     * Given taht this end-point isn't mapped in swagger follow this steps:
```
rushtong: Typos: 
* `taht` -> that
* `this` -> these

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
```diff
@@ -85,4 +99,59 @@ protected Logger logger() {
...
+     * 1. Works only for Admin users. It should be used via a REST client, such as Postman.
+     * What it's needed?
+     *     Admin user token
+     *     point to /api/consents/group-names/
+     *     Headers: Accept: application/json Authorization: Bearer token_admin Content-Type: multipart/form-data
```
rushtong: I missed this in the first review. Why are we requiring a file upload? Can't this simply be a POST with json content? I think if you did that, then several things would be easier: 

* This endpoint could just be a simple curl call.  
* You wouldn't need an `InputStream` in the endpoint method. 
* The endpoint method could use Dropwizard's default type coercion on the input object instead of parsing it out via objectMapper.
* The tests that rely on files in test/resources could instead just do normal posts instead of file uploads.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
```diff
@@ -85,4 +99,59 @@ protected Logger logger() {
...
+     * @param user
+     * @param uploadedDataSet
+     * @return
+     */
+    @POST
+    @Path("group-names")
+    @Produces(value = {MediaType.APPLICATION_JSON, MediaType.MULTIPART_FORM_DATA})
+    @Consumes(value = {MediaType.MULTIPART_FORM_DATA, MediaType.APPLICATION_JSON})
+    @RolesAllowed("ADMIN")
+    public Response updateGroupName(@Context UriInfo info, @Auth User user,
```
rushtong: Nit: pluralize to more clearly represent the API path.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseConsentAPI.java
```diff
@@ -485,4 +479,28 @@ public Election retrieveElection(Integer electionId, String consentId) {
         }
         return election;
     }
+
+    @Override
+    public List<ConsentGroupNameDTO> verifyAndUpdateConsentGroupNames(List<ConsentGroupNameDTO> consentGroupNames) {
```
rushtong: From a testing and maintenance perspective, I prefer not to have dual-purpose methods like this. There are two distinct functions here - validate data and update data. To me, bundling them up is more difficult to both test and to maintain. If you split those out, both the test code and the API could deal with the response more cleanly.

consent-ws/src/test/java/org/broadinstitute/consent/http/AbstractTest.java
```diff
@@ -209,4 +209,17 @@ protected void shutDownMongo() {
...
+                setRequiresManualReview(false).
+                setUseRestriction(useRestriction).
+                setDataUse(dataUse).
+                setName(UUID.randomUUID().toString()).
+                setCreateDate(createDate).
+                setLastUpdate(createDate).
+                setSortDate(createDate).
+                build();
+    }
 }
```
rushtong: Nit: not your code, but could use a newline here.

consent-ws/src/test/java/org/broadinstitute/consent/http/ConsentResourceTest.java
```diff
@@ -143,4 +154,65 @@ private Election updateElection(String consentId, Integer electionId) throws IOE
         return updated;
     }
 
+    @Test
+    public void testUpdateConsentGroupName() throws Exception {
+        Client client = ClientBuilder.newBuilder().register(MultiPartFeature.class).build();
+
+
```
rushtong: Nit: extra lines

consent-ws/src/test/java/org/broadinstitute/consent/http/ConsentResourceTest.java
```diff
@@ -143,4 +154,65 @@ private Election updateElection(String consentId, Integer electionId) throws IOE
...
+                .header("Authorization", "Bearer access-token")
+                .post(Entity.entity(multipart, multipart.getMediaType()));
+
+        assertEquals(OK, response.getStatus());
+    }
+
+    @Test
+    public void testUpdateConsentGroupNameBadDuplicated() throws Exception {
+        String invalidJson = "[{\"consentId\":\"testId3\",\"groupName\":\"wrong group name\"}," +
+                "{\"consentId\":\"\",\"groupName\":\"empty consent Id\"}]";
```
rushtong: Since you have a DTO for this class, I think it would make sense to use that and convert it to json for the test. String values like this are hard to maintain in test code.

consent-ws/src/test/resources/json/consent-group-name-duplicated.json
```diff
@@ -0,0 +1,14 @@
...
+  },
+  {
+    "consentId": "testId3",
+    "groupName": "wrong group name"
+  },
+  {
+    "consentId": "",
+    "groupName": "empty consent Id"
+  }
+]
```
rushtong: This test case doesn't actually have duplicated group names. It does have duplicated consent IDs, so I think you might need to add a separate test for the duplicate group name case.

rushtong (changes requested):
Nice work adding test coverage! I think this can still use some work though. Comments inline.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
```diff
@@ -85,4 +94,55 @@ protected Logger logger() {
         return Logger.getLogger("ConsentsResource");
     }
 
+
+    /**
+     * Given that these end-point isn't mapped in swagger follow this steps:
```
rushtong: Should be: 
`Given that this end-point isn't mapped in swagger, follow these steps`

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
```diff
@@ -85,4 +94,55 @@ protected Logger logger() {
...
+     * 1. Works only for Admin users. It should be used via a REST client, such as Postman.
+     * What it's needed?
```
rushtong: Should be:
`Requirements:`

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
```diff
@@ -1,15 +1,24 @@
...
 import org.broadinstitute.consent.http.service.AbstractConsentAPI;
 import org.broadinstitute.consent.http.service.ConsentAPI;
 
 import javax.annotation.security.PermitAll;
+import javax.annotation.security.RolesAllowed;
 import javax.ws.rs.*;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.MediaType;
 import java.util.*;
+import java.util.List;
```
ahaessly: you can remove this import since you have java.util.*

tadeork: This was a change made by the IDE. I was thinking to put as it was, what is the best option?

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseConsentAPI.java
```diff
@@ -485,4 +479,34 @@ public Election retrieveElection(Integer electionId, String consentId) {
         }
         return election;
     }
+
+    @Override
+    public List<ConsentGroupNameDTO> updateConsentGroupNames(List<ConsentGroupNameDTO> consentGroupNames) {
```
ahaessly: I still think the names of this method is misleading as well as updateConsentGroupName. If I just look at the name of this method, it doesn't indicate anything about verification and I might think the updated consents are returned instead returned values indicating an error.

This is one case where in the calling code, it might be more straightforward to call verify separately and return an error if there are values. and then call update.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
rushtong: I'm pretty sure Belatrix has been avoiding `*` imports as a policy. I know that in AoU, we don't use them at all, and I can see the merit to that.

ahaessly (approved):
just one suggestion.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseConsentAPI.java
tadeork: You're right! I didn't realise that with this change is less code and more intuitive.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
```diff
@@ -85,4 +97,56 @@ protected Logger logger() {
         return Logger.getLogger("ConsentsResource");
     }
 
+
+    /**
+     * Given that this end-point isn't mapped in swagger follow this steps:
```
rushtong: Grammar nit: 'follow this steps' should be `follow these steps`

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
```diff
@@ -85,4 +97,56 @@ protected Logger logger() {
...
+     *     Body of the request: Key -> data; Value -> file
+     *
+     * 2. The info of the group names comes from the ORSP db. Run this query:
+     * <code>select  dur.vault_consent_id as consentId, i.summary as groupName
+     * from issue i
+     * inner join data_use_restriction dur on dur.consent_group_key = i.project_key
+     * where i.type = 'Consent Group'
+     * and dur.vault_export_date is not null;</code>
+     *
+     * 3. Export to csv file. This file should be exported to JSON with the following format:
```
rushtong: Nit: Is this still true now that you're using a POST with structured json?

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
```diff
@@ -85,4 +97,56 @@ protected Logger logger() {
...
     }
 
+
+    /**
+     * Given that this end-point isn't mapped in swagger follow this steps:
+     * 1. Works only for Admin users. It should be used via a REST client, such as Postman.
+     * Requirements:
+     *     Admin user token
+     *     point to /api/consents/group-names/
+     *     Headers: Accept: application/json Authorization: Bearer token_admin Content-Type: multipart/json
```
rushtong, line 107: Is ` Content-Type: multipart/json` still true?

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
```diff
@@ -85,4 +97,56 @@ protected Logger logger() {
...
+    @Produces(MediaType.APPLICATION_JSON)
+    @RolesAllowed("ADMIN")
+    public Response updateGroupNames(@Context UriInfo info, @Auth User user, List<ConsentGroupNameDTO> data) {
+        try {
+            List<ConsentGroupNameDTO> errors = api.verifyConsentGroupNames(data);
+            if (errors.isEmpty()) {
+                api.updateConsentGroupNames(data);
+                return Response.status(Response.Status.OK).build();
+            } else {
+                return Response.status(Response.Status.BAD_REQUEST).entity(errors).build();
```
rushtong, line 144: This could be simpler by throwing a BadRequestException and let the superclass handle it with `createExceptionResponse`. That's a minor nit, however - don't hold up this PR on this - it's more for general knowledge sharing.

consent-ws/src/test/java/org/broadinstitute/consent/http/ConsentResourceTest.java
```diff
@@ -1,31 +1,47 @@
 package org.broadinstitute.consent.http;
 
+import com.google.common.io.Resources;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.stream.JsonReader;
```
rushtong: Some of these imports are unused.

consent-ws/src/test/java/org/broadinstitute/consent/http/ConsentResourceTest.java
```diff
@@ -143,4 +159,103 @@ private Election updateElection(String consentId, Integer electionId) throws IOE
         return updated;
     }
 
+    @Test
+    public void testUpdateConsentGroupName() throws Exception {
+        Client client = ClientBuilder.newBuilder().register(MultiPartFeature.class).build();
```
rushtong: Do you really need `MultiPartFeature`? Tests seem to run fine without that - I think it was necessary for file uploading.

rushtong (approved):
👍 Looks good! Some minor grammar comments and some places where there's unused code/imports. If you make those changes, no need to re-review.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
rushtong: Please ignore ... this is important.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
tadeork: I don't quite get this change. It will still pass the PR check if I let this as is?

rushtong: Correct, this is not something I'm asking you to do, just a note that there is more than one way to return an error response. We have some ease-of-use things that make this a little simpler. Here's an example of letting the superclass handle this exact exception:

https://github.com/DataBiosphere/consent/blob/4b03066106e4a22290b3e62e440d55557a715c32/consent-ws/src/main/java/org/broadinstitute/consent/http/resources/IndexerResource.java#L65

In this class, it could be something like this:
```
return createExceptionResponse(new BadRequestException(errors.toString()));
```

But, again, don't feel obligated to make this change.

tadeork: yes! I just understand now. It looks better this way. Thanks for taking the time to explain this to me.

PR #295, opened 2018-08-17
BTRX- 268 - DUOS: Authenticate Researcher Data Access Request w/ ERA Commons ID

LeoForconesi:
## Addresses Jira Story:
[BTRX-268](https://broadinstitute.atlassian.net/browse/BTRX-268) : _Authenticate Researcher Data Access Request w/ ERA Commons ID_ 
## Changes in this PR:
- This story and code is related to [BTRX-268](https://github.com/DataBiosphere/consent-ui/pull/202) Ui branch
- Implement JWT authentication for shibboleth's token. We are using [JJWT](https://www.google.com.ar/search?q=jjwt&oq=jjwt&aqs=chrome..69i57j0j69i60l3j69i59.2231j0j9&sourceid=chrome&ie=UTF-8) library for this purpose
- Save, Map and return ERA Commons values, such as Authorized status, Expiration Date and Nih User Name
- Modified logic of how a Researcher Profile is updated. This way we can update without deleting all properties by User Id
- Replace Data Access Requests values that were mapped from MySql to MongoDb so later in the Ui we show a snapshot of these values at the moment of its creation
- Some minor code refactor.

_UPDATE_
- We have included PDFs missing data fields corresponding to [BTRX-289](https://broadinstitute.atlassian.net/browse/BTRX-289) - Missing information in downloadable Full DAR Application PDF 

_NOTE_
- Before merging we need to add Nih Jwt secret to Vault (test & prod) and modify Docker config template version to include the new yaml field.

---

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


consent-ws/src/main/java/org/broadinstitute/consent/http/db/ResearcherPropertyDAO.java
```diff
@@ -37,6 +37,9 @@
     @SqlUpdate("delete from researcher_property where  userId = :userId")
     void deleteAllPropertiesByUser(@Bind("userId") Integer userId);
 
+    @SqlBatch("delete from researcher_property where userId = :userId and propertyKey = :propertyKey")
+    void deletePropertieByUserAndKey(@BindBean Collection<ResearcherProperty> researcherProperties);
```
rushtong: Nit: typo, `Propertie` should be pluralized `deletePropertiesByUserAndKey`

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/NihAccountResource.java
```diff
@@ -0,0 +1,46 @@
+package org.broadinstitute.consent.http.resources;
+
+import org.broadinstitute.consent.http.service.NihAuthApi;
+
+import javax.annotation.security.RolesAllowed;
+import javax.ws.rs.*;
+import javax.ws.rs.core.Response;
+import java.util.Map;
+
+@Path("{api : (api/)?}nih-login/")
```
rushtong: Nit: You probably don't need this regex in the path. We used this technique for handling basic authentication and I think the pattern stuck. It can probably just be `@Path("api/nih-login/")`. This should be true in lots of other cases too.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/NihAccountResource.java
```diff
@@ -0,0 +1,46 @@
...
+        this.nihAuthApi = nihAuthApi;
+    }
+
+    @POST
+    @Path("{userId}/{token}")
+    @Produces("application/json")
+    @RolesAllowed("RESEARCHER")
+
+    public Response registerResearcher(@PathParam("userId") Integer userId, @PathParam("token") String jwt) {
+        try{
```
rushtong: Nit: spacing here and on the catch line as well.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/NihServiceAPI.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Map;
+import java.util.List;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Calendar;
+import java.util.Date;
+
+
+
+
```
rushtong: Nit: several extra lines here, and below.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/NihServiceAPI.java
```diff
@@ -0,0 +1,87 @@
...
+        if (nihUserName != null && !nihUserName.isEmpty()) {
+            nihComponents.put(ResearcherFields.ERA_STATUS.getValue(), Boolean.TRUE.toString());
+            nihComponents.put(ResearcherFields.ERA_USERNAME.getValue(), nihUserName);
+            nihComponents.putAll(updateAndExpirationDates());
+            researcherAPI.updateResearcher(nihComponents, userId, false);
+        }
+        return nihComponents;
+    }
+
+    private Map<String, String> updateAndExpirationDates() {
```
rushtong: Nit, naming. Should this be `updateEraExpirationDates` instead? Also, the `update` doesn't make sense here because we're not actually updating, just generating a new map with a single date value.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/users/handler/DatabaseResearcherAPI.java
```diff
@@ -59,30 +58,31 @@ public DatabaseResearcherAPI(ResearcherPropertyDAO researcherPropertyDAO, DACUse
...
-        if(validate) validateRequiredFields(researcherPropertiesMap);
+        if (validate) validateRequiredFields(researcherPropertiesMap);
         validateExistentFields(researcherPropertiesMap);
         Boolean isUpdatedProfileCompleted = Boolean.valueOf(researcherPropertiesMap.get(ResearcherFields.COMPLETED.getValue()));
         String completed = researcherPropertyDAO.isProfileCompleted(userId);
         Boolean isProfileCompleted = StringUtils.isEmpty(completed) ? false : Boolean.valueOf(completed);
         List<ResearcherProperty> properties = getResearcherProperties(researcherPropertiesMap, userId);
-        if(!isProfileCompleted && isUpdatedProfileCompleted){
-            saveProperties(userId, properties);
+        if (!isProfileCompleted && isUpdatedProfileCompleted){
```
rushtong: Nice spacing fix, but need one before the `{` as well.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/users/handler/DatabaseResearcherAPI.java
```diff
@@ -97,12 +97,18 @@ public void deleteResearcherProperties(Integer userId) {
...
 
+    @Override
+    public void deleteResearcherSpecificProperties(List<ResearcherProperty> properties) {
+        researcherPropertyDAO.deletePropertieByUserAndKey(properties);
+    }
+
     @Override
     public Map<String, String> describeResearcherPropertiesForDAR(Integer userId) {
-        Map<String, String> properties = describeResearcherPropertiesMap(userId);
+         Map<String, String> properties = describeResearcherPropertiesMap(userId);
```
rushtong: Nit: extra space.

rushtong:
Do you want to document this in `api-docs.yaml` or is that intentionally not there?

LeoForconesi:
Thanks for your review Greg! I just updated `api-docs.yaml`, i was missing that.

rushtong:
Caveat, I haven't tested anything functionally, so, pending others' approval, 👍

LeoForconesi:
@ahaessly 
We have updated this PR. When you have some moment, please review.

LeoForconesi:
@rushtong if you want to take a look at this code again it would really helpfull. It hasn't changed a lot though since the last time you reviewed it, only develop branch updates and adding pdf corresponding fields.

rushtong:
With respect to `nihAuthentication`, we'll need a corresponding change to the config similar to the one we did for the [dataset configs](https://github.com/broadinstitute/firecloud-develop/pull/1408)

As a practical note, is there a way we can write the `NihConfiguration` such that it can be null in the configs in the deployed app? If we can do that, and log (log level severe) the fact that it doesn't exist, then the consent PR and the fc-dev PR can be decoupled. This would be a nice approach for any config change. Currently, DropWizard config reader complains if there's any kind of mis-match at all, likely because it's using jackson to model the config class alongside the config file. It would be a *lot* easier to deploy if this can be done.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataRequestReportsResource.java
```diff
@@ -79,4 +89,34 @@ public Response downloadReviewedDARs() {
...
+    private Boolean requiresManualReview(Document dar) throws IOException {
+        Map<String, Object> form = parseAsMap(dar.toJson());
+        for (String field: fieldsForManualReview) {
+            if (form.containsKey(field)) {
+                if (Boolean.valueOf(form.get(field).toString())) {
+                    return true;
+                }
+            }
+        }
+        return false;
```
rushtong: Could be a lot more compact with a stream: 
```
        return Arrays.stream(fieldsForManualReview).
            filter(form::containsKey).
            map(field -> Boolean.valueOf(form.get(field).toString())).
            collect(Collectors.toSet()).
            contains(Boolean.TRUE);
```

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/NihAccountResource.java
```diff
@@ -0,0 +1,46 @@
+package org.broadinstitute.consent.http.resources;
+
+import org.broadinstitute.consent.http.service.NihAuthApi;
+
+import javax.annotation.security.RolesAllowed;
+import javax.ws.rs.*;
+import javax.ws.rs.core.Response;
+import java.util.Map;
```
rushtong: unused

rushtong:
FYI, I have not looked into this, but we should do a security review due to JWT. Please review https://www.owasp.org/index.php/JSON_Web_Token_(JWT)_Cheat_Sheet_for_Java for any potential issues. I'll ping our devops security to see if they want to review.

rushtong:
@adrazhi - Can I ask for your review on this due to JWT? Let's meet to discuss if you think we need to.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/Resource.java
```diff
@@ -67,6 +72,14 @@ protected Response createExceptionResponse(Exception e) {
...
         dispatch.put(UpdateConsentException.class, e ->
                 Response.status(Response.Status.BAD_REQUEST).entity(new Error(e.getMessage(), Response.Status.BAD_REQUEST.getStatusCode())).build());
+        dispatch.put(SignatureException.class, e->
+                Response.status(Response.Status.BAD_REQUEST).entity(new Error(e.getMessage(), Response.Status.BAD_REQUEST.getStatusCode())).build());
+        dispatch.put(MalformedJwtException.class, e->
+                Response.status(Response.Status.BAD_REQUEST).entity(new Error(e.getMessage(), Response.Status.BAD_REQUEST.getStatusCode())).build());
+        dispatch.put(UnsupportedJwtException.class, e->
+                Response.status(Response.Status.BAD_REQUEST).entity(new Error(e.getMessage(), Response.Status.BAD_REQUEST.getStatusCode())).build());
+        dispatch.put(DecodingException.class, e->
+                Response.status(Response.Status.BAD_REQUEST).entity(new Error(e.getMessage(), Response.Status.BAD_REQUEST.getStatusCode())).build());
```
rushtong: I question whether all of these are really bad requests. Some seem like they should at least be internal server errors. Additionally, I think it might make more sense to catch these in `NihServiceAPI` instead and log them since we should be alerted via sentry if we see these kinds of errors.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DataAccessParser.java
```diff
@@ -194,6 +195,43 @@ public PDAcroForm fillDARForm(Document dar, Map<String, String> researcherProper
                     field.setValue(getYesOrNoValue(dar.getBoolean(DarConstants.POA)));
                     break;
                 }
+                case DarConstants.CHECK_COLLABORATOR: {
+                    field.setValue(getYesOrNoValue(dar.getBoolean(DarConstants.CHECK_COLLABORATOR)));
+                    break;
+                }
+                case "userStatus": {
```
rushtong: Any reason we're not using constants here and below?

consent-ws/src/main/java/org/broadinstitute/consent/http/service/NihServiceAPI.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public Map<String, String> authenticateNih(String jwt, Integer userId) throws SignatureException, MalformedJwtException, UnsupportedJwtException, DecodingException {
+
+        Key key = Keys.hmacShaKeyFor(nihConfiguration.getSigningSecret());
+
+        String nihUserName = Jwts.parser()
+                .setSigningKey(key)
+                .parsePlaintextJws(jwt).getBody();
```
rushtong: See my comment about exceptions above. I think this block should catch and log (probably at the warning level for client-provided `jwt` tokens) all of the exceptions possible with some explanatory language for each.

consent-ws/src/main/resources/assets/api-docs.yaml
```diff
@@ -2049,6 +2049,55 @@ paths:
...
+      parameters:
+        - name: userId
+          in: path
+          required: true
+          type: integer
+          format: int32
+      tags:
+        - User
+        - Researcher
+        - Era Commons
```
rushtong: nit: Technically, it's `eRA Commons`. See https://public.era.nih.gov/commons/public/login.do and  https://era.nih.gov/

rushtong:
With respect to `RequestApplication.pdf` - instead of checking that into code, I think a better approach is to put it in a consent-accessible bucket and render it via the GCS api. This way we don't need to do a code release for a simple change in the template file.

consent-ws/src/test/java/org/broadinstitute/consent/http/NihAccountTest.java
```diff
@@ -0,0 +1,98 @@
...
+        // Check null, not found, and not allowed users
+        checkStatus(NOT_FOUND, post(client, path2Url(String.format(AUTHENTICATE_NIH_TOKEN, null, VALID_TOKEN)), val));
+        checkStatus(NOT_FOUND, post(client, path2Url(String.format(AUTHENTICATE_NIH_TOKEN, 10101010, VALID_TOKEN)), val));
+
+        // Check different invalid tokens
+        checkStatus(BAD_REQUEST, post(client, path2Url(String.format(AUTHENTICATE_NIH_TOKEN, RESEARCHER_USER_ID, INVALID_TOKEN_SIGNATURE)), val));
+        checkStatus(BAD_REQUEST, post(client, path2Url(String.format(AUTHENTICATE_NIH_TOKEN, RESEARCHER_USER_ID, INVALID_TOKEN_MALFORMED)), val));
+        checkStatus(BAD_REQUEST, post(client, path2Url(String.format(AUTHENTICATE_NIH_TOKEN, RESEARCHER_USER_ID, null)), val));
+
+
```
rushtong: nit: unnecessary extra lines

consent-ws/src/test/java/org/broadinstitute/consent/http/NihAccountTest.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+
+public class NihAccountTest extends AbstractTest{
```
rushtong: I don't have a suggestion, but would love to see if there's a way to test a security scenario. Will try to spend some time digesting the OWASP doc and see if there's something more we can look for.

rushtong (changes requested):
Looks good, but I definitely have some concerns that are beyond my knowledge at the moment.

consent-ws/pom.xml
```diff
@@ -215,6 +215,27 @@
 
     <dependencies>
 
+
+        <dependency>
+            <groupId>io.jsonwebtoken</groupId>
+            <artifactId>jjwt-api</artifactId>
+            <version>0.10.1</version>
```
rushtong: Latest version here is `0.10.5` ... there have been a bunch of releases in August.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/NihServiceAPI.java
rushtong: See the [quickstart example](https://github.com/jwtk/jjwt#quickstart): 
```
try {
    Jwts.parser().setSigningKey(key).parseClaimsJws(compactJws);
    //OK, we can trust this JWT
} catch (JwtException e) {
    //don't trust the JWT!
}
```
In our implementation, `parsePlaintextJws` is where we care about catching any kind of error. This is where we'll find any kind of security problem with the token. These should be logged as severe (with some more info about user, request, etc.,) so we can follow up with those users via our normal communications process.

adrazhi:
> @adrazhi - Can I ask for your review on this due to JWT? Let's meet to discuss if you think we need to.

@rushtong 
So the first concern related to JWT is the integrity protection and to ensure JWT is protected by either a signature or a MAC. Do not allow the unsecured JWTs: {"alg":"none"}.  I think you're going the signature direction which is great and I think it's preferred over HMAC. 

Few other things I'd recommend watching out for: 
1. Use a cryptographically random JWT secret. You are using [this library](https://github.com/jwtk/jjwt) to do that, correct?
2. What is the current token expiration time? 
3. Obvious but worth mentioning to not store any sensitive data in the JWT payload as it can be decoded [easily.](https://jwt.io/#debugger-io) 
4. On headers: Ensure embedded "jwk elements are not trusted". 

Let me know if you have any questions.

src/main/java/org/broadinstitute/consent/http/resources/NihAccountResource.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @POST
+    @Path("{userId}")
+    @Produces("application/json")
+    @RolesAllowed("RESEARCHER")
+
+    public Response registerResearcher(@PathParam("userId") Integer userId, NIHUserAccount nihAccount) {
+        try {
+            return Response.status(Response.Status.OK).entity(nihAuthApi.authenticateNih(nihAccount, userId)).build();
+
```
rushtong: nit: extra line

src/main/java/org/broadinstitute/consent/http/resources/NihAccountResource.java
```diff
@@ -0,0 +1,52 @@
...
+    public NihAccountResource(NihAuthApi nihAuthApi) {
+        this.nihAuthApi = nihAuthApi;
+    }
+
+    @POST
+    @Path("{userId}")
+    @Produces("application/json")
+    @RolesAllowed("RESEARCHER")
+
+    public Response registerResearcher(@PathParam("userId") Integer userId, NIHUserAccount nihAccount) {
```
rushtong: I'm still worried that this takes a userid as a path param. I think it would be much safer to extract this using `@Auth User user`

src/main/java/org/broadinstitute/consent/http/service/NihServiceAPI.java
```diff
@@ -0,0 +1,55 @@
...
+        properties.add(new ResearcherProperty(userId, ResearcherFields.ERA_STATUS.getValue()));
+        properties.add(new ResearcherProperty(userId, ResearcherFields.ERA_USERNAME.getValue()));
+
+        researcherAPI.deleteResearcherSpecificProperties(properties);
+    }
+
+
+    private String generateEraExpirationDates() {
+        Date currentDate = new Date();
+
```
rushtong: nit: extra line

src/main/resources/assets/api-docs.yaml
```diff
@@ -2049,6 +2049,51 @@ paths:
           description: The requested user doesnt exist in the system
           schema:
             $ref: '#/definitions/ErrorResponse'
+  /api/nih-register/{userId}:
```
rushtong: nit: I can't remember this or find it in the comments, but I thought we discussed making this just `/api/nih` instead of `/api/nih-register`

rushtong (changes requested):
Two change requests, some nits, otherwise looks good!

LeoForconesi:
Closing this PR. Review comments are solved in [PR-309](https://github.com/DataBiosphere/consent/pull/309).

PR #297, opened 2018-10-18
BTRX-394 Add 'Send Notifications To' field to User Profile in DUOS to enable notifications to an additional email

vvicario:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


consent-ws/src/main/java/org/broadinstitute/consent/http/db/DACUserDAO.java
```diff
@@ -51,11 +51,12 @@ Integer insertDACUser(@Bind("email") String email,
                           @Bind("displayName") String displayName,
                           @Bind("createDate") Date createDate);
 
-    @SqlUpdate("update dacuser set email=:email, displayName=:displayName where dacUserId=:id")
+    @SqlUpdate("update dacuser set email=:email, displayName=:displayName, additional_email=:additionalEmail where dacUserId=:id")
```
rushtong, line 54: Similarly to PR 298, we now have mixed formats for table column names. We should standardize (via a new story) changing columns from camel-case format to underscore format.

rushtong:
Note: Is `protected Mail generateEmailMessage` in `MailMessage` used any longer? If not, we should remove it. It looks like all usages of that moved to `protected List<Mail> generateEmailMessage`

Additionally, `protected List<Mail> generateEmailMessage` should probably be renamed to be plural.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/EmailNotifierService.java
```diff
@@ -3,25 +3,19 @@
 import com.mongodb.BasicDBObject;
 import freemarker.template.TemplateException;
 import org.apache.commons.collections.CollectionUtils;
-import org.broadinstitute.consent.http.db.DACUserDAO;
-import org.broadinstitute.consent.http.db.ElectionDAO;
-import org.broadinstitute.consent.http.db.MailMessageDAO;
-import org.broadinstitute.consent.http.db.MailServiceDAO;
-import org.broadinstitute.consent.http.db.VoteDAO;
+import org.apache.commons.lang3.StringUtils;
+import org.broadinstitute.consent.http.db.*;
```
rushtong: revert the * imports

consent-ws/src/main/java/org/broadinstitute/consent/http/service/EmailNotifierService.java
```diff
@@ -255,6 +280,10 @@ public void sendNewRequestHelpMessage(HelpReport helpReport) throws MessagingExc
             List<DACUser> users = dacUserDAO.describeUsersByRoleAndEmailPreference(DACUserRoles.ADMIN.getValue(), true);
             Writer template = templateHelper.getHelpReportTemplate(helpReport, SERVER_URL);
             List<String> usersEmail = users.stream().map(DACUser::getEmail).collect(Collectors.toList());
+            List<String> additionEmails = users.stream().map(DACUser::getAdditionalEmail).collect(Collectors.toList());
+            if(CollectionUtils.isNotEmpty(additionEmails)) usersEmail.addAll(additionEmails);
+            List<String> academicEmails =  getAcademicEmails(users);
+            if(CollectionUtils.isNotEmpty(academicEmails)) usersEmail.addAll(academicEmails);
```
rushtong: There is a ton of repetition of collecting academic and additional email addresses. These should all be refactored to a common method to get the valid email addresses for a user. 

I am confused on a couple points that aren't clear in the story. This code sends email to **all** addresses found. 
* The story doesn't seem to address the point of if a user wants email to only go to one or the other address. 
* I don't see any logic here about honoring the user's selection for not getting notifications. This is actually a complaint from lots of dev users now. Many, many folks on dev are getting spammed by testing. We should have something that does that for all users for all message types, not just for admins.

rushtong (changes requested):
Code looks good, but I have some questions and at least one functional request.

* This code sends email to all addresses found for a user.
* The story isn't clear to me if the request is for email to go to two email addresses, or if the second one is the **real one** and email should only go there. If that is the case (and I think that's the more common use case), this might be better named contact email instead of additional email.
* I don't see any mechanism for opting out. We have this complaint now in dev - everyone in the system gets spammed. We need something that checks for a user's email opt-in/out status.

wloforteBroad:
@rushtong 

As per request by story https://broadinstitute.atlassian.net/browse/BTRX-394:
*Future State:*
*In the future, DUOS will provide users with the _option _ to add another email which will also receive notifications. DUOS users can either enter this information in their profile page in a 'Send additional email notifications to this email:' field.*
*Additionally, DUOS Researcher users who are required to input an 'Academic/Business Email Address' (required field) can click a checkbox to have notifications sent to their 'Academic/Business Email Address'.*

- This means that the new e-mail will also get the notifications the user email is currently receiving. And If the user is a Researcher, the checkbox is to also send the email to that address.
- About opting out, the only users that can do that are Admins that can "switch off" the notifications, but only the ones they get for being admins, they will still get notifications sent to them for having another role, for example to vote as a Data Owner, among others.
- This was our understanding of the ticket, If the request is different we will need to discuss this, edit the ticket and work on it again.

rushtong:
Thanks @wloforte - that's helpful. I would recommend reviewing the requirements here - email is already a problem and this risks increasing the scope of it.

vvicario:
Hi @rushtong I worked on your feedback, please let me know if you have any other comment

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DACUserResource.java
```diff
@@ -98,6 +98,21 @@ public Response update(@Context UriInfo info, Map<String, DACUser> dac, @PathPar
         }
     }
 
+    @PUT
+    @Path("/main_fields/{id}")
```
rushtong: Nit: We generally camel case paths - the only exception I see in the code is `group-names`. We should try to be consistent here. 
```suggestion
    @Path("/mainFields/{id}")
```

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseElectionAPI.java
```diff
@@ -8,12 +8,12 @@
...
 import org.apache.commons.lang3.StringUtils;
+import org.broadinstitute.consent.http.db.MailMessageDAO;
+import org.broadinstitute.consent.http.db.ElectionDAO;
 import org.broadinstitute.consent.http.db.ConsentDAO;
+import org.broadinstitute.consent.http.db.VoteDAO;
 import org.broadinstitute.consent.http.db.DACUserDAO;
 import org.broadinstitute.consent.http.db.DataSetDAO;
-import org.broadinstitute.consent.http.db.ElectionDAO;
-import org.broadinstitute.consent.http.db.MailMessageDAO;
-import org.broadinstitute.consent.http.db.VoteDAO;
```
rushtong: It would be a cleaner PR if this file were reverted completely since there are no changes.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/EmailNotifierService.java
```diff
@@ -253,12 +262,22 @@ public void sendNewResearcherCreatedMessage(Integer researcherId, String action)
...
     }
 
+    private List<String> getEmails(List<DACUser> users) {
+        List<String> usersEmail = users.stream().map(DACUser::getEmail).collect(Collectors.toList());
+        List<String> additionalEmails = users.stream().filter(user -> StringUtils.isNotEmpty(user.getAdditionalEmail())).map(DACUser::getAdditionalEmail).collect(Collectors.toList());
+        if(CollectionUtils.isNotEmpty(additionalEmails)) usersEmail.addAll(additionalEmails);
+        List<String> academicEmails =  getAcademicEmails(users);
+        if(CollectionUtils.isNotEmpty(academicEmails)) usersEmail.addAll(additionalEmails);
+        return usersEmail;
+    }
```
rushtong, line 284: I think this would be safer as a set, not a list, so you don't get duplicates. Also, some of this can be done more legibly with a stream. The logic in line 274 is pretty unintuitive. You could replace lines 273 - 275 with 
```
        Set<String> emails = users.stream()
            .map(u -> new ArrayList<String>(){{add(u.getEmail()); add(u.getAdditionalEmail());}})
            .flatMap(Collection::stream)
            .filter(s -> !s.isEmpty())
            .collect(Collectors.toSet());
```
Then add in the non-empty academicEmails.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/EmailNotifierService.java
```diff
@@ -478,4 +499,22 @@ private String retrieveElectionTypeStringCollect(String electionType) {
...
+                if(checkNotification.isPresent()){
+                    if (checkNotification.get().getPropertyValue().equals("true")) {
+                        ResearcherProperty academicEmailRP = researcherProperties.stream().filter(rp -> rp.getPropertyKey().equals(ResearcherFields.ACADEMIC_BUSINESS_EMAIL.getValue())).findFirst().get();
+                        academicEmails.add(academicEmailRP.getPropertyValue());
+                    }
+                }
+
+            }
+        }
+        return academicEmails;
```
rushtong, line 523: The structure here makes me nervous. Inside a loop, we make a db call to check the role, then we make another db call to found the researcher properties. Is the `checkResearcherRole` call necessary if we're following this up with a query on researcher properties? Doesn't that second query mean the properties are for a researcher? It would make more sense to me to: 

- Batch query for all researcher properties for user ids in the DACUser list
- Inner join that on the role table to enforce researcher role
- Filter the query on property key in (CHECK_NOTIFICATIONS, ACADEMIC_BUSINESS_EMAIL)
- Iterate over the users
  - For properties that match check notifications and have a non-empty email, add to list.

consent-ws/src/test/java/org/broadinstitute/consent/http/mail/message/DelegateResponsibilitiesMessageTest.java
```diff
@@ -23,8 +25,8 @@ public void setUp(){
 
     @Test
     public void testMessageSubject() throws MessagingException {
-        Mail message = new DelegateResponsibilitiesMessage().delegateResponsibilitiesMessage("to@address.com", "from@address.com", template);
-        assertTrue(message.getSubject().equals("You have been assigned a New Role in DUOS."));
+        List<Mail> messages = new DelegateResponsibilitiesMessage().delegateResponsibilitiesMessage(Collections.singletonList("to@address.com"), "from@address.com", template);
+        assertTrue(messages.get(0).getSubject().equals("You have been assigned a New Role in DUOS."));
     }
 
 }
```
rushtong: nit: newline

consent-ws/src/test/java/org/broadinstitute/consent/http/mail/message/DisabledDatasetMessageTest.java
```diff
@@ -23,8 +25,8 @@ public void setUp(){
 
     @Test
     public void testMessageSubject() throws MessagingException {
-        Mail message = new DisabledDatasetMessage().disabledDatasetMessage("to@address.com", "from@address.com", template, "DAR-123", "SomeType");
-        assertTrue(message.getSubject().equals("Datasets not available for Data Access Request Application id: DAR-123."));
+        List<Mail> messages = new DisabledDatasetMessage().disabledDatasetMessage(Collections.singletonList("to@address.com"), "from@address.com", template, "DAR-123", "SomeType");
+        assertTrue(messages.get(0).getSubject().equals("Datasets not available for Data Access Request Application id: DAR-123."));
     }
 
 }
```
rushtong: nit: newline

consent-ws/src/test/java/org/broadinstitute/consent/http/mail/message/FlaggedDarApprovedMessageTest.java
```diff
@@ -24,8 +26,8 @@ public void setUp(){
 
     @Test
     public void testMessageSubject() throws MessagingException {
-        Mail message = new FlaggedDarApprovedMessage().flaggedDarMessage("to@address.com", "from@address.com", template, "DS-123", "SomeType");
-        assertTrue(message.getSubject().equals("DS-123 that requires data owners reviewing approved."));
+        List<Mail> messages = new FlaggedDarApprovedMessage().flaggedDarMessage(Collections.singletonList("to@address.com"), "from@address.com", template, "DS-123", "SomeType");
+        assertTrue(messages.get(0).getSubject().equals("DS-123 that requires data owners reviewing approved."));
     }
 
 }
```
rushtong: nit: newline

consent-ws/src/test/java/org/broadinstitute/consent/http/mail/message/NewCaseMessageTest.java
```diff
@@ -24,10 +26,10 @@ public void setUp(){
...
-        assertTrue(message.getSubject().equals("Log vote on Data Use Limitations case id: DUL-123."));
-        message = new NewCaseMessage().newCaseMessage("to@address.com", "from@address.com", template, "DAR-123", "Data Access");
-        assertTrue(message.getSubject().equals("Log votes on Data Access Request case id: DAR-123."));
+        List<Mail> messages = new NewCaseMessage().newCaseMessage(Collections.singletonList("to@address.com"), "from@address.com", template, "DUL-123", "Data Use Limitations");
+        assertTrue(messages.get(0).getSubject().equals("Log vote on Data Use Limitations case id: DUL-123."));
+        messages = new NewCaseMessage().newCaseMessage(Collections.singletonList("to@address.com"), "from@address.com", template, "DAR-123", "Data Access");
+        assertTrue(messages.get(0).getSubject().equals("Log votes on Data Access Request case id: DAR-123."));
     }
 
 }
```
rushtong: nit: newline

consent-ws/src/test/java/org/broadinstitute/consent/http/mail/message/ReminderMessageTest.java
```diff
@@ -23,12 +25,12 @@ public void setUp(){
...
-        assertTrue(message.getSubject().equals("Urgent: Log votes on Research Purpose Review case id: RP-123."));
+        List<Mail> messages = new ReminderMessage().reminderMessage(Collections.singletonList("to@address.com"), "from@address.com", template, "DUL-123", "Data Use Limitations");
+        assertTrue(messages.get(0).getSubject().equals("Urgent: Log vote on Data Use Limitations case id: DUL-123."));
+        messages = new ReminderMessage().reminderMessage(Collections.singletonList("to@address.com"), "from@address.com", template, "DAR-123", "Data Access Request");
+        assertTrue(messages.get(0).getSubject().equals("Urgent: Log votes on Data Access Request case id: DAR-123."));
+        messages = new ReminderMessage().reminderMessage(Collections.singletonList("to@address.com"), "from@address.com", template, "RP-123", "Research Purpose");
+        assertTrue(messages.get(0).getSubject().equals("Urgent: Log votes on Research Purpose Review case id: RP-123."));
     }
 
 }
```
rushtong: nit: newline

rushtong (changes requested):
Some nits, some suggestions. Let me know if you have questions!

consent-ws/src/main/java/org/broadinstitute/consent/http/service/EmailNotifierService.java
vvicario: I completely agree with you, I changed it

vvicario:
Hi @rushtong i worked on your feedback!

consent-ws/src/main/java/org/broadinstitute/consent/http/service/EmailNotifierService.java
```diff
@@ -141,7 +134,7 @@ public void sendCollectMessage(Integer electionId) throws MessagingException, IO
             Map<String, String> data = retrieveForCollect(electionId);
             String collectUrl = generateCollectVoteUrl(SERVER_URL, data.get("electionType"), data.get("entityId"), data.get("electionId"));
             Writer template = templateHelper.getCollectTemplate(data.get("userName"), data.get("electionType"), data.get("entityName"), collectUrl);
-            List<String> emails = StringUtils.isNotEmpty(data.get("additionalEmail")) ? Arrays.asList(data.get("additionalEmail"), data.get("email")) : Collections.singletonList(data.get("email"));
+            Set<String> emails = StringUtils.isNotEmpty(data.get("additionalEmail")) ? new HashSet<>(Arrays.asList(data.get("additionalEmail"), data.get("email"))) : new HashSet<>(Collections.singletonList(data.get("email")));
```
rushtong, line 147: All of the `new HashSet<>(Collections.singletonList(data.get("email")))` should be replaceable with `Collections.singleton(data.get("email"))`

See https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#singleton-T-

rushtong (approved):
One minor optimization. Other than that, 👍

consent-ws/src/main/java/org/broadinstitute/consent/http/service/EmailNotifierService.java
vvicario: Done!  Thanks for your comments were really helpful.

rushtong (approved):
👍 Thank you!

PR #298, opened 2018-10-23
BTRX-323 - DUOS: Allow Dataset Catalog Upload with either a valid DatasetID OR ConsentID (one or the other is req'd)

vvicario:
Requires FC Dev PR: https://github.com/broadinstitute/firecloud-develop/pull/1408

---
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


vvicario:
It's needed to run PUT: api/dar/dataset after deploy  in order to apply datasets correspondent alias Id in MySql and replace objectId to datasetId n Mongodb

consent-ws/src/main/java/org/broadinstitute/consent/http/models/darsummary/DARModalDetailsDTO.java
```diff
@@ -261,7 +261,10 @@ public void setIsTherePurposeStatements(boolean isTherePurposeStatements) {
 
     public void setDatasetDetail(ArrayList<Document> datasetDetail) {
         Map<String, String> datasetDetailMap = new HashMap<>();
-        datasetDetail.forEach((doc) -> datasetDetailMap.put(doc.getString(DarConstants.DATASET_ID),doc.getString("name")));
+        datasetDetail.forEach((doc) -> {
+            String objectId = doc.getString(DarConstants.OBJECT_ID) != null ? doc.getString(DarConstants.OBJECT_ID) : "";
+            datasetDetailMap.put(objectId ,doc.getString("name"));
```
rushtong: Potential for having an empty string as the key in `datasetDetailMap` which doesn't seem like a great idea here. Maybe you want to filter out the docs that don't have a `DarConstants.OBJECT_ID` property?
Also, spacing nit:
```suggestion
            datasetDetailMap.put(objectId, doc.getString("name"));
```

consent-ws/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -26,22 +26,33 @@
 public interface DataSetDAO extends Transactional<DataSetDAO> {
 
     String CHAIRPERSON = "CHAIRPERSON";
+    String SECOND_DATASET_NAME = "Melanoma_Regev";
+    String FIRST_DATASET_NAME = "Melanoma-Regev-Izar-Garraway-DFCI-ICR";
```
rushtong: This approach doesn't seem right to me. We're hard-coding dataset names into the codebase for some kind of logical reason. These should be in a configuration somewhere so the code can be flexible about the logic its using for these fields.

consent-ws/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -92,34 +112,65 @@
     List<Dictionary> getMappedFieldsOrderByReceiveOrder();
 
     @RegisterMapper({DictionaryMapper.class})
-    @SqlQuery("SELECT * FROM dictionary d order by displayOrder")
+    @SqlQuery("SELECT * FROM dictionary d WHERE d.displayOrder != 10 order by displayOrder")
```
rushtong: I don't understand this. Should we really be ignoring a specific dictionary value all of the time? If that's so, why does it need to be there at all? Shouldn't it just be removed from the database?

consent-ws/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -92,34 +112,65 @@
...
 
-    @SqlQuery("SELECT COUNT(*) FROM consentassociations ca WHERE ca.objectId IN (<objectIdList>)")
-    Integer consentAssociationCount(@BindIn("objectIdList") List<String> objectIdList);
-
     @SqlQuery(" SELECT * FROM dataset d WHERE d.objectId IN (<objectIdList>)")
     List<DataSet> getDataSetsForObjectIdList(@BindIn("objectIdList") List<String> objectIdList);
 
-    @SqlQuery("SELECT * FROM dataset d WHERE d.objectId IN (Select objectId FROM consentassociations Where consentId = :consentId)")
-    List<DataSet> getDataSetsForConsent(@Bind("consentId") String consentId);
+    @SqlQuery(" SELECT * FROM dataset d WHERE d.name IN (<names>)")
```
rushtong: nit spacing. Try to be consistent - there are cases where there are spaces between the quote and `SELECT` and others where there are none. I think the more consistent way is to remove it.

consent-ws/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -92,34 +112,65 @@
...
+    @SqlQuery("select dataSetId, name  from dataset where name in (<nameList>)")
+    List<Map<String,Integer>> searchByNameIdList(@BindIn("nameList") List<String> nameList);
+
+    @SqlQuery(" SELECT * FROM dataset d WHERE d.objectId IN (<objectIdList>) AND d.name is not null")
+    List<DataSet> getDataSetsWithValidNameForObjectIdList(@BindIn("objectIdList") List<String> objectIdList);
+
+    @SqlQuery("select *  from dataset where dataSetId in (<dataSetIds>) ")
+    List<DataSet> searchDataSetsByIds(@BindIn("dataSetIds") List<Integer> dataSetIds);
+
+    @SqlQuery("select MAX(alias) from dataset where name !=  '" + FIRST_DATASET_NAME  + "' AND name !=  '" + SECOND_DATASET_NAME + "'")
```
rushtong: Similar to another comment. I think a better approach is either to remove the datasets from the table, or if they are needed, add a column that specified they should not be returned i.e., `non_public` or something that indicates why they are not being returned in this query. 

Alternatively, the values could be in a config somewhere.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -48,13 +49,10 @@
 import javax.ws.rs.core.UriInfo;
 import java.io.IOException;
 import java.net.URI;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Date;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
rushtong: Revert star import.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -286,13 +285,19 @@ public Response createPartialDataAccessRequest(@Context UriInfo info, Document d
     @Produces("application/json")
     @Path("/partial/datasetCatalog")
     @RolesAllowed("RESEARCHER")
-    public Response createPartialDataAccessRequestFromCatalog(@QueryParam("userId") Integer userId, List<String> datasetIds) {
+    /*
+     * Note: Run this endpoint only once, in order to apply datasets correspondent alias Id
+     * in MySql and replace objectId to datasetId in Mongodb
```
rushtong, line 295: If this really is meant to only be run once, please add a story for removing this code once it has been run and reference the story in this comment.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -371,8 +376,8 @@ public Response cancelDataAccessRequest(@PathParam("referenceId") String referen
     }
 
     @GET
-    @Produces("application/json")
-    @Path("/hasUseRestriction/{referenceId}")
+    @Consumes("application/json")
+    @Produces("application/json")@Path("/hasUseRestriction/{referenceId}")
```
rushtong: nit: split into two lines
```suggestion
    @Produces("application/json")
    @Path("/hasUseRestriction/{referenceId}")
```

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -402,6 +407,19 @@ public Response getUseRestrictionFromQuestions(Document dar) {
...
 
+    @PUT
+    @Produces("application/json")
+    @Path("/dataset")
+    @PermitAll
+    public Response updateObjectIdByDataset() {
+        try{
+            dataSetAPI.updateDataSetAlias();
+            return Response.ok(dataSetAPI.updateDataSetIdToDAR()).build();
+        }catch(Exception e){
```
rushtong: nit spacing
```suggestion
        } catch(Exception e) {
```

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -402,6 +407,19 @@ public Response getUseRestrictionFromQuestions(Document dar) {
...
+            return Response.status(Response.Status.INTERNAL_SERVER_ERROR).entity(new Error(e.getMessage(), Response.Status.INTERNAL_SERVER_ERROR.getStatusCode())).build();
```
rushtong: Probably easier to use the super class's `createExceptionResponse` method instead of manually constructing it here.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -109,7 +108,7 @@ public Response createDataSet(
     public Response describeDataSets(@Context HttpServletRequest request , @QueryParam("dacUserId") Integer dacUserId){
         if (StringUtils.isEmpty(request.getParameter("dacUserId"))) {
             return Response.status(Response.Status.NOT_FOUND).build();
-        }else{
+        } else{
```
rushtong: nit ... halfway there on the spacing:
```suggestion
        } else {
```

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DataAccessParser.java
```diff
@@ -197,6 +202,17 @@ public PDAcroForm fillDARForm(Document dar, Map<String, String> researcherProper
         return acroForm;
     }
 
+    private Map<String, String> parseDatasetDetail(Document dar) {
+        ArrayList<Document> datasetDetail = (ArrayList<Document>) dar.get(DarConstants.DATASET_DETAIL);
+        Map<String, String> datasetDetailMap = new HashMap<>();
+        datasetDetail.forEach((doc) -> {
+            String objectId = doc.getString(DarConstants.OBJECT_ID) != null ? doc.getString(DarConstants.OBJECT_ID) : "";
+            datasetDetailMap.put("dataSetName", doc.getString("name"));
+            datasetDetailMap.put("objectId" ,objectId);
```
rushtong: Similar to comment above ... potential here for putting an empty string as a key in `datasetDetailMap` which seems brittle.
Also spacing:
```suggestion
            datasetDetailMap.put("objectId", objectId);
```

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DataAccessReportsParser.java
```diff
@@ -15,7 +16,7 @@
 
 public class DataAccessReportsParser {
 
-    private static final String DEFAULT_SEPARATOR =  "\t";;
+    private static final String DEFAULT_SEPARATOR =  "\t";
```
rushtong, line 19: Nice!

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DataSetFileParser.java
```diff
@@ -14,13 +15,18 @@
...
+    private final String BLANK_REQUIRED_FIELDS = "Dataset %s - Consent ID or Sample Collection ID is required";
     private final String MISSING_COLUMNS = "Your file has more/less columns than expected. Expected quantity: %s";
     private final String MISSING_MISPLACED_HEADER = "The uploaded file does not comply with the accepted fields. Field: (%s)%s is not recognized/ordered correctly. It should be '%s'";
     private final String PLEASE_DOWNLOAD = "Please download the Dataset Spreadsheet Model from the 'Add Datasets' window.";
-    private int DATASET_ID_INDEX = 9;
+    private int DATASET_NAME_INDEX = 0;
+    private int SAMPLE_COLLECTION_INDEX = 9;
+    private int CONSENT_ID_INDEX = 10;
+    private String SECOND_DATASET_NAME = "Melanoma_Regev";
+    private String FIRST_DATASET_NAME = "Melanoma-Regev-Izar-Garraway-DFCI-ICR";
```
rushtong: See other comments about hard-coding dataset names. I think this could be less brittle with either a configuration or a new field on the dataset table.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseConsentAPI.java
```diff
@@ -230,30 +230,44 @@ public void logicalDelete(String id) throws UnknownIdentifierException {
...
 
+    private void processAssociation(List<String> objectIds) {
+        if (CollectionUtils.isNotEmpty(objectIds)) {
+            objectIds.stream().forEach(objectId -> {
+                Integer dsId = dataSetDAO.findDataSetIdByObjectId(objectId);
+                if (dsId == null) {
+                    dataSetDAO.insertAll(Arrays.asList(new DataSet(objectId)));
+                }
+
+            });
```
rushtong: two dao calls in a loop make me nervous. Could these be re-written as a query for all datasets not in <objectIds> and then bulk adding?

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseMatchingServiceAPI.java
```diff
@@ -20,10 +20,7 @@
 import javax.ws.rs.core.GenericType;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
rushtong: revert star import

consent-ws/src/main/resources/changesets/changelog-consent-43.0.xml
```diff
@@ -0,0 +1,123 @@
...
+
+    <changeSet author="vvicario" id="131">
+        <sql dbms="hsql, mysql">
+            UPDATE dataset SET alias = 2 where name = 'Melanoma_Regev';
+            UPDATE dataset SET alias = 1 where name = 'Melanoma-Regev-Izar-Garraway-DFCI-ICR';
+        </sql>
+    </changeSet>
+
+
+    <!--<changeSet author="vvicario" id="132">-->
```
rushtong: Remove commented out code?

consent-ws/src/test/java/org/broadinstitute/consent/http/AssociationTest.java
```diff
@@ -13,6 +13,7 @@
 import org.broadinstitute.consent.http.models.grammar.Everything;
 import org.junit.Before;
 import org.junit.ClassRule;
+import org.junit.Ignore;
```
rushtong: unused?

consent-ws/src/test/java/org/broadinstitute/consent/http/DatasetAssociationTest.java
```diff
@@ -3,6 +3,7 @@
 import io.dropwizard.testing.junit.DropwizardAppRule;
 import org.broadinstitute.consent.http.configurations.ConsentConfiguration;
 import org.junit.ClassRule;
+import org.junit.Ignore;
```
rushtong: unused?

consent-ws/src/test/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceTest.java
```diff
@@ -56,64 +56,66 @@ public void teardown() {
...
+//        assertTrue(manageDars.size() == 1);
+//    }
+//
+//    @Test
+//    public void testRestrictionFromQuestions() throws IOException {
+//        Client client = ClientBuilder.newClient();
+//        Response response = checkStatus(OK, put(client, restrictionFromQuestionsUrl(), DataRequestSamplesHolder.getSampleDar()));
+//        String res = response.readEntity(String.class);
+//        assertTrue(res.equals("{\"useRestriction\":\"Manual Review\"}"));
+//    }
```
rushtong: Please fix tests, or rewrite as necessary.

rushtong (commented):
Oh my goodness, this is huge. I did not test this locally. Please see comments inline and I think this could use a review from someone on the C&D team, maybe @ahaessly

consent-ws/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
vvicario: we use it only when we upload datasets in order to check the specific key and receive order in the  Dataset Spreadsheet, this field is new. The user has the possibility to upload with either a valid DatasetID OR ConsentID. But we don't need to return this on dataset catalog. I will put display order in null for this specific case.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
vvicario: I created https://broadinstitute.atlassian.net/browse/BTRX-404

LeoForconesi:
@rushtong  
We have made corrections on the code. Could you please take a look again?

consent-ws/src/main/java/org/broadinstitute/consent/http/configurations/DatasetConfiguration.java
```diff
@@ -0,0 +1,28 @@
...
+        this.duos1 = duos1;
+    }
+
+    public String getDuos2() {
+        return duos2;
+    }
+
+    public void setDuos2(String duos2) {
+        this.duos2 = duos2;
+    }
```
ahaessly: why do we have 2 duos strings? maybe add a comment to the class to help explain

vvicario:
@rushtong && @ahaessly  we added new properties in config file, we will need some help to add this on firecloud-develop and vault!
Is there a possibility to merge today? So,  Moran can start testing

rushtong:
@vvicario - I am swamped today - I can find some time to look at it tomorrow or this weekend :-(

consent-ws/src/main/java/org/broadinstitute/consent/http/db/ConsentDAO.java
```diff
@@ -27,23 +27,26 @@
 
     @SqlQuery("SELECT c.* " +
             "FROM consents c INNER JOIN consentassociations cs ON c.consentId = cs.consentId "+
-            "WHERE cs.objectId = :datasetId")
-    Consent findConsentFromDatasetID(@Bind("datasetId") String datasetId);
+            "WHERE cs.dataSetId = :datasetId")
+    Consent findConsentFromDatasetID(@Bind("datasetId") Integer datasetId);
```
ahaessly, line 31: why is the datasetId now an integer? is this now representing the DUOS_000001 id?

consent-ws/src/main/java/org/broadinstitute/consent/http/db/ConsentDAO.java
```diff
@@ -27,23 +27,26 @@
...
 
     @SqlQuery("select * from consents  where consentId in (<consentIds>)")
     Collection<Consent> findConsentsFromConsentsIDs(@BindIn("consentIds") List<String> consentIds);
 
+    @SqlQuery("select * from consents  where name in (<names>)")
+    List<Consent> findConsentsFromConsentNames(@BindIn("names") List<String> names);
+
     @Mapper(ConsentDataSetMapper.class)
-    @SqlQuery("SELECT c.consentId, cs.objectId, ds.name " +
+    @SqlQuery("SELECT c.consentId, cs.dataSetId, ds.name, ds.objectId " +
```
ahaessly, line 45: also, is the objectId still the sample collection (SC) number?

vvicario: objectId is the sample collection id but we save it in dataset table, consent association has a relation with dataset

consent-ws/src/main/java/org/broadinstitute/consent/http/db/ConsentDAO.java
rushtong: I think I missed this. If Andrea is right about this question, then by truncating the integer component, you are losing valuable information. These term IDs can all be valid and distinctly different, DOID_001, DOID_0001, DOID_00001, but would all be coerced to the same digit, `1`

vvicario: No, we associated consent association with dataset. datasetId is the id of dataset where we save the rest of information related to dataset like object id, name, etc.

vvicario: we created a new field on dataset called alias where we save a number, dataset UID  is generated by the system, Moran asked to set some specific values for the first and second datasets with specific name, and the rest should be incremental

vvicario: Greg this cases DOID_001, DOID_0001, DOID_00001,  wont happen because the story says that we have specific pattern for this.

"We want to assign the DatasetID 'DUOS-000001' to a dataset which has not yet been uploaded to DUOS, but will be uploaded shortly.
We want to assign the DatasetID 'DUOS-000002' to the Melanoma_Regev dataset already in the Dataset Catalog in DUOS.
Please do not assign either of these UIDs to other datasets.
The remaining datasets already in the Dataset Catalog can be assigned the following DatasetID UIDs 'DUOS-000003', 'DUOS-000004' etc.

consent-ws/src/main/java/org/broadinstitute/consent/http/mail/freemarker/DatasetOwnerListModel.java
```diff
@@ -9,7 +11,7 @@
     public DatasetOwnerListModel(String ownerName, String datasetName, String datasetId) {
         this.ownerName = ownerName;
         this.datasetName = datasetName;
-        this.datasetId = datasetId;
+        this.datasetId = StringUtils.isEmpty(datasetId) ? "--" : datasetId;
```
ahaessly: if the dataset is not empty are we expecting this to be an integer? is this the same value as the dataSetId used in the association?

consent-ws/src/main/java/org/broadinstitute/consent/http/models/ConsentDataSet.java
```diff
@@ -9,12 +9,15 @@
...
+
     @JsonProperty
     public Map<String, String> dataSets;
-;
 
-    public ConsentDataSet(String consentId,  Map<String, String> dataSets) {
+    public ConsentDataSet(String consentId,  Map<String, String> dataSets, String objectId) {
         this.consentId = consentId;
+        this.objectId = objectId;
         this.dataSets = dataSets;
```
ahaessly, line 25: what are the keys and values of this map? maybe a comment in the code.

consent-ws/src/main/java/org/broadinstitute/consent/http/models/DataSet.java
```diff
@@ -22,14 +22,29 @@
...
 
+    @JsonProperty
+    private String consentName;
+
     @JsonProperty
     private Boolean needsApproval;
 
+    @JsonProperty
+    private Integer alias;
+
```
ahaessly, line 33: what is the alias typically used for?

vvicario: we use alias to create a user-friendly dataset UID for each dataset

consent-ws/src/main/java/org/broadinstitute/consent/http/mail/freemarker/DatasetOwnerListModel.java
LeoForconesi: Actually this is a wrong variable name, it should be renamed objectId, which is a string type. This value can be empty, so we replace it with "-- " on the e-mail template. 
I'll make this change.

wloforteBroad:
@ahaessly , @rushtong  Here's the link to the Jira ticket in case you need more information: https://broadinstitute.atlassian.net/browse/BTRX-323

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ElectionReviewResource.java
```diff
@@ -64,7 +64,7 @@ public ElectionReview getElectionReviewByElectionId(@PathParam("electionId") Int
     public ElectionReview getAccessElectionReviewByReferenceId(@PathParam("electionId") Integer electionId, @QueryParam("isFinalAccess") Boolean isFinalAccess) {
         Election election = electionAPI.describeElectionById(electionId);
         Election consentElection = electionAPI.getConsentElectionByDARElectionId(election.getElectionId());
-        List<String> dataSetId = accessRequestAPI.describeDataAccessRequestFieldsById(election.getReferenceId(), Arrays.asList(DarConstants.DATASET_ID)).get(DarConstants.DATASET_ID, List.class);
+        List<Integer> dataSetId = accessRequestAPI.describeDataAccessRequestFieldsById(election.getReferenceId(), Arrays.asList(DarConstants.DATASET_ID)).get("datasetId", List.class);
```
ahaessly, line 67: i think you can still use the DATASET_ID constant here

vvicario: yes, we are saving the dataset id in Data Access Request Application,  not the object id any more because it's optional now.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -247,7 +248,7 @@ public Response updateNeedsReviewDataSets(@QueryParam("dataSetId") String dataSe
     @Produces(MediaType.APPLICATION_OCTET_STREAM)
     @PermitAll
     @Path("/{datasetId}/approved/users")
-    public Response downloadDatasetApprovedUsers(@PathParam("datasetId") String dataSetId) {
+    public Response downloadDatasetApprovedUsers(@PathParam("datasetId") Integer dataSetId) {
```
ahaessly: there are 2 different cases in the parameters here and above. above it is dataSetId, here it is datasetId. can we standardize on one?

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DataAccessParser.java
```diff
@@ -197,6 +199,20 @@ public PDAcroForm fillDARForm(Document dar, Map<String, String> researcherProper
...
+        Map<String, String> datasetDetailMap = new HashMap<>();
+
+        datasetDetail.forEach((doc) -> {
+            String objectId = doc.getString(DarConstants.OBJECT_ID) != null ? " | ".concat(doc.getString(DarConstants.OBJECT_ID)) : "";
+            datasetDetailMap.put(doc.getString("name"), objectId);
+        });
+
+        return datasetDetailMap.entrySet().stream()
+                .map(name -> name.getKey() + name.getValue())
+                .collect(Collectors.joining("; "));
```
ahaessly: previously joined ids with ", " now joining with "; "
was this done on purpose?

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
vvicario: Sure!

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DataAccessReportsParser.java
```diff
@@ -95,14 +96,22 @@ private String formatTimeToDate(long time) {
     }
 
     private void addDARLine(FileWriter darWriter, Document dar, String customContent1, String customContent2, String consentName, String translatedUseRestriction) throws IOException {
-        List<Document> dataSetDetail = dar.get(DarConstants.DATASET_DETAIL, ArrayList.class);
-        String dataSetName = CollectionUtils.isNotEmpty(dataSetDetail) ? dataSetDetail.get(0).getString("name") : " ";
+        List<Document> dataSetDetails = dar.get(DarConstants.DATASET_DETAIL, ArrayList.class);
+        List<String> datasetNames = new ArrayList<>();
+        for(Document detail : dataSetDetails) {
+            datasetNames.add(StringUtils.isNotEmpty(detail.getString("name")) ? detail.getString("name") : " ");
```
ahaessly: you can use:
datasetNames.add(StringUtils.defaultString(detail.getString("name")));

vvicario: Nice, we will change it

LeoForconesi:
@ahaessly 
We just updated the code with your reviews. could you please check them?

consent-ws/src/main/java/org/broadinstitute/consent/http/models/ConsentDataSet.java
LeoForconesi: I hope the comment we've just added clarifies this.

consent-ws/src/main/java/org/broadinstitute/consent/http/configurations/DatasetConfiguration.java
rushtong: Instead of two separate values, this should be a list instead. As an example, look at `ElasticSearchConfiguration.servers`. That way, you can define the config once and add more datasets via config instead of adding more code.

consent-ws/src/main/java/org/broadinstitute/consent/http/db/ConsentDAO.java
rushtong: 👍

consent-ws/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -33,15 +33,24 @@
...
-    @SqlQuery("select * from dataset where objectId in (<objectIdList>) and needs_approval = true")
-    List<DataSet> findNeedsApprovalDataSetByObjectId(@BindIn("objectIdList") List<String> objectIdList);
+    @SqlQuery("select objectId from dataset where dataSetId = :dataSetId")
+    String findObjectIdByDataSetId(@Bind("dataSetId") Integer dataSetId);
 
-    @SqlBatch("insert into dataset (name, createDate, objectId, active) values (:name, :createDate, :objectId, :active)")
+    @SqlQuery("select * from dataset where objectId = :objectId")
+    Integer findDataSetIdByObjectId(@Bind("objectId") String objectId);
+
+    @SqlQuery("select * from dataset where dataSetId in (<dataSetIdList>) and needs_approval = true")
```
rushtong, line 42: Definitely not for this PR, but we should add a tech debt story to rename a lot of our db fields consistently. `dataSetId` should be `dataset_id`.

consent-ws/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -92,34 +110,73 @@
...
+
+    @SqlQuery("select *  from dataset where dataSetId in (<dataSetIds>) ")
+    List<DataSet> searchDataSetsByIds(@BindIn("dataSetIds") List<Integer> dataSetIds);
+
+    @SqlQuery("select MAX(alias) from dataset")
+    Integer findLastAlias();
+
+    @SqlQuery("select *  from dataset where alias is null or alias = 0 ")
+    List<DataSet> getDataSetsWithoutAlias();
 }
```
rushtong: nit: add newline

consent-ws/src/main/java/org/broadinstitute/consent/http/mail/freemarker/DatasetOwnerListModel.java
```diff
@@ -20,8 +22,8 @@ public String getDatasetName() {
         return datasetName;
     }
 
-    public String getDatasetId() {
-        return datasetId;
+    public String getObjectId() {
+        return objectId;
     }
 
 }
```
rushtong: nit: add newline

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -219,10 +222,11 @@ public Document describeSpecificFields(@PathParam("id") String id, @QueryParam("
     @Produces("application/json")
     @PermitAll
     public Consent describeConsentForDAR(@PathParam("id") String id) {
-        List<String> datasetId = (dataAccessRequestAPI.describeDataAccessRequestFieldsById(id, Arrays.asList(DarConstants.DATASET_ID))).get("datasetId", List.class);
+        List<Integer> dataSetIdList = (dataAccessRequestAPI.describeDataAccessRequestFieldsById(id, Arrays.asList(DarConstants.DATASET_ID))).get("datasetId", List.class);
```
rushtong, line 225: Note that in different objects, we have `"dataSetId"` vs `"datasetId"`. That refactor might be scope creep for this PR. I leave it to you as to whether we should do some renaming in this PR or address in a tech-debt story.

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -74,9 +73,9 @@ public Response createDataSet(
                 inputFile = new File(UUID.randomUUID().toString());
                 FileUtils.copyInputStreamToFile(uploadedDataSet, inputFile);
                 ParseResult result;
-                if(overwrite) {
+                if (overwrite) {
                     result = api.overwrite(inputFile, userId);
-                }else{
+                } else{
```
rushtong, line 78: nit: spacing
```suggestion
                } else {
```

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -119,7 +118,7 @@ public Response describeDataSets(@Context HttpServletRequest request , @QueryPar
     @Path("/{datasetId}")
     @Produces("application/json")
     @PermitAll
-    public Response describeDataSet( @PathParam("datasetId") String datasetId){
+    public Response describeDataSet( @PathParam("datasetId") Integer datasetId){
```
rushtong, line 121: Another case of `"datasetId"` vs `"dataSetId"`

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseConsentAPI.java
```diff
@@ -230,30 +230,52 @@ public void logicalDelete(String id) throws UnknownIdentifierException {
...
-            insertBatch.add(consentId, associationType, id);
+        PreparedBatch insertBatch = h.prepareBatch("insert into consentassociations (consentId, associationType, dataSetId) values (?, ?, ?)");
+        for (String id : ids) {
+            insertBatch.add(consentId, associationType, dataSetDAO.findDataSetIdByObjectId(id));
         }
         List<Long> insertedIds = insertBatch.executeAndGenerateKeys(LongColumnMapper.PRIMITIVE).list();
         h.close();
         List<String> stringsList = new ArrayList<>();
-        for (Long id: insertedIds) stringsList.add(id.toString());
+        for (Long id : insertedIds) stringsList.add(id.toString());
```
rushtong, line 245: Could be streamed a little more efficiently:
```suggestion
        List<String> stringsList = insertedIds.
            stream().
            map(String::valueOf).
            collect(Collectors.toList());
```

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseConsentAPI.java
```diff
@@ -230,30 +230,52 @@ public void logicalDelete(String id) throws UnknownIdentifierException {
...
         }
     }
 
+    private void processAssociation(List<String> objectIds) {
+        if (CollectionUtils.isNotEmpty(objectIds)) {
+            List<DataSet> dataSets = dataSetDAO.getDataSetsForObjectIdList(objectIds);
+            List<String> existentObjectsId = dataSets.stream().map(DataSet::getObjectId).collect(Collectors.toList());
+            List<DataSet> dataSetsToCreate = new ArrayList<>();
+            if(CollectionUtils.isNotEmpty(dataSets)) {
+                objectIds.stream().forEach(objectId -> {
```
rushtong, line 265: don't need the `stream()` here (and in else block), `forEach` is on `List`

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseDataSetAPI.java
```diff
@@ -29,51 +33,84 @@
...
         if (CollectionUtils.isNotEmpty(dataSets)) {
-            if (completeDBCheck(dataSets)) {
-                dsDAO.insertAll(dataSets);
-                List<DataSetProperty> dataSetProperties = insertProperties(dataSets);
-                insertDataSetAudit(dataSets, CREATE, userId, dataSetProperties);
+            if (isValid(dataSets, false)) {
+
+                List<DataSet> existentdataSets = dsDAO.getDataSetsForObjectIdList(dataSets.stream().map(DataSet::getObjectId).collect(Collectors.toList()));
+
+                List<DataSet> dataSetsToUpdate = new ArrayList<>();
```
rushtong: nit: unnecessary newlines

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseDataSetAPI.java
```diff
@@ -82,26 +119,35 @@ public ParseResult create(File dataSetFile, Integer userId) {
...
-
-            List<DataSet> existentDataSets = dsDAO.searchDataSetsByObjectIdList(objectIdList);
-            Map<String,DataSet> dataSetMap = new HashMap<>();
+            List<DataSet> existentDataSets = dsDAO.searchDataSetsByNameList(nameList);
+            if (CollectionUtils.isNotEmpty(objectIdList)) {
+                existentDataSets.addAll(dsDAO.searchDataSetsByObjectIdList(objectIdList));
+            }
+            Map<String, DataSet> dataSetMap = new HashMap<>();
 
             existentDataSets.stream().forEach(dataSet -> {
```
rushtong, line 132: nit: don't need `stream()` here.

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseDataSetAPI.java
```diff
@@ -124,53 +170,51 @@ public ParseResult overwrite(File dataSetFile, Integer userId) {
...
                 List<Document> accessRequests = accessAPI.describeDataAccessRequests();
-                List<String> dataSetObjectIdList = new ArrayList<>();
-                dataSetDTOList.stream().forEach(dataSet -> {
-                    Map<String,String> dataSetProperties = dataSet.getProperties().stream().collect(Collectors.toMap(DataSetPropertyDTO::getPropertyName,DataSetPropertyDTO::getPropertyValue));
-                    dataSetObjectIdList.add(dataSetProperties.get(DATA_SET_ID));
+                List<Integer> dataSetIdList = new ArrayList<>();
+
+                dataSetDTOList.stream().forEach(dataSet ->{
+                    dataSetIdList.add(dataSet.getDataSetId());
                 });
```
rushtong: `List<Integer> dataSetIdList` could be a one-liner
```suggestion
                List<Integer> dataSetIdList = dataSetDTOList.
                    stream().
                    map(DataSetDTO::getDataSetId).
                    collect(Collectors.toList());
```

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseDataSetAPI.java
```diff
@@ -195,127 +239,212 @@ public ParseResult overwrite(File dataSetFile, Integer userId) {
...
         try {
             dsDAO.begin();
             dataSetAuditDAO.begin();
-            DataSet dataset = dsDAO.findDataSetByObjectId(datasetObjectId);
-            Collection<Integer> dataSetId = new ArrayList<>();
-            dataSetId.add(dataset.getDataSetId());
-            if(checkDatasetExistence(dataset.getDataSetId())){
+            DataSet dataset = dsDAO.findDataSetById(dataSetId);
+            Collection<Integer> dataSetsId = new ArrayList<>();
+            dataSetsId.add(dataset.getDataSetId());
```
rushtong: Could be combined with line above for a one-liner
```suggestion
            Collection<Integer> dataSetsId = Collections.singletonList(dataset.getDataSetId());
```

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseDataSetAPI.java
```diff
@@ -370,68 +499,73 @@ private void setConsentNameDTOList(Collection<DataSetDTO> dataSetDTOList) {
...
+            List<DataSet> existentDataSets = dsDAO.getDataSetsForNameList(new ArrayList(dataSetObjectIdMap.keySet()));
+            createDataSetAudit(existentDataSets, userId, action, properties);
+        }
+    }
+
+
+    private void createDataSetAudit(List<DataSet> dataSetList, Integer userId, String action, List<DataSetProperty> properties) {
+        if (CollectionUtils.isNotEmpty(dataSetList)) {
+            Map<Integer, List<DataSetProperty>> dataSetPropertyMap = new HashMap<>();
+            properties.stream().forEach(property -> {
```
rushtong, line 552: Should probably to a search-replace for `stream().forEach` -> `.forEach`

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseDataSetAPI.java
```diff
@@ -445,5 +579,4 @@ private void createDataSetAudit(List<DataSet> dataSetList, Integer userId, Strin
         });
         return auditProperties;
     }
-
 }
```
rushtong: nit: newline

consent-ws/src/test/java/org/broadinstitute/consent/http/service/DataSetFileParserTest.java
```diff
@@ -0,0 +1,58 @@
+package org.broadinstitute.consent.http.service;
+
+import org.broadinstitute.consent.http.models.DataSet;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class DataSetFileParserTest {
```
rushtong, line 11: Thank you for this!

rushtong (changes requested):
Lots of nits, some suggested optimizations, and a recommended change to the dataset configuration. 
In general, there are a lot of `dataset` and `datasetId` casing problems. Since this PR is in a rush, I recommend adding a story to the backlog for some renaming to avoid confusion:
* `DataSet` -> `Dataset`
* `dataSet` -> `dataset`
* `dataSetId` -> `datasetId`

consent-ws/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
LeoForconesi: I agree with you on this changes, and the way to handle them. 
I created a new story for this technical debt, this is the link ; https://broadinstitute.atlassian.net/browse/BTRX-405

vvicario:
@rushtong We worked on your feedback, please let us know if you have any other comments

rushtong (approved):
👍

consent-ws/src/main/java/org/broadinstitute/consent/http/models/DataSet.java
ahaessly: so which has the value represented by DUOS_00000x
is the alias=x?
or is the datasetId = x?
or both?

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ElectionReviewResource.java
ahaessly: i was just saying that DarConstants.DATASET_ID was changed to "datasetId" in the get call. I think you should still be able to use the constant instead of the string

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DataSetFileParser.java
```diff
@@ -68,25 +73,54 @@ public ParseResult parseTSVFile(File file, List<Dictionary> allFields) {
...
+                ds.setAlias(1);
+            }
+            else if(lastAlias == null || lastAlias < 3) {
+                ds.setAlias(initialAlias);
+                ++initialAlias;
+            }
+            else if(ds.getAlias() == null || ds.getAlias() == 0){
+                ds.setAlias(++lastAlias);
+            }
+        }
```
ahaessly, line 122: is this code thread safe?

rushtong: My first guess without evidence is that it isn't - there is a single final instance on the service class which is reused for all API calls. That's an existing issue and should be storied out as tech-debt.

vvicario: it isn't, I agree we have to handle it, we can create an story for that

LeoForconesi: Here's the story to handle this bug.
https://broadinstitute.atlassian.net/browse/BTRX-407

consent-ws/src/main/java/org/broadinstitute/consent/http/models/DataSet.java
vvicario: it's alias

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/ElectionReviewResource.java
LeoForconesi: :+1:

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseDataAccessRequestAPI.java
```diff
@@ -116,15 +116,15 @@ public void setMongoDBInstance(MongoConsentDB mongo) {
             dataAccessRequest.remove(DarConstants.ID);
             dataAccessRequest.remove(DarConstants.PARTIAL_DAR_CODE);
         }
-        List<String> dataSets =  dataAccessRequest.get(DATA_SET_ID, List.class);
-        dataAccessRequest.remove(DATA_SET_ID);
+        List<Integer> dataSets =  dataAccessRequest.get(DATA_SET_ID, List.class);
         if (CollectionUtils.isNotEmpty(dataSets)) {
-            Set<ConsentDataSet> consentDataSets = consentDAO.getConsentIdAndDataSets(dataSets);
+            Set<ConsentDataSet> consentDataSets = consentDAO.getConsentIdAndDataSets(dataAccessRequest.get(DATA_SET_ID, List.class));
```
ahaessly: why are you making the get request again instead of just using the dataSets variable like before?

consent-ws/src/main/java/org/broadinstitute/consent/http/service/DatabaseDataSetAPI.java
```diff
@@ -29,51 +31,84 @@
...
     private final String OVERWRITE_ON = "If you wish to overwrite DataSet values, you can turn OVERWRITE mode ON.";
-    private final String DATASETID_PROPERTY_NAME = "Dataset ID";
+    private final String DATASETID_PROPERTY_NAME = "Sample Collection ID";
+    private final String CONFLICT_IDS = "Conflict in dataset association identificator,  %s - %s, use either Sample Collection ID or Consent ID";
     private final String CREATE = "CREATE";
     private final String UPDATE = "UPDATE";
     private final String DELETE = "DELETE";
+    private final  List<String> predefinedDatasets;
 
+    private final Integer CONSENT_ID = 11;
```
ahaessly: why is this hardcoded to 11?

ahaessly (changes requested):
most of it looks good. but i think the code that increments the alias integer is not thread safe.

LeoForconesi:
@ahaessly we have worked on your review. We will be pending if you have any other comment.

ahaessly (approved):
👍

consent-ws/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
rushtong: Following up on the change from String to Integer ... legacy DARs that used the full SC-XXX number as the dataset ID causes this to throw a class cast exception. Adding one more reason I do not like how we're decomposing mongo objects ... lots of room for bugs here.

rushtong: See https://broadinstitute.atlassian.net/browse/DUOS-390 for fixing this.

PR #301, opened 2018-11-09
[BTRX-109][risk=no] Move consent out of module

rushtong:
## Addresses 
https://broadinstitute.atlassian.net/browse/BTRX-109

Requires https://github.com/broadinstitute/firecloud-develop/pull/1439
Requires https://github.com/broadinstitute/dsp-jenkins/pull/160

Moves consent from a module to the top level of the project.
Includes fixes to the build/test process.

## Testing
Have tested this locally against dev (using [configs generated from this PR](https://github.com/broadinstitute/firecloud-develop/pull/1440)) and it is working as expected with no changes to functionality.

Cannot test this on a FiaB due to the change in https://github.com/broadinstitute/dsp-jenkins/pull/160. That change, should it be applied, would break all current FiaBs as well as dev deployments since they use consent. These three PRs will need to be merged in succession.


vvicario (approved):
Nice job! I tested it locally and it works fine!

dags (approved):

PR #305, opened 2018-11-14
[BTRX-421][risk=no] Get Swagger lib from maven

rushtong:
## Addresses 
Consent side of https://broadinstitute.atlassian.net/browse/BTRX-421
Does not complete this story though. Will require a corresponding PR in ontology.
See also: https://github.com/DataBiosphere/consent-ontology/pull/108

## Changes
* Add some mvn magic to convert pom properties to a properties file
* Keep configurations wrt swagger in pom properties
* Read the properties file for the right swagger library in swagger resource.

### DIFF
--- a/src/test/resources/mvn.properties
+++ b/src/test/resources/mvn.properties
@@ -1,3 +1,2 @@
 #Properties
-swagger.ui.version=2.2.8
 swagger.ui.path=META-INF/resources/webjars/swagger-ui/2.2.8/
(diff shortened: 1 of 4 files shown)


rushtong:
@vvicario - What do you think of this approach? I like this better than using a config because it keeps the jar version and path close together in the code and should be easier to update.

src/test/resources/mvn.properties
```diff
@@ -0,0 +1,3 @@
+#Properties
+swagger.ui.version=2.2.8
```
vvicario: is it used?

vvicario (approved):
I like this approach!

src/test/resources/mvn.properties
rushtong: Technically, no - will remove.

rushtong: Done

rushtong:
Thank you! I'll do the same for ontology soon.

PR #326, opened 2018-12-12
[BTRX-468][risk=no] Remove unnecessary translate calls

rushtong:
First step in addressing https://broadinstitute.atlassian.net/browse/BTRX-468

Currently, we make unnecessary translate calls on the structured use restriction. We don't need to do that - we can rely on the translation that comes from the client. This PR removes that and also cleans up the service naming.


src/main/java/org/broadinstitute/consent/http/resources/ConsentResource.java
```diff
@@ -70,9 +87,6 @@ public Response describeInvalidConsents() {
     public Response createConsent(@Context UriInfo info, Consent rec, @Auth User user) {
         try {
             DACUser dacUser = dacUserAPI.describeDACUserByEmail(user.getName());
-            if (rec.getTranslatedUseRestriction() == null) {
-                rec.setTranslatedUseRestriction(translateServiceAPI.translate(TranslateType.SAMPLESET.getValue(),rec.getUseRestriction()));
-            }
```
rushtong, line 75: This call, and the `update` below, are the real meat of this PR. We don't need to call out to ontology for this.

dags (approved):

PR #333, opened 2019-03-29
[DUOS-218] Serve swagger from root

rushtong:
See https://broadinstitute.atlassian.net/browse/DUOS-218
See also https://github.com/DataBiosphere/consent-ontology/pull/132

Currently, the root url is a 404. Instead, serve the swagger page.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.broadinstitute</groupId>
   <artifactId>consent</artifactId>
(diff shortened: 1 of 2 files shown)


tadeork (approved):
It's OK.
The idea is that when received a GET to ``/`` the client should get swagger page. Currently to see swagger page the client has to GET ``/swagger``?

rushtong:
Thanks @tadeork - yes, correct. Right now, if you to to consent's root url, you get a 404 which is a pretty poor experience.

PR #334, opened 2019-03-29
[DUOS-228] Handle invalid dar ids better

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-228

## Changes
* Handle the lookup of the dar by document id better. If we can't find it, or if it is not an id we can look up, return a 404. Otherwise, return the exception through the normal exception response mechanism. 
* Added tests.
* Lots of minor formatting cleanup for the files touched.
* While I'm at it ... addressing some long-standing test logging problems in mongo and mockserver.

## Manual Testing
* Spin up and go to the dar lookup page which is `Data_Access_Request/get_api_dar_id`
* Try entering gibberish into the id field and ensure that you get a 404. 
* Enter a real id (look one up from duos dev) and ensure that you still get a valid document.

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -186,12 +186,11 @@ public class DataAccessRequestResource extends Resource {
         try {
             new ObjectId(id);
         } catch (IllegalArgumentException e) {
-            logger.log(Level.INFO, "Returning DAR not found response to client from exception: " + e.getMessage());
+            String message = "The provided id is not a valid Data Access Request Id: " + id;
+            logger.log(Level.INFO, message + "; " + e.getMessage());
             return Response
                     .status(Response.Status.BAD_REQUEST)
-                    .entity(new Error(
-                            "The provided id is not a valid Data Access Request Id: " + id,
-                            Response.Status.BAD_REQUEST.getStatusCode()))
+                    .entity(new Error(message, Response.Status.BAD_REQUEST.getStatusCode()))
                     .build();
         }
         try {
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -184,15 +184,22 @@ public class DataAccessRequestResource extends Resource {
     public Response describe(@PathParam("id") String id) {
         try {
             Document document = dataAccessRequestAPI.describeDataAccessRequestById(id);
-            return Response.status(Response.Status.OK).entity(document).build();
+            if (document != null) {
+                return Response.status(Response.Status.OK).entity(document).build();
+            }
+            return Response
+                    .status(Response.Status.NOT_FOUND)
+                    .entity("Unable to find Data Access Request with id: " + id)
+                    .build();
         } catch (IllegalArgumentException e) {
+            // This exception is thrown when mongo can't coerce the string value to an ObjectId.
             logger.log(Level.INFO, "Returning DAR not found response to client from exception: " + e.getMessage());
             return Response
                     .status(Response.Status.NOT_FOUND)
                     .entity("Unable to find Data Access Request with id: " + id)
                     .build();
         } catch (Exception e) {
-            return createExceptionResponse(new NotFoundException("Unable to find Data Access Request with id: " + id));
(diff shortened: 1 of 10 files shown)


src/test/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceTest.java
```diff
@@ -117,9 +125,50 @@ public void testRestrictionFromQuestions() throws IOException {
...
+        // Create a DAR
+        Client client = ClientBuilder.newClient();
+        Document sampleDar = DataRequestSamplesHolder.getSampleDar();
+        checkStatus(CREATED, post(client, darPath(), sampleDar));
+        List<Document> created = retrieveDars(client, darPath());
+        System.out.println("Created Size: " + created.size());
+        assertEquals(1, created.size());
+        Document newDoc = created.get(0);
+
+        // TODO: Figure out how to
```
rushtong: @dags - Here (and below) is where I'm struggling. I can't figure out how to get the hex string value of the saved Data Access Request id. I know it's there, it looks something like 
```
{timestamp=1553890382, machineIdentifier=6733188, processIdentifier=17821, counter=13091050, timeSecond=1553890382, time=1553890382000, date=1553890382000}
```
But ... how do I turn that into the ObjectId.toString value which is what we would use in the API call to get a DAR by id? I can't find anything in the docs or in stack overflow that address this. 

Thank you!

rushtong: Ok, I think I've figured out something super painful. This gives me a real hex string from manually constructing the ObjectId from the json version of the document's `_id`. This is super ugly, painful, and should not be necessary.
```
        Document newDoc = created.get(0);
        LinkedHashMap id = (LinkedHashMap) newDoc.get(DarConstants.ID);
        ObjectId objectId = new ObjectId(
                Integer.valueOf(id.get("timestamp").toString()),
                Integer.valueOf(id.get("machineIdentifier").toString()),
                Short.valueOf(id.get("processIdentifier").toString()),
                Integer.valueOf(id.get("counter").toString())
        );
```

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -182,11 +182,33 @@ public Response getInvalidDataAccessRequest() {
...
+                    .status(Response.Status.BAD_REQUEST)
+                    .entity(new Error(message, Response.Status.BAD_REQUEST.getStatusCode()))
+                    .build();
+        }
+        try {
+            Document document = dataAccessRequestAPI.describeDataAccessRequestById(id);
+            if (document != null) {
+                return Response.status(Response.Status.OK).entity(document).build();
+            }
+            return Response
```
vvicario, line 201:
```suggestion
          return createExceptionResponse(new NotFoundException("Unable to find Data Access Request with id: " + id));
```
Maybe it could be simplified like this

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -182,11 +182,33 @@ public Response getInvalidDataAccessRequest() {
...
     @PermitAll
-    public Document describe(@PathParam("id") String id) {
-        return dataAccessRequestAPI.describeDataAccessRequestById(id);
+    public Response describe(@PathParam("id") String id) {
+        try {
+            new ObjectId(id);
+        } catch (IllegalArgumentException e) {
+            String message = "The provided id is not a valid Data Access Request Id: " + id;
+            logger.log(Level.INFO, message + "; " + e.getMessage());
+            return Response
```
vvicario, line 191:
```suggestion
            return createExceptionResponse(new IllegalArgumentException(message));
```
Maybe it could be simplified like this

vvicario (approved):
:+1:

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
rushtong: I considered that - but I didn't what the logging that comes along with `createExceptionResponse` In these cases, a 400 or 404 is not really something we need to be logging at the warning level and sending to Sentry - it's somewhat expected that these kinds of things will happen and don't need any followup action. 

I think it might make sense to rename `createExceptionResponse` to something more in line with what it does ... log the error and try to decode the unexpected exception.

PR #335, opened 2019-04-24
[DUOS-283] HMB Manual Review Bug Fix

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-283

## Changes
* The modal summary API was incorrectly setting this value for HMB research. This led to a simple display bug that this should fix.


vvicario (approved):
:+1:

PR #336, opened 2019-04-24
[DUOS-283] Another HMB is not manual review case

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-283

## Changes
* The DAR summary was incorrectly setting this value for HMB research. This led to a simple display bug that this should fix. 

## See Also
Filed https://broadinstitute.atlassian.net/browse/DUOS-335 to cover the cleanup here - we should not be setting this value in two different places - the UI should read it from the same place consistently.


dags (approved):

vvicario (approved):
:+1:

rushtong:
Thank you @vvicario and @dags!

PR #338, opened 2019-05-07
[DUOS-348][risk=no] Update the logic for the number of DAC members

rushtong:
## Addresses
See https://broadinstitute.atlassian.net/browse/DUOS-348

## Changes
* Change the validate number of DAC members such that there is only a requirement that there be a single Chairperson and no minimum number of additional DAC members.
* There were no tests for this section of the code so added conditional tests for consent elections that covers the following cases:
  * Single chairperson and no DAC members (success)
  * No chairperson or DAC members (failure)
  * No chairperson (failure)

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/validate/UseRestrictionValidator.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/validate/UseRestrictionValidator.java
@@ -1,9 +1,5 @@
 package org.broadinstitute.consent.http.service.validate;
 
-import javax.ws.rs.client.Client;
-import javax.ws.rs.client.ClientBuilder;
-import javax.ws.rs.client.Entity;
-import javax.ws.rs.core.Response;
 import com.google.gson.Gson;
 import com.mongodb.Block;
 import com.mongodb.client.FindIterable;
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/consent/http/service/validate/UseRestrictionValidator.java
```diff
@@ -53,8 +53,8 @@ public void validateUseRestriction(String useRestriction) throws IllegalArgument
             if (!entity.isValid()) {
                 throw new IllegalArgumentException(entity.getErrors().toString());
             }
-        }else{
-            throw new IllegalArgumentException("There was an error posting the use restriction" + (res.readEntity(ValidateResponse.class)).getErrors().toString());
+        } else {
+            throw new IllegalArgumentException("There was an error posting the use restriction" + (res.readEntity(String.class)));
```
rushtong, line 57: This is not technically related to this PR, but this bug added all sorts of error messages to the logs which made debugging much harder than it needed to be.

vvicario (approved):
:+1:

dags (approved):

PR #339, opened 2019-05-09
[risk=no] Refactor Role Values

rushtong:
No Jira ticket. Note that this is already a fairly large PR so I am intentionally not trying to refactor everything role-based. Will address more cleanup in separate PRs. 

## Changes
* Moved all references to the current user roles to one place so we're not copy/pasting strings everywhere
* Move `DACUserRoles` -> `UserRoles` since that's what they really are, roles applicable to any user, not just DAC members

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/enumeration/VoteType.java
+++ b/src/main/java/org/broadinstitute/consent/http/enumeration/VoteType.java
@@ -1,8 +1,10 @@
 package org.broadinstitute.consent.http.enumeration;
 
+import org.broadinstitute.consent.http.resources.Resource;
+
 public enum VoteType {
 
-    DAC("DAC"), FINAL("FINAL"), AGREEMENT("AGREEMENT"), CHAIRPERSON("CHAIRPERSON"), DATA_OWNER("DATA_OWNER");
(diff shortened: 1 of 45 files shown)


src/main/java/org/broadinstitute/consent/http/enumeration/VoteType.java
```diff
@@ -1,8 +1,10 @@
 package org.broadinstitute.consent.http.enumeration;
 
+import org.broadinstitute.consent.http.resources.Resource;
+
 public enum VoteType {
 
-    DAC("DAC"), FINAL("FINAL"), AGREEMENT("AGREEMENT"), CHAIRPERSON("CHAIRPERSON"), DATA_OWNER("DATA_OWNER");
+    DAC("DAC"), FINAL("FINAL"), AGREEMENT("AGREEMENT"), CHAIRPERSON(Resource.CHAIRPERSON), DATA_OWNER("DATA_OWNER");
```
rushtong, line 7: Note that the vote type's `DATA_OWNER` is not consistent with the role we define in the roles table.

vvicario (approved):
Nice refactor!

dags (approved):
Nice refactor !

rushtong:
Thank you!

PR #340, opened 2019-05-10
[risk=no] Fix election archive status bug

rushtong:
## Changes
Update the query used to get the last election for entities when determining their archive status. The removed query was not reliable and sometimes returned incorrect results.


src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
```diff
@@ -186,12 +186,6 @@ void bulkCancelOpenElectionByReferenceIdAndType(@Bind("electionType") String ele
     @Mapper(DatabaseElectionMapper.class)
     Election findLastElectionByReferenceIdAndType(@Bind("referenceId") String referenceId, @Bind("type") String type);
```
rushtong, line 187: This is the query that we should be using.

dags (approved):

PR #345, opened 2019-06-20
[DUOS-390][risk=no] Handle legacy dataset id formats

rushtong:
## Addresses
Ticket: https://broadinstitute.atlassian.net/browse/DUOS-390

## Changes
This PR checks for the type of dataset id the Data Access Request has stored and then queries appropriately for it, whether it is an integer value or a string value.

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DatabaseDataSetAPI.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DatabaseDataSetAPI.java
@@ -1,24 +1,43 @@
 package org.broadinstitute.consent.http.service;
 
-import java.io.File;
-import java.util.*;
-import java.util.stream.Collectors;
-import javax.ws.rs.NotFoundException;
-
 import org.apache.commons.collections.CollectionUtils;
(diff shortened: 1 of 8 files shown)


src/main/java/org/broadinstitute/consent/http/service/DatabaseDataSetAPI.java
```diff
@@ -18,6 +18,7 @@
 import org.broadinstitute.consent.http.util.DarConstants;
 
 import org.bson.Document;
+import org.skife.jdbi.v2.sqlobject.Bind;
```
LeoForconesi: It seems that this isn't used anywhere

LeoForconesi (approved):
Other than an unnecessary import, everything is looking good for me.

src/main/java/org/broadinstitute/consent/http/service/DatabaseDataSetAPI.java
rushtong: Yup ... missed that, thanks!

PR #346, opened 2019-06-24
[DUOS-340][risk=no] Basic DAC API functionality

rushtong:
## Addresses 
https://broadinstitute.atlassian.net/browse/DUOS-340

## Changes
* new `Dac` class for the data access committee
* apis for CRUD
* swagger docs
* unit tests for all resource methods

Note that `DacService` does not follow the old pattern of an API holder, API interface, and API implementation. I hope to move away from that pattern for the much simpler one used in this PR. 

On the testing front, I'm hoping to mock all resource test dependencies and do functional tests on the database. That should reduce the amount of test-harness hoop jumping we go through in the future.

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
@@ -1,9 +1,7 @@
 package org.broadinstitute.consent.http.resources;
 
 import com.google.inject.Inject;
-import io.dropwizard.auth.Auth;
 import org.broadinstitute.consent.http.models.Dac;
-import org.broadinstitute.consent.http.models.User;
 import org.broadinstitute.consent.http.service.DacService;
 
 import javax.annotation.security.RolesAllowed;
@@ -16,9 +14,7 @@ import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
-import javax.ws.rs.core.Context;
 import javax.ws.rs.core.Response;
-import javax.ws.rs.core.UriInfo;
 import java.util.List;
 
 @Path("api/dac")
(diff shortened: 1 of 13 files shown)


dags (approved):

src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
```diff
@@ -0,0 +1,105 @@
...
+    @DELETE
+    @Path("{dacId}")
+    @Produces("application/json")
+    @RolesAllowed({ADMIN})
+    public Response deleteDac(@PathParam("dacId") Integer dacId) {
+        Dac dac = dacService.findById(dacId);
+        if (dac == null) {
+            throw new NotFoundException("Unable to find Data Access Committee with the provided id: " + dacId);
+        }
+        dacService.deleteDac(dacId);
```
ahaessly: probably a minor threading issue because it seems unlikely, but if two people try to delete the dac at the same time, there is a potential for calling deleteDac for an id that has already been deleted. i'm not sure if that would cause an exception, but if so you probably want to catch it and handle it.

src/test/java/org/broadinstitute/consent/http/resources/DacResourceTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+        Response response = dacResource.createDac(dac);
+        Assert.assertEquals(200, response.getStatus());
+    }
+
+    @Test(expected = BadRequestException.class)
+    public void testCreateDac_badRequest_1() throws Exception {
+        Dac dac = null;
+        when(dacService.createDac(any(), any())).thenReturn(1);
+        when(dacService.findById(1)).thenReturn(dac);
```
ahaessly: do you need these 2 lines of code? the createDac request doesn't get to either call. same with next couple of tests below

ahaessly (approved):
2 minor things

src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
rushtong: Great idea ... will do that.

src/test/java/org/broadinstitute/consent/http/resources/DacResourceTest.java
rushtong: Maybe not ... probably a copy/paste thing. Will fix, thank you!

PR #351, opened 2019-07-09
[DUOS-343][risk=no] Move email preference to user

rushtong:
## Addresses
Partially addresses https://broadinstitute.atlassian.net/browse/DUOS-343
There will be several PRs to address this story, one field at a time to limit the scope of the PRs. 
Requires: https://github.com/DataBiosphere/duos-ui/pull/179

## Changes
* Moved email preference to the user instead of the user's role.
* Updated APIs that enforced a DACUser or UserRole as input to be more flexible so we don't break legacy clients' API calls


src/main/java/org/broadinstitute/consent/http/service/users/handler/DACUserRolesHandler.java
```diff
@@ -185,31 +187,14 @@ public void updateRoles(Map<String, DACUser> usersMap) throws UserRoleHandlerExc
         }
     }
 
-    private void updateAdminEmailPreference(List<UserRole> originalRoles, List<UserRole> updatedRoles, Integer dacUserId) throws UserRoleHandlerException {
```
rushtong, line 188: This method, and all attempts to manage email by role membership is no longer a valid approach. Instead, we manage the user's email settings directly from the UI.

src/main/java/org/broadinstitute/consent/http/db/DACUserDAO.java
```diff
@@ -75,7 +75,7 @@ void updateDACUser(@Bind("email") String email,
     Integer verifyAdminUsers();
 
     @Mapper(DACUserRoleMapper.class)
-    @SqlQuery("select u.*, r.roleId, r.name, du.status from dacuser u inner join user_role du on du.user_id = u.dacUserId inner join roles r on r.roleId = du.role_id where r.name = :roleName and du.email_preference = :emailPreference")
+    @SqlQuery("select u.*, r.roleId, r.name, du.status from dacuser u inner join user_role du on du.user_id = u.dacUserId inner join roles r on r.roleId = du.role_id where r.name = :roleName and u.email_preference = :emailPreference")
```
thathert: I think this would be more readable if `dacuser` was aliased with `du`, `user_role` with `ur`, and keeping `roles` as `r`.

src/main/java/org/broadinstitute/consent/http/db/DACUserMapper.java
```diff
@@ -10,12 +10,13 @@
...
-        return new DACUser(
-                r.getInt("dacUserId"),
-                r.getString("email"),
-                r.getString("displayName"),
-                r.getDate("createDate"),
-                r.getString("additional_email"));
+        DACUser user = new DACUser();
+        user.setDacUserId(r.getInt("dacUserId"));
+        user.setEmail(r.getString("email"));
+        user.setEmailPreference(r.getBoolean("email_preference"));
```
thathert, line 16: Would be nice to consistently have camelCase or snake_case.. not sure if it's worth the migration here. I know in Postgres the convention is to go with snake_case for column/table names.

src/main/java/org/broadinstitute/consent/http/models/DACUser.java
```diff
@@ -1,12 +1,22 @@
...
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
 import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class DACUser {
 
+    private static final Logger logger = LoggerFactory.getLogger(DACUser.class.getName());
```
thathert: Shouldn't need the call to `.getName()` here -- I prefer to pass the class in and let slf4j figure out the formatting.

src/main/java/org/broadinstitute/consent/http/models/UserRole.java
```diff
@@ -24,29 +27,72 @@
...
+     *
+     * @param json A json string that may or may not be correctly structured as a UserRole
+     */
+    public UserRole(String json) {
+        JsonObject jsonObject = new JsonParser().parse(json).getAsJsonObject();
+        Logger log = LoggerFactory.getLogger(UserRole.class.getName());
+        if (jsonObject.has("roleId") && !jsonObject.get("roleId").isJsonNull()) {
+            try {
+                this.setRoleId(jsonObject.get("roleId").getAsInt());
+            } catch (Exception e) {
```
thathert: Is this idiomatic Gson? I'm used to Jackson and not familiar with this pattern -- wondering if there are particular cases where you'd want to raise an exception instead of logging?

src/test/java/org/broadinstitute/consent/http/db/UserRoleDAOTest.java
```diff
@@ -44,19 +44,19 @@ private void resetUserRoleTable() {
         // TODO: In concept, I don't like this approach and hope to migrate all db tests to be
         // independant of one another
         userRoleDAO.deleteAllUserRoles();
-        userRoleDAO.insertSingleUserRole(2, 1, false);
-        userRoleDAO.insertSingleUserRole(5, 1, false);
+        userRoleDAO.insertSingleUserRole(2, 1);
```
thathert: Having constants for these arguments would help make this more readable.

thathert (approved):
LGTM. Given I'm not as familiar with the system most of my comments are style related, but the approach and refactor seems sound.

src/main/java/org/broadinstitute/consent/http/db/DACUserDAO.java
rushtong: Will do.

src/main/java/org/broadinstitute/consent/http/db/DACUserMapper.java
rushtong: Yes, you're seeing the reason why I'm going through a lot of refactoring. I have https://broadinstitute.atlassian.net/browse/DUOS-282 to deal with this more generically, but am addressing these problems in a piecemeal fashion. We're moving to snake case db fields as we add new ones and there will be more refactoring to move camel cased columns to snake case.

src/main/java/org/broadinstitute/consent/http/models/DACUser.java
rushtong: Good catch - will double check that that works.

src/main/java/org/broadinstitute/consent/http/models/UserRole.java
rushtong: Pretty sure this is not idiomatic anything, unfortunately. The goal here is to do JSON conversion in the case where we don't have a json structure that matches the POJO. I think maybe logging a warning here might be better, but I expect that the older (deployed) client code will be triggering that pretty regularly and really shouldn't be an error in that case. The non-standard JSON is picked apart in other places to handle the legacy clients.

src/test/java/org/broadinstitute/consent/http/db/UserRoleDAOTest.java
rushtong: 👍

rushtong: We actually have constants for them that were added in after these tests were updated. Will make this cleaner.

src/main/java/org/broadinstitute/consent/http/models/UserRole.java
rushtong: This can be done a lot more cleanly, thanks for the advice.

src/main/java/org/broadinstitute/consent/http/models/DACUser.java
```diff
@@ -53,6 +60,41 @@ public DACUser(Integer dacUserId, String email, String displayName,
...
+        if (u.getDisplayName() != null) {
+            this.setDisplayName(u.getDisplayName());
+        }
+        if (u.getCreateDate() != null) {
+            this.setCreateDate(u.getCreateDate());
+        }
+        if (u.getAdditionalEmail() != null) {
+            this.setAdditionalEmail(u.getAdditionalEmail());
+        }
+        if (u.getEmailPreference() != null) {
```
vvicario, line 91: Can be empty email, displayName or additionalEmail ? If this is possible, we can validate using !StringUtils.isEmpty

vvicario (approved):

src/main/java/org/broadinstitute/consent/http/models/DACUser.java
rushtong: Nice catch - will look at improving these checks.

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -46,28 +47,4 @@ public Response createUser(@Context UriInfo info, DACUser userToCreate, @Auth Au
...
-    @Produces("application/json")
-    @PermitAll
-    public Response partialUpdate(List<PatchOperation> patchOperations, @Auth AuthUser user) {
-        try {
-            return Response.ok().entity(userAPI.updatePartialUser(patchOperations, user.getName())).build();
-        } catch (Exception e){
-            return createExceptionResponse(e);
-        }
-    }
-
```
rushtong: Both of these endpoints were unused.

PR #352, opened 2019-07-17
[DUOS-343][risk=no] Move status, rationale, profile complete to user

rushtong:
## Addresses
Finishes up https://broadinstitute.atlassian.net/browse/DUOS-343
As a side effect, this also completes https://broadinstitute.atlassian.net/browse/DUOS-393
Requires: 
* https://github.com/DataBiosphere/consent/pull/351
* https://github.com/DataBiosphere/duos-ui/pull/179

## Changes
Moved status, rationale, profile complete to the user instead of the user's role.

# Testing 
Have tested this manually with the corresponding UI PR.
* Admin -> add new user
* Admin -> edit user
* Admin -> change user roles (chair, admin, member, alumni, researcher, data owner)
* Admin -> change admin email notification preference
* Admin -> manage users -> appove pending researcher
* Admin -> manage users -> reject approved researcher
* Admin -> manage users -> approve rejected researcher


src/main/java/org/broadinstitute/consent/http/resources/DACUserResource.java
```diff
@@ -162,33 +165,49 @@ public Response validateDelegation(@QueryParam("role") String role, String json)
...
+                user = dacUserAPI.updateUserStatus(statusOpt.get(), userId);
+            } catch (Exception e) {
+                return createExceptionResponse(e);
+            }
         }
+        if (rationaleOpt.isPresent()) {
+            try {
+                user = dacUserAPI.updateUserRationale(rationaleOpt.get(), userId);
+            } catch (Exception e) {
+                return createExceptionResponse(e);
```
tadeork, line 189: I think this should be on a service or be transactional. 
What if `dacUserAPI.updateUserRationale(rationaleOpt.get(), userId);`fails?
Then the `dacUserAPI.updateUserStatus(statusOpt.get(), userId);` will be updated but not the `dacUserAPI.updateUserRationale(rationaleOpt.get(), userId);`

tadeork (approved):
Just one minor comment.

src/main/java/org/broadinstitute/consent/http/resources/DACUserResource.java
rushtong: Good point - I would normally agree with that suggestion. The intention is to deal with them individually if and only if they are provided in the request. One of the issues we're dealing with on the API side is that we have many overlapping and duplicative calls that really should be condensed. In the case of this API, the underlying method was performing partial updates depending on what was in the request, but it was hidden. This just makes that explicit. Given that this API is being deprecated in favor of a real update API, I lean towards leaving this as is for now.

tadeork: Yes, I thought that could be a case like this but I didn't know the context. It's OK then.

PR #355, opened 2019-07-23
[DUOS-411][risk=no] Admin api for advancing an election

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-411

## Changes
Admin only endpoint that allows for setting the empty DAC votes for an election.

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/ElectionResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/ElectionResource.java
@@ -1,26 +1,73 @@
 package org.broadinstitute.consent.http.resources;
 
+import org.broadinstitute.consent.http.enumeration.VoteType;
 import org.broadinstitute.consent.http.models.Election;
+import org.broadinstitute.consent.http.models.Vote;
 import org.broadinstitute.consent.http.models.dto.Error;
 import org.broadinstitute.consent.http.service.AbstractElectionAPI;
+import org.broadinstitute.consent.http.service.AbstractVoteAPI;
 import org.broadinstitute.consent.http.service.ElectionAPI;
+import org.broadinstitute.consent.http.service.VoteAPI;
 
 import javax.annotation.security.PermitAll;
 import javax.annotation.security.RolesAllowed;
-import javax.ws.rs.*;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.GET;
+import javax.ws.rs.POST;
+import javax.ws.rs.PUT;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriInfo;
+import java.util.Date;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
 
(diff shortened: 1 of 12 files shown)


rushtong:
There is some kind of bug that needs to be dug through. The createDate field is not getting populated in the database even though the value is being set in the DAO call. Need to run a post-fix up query after using this:
```
select * from vote where rationale = 'Advanced by administrator';
// followed by
update vote set createDate = updateDate where rationale = 'Advanced by administrator';
```

src/main/java/org/broadinstitute/consent/http/resources/ElectionResource.java
```diff
@@ -1,26 +1,73 @@
...
+import javax.ws.rs.PUT;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriInfo;
+import java.util.Date;
+import java.util.List;
+import java.util.Objects;
```
vvicario: It's not used

src/main/java/org/broadinstitute/consent/http/resources/ElectionResource.java
```diff
@@ -1,26 +1,73 @@
...
+    @Produces("application/json")
+    @Path("/{referenceId}/advance/{vote}")
+    @RolesAllowed({ADMIN})
+    public Response advanceElection(@PathParam("referenceId") String referenceId, @PathParam("vote") String vote) {
+        try {
+            List<Vote> voteList = voteApi.describeVotes(referenceId);
+            List<Integer> electionIds = voteList.stream().map(Vote::getElectionId).collect(Collectors.toList());
+            Date now = new Date();
+            Boolean voteValue = vote.equalsIgnoreCase("yes");
+            voteList.
```
vvicario: Just a thought. Maybe this logic can be moved to a service

vvicario (approved):

LeoForconesi (approved):

src/main/java/org/broadinstitute/consent/http/resources/ElectionResource.java
rushtong: Thank you - good catch.

src/main/java/org/broadinstitute/consent/http/resources/ElectionResource.java
rushtong: Yes, the should, but I hesitate to add to current services as they are constructed now - the election and vote api classes use the old pattern and I don't want to tackle refactoring those in this PR. I'm on the fence, but lean towards leaving them as is here and tackling the refactor later. The API classes are so convoluted and inefficient, they really need a complete re-write.

rushtong: Ok, you convinced me @vvicario - I'm going to take a shot at extracting out a new service class and add some TODOs for migrating older code to newer.

rushtong:
@vvicario or @LeoForconesi - made some relatively significant refactorings and also fixed the weird create date bug that was happening under the previous iteration that used the legacy `VoteAPI` implementation. Please feel free to re-review if you have time, thank you!

PR #362, opened 2019-09-16
[DUOS-444][risk=low] Remove deprecated APIs

rushtong:
## Addresses
Partially addresses https://broadinstitute.atlassian.net/browse/DUOS-444

This PR removes two unused endpoints that have security vulnerabilities. The only API in use in this resource (POST to `/api/user`) is described here: https://github.com/DataBiosphere/duos-ui/blob/develop/src/libs/ajax.js#L930

A consequence of this is that there is a ton of supporting infrastructure code that was unused. Taking the opportunity here to remove anything that touched/tested these unused APIs.


rushtong:
@dags and @vvicario - Requesting review from you two as the original implementers of a lot of this code that I think was intended to support researcher user updates. Turns out we never used two endpoints in `UserResource` so this PR strips out all of that along with all of the additional support code for it.

rushtong:
@dags and @vvicario - Just a reminder about this PR.

vvicario (approved):

dags (approved):

PR #364, opened 2019-09-17
[DUOS-452][risk=no] Stream content instead of writing files

rushtong:
## Addresses
This covers the API part of https://github.com/DataBiosphere/consent/pull/new/DUOS-452
It does not cover `DataSetResource.createDataSet()` ... I will deal with that in a different story that we have for revamping the data set upload/download feature.

## Changes
Use a stream to capture the output from GCS instead of writing to a temp file.


dags (approved):

vvicario (approved):
:+1:

dags:
nice use of streams !

PR #368, opened 2019-10-03
[DUOS-395][risk=no] API Support

rushtong:
## Addresses
API changes in support of https://broadinstitute.atlassian.net/browse/DUOS-395
See also https://github.com/DataBiosphere/duos-ui/pull/171

## Changes
* Due to security testing, we have some invalid data that presents an opportunity to better handle error cases. Log the case of invalid users on a DAR instead of throwing a 404.

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DatabaseDataAccessRequestAPI.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DatabaseDataAccessRequestAPI.java
@@ -28,7 +28,6 @@ import org.broadinstitute.consent.http.models.DataSet;
 import org.broadinstitute.consent.http.models.DataUseDTO;
 import org.broadinstitute.consent.http.models.Election;
 import org.broadinstitute.consent.http.models.ResearcherProperty;
-import org.broadinstitute.consent.http.models.UserRole;
 import org.broadinstitute.consent.http.models.Vote;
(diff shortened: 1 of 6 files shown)


src/main/java/org/broadinstitute/consent/http/service/DatabaseDataAccessRequestAPI.java
```diff
@@ -562,12 +561,18 @@ public File createDataSetApprovedUsersDocument(Integer dataSetId) throws IOExcep
     @Override
     public DARModalDetailsDTO DARModalDetailsDTOBuilder(Document dar, DACUser dacUser, ElectionAPI electionApi) {
         DARModalDetailsDTO darModalDetailsDTO = new DARModalDetailsDTO();
-        List<Document> datasetDetails = populateDatasetDetailDocuments(dar.get(DarConstants.DATASET_DETAIL));
+        List<DataSet> datasets = populateDatasets(dar.get(DarConstants.DATASET_DETAIL));
+        Optional<DACUser> optionalUser = Optional.ofNullable(dacUser);
```
rushtong, line 565: Technically, this shouldn't be necessary, but through security testing, we've found that it's possible to inject invalid data (i.e. fake user ids) which throws off a lot of the API functions.

dags (approved):

PR #379, opened 2019-11-18
[DUOS-346][risk=no] Create DUL Election with a Dac

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-346

## Changes
New endpoint for creating a consent election with a dac
Requires https://github.com/DataBiosphere/duos-ui/pull/214

Lots and lots of test updates. There is a very large testing gap in this area. This PR removes some of the consent election resource integration tests in favor of unit tests that test the resource, service, and dao layers separately.


src/test/java/org/broadinstitute/consent/http/AbstractTest.java
```diff
@@ -247,4 +251,57 @@ protected DBI getApplicationJdbi() {
         return new DBIFactory().build(environment, configuration.getDataSourceFactory(), "mysql");
     }
 
+    String consentPath() {
```
rushtong, line 269: All of these utility methods are copied from what was `ConsentServiceTest`. That class now actually tests consent service methods instead of being just a group of utility methods.

PR #380, opened 2019-11-18
[no jira][risk=no] Address RDF4J Security Vulnerability

rushtong:
## Addresses
No Jira

## Changes
Library update due to SC report:

https://www.sourceclear.com/vulnerability-database/vulnerabilities/8050
> Directory Traversal
> 
> rdf4j-util is vulnerable to directory traversal. An attacker is able to overwrite arbitrary files using the characters `../` as an entry in a ZIP archive. The overwrite occurs during decompressing of the ZIP file.
> 
> Update
> 
> This issue was fixed in version 2.4.3 of RDF4J: util. That version is currently considered safe, we suggest that you upgrade to the fixed version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -795,13 +795,13 @@
     <dependency>
       <groupId>org.eclipse.rdf4j</groupId>
       <artifactId>rdf4j-rio-rdfxml</artifactId>
-      <version>2.4.2</version>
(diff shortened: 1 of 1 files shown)


MatthewBemis (approved):

rushtong:
Thank you @MatthewBemis!

PR #385, opened 2019-12-03
[no jira][risk=no] Fix maven enforcer warning

rushtong:
## Addresses
Fixes:

> [WARNING] The project org.broadinstitute:consent:jar:1.0.1 uses prerequisites which is only intended for maven-plugin projects but not for non maven-plugin projects. For such purposes you should use the maven-enforcer-plugin. See https://maven.apache.org/enforcer/enforcer-rules/requireMavenVersion.html

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,10 +9,6 @@
   <packaging>jar</packaging>
   <name>Consent Management Services</name>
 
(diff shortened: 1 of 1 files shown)


abaumann:
👍

PR #392, opened 2019-12-16
[DUOS-487][risk=no] Create votes for elections for DAC members

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-487

## Changes
* Move create votes logic to service class
* Filter vote creation by DAC user if the election is associated to a DAC
* More unit tests for service classes and daos


rushtong:
Received a 👍from @Marianie-Simeon in slack

PR #394, opened 2019-12-18
[DUOS-345][risk=no] Some fixes for add/edit user api

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-345
Requires UI PR: https://github.com/DataBiosphere/duos-ui/pull/223
Also depends on: https://github.com/DataBiosphere/consent/pull/393

## Changes
* Minor refactoring mainly around role removal and extracting the email preference.


src/main/java/org/broadinstitute/consent/http/service/users/handler/DACUserRolesHandler.java
```diff
@@ -304,10 +305,6 @@ private void changeRolesWithoutDelegation(DACUser updatedUser, List<UserRole> re
         return roles.stream().filter(rol -> !toSubstractRolesNames.contains(rol.getName().toUpperCase())).collect(Collectors.toList());
     }
 
-    private void removeRole(Integer dacUserId, String role) {
-        userRoleDAO.removeSingleUserRole(dacUserId, roleIdMap.get(role));
```
rushtong, line 308: This was the root of the bug we saw where users were not being removed. Accessing the string name via the `roleIdMap` was brittle so refactored to be more clear (and consistent using the `UserRoles` enum) about where the ID is coming from.

PR #395, opened 2019-12-19
[DUOS-392][risk=no] Remove Deprecated Code

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-392
Depends on https://github.com/DataBiosphere/consent/pull/394 which also depends on https://github.com/DataBiosphere/consent/pull/393

## Changes
The `DACUserDAO.findDACUserIdByRole` method was in use only to move a chair from chair to alumni when adding a new chairperson. That logic is no longer valid and will be addressed differently in [DUOS-491](https://broadinstitute.atlassian.net/browse/DUOS-491) so we're removing it altogether here. Lots of dead code went along with that.

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DACUserResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DACUserResource.java
@@ -3,16 +3,11 @@ package org.broadinstitute.consent.http.resources;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
-import com.google.inject.Inject;
 import org.apache.log4j.Logger;
 import org.broadinstitute.consent.http.models.DACUser;
-import org.broadinstitute.consent.http.models.Election;
 import org.broadinstitute.consent.http.models.UserRole;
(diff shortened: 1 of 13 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DACUserResource.java
```diff
@@ -63,11 +53,6 @@ public DACUserResource(VoteService voteService) {
     public Response createDACUser(@Context UriInfo info, String json) {
         try {
             DACUser dacUser = dacUserAPI.createDACUser(new DACUser(json));
-            if (isChairPerson(dacUser.getRoles())) {
-                dacUserAPI.updateExistentChairPersonToAlumni(dacUser.getDacUserId());
-                List<Election> elections = electionAPI.cancelOpenElectionAndReopen();
-                voteService.createVotesForElections(elections);
-            }
```
rushtong, line 70: This functionality is being replaced with https://broadinstitute.atlassian.net/browse/DUOS-491

gordonwade (approved):

rushtong:
Thank you @gordonwade!

PR #407, opened 2020-01-16
[DUOS-376][risk=no] Update dac-for-election queries

rushtong:
## Addresses
Minor fixup for https://broadinstitute.atlassian.net/browse/DUOS-376
Does not complete the above epic - task is too small for it's own story.

## Changes
The query for dacs by election was not sufficient. We need to be able to find elections that may not have had a dataset id associated to it yet. This happens in the case of DUL elections. In that case, we first vote on the DUL, then we make the dataset association later on. This update allows us to find elections for dacs where that association hasn't been made yet.

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
@@ -272,17 +272,23 @@ public interface ElectionDAO extends Transactional<ElectionDAO> {
     Boolean findFinalAccessVote(@Bind("electionId") Integer electionId);
 
     /**
-     * Find the Dac for this election
+     * Find the Dac for this election. Looks across associations to a dac via dataset and those
+     * associated via the consent.
      *
      * @param electionId The election id
      * @return Dac for this election
      */
     @Mapper(DacMapper.class)
-    @SqlQuery("select d.* from dac d " +
-            " inner join consents c on d.dac_id = c.dac_id " +
-            " inner join consentassociations a on a.consentId = c.consentId " +
-            " inner join election e on e.datasetId = a.dataSetId and e.electionId = :electionId " +
-            " limit 1 ")
+    @SqlQuery("select d0.* from ( " +
+            "   select d1.* from dac d1 " +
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
```diff
@@ -272,17 +272,23 @@ void updateElectionById(@Bind("electionId") Integer electionId,
...
      * @param electionId The election id
      * @return Dac for this election
      */
     @Mapper(DacMapper.class)
-    @SqlQuery("select d.* from dac d " +
-            " inner join consents c on d.dac_id = c.dac_id " +
-            " inner join consentassociations a on a.consentId = c.consentId " +
-            " inner join election e on e.datasetId = a.dataSetId and e.electionId = :electionId " +
-            " limit 1 ")
+    @SqlQuery("select d0.* from ( " +
```
Marianie-Alphonse, line 282: You are joining with union an election/dac table with dataset Id and one without. How are you going to later assign the elections with dataset Ids.

rushtong: There are three kinds of elections we currently support:
* Consent (DUL) Elections. Don't have a dataset yet (or ever) because the consent has to be voted/approved before we associate datasets to the consent.
* Data Access Request (DAR) Elections. We have a dataset here as this is when the researcher is requesting access to a dataset. In this case, we set the dataset id.
* Dataset Elections. We have a dataset here too since this is when a data owner is granting permission to a researcher to use the owner's dataset.

For consent elections, we don't necessarily ever need to associate a dataset to the election. In the case that have a DAC and the DAC does have datasets, we grab the first one and associate to the [election here](https://github.com/DataBiosphere/consent/blob/develop/src/main/java/org/broadinstitute/consent/http/resources/ConsentElectionResource.java#L143)

Marianie-Alphonse: That makes sense! Thank you for explaining.

Marianie-Alphonse (approved):

PR #515, opened 2020-02-07
[DUOS-225][risk=no] Use Postgres

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-225

Requires: https://github.com/broadinstitute/firecloud-develop/pull/2019

New databases are set up and configured for `dev`, `staging`, `perf`, `alpha`, and `prod`. Have already run migrations from `mysql` -> `postgres` using instructions from https://github.com/DataBiosphere/consent/pull/486. When FC does a release, I will coordinate a new `pgloader` run so prod is up to date.

### Code
Code changes are pretty minimal. 
* The dependencies
* Minor query update for the unsupported `datediff` mysql function
* Minor refactoring around the db name that we use in a few places

### Liquibase
Liquibase is another story and deserves some explanation. The `pgloader` migration down-cases all table and column names in the migration which works just fine because our camel cased queries via jdbi are handled in a down-cased manner at the db level. Future PRs will address this technical debt (see list). 

However, starting up the database from scratch (e.g. FiaBs), all of the create/update/insert statements are *not* down-cased and retain their original mixed case naming similar to what raw sql in the code is expecting. That raw sql is down-cased when it gets to the database and nothing works. So, the more concise solution here is to update all of the changesets (where necessary) such that we are only using lower-cased tables and columns. Constraints are fine - they are not referenced in code. 

## Tech Debt
* https://broadinstitute.atlassian.net/browse/DUOS-282
* https://broadinstitute.atlassian.net/browse/DUOS-397
* https://broadinstitute.atlassian.net/browse/DUOS-344
* https://broadinstitute.atlassian.net/browse/DUOS-569
* https://broadinstitute.atlassian.net/browse/DUOS-570


src/main/resources/changesets/changelog-consent-36.0.xml
```diff
@@ -5,11 +5,12 @@
...
     <changeSet author="grushton" id="114">
+        <validCheckSum>ANY</validCheckSum>
         <addColumn tableName="consents">
-            <column name="dataUse" type="text"/>
+            <column name="datause" type="text"/>
         </addColumn>
-        <sql dbms="hsql, mysql">UPDATE consents SET dataUse = ''</sql>
-        <sql dbms="hsql, mysql">ALTER TABLE consents MODIFY dataUse text NOT NULL</sql>
+        <sql dbms="mysql">UPDATE consents SET datause = ''</sql>
+        <sql dbms="mysql">ALTER TABLE consents MODIFY datause text NOT NULL</sql>
```
ahaessly, line 13: other places where you had dbms="hsql, mysql" you remove both... do you need mysql here?

src/main/resources/changesets/changelog-consent-15.0.xml
```diff
@@ -5,61 +5,69 @@
...
+            <column name="type"  type="varchar(255)"/>
         </addColumn>
     </changeSet>
 
     <changeSet author="vvicario" id="66">
-        <sql dbms="hsql, mysql">
-            update vote set type = 'DAC' where isFinalAccessVote = 0;
-            update vote set type = 'FINAL' where isFinalAccessVote = 1;
+        <validCheckSum>ANY</validCheckSum>
+        <sql dbms="hsqldb, mysql">
```
ahaessly, line 62: do you still need dbms="hsql, mysql" here?

src/main/resources/changesets/changelog-consent-48.0.xml
```diff
@@ -5,12 +5,13 @@
...
     <changeSet author="grushton" id="48.0-1">
+        <validCheckSum>ANY</validCheckSum>
         <dropAllForeignKeyConstraints baseTableName="user_role_bak"/>
         <addColumn tableName="dacuser">
             <column name="email_preference" type="boolean" defaultValueBoolean="true" />
         </addColumn>
-        <sql>
-            update dacuser set email_preference = false where dacUserId in (select distinct user_id from user_role where email_preference = false)
+        <sql dbms="hsqldb, mysql">
+            update dacuser set email_preference = false where dacuserid in (select distinct user_id from user_role where email_preference = false)
```
ahaessly, line 14: do you still need dbms="hsql, mysql" here?

src/main/resources/changesets/changelog-consent-49.0.xml
```diff
@@ -5,25 +5,26 @@
...
 
     <changeSet author="grushton" id="49.0">
+        <validCheckSum>ANY</validCheckSum>
         <addColumn tableName="dacuser">
             <column name="status" type="int" defaultValue="0"/>
             <column name="rationale" type="text"/>
         </addColumn>
-        <sql>
-            update dacuser set status = 0 where dacUserId in (select user_id from user_role where status = 0 and role_id = 5)
+        <sql dbms="hsqldb, mysql">
```
ahaessly, line 13: again here and below: do you still need dbms="hsql, mysql" here?

src/main/resources/changesets/changelog-consent-5.0.xml
```diff
@@ -5,20 +5,22 @@
...
             </column>
         </addColumn>
     </changeSet>
 
 
     <changeSet author="ssaucedo" id="31">
-        <sql dbms="hsql, mysql">
-            update consents t1, consents t2 set t1.name = t2.consentId where t1.consentId = t2.consentId
+        <validCheckSum>ANY</validCheckSum>
+        <sql dbms="mysql">
```
ahaessly, line 22: again dbms="mysql"

ahaessly (approved):
just a few places you might have missed something, but otherwise 👍

src/main/resources/changesets/changelog-consent-36.0.xml
rushtong: This could have used some more explanation in the PR description. In all of the cases where we changed the `dbms` field, these are my reasons. 

#### `dbms="mysql"` cases
These are the cases where I wanted to preserve some history since we made real changes to the mysql database. We need that field to restrict the sql so it will only run when in a mysql environment. We don't want these to run in either hsqldb or postgres for two reasons. When populating a new db (hsqldb, posgres), that data doesn't exist and never did. It was only ever used to migrate mysql data. There's no harm in running it for hsqldb, it works, but in a lot of cases, it fails in postgres due to syntax differences. In this case, the alter table fails. 

#### `dbms="hsqldb, mysql"` cases
In these cases, my intention was also to preserve some history. To be honest, some of the following ones could just be restricted to mysql as well. They either fail in postgres as they are written or are not necessary. For example, this fails in postgres (see next comment): `update vote set type = 'DAC' where isFinalAccessVote = 0;`. And like the above reason, we need to exclude these changes from postgres, so we need to specify which environments they should be run in.

src/main/resources/changesets/changelog-consent-15.0.xml
rushtong: Yes, to restrict them from failing in postgres.

src/main/resources/changesets/changelog-consent-48.0.xml
rushtong: Yes, to restrict them from failing in postgres.

src/main/resources/changesets/changelog-consent-49.0.xml
rushtong: Yes, to restrict them from failing in postgres.

src/main/resources/changesets/changelog-consent-5.0.xml
rushtong: Could have also been `dbms="hsqldb, mysql"`, but yes, to restrict them from failing in postgres.

Marianie-Alphonse (approved):

src/main/java/org/broadinstitute/consent/http/db/ConsentDAO.java
```diff
@@ -45,7 +45,7 @@
     @SqlQuery("SELECT c.consentId, cs.dataSetId, ds.name, ds.objectId " +
             "FROM consents c INNER JOIN consentassociations cs ON c.consentId = cs.consentId " +
             "INNER JOIN dataset ds on cs.dataSetId = ds.dataSetId "+
-            "WHERE cs.dataSetId IN (<datasetId>)")
+            "WHERE cs.dataSetId::text IN (<datasetId>)")
```
rushtong, line 48: This is unintuitive and took a good amount of time to debug. JDBI is quoting the integer values in `<datasetId>` in this query which fails against the `bigint` `datasetid` column. This doesn't seem to be a problem in other cases but is in this one.

src/main/resources/changesets/changelog-consent-51.0.xml
```diff
@@ -0,0 +1,13 @@
+<databaseChangeLog
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet author="grushton" id="51.0">
+        <sql dbms="postgres">
+            alter table election alter column archived set default false;
+        </sql>
```
rushtong: This should have been covered in the `pgloader` migration, but isn't in the final result.

PR #530, opened 2020-02-21
[DUOS-569][risk=no] Remove Unused Endpoints

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-569

While tracing through the service call that was subject to sql injection, I found several bugs and unused endpoints in the resource that called it. Also found that we don't use this in the UI at all.

## Changes
* Remove all related resource, api, dao calls, and tests.


src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
```diff
@@ -1,152 +1,87 @@
...
-                throw new NotFoundException("Cannot find some consents ids: " + StringUtils.join(diffIds, ","));
+    public Response findByIds(@QueryParam("ids") String ids) {
+        if (StringUtils.isBlank(ids)) {
+            throw new NotFoundException("Cannot find any consents without ids");
+        }
+        List<String> splitIds = new ArrayList<>();
+        for (String id : ids.split(",")) {
+            try {
+                //noinspection ResultOfMethodCallIgnored
+                UUID.fromString(id);
```
adrazhi: @rushtong If you need strong security around this uuid consider using the java.secureRandom would be a better choice.   But if it's related to a non-critical part then no need to worry about it.

adrazhi (approved):
👍

src/main/java/org/broadinstitute/consent/http/resources/ConsentsResource.java
rushtong: Thanks for the tip - I will definitely look to replace this pattern in other parts of the code.

PR #550, opened 2020-03-09
[DUOS-571][risk=no] Postgres Unit Testing

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-571

## Changes
* Use a real postgres db to do unit testing. 
* Remove all old hsqldb stuff
* Migrate lots of integration tests to unit tests
* Delete several tests that are either duplicated or deprecated
* Fixed query bugs encountered during testing
  * See, for example, `ConsentDAO`
  * Significant test coverage (and fixes) added to `VoteDAO` due to some casing bugs I missed in the mysql -> postgres conversion


src/main/java/org/broadinstitute/consent/http/db/ConsentDAO.java
```diff
@@ -31,7 +31,7 @@
     @SqlQuery("SELECT c.name " +
             "FROM consents c INNER JOIN consentassociations cs ON c.consentId = cs.consentId "+
             "WHERE cs.dataSetId = :datasetId")
-    String findConsentNameFromDatasetID(@Bind("datasetId") String datasetId);
+    String findConsentNameFromDatasetID(@Bind("datasetId") Integer datasetId);
```
Marianie-Alphonse, line 34: What is the purpose of making the dataset Id an Integer instead of a String?

rushtong: `datasetId` is actually an integer in the database. This was a pre-existing bug that we never saw and didn't break in mysql. In postgres, it breaks :-(

src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
```diff
@@ -18,54 +20,81 @@
...
+    static DacDAO dacDAO;
+    static DACUserDAO userDAO;
+    static DataSetDAO dataSetDAO;
+    static ElectionDAO electionDAO;
+    static UserRoleDAO userRoleDAO;
+    static VoteDAO voteDAO;
+
+    private static List<Integer> createdDataSetIds = new ArrayList<>();
+    private static List<Integer> createdDacIds = new ArrayList<>();
+    private static List<String> createdConsentIds = new ArrayList<>();
```
Marianie-Alphonse, line 52: Why is thecreatedConsentIds type String and the rest are Integers?

rushtong: Consent IDs are UUIDs, so they are strings. The others are auto-incremented IDs and are really ints.

src/main/java/org/broadinstitute/consent/http/db/ConsentDAO.java
Marianie-Alphonse: That makes sense! Thank you!

Marianie-Alphonse (approved):

PR #564, opened 2020-03-25
[DUOS-611][risk=no] Chair console fixes

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-611

## Changes
* Lower case all of the election dao queries. There are case inconsistencies between several of the enum values that we use for elections and votes so lower casing both sides of the equality checks resolves that. This was the primary cause of the filed bug in that there was a type mismatch that caused an election query to return no results.
  * Discovered secondary bug while testing that when adding a user as a dac member, votes would sometimes not be created properly.
  * When creating votes for a dac chair/member, ensure that we're looking at the user with updated roles.
  * Update the primary get user by id query so it gets the roles.
  * Removed duplicate find user by id method.


rushtong:
Merging this - it needs to be in the next release.

PR #571, opened 2020-03-31
[DUOS-614] Fix query to match running environments

rushtong:
## Addresses
https://github.com/DataBiosphere/consent/pull/new/DUOS-614

This error shows up in the UI as an error when updating the consent:
Screen Shot 2020-03-31 at 1 08 24 PM
And when that happens, the endpoint creates an election, but doesn't create votes for the election. See https://github.com/DataBiosphere/consent/pull/new/DUOS-616 for handling that error case.

## Changes
The query was written to pass tests, but the test db was not being set up correctly. It was being converted to bit instead of what it is in production environments, boolean.

### DIFF
--- a/src/main/resources/changesets/changelog-consent-42.0.xml
+++ b/src/main/resources/changesets/changelog-consent-42.0.xml
@@ -5,8 +5,9 @@
     http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
 
     <changeSet author="lforconesi" id="120">
+        <validCheckSum>ANY</validCheckSum>
         <addColumn tableName="consents">
-            <column name="updated" type="BIT(1)">
+            <column name="updated" type="boolean">
                 <constraints nullable="true"/>
             </column>
         </addColumn>
--- a/src/main/java/org/broadinstitute/consent/http/db/ConsentDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/ConsentDAO.java
@@ -86,7 +86,7 @@ public interface ConsentDAO extends Transactional<ConsentDAO> {
             " sortDate = :sortDate, " +
             " translatedUseRestriction = :translatedUseRestriction, " +
             " groupName = :groupName, " +
-            " updated = (:updated::int)::bit(1), " +
+            " updated = :updated, " +
             " dac_id = :dacId " +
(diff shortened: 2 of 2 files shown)


src/main/resources/changesets/changelog-consent-42.0.xml
```diff
@@ -5,8 +5,9 @@
     http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
 
     <changeSet author="lforconesi" id="120">
+        <validCheckSum>ANY</validCheckSum>
         <addColumn tableName="consents">
-            <column name="updated" type="BIT(1)">
+            <column name="updated" type="boolean">
```
rushtong, line 10: I missed this changelog in the original mysql->postgres conversion. This is what we actually have in all running environments:
```
consent=> \d+ consents 
                                                   Table "consent.consents"
          Column          |           Type           | Collation | Nullable | Default | Storage  | Stats target | Description 
--------------------------+--------------------------+-----------+----------+---------+----------+--------------+-------------
 consentid                | character varying(255)   |           | not null |         | extended |              | 
 requiresmanualreview     | boolean                  |           | not null | true    | plain    |              | 
 userestriction           | text                     |           | not null |         | extended |              | 
 active                   | boolean                  |           | not null | true    | plain    |              | 
 datauseletter            | text                     |           |          |         | extended |              | 
 name                     | character varying(255)   |           | not null |         | extended |              | 
 dulname                  | character varying(255)   |           |          |         | extended |              | 
 createdate               | timestamp with time zone |           |          |         | plain    |              | 
 lastupdate               | timestamp with time zone |           |          |         | plain    |              | 
 sortdate                 | timestamp with time zone |           |          |         | plain    |              | 
 translateduserestriction | text                     |           |          |         | extended |              | 
 valid_restriction        | boolean                  |           |          | true    | plain    |              | 
 datause                  | text                     |           | not null |         | extended |              | 
 groupname                | character varying(255)   |           |          |         | extended |              | 
 updated                  | boolean                  |           |          |         | plain    |              | 
 dac_id                   | bigint                   |           |          |         | plain    |              | 
...
```

gordonwade (approved):

rushtong:
Thank you @gordonwade!

PR #572, opened 2020-03-31
[DUOS-611][risk=no] Fix DUL Pending Cases

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-611

## Changes
* Due to other bugs, there are cases where there are missing DUL election votes for a user.
* When we find those cases, create votes for them on the fly.
* Other minor enhancements

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/ConsentDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/ConsentDAO.java
@@ -170,9 +170,9 @@ public interface ConsentDAO extends Transactional<ConsentDAO> {
     @SqlQuery("select consentId, name, useRestriction from consents where valid_restriction = false ")
     List<UseRestrictionDTO> findInvalidRestrictions();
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/consent/http/db/ConsentDAO.java
```diff
@@ -170,9 +170,9 @@ Consent findConsentByAssociationAndObjectId(@Bind("associationType") String asso
     @SqlQuery("select consentId, name, useRestriction from consents where valid_restriction = false ")
     List<UseRestrictionDTO> findInvalidRestrictions();
 
-    @SqlUpdate("update consents set updated = (:consentStatus::int)::bit(1) where consentId = :referenceId")
+    @SqlUpdate("update consents set updated = :updated where consentId = :referenceId")
```
rushtong, line 173: I should have caught this in https://github.com/DataBiosphere/consent/pull/571 but missed it.

rushtong:
Merging so we can continue staging testing.

PR #575, opened 2020-04-01
[DUOS-618][risk=no] Handle Unescaped Unicode in DARs

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-618

## Changes
* Offload some string manipulation of the dar to the dao queries using the `(data #>> '{}')::jsonb` syntax
* Double unescape the double-escaped content before  trying to deserialize via `Gson`
* Added some sample content that comes from current data and tests to ensure we're decoding things correctly

### Old View:
Screen Shot 2020-04-01 at 3 24 05 PM

### New View
Screen Shot 2020-04-01 at 3 24 28 PM


solideoglori (approved):
UI changes look good to me

rushtong:
Thank you @solideoglori !

PR #576, opened 2020-04-02
[DUOS-404][risk=no] Fix describe all users call

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-404

Mostly a refactoring PR but necessary in the. multi-DAC world.

## Changes
* Refactor call from api to service
* Refactor query to get completed from profile
* Refactor user mapper so we don't add null user roles
* Additional tests
* Remove Moran from the default list of admin users while testing
  * This required a test update since there was an admin user set up by default.

### DIFF
--- a/src/main/resources/changesets/changelog-consent-10.0.xml
+++ b/src/main/resources/changesets/changelog-consent-10.0.xml
@@ -7,16 +7,6 @@
     <!--Admin user-->
     <changeSet author="vvicario" id="41">
         <validCheckSum>ANY</validCheckSum>
-        <insert tableName="dacuser">
-            <column name="dacuserid" value="3333"/>
-            <column name="displayname" value="Moran Cabili-Kalmar"/>
-            <column name="email" value="nmcabili@broadinstitute.org"/>
-            <column name="createdate" value="2015-08-21 13:58:50"/>
-        </insert>
-        <insert tableName="user_role">
-            <column name="roleid" value="4"/>
-            <column name="dacuserid" value="3333"/>
-        </insert>
     </changeSet>
 
 </databaseChangeLog>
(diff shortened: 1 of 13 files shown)


src/test/java/org/broadinstitute/consent/http/db/UserDAOTest.java
```diff
@@ -47,38 +55,38 @@ public void testDescribeUsersByRole() {
...
+        assertFalse(chairs.isEmpty());
 
         // Only case where we don't set up users by default.
         List<DACUser> alumni = userDAO.describeUsersByRole(UserRoles.ALUMNI.getRoleName());
-        Assert.assertTrue(alumni.isEmpty());
+        assertTrue(alumni.isEmpty());
 
         List<DACUser> admins = userDAO.describeUsersByRole(UserRoles.ADMIN.getRoleName());
-        Assert.assertFalse(admins.isEmpty());
+        assertTrue(admins.isEmpty());
```
rushtong, line 68: This change was necessary after removing Moran from `changelog-consent-10.0.xml`

src/main/resources/changesets/changelog-consent-10.0.xml
```diff
@@ -7,16 +7,6 @@
...
-        <insert tableName="dacuser">
-            <column name="dacuserid" value="3333"/>
-            <column name="displayname" value="Moran Cabili-Kalmar"/>
-            <column name="email" value="nmcabili@broadinstitute.org"/>
-            <column name="createdate" value="2015-08-21 13:58:50"/>
-        </insert>
-        <insert tableName="user_role">
-            <column name="roleid" value="4"/>
-            <column name="dacuserid" value="3333"/>
-        </insert>
```
rushtong, line 19: This was unnecessary ... I think it was used when we needed an admin for fiab testing.

PR #584, opened 2020-04-07
[DUOS-632][risk=no] Handle some real and some potential NPEs

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-632

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/VoteService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/VoteService.java
@@ -17,6 +17,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Date;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/consent/http/service/VoteService.java
```diff
@@ -177,13 +178,15 @@ private boolean isDacChairPerson(Dac dac, DACUser user) {
         if (dac != null) {
             return user.getRoles().
                     stream().
-                    anyMatch(userRole ->
+                    anyMatch(userRole -> Objects.nonNull(userRole.getRoleId()) &&
+                            Objects.nonNull(userRole.getDacId()) &&
                             userRole.getRoleId().equals(UserRoles.CHAIRPERSON.getRoleId()) &&
-                                    userRole.getDacId().equals(dac.getDacId()));
+                            userRole.getDacId().equals(dac.getDacId()));
```
rushtong, line 184: This block was the trigger for the problem we saw in staging. In the failure case, the user had a chairperson role with no associated dac. That threw  an NPE and failed to create votes, but left the user in a state where they were a valid dac member, but didn't have a vote to display in the UI.

PR #691, opened 2020-07-30
[DUOS-388][risk=no] Don't delete DAA on exception

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-388

Also, one minor dataset query update for consistency.

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
@@ -100,7 +100,7 @@ public interface DataSetDAO extends Transactional<DataSetDAO> {
     Set<DataSetDTO> findDataSetWithPropertiesByDataSetId(@Bind("dataSetId") Integer dataSetId);
 
     @UseRowMapper(DataSetPropertiesMapper.class)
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -100,7 +100,7 @@
     Set<DataSetDTO> findDataSetWithPropertiesByDataSetId(@Bind("dataSetId") Integer dataSetId);
 
     @UseRowMapper(DataSetPropertiesMapper.class)
-    @SqlQuery(" select d.*, k.key, dp.propertyValue, ca.consentId , c.translatedUseRestriction from dataset  d inner join datasetproperty dp on dp.dataSetId = d.dataSetId " +
+    @SqlQuery(" select d.*, k.key, dp.propertyValue, ca.consentId, c.dac_id, c.translatedUseRestriction from dataset  d inner join datasetproperty dp on dp.dataSetId = d.dataSetId " +
```
rushtong, line 103: Technically unrelated to the story, but an easy fix to keep the query's consistent.

PR #692, opened 2020-08-04
[no-jira][risk=no] Handle null dataset details

rushtong:
## Addresses
Handles the case seen only on dev where we have DARs with invalid null data.


rushtong:
Merging to unblock dev

PR #693, opened 2020-08-04
[DUOS-684][risk=no] Remove old help APIs, models, emails

rushtong:
## Addresses
Partially addresses https://broadinstitute.atlassian.net/browse/DUOS-684
See also: https://github.com/DataBiosphere/duos-ui/pull/545


lindojong (commented):
Took a quick look at the branch and found `src\main\resources\freemarker\new-help-report.html`  and some mentions of of the consent.help_report table in `migration\postgres`. The html template message could also probably be removed safely. Asking more for my own knowledge but do the references to the db table still need to exist until those are removed too?

rushtong:
@lindojong Good catch on the template, I'll remove that. For the db tables, I broke that into a separate story: https://broadinstitute.atlassian.net/browse/DUOS-715

lindojong (approved):

PR #694, opened 2020-08-05
[no-jira][risk=no] Handle null election id

rushtong:
## Addresses
NPE when getting a DUL for a consent that doesn't have an election.
Staging: https://sentry.io/organizations/broad-institute/issues/1825784751

The UI calls this API method with and without an election id, update API and docs to handle that appropriately.


lindojong (approved):
LGTM!

PR #698, opened 2020-08-11
[DUOS-660][risk=no] V2 endpoint for create/update/get DAR

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-660
Requires: https://github.com/DataBiosphere/duos-ui/pull/562
Recommended that you hide white space changes. There is a good amount of reformatting in some of the classes.

## Changes
* v2 POST for new submitted DAR
* v2 GET for existing submitted DAR
* v2 PUT for existing submitted DAR
* Simplified, streamlined DAR creation
* Simplified DAR output
  * Top level and legacy values are merged into single object
  * Deprecated terms not returned
  * Null values not returned
* New field support
  * Rename `datasetId` -> `datasetIds`
    * Also, update serialized names such that a client can specify the old form of the field name and it will still deserialize into the new format correctly
  * Integer userId (top level field)
  * Date createDate (top level field)
  * Date updateDate (top level field)
  * Date submissionDate (top level field)
  * Boolean cloudUse
  * Boolean anvilUse
  * String cloudProvider
  * String cloudProviderType
  * Boolean geneticStudiesOnly
  * Boolean irb
  * String irbDocumentLocation
  * String irbProtocolExpiration
  * Boolean publication
  * Boolean collaboration
  * String collaborationLetterLocation
  * Boolean forensicActivities
  * Boolean sharingDistribution

## Example Data
Post this:
```
{
  "datasetIds": [385],
  "projectTitle": "Minimal DAR Submission",
  "hmb": true,
  "rus": "Minimal DAR Submission",
  "nonTechRus": "Minimal DAR Submission"
}
```
And get back a list of created DARs (there will be N dars created, 1 per dataset id)
```
[
  {
    "updateDate": 1597363200000,
    "sortDate": 1597363200000,
    "submissionDate": 1597363200000,
    "id": 312,
    "referenceId": "02e662e4-0b70-48af-bbd4-d5afc1df1e59",
    "createDate": 1597363200000,
    "projectTitle": "Minimal DAR Submission",
    "rus": "Minimal DAR Submission",
    "nonTechRus": "Minimal DAR Submission",
    "hmb": true,
    "darCode": "DAR-259",
    "datasetIds": [
      385
    ]
  }
]
```

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
@@ -1,5 +1,6 @@
 package org.broadinstitute.consent.http.db;
 
+import java.util.Date;
 import org.broadinstitute.consent.http.db.mapper.DataAccessRequestMapper;
 import org.broadinstitute.consent.http.models.DataAccessRequest;
 import org.broadinstitute.consent.http.models.DataAccessRequestData;
@@ -26,7 +27,7 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
      * Find all non-draft/partial DataAccessRequests
      * @return List<DataAccessRequest>
      */
-    @SqlQuery("SELECT id, reference_id, draft, (data #>> '{}')::jsonb AS data FROM data_access_request " +
+    @SqlQuery("SELECT id, reference_id, draft, user_id, create_date, sort_date, (data #>> '{}')::jsonb AS data FROM data_access_request " +
             "  WHERE not (data #>> '{}')::jsonb ??| array['partial_dar_code', 'partialDarCode'] " +
             "  AND draft != true ")
     List<DataAccessRequest> findAllDataAccessRequests();
@@ -35,7 +36,7 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
      * Find all draft/partial DataAccessRequests, sorted descending order
      * @return List<DataAccessRequest>
      */
-    @SqlQuery("SELECT id, reference_id, draft, (data #>> '{}')::jsonb AS data FROM data_access_request " +
+    @SqlQuery("SELECT id, reference_id, draft, user_id, create_date, sort_date, (data #>> '{}')::jsonb AS data FROM data_access_request " +
             "  WHERE (data #>> '{}')::jsonb ??| array['partial_dar_code', 'partialDarCode'] " +
             "  OR draft = true " +
             "  ORDER BY ((data #>> '{}')::jsonb->>'sortDate')::numeric DESC")
@@ -45,7 +46,7 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
      * Find all draft/partial DataAccessRequests by user id, sorted descending order
      * @return List<DataAccessRequest>
      */
-    @SqlQuery("SELECT id, reference_id, draft, (data #>> '{}')::jsonb AS data FROM data_access_request " +
+    @SqlQuery("SELECT id, reference_id, draft, user_id, create_date, sort_date, (data #>> '{}')::jsonb AS data FROM data_access_request " +
             "  WHERE ( (data #>> '{}')::jsonb ??| array['partial_dar_code', 'partialDarCode'] " +
             "          OR draft = true ) " +
             "  AND ((data #>> '{}')::jsonb->>'userId')::numeric = :userId " +
@@ -57,7 +58,7 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
      * @param referenceId String
      * @return DataAccessRequest
      */
-    @SqlQuery("SELECT id, reference_id, draft, (data #>> '{}')::jsonb AS data FROM data_access_request WHERE reference_id = :referenceId limit 1")
+    @SqlQuery("SELECT id, reference_id, draft, user_id, create_date, sort_date, (data #>> '{}')::jsonb AS data FROM data_access_request WHERE reference_id = :referenceId limit 1")
     DataAccessRequest findByReferenceId(@Bind("referenceId") String referenceId);
 
     /**
@@ -65,7 +66,7 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
      * @param referenceIds List of Strings
      * @return List<DataAccessRequest>
      */
-    @SqlQuery("SELECT id, reference_id, draft, (data #>> '{}')::jsonb AS data FROM data_access_request WHERE reference_id IN (<referenceIds>)")
+    @SqlQuery("SELECT id, reference_id, draft, user_id, create_date, sort_date, (data #>> '{}')::jsonb AS data FROM data_access_request WHERE reference_id IN (<referenceIds>)")
     List<DataAccessRequest> findByReferenceIds(@BindList("referenceIds") List<String> referenceIds);
 
     /**
@@ -76,6 +77,24 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
     @RegisterArgumentFactory(JsonArgumentFactory.class)
     @SqlUpdate("UPDATE data_access_request SET data = to_jsonb(:data) WHERE reference_id = :referenceId")
     void updateDataByReferenceId(@Bind("referenceId") String referenceId, @Bind("data") @Json DataAccessRequestData data);
+
+    /**
+     * Update DataAccessRequest by reference id and provided DataAccessRequestData
+     * @param referenceId String
+     * @param userId Integer
+     * @param createDate Date
+     * @param sortDate Date
+     * @param data DataAccessRequestData
+     */
+    @RegisterArgumentFactory(JsonArgumentFactory.class)
+    @SqlUpdate("UPDATE data_access_request SET data = to_jsonb(:data), user_id = :userId, create_date = :createDate, sort_date = :sortDate WHERE reference_id = :referenceId")
+    void updateDataByReferenceIdVersion2(
+        @Bind("referenceId") String referenceId,
+        @Bind("userId") Integer userId,
+        @Bind("createDate") Date createDate,
+        @Bind("sortDate") Date sortDate,
+        @Bind("data") @Json DataAccessRequestData data
+    );
 
     /**
      * Delete DataAccessRequest by reference id
@@ -92,6 +111,23 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
     @RegisterArgumentFactory(JsonArgumentFactory.class)
     @SqlUpdate("INSERT INTO data_access_request (reference_id, data) VALUES (:referenceId, to_jsonb(:data)) ")
     void insert(@Bind("referenceId") String referenceId, @Bind("data") @Json DataAccessRequestData data);
+
+    /**
+     * Insert DataAccessRequest by reference id and provided DataAccessRequestData
+     * @param referenceId String
+     * @param userId Integer
+     * @param createDate Date
+     * @param sortDate Date
+     * @param data DataAccessRequestData
+     */
+    @RegisterArgumentFactory(JsonArgumentFactory.class)
+    @SqlUpdate("INSERT INTO data_access_request (reference_id, user_id, create_date, sort_date, data) VALUES (:referenceId, :userId, :createDate, :sortDate, to_jsonb(:data)) ")
+    void insertVersion2(
+        @Bind("referenceId") String referenceId,
+        @Bind("userId") Integer userId,
+        @Bind("createDate") Date createDate,
+        @Bind("sortDate") Date sortDate,
+        @Bind("data") @Json DataAccessRequestData data);
 
     /**
      * Insert DataAccessRequest by reference id and provided DataAccessRequestData
--- a/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
@@ -1,6 +1,7 @@
 package org.broadinstitute.consent.http.db;
 
 import java.util.Date;
+import java.util.List;
 import org.broadinstitute.consent.http.db.mapper.DataAccessRequestMapper;
 import org.broadinstitute.consent.http.models.DataAccessRequest;
 import org.broadinstitute.consent.http.models.DataAccessRequestData;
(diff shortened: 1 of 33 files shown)


src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -13,100 +15,165 @@
...
-            "  AND draft != true ")
-    List<DataAccessRequest> findAllDataAccessRequests();
+  /**
+   * Find all non-draft/partial DataAccessRequests
+   *
+   * @return List<DataAccessRequest>
+   */
+  @SqlQuery(
+      "SELECT id, reference_id, draft, user_id, create_date, sort_date, submission_date, update_date, (data #>> '{}')::jsonb AS data FROM data_access_request "
+          + "  WHERE not (data #>> '{}')::jsonb ??| array['partial_dar_code', 'partialDarCode'] "
```
JVThomas, line 33: Just asking for clarification, is the WHERE statement  looking for records whose data column does not contain either ```partial_dar_code``` or ```partialDarCode``` in the jsonb data?

rushtong: Exactly. Now that I look at this, I'm not sure why `??|` works. Table [9.41 in the docs](https://www.postgresql.org/docs/9.4/functions-json.html) shows `?|`. There had to be a SO post that I pulled this syntax from.

rushtong: Oops, wrong link ... should be version 11: https://www.postgresql.org/docs/11/functions-json.html

rushtong: Aha!!! I remember now. The double `??` is an artifact of jdbi3: https://jdbi.org/#_postgresql where some operators need to be escaped.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -13,100 +15,165 @@
...
-    /**
-     * Update DataAccessRequest by reference id and provided DataAccessRequestData
-     * @param referenceId String
-     * @param data DataAccessRequestData
-     */
-    @RegisterArgumentFactory(JsonArgumentFactory.class)
-    @SqlUpdate("UPDATE data_access_request SET data = to_jsonb(:data) WHERE reference_id = :referenceId")
-    void updateDataByReferenceId(@Bind("referenceId") String referenceId, @Bind("data") @Json DataAccessRequestData data);
+  /**
+   * Update DataAccessRequest by reference id and provided DataAccessRequestData
```
JVThomas: This is me nitpicking a bit, but could you expand on this comment a bit so it distinguishes the difference between this and the version 1 variant?

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -13,100 +15,165 @@
...
-    /**
-     * Insert DataAccessRequest by reference id and provided DataAccessRequestData
-     * @param referenceId String
-     * @param data DataAccessRequestData
-     */
-    @RegisterArgumentFactory(JsonArgumentFactory.class)
-    @SqlUpdate("INSERT INTO data_access_request (reference_id, draft, data) VALUES (:referenceId, true, to_jsonb(:data)) ")
-    void insertDraft(@Bind("referenceId") String referenceId, @Bind("data") @Json DataAccessRequestData data);
+  /**
+   * Insert DataAccessRequest by reference id and provided DataAccessRequestData
```
JVThomas: Similar nitpick to the comment regarding Update version 2

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -13,100 +15,165 @@
...
-            "  ORDER BY ((data #>> '{}')::jsonb->>'sortDate')::numeric DESC")
-    List<DataAccessRequest> findAllDraftsByUserId(@Bind("userId") Integer userId);
+  /**
+   * Find DataAccessRequest by reference id
+   *
+   * @param referenceId String
+   * @return DataAccessRequest
+   */
+  @SqlQuery(
+      "SELECT id, reference_id, draft, user_id, create_date, sort_date, submission_date, update_date, (data #>> '{}')::jsonb AS data FROM data_access_request WHERE reference_id = :referenceId limit 1")
```
JVThomas, line 69: Quick question: are reference ids not unique in the db?

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
JVThomas: Oh I see. There's an operator collision for ```?``` so you have to use ```??``` instead. I was going to ask you about that next, but you've pretty much anticipated it.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
```diff
@@ -1,54 +1,138 @@
...
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
 
 public class DataAccessRequest {
 
-    @JsonProperty
-    public Integer id;
+  private static final List<String> DEPRECATED_PROPS = Arrays.asList("referenceId", "investigator",
+      "institution", "department", "address1", "city", "zipcode", "zipCode", "state", "country",
```
JVThomas: Wasn't there an ```address2``` property as well?

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
```diff
@@ -1,54 +1,138 @@
...
 
-    public Integer getId() {
-        return id;
-    }
+  @JsonProperty public Boolean draft;
 
-    public void setId(Integer id) {
-        this.id = id;
-    }
+  @JsonProperty public Integer userId;
```
JVThomas, line 27: ```userId``` is listed within ```DEPRECATED_PROPS```. Should it be assigned as a JsonProperty (or vice versa)?

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequestData.java
```diff
@@ -2,92 +2,103 @@
...
+    private String nonTechRus;
+    private Boolean diseases;
+    private Boolean methods;
+    private Boolean controls;
+    private Boolean population;
+    private Boolean other;
+    private String otherText;
+    private List<OntologyEntry> ontologies;
+    private Boolean forProfit;
+    @SerializedName(value = "oneGender", alternate = "onegender")
```
JVThomas, line 68: Nothing wrong here, just wanted to mention that I support these new variable names and would like to do the same on the front end.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
rushtong: They are unique. I'm anticipating that you'll ask ... why not just use the id? They also have a bit of history. When they were mongo objects, they were created by converting the hex `_id` to a string. Since moving to postgres, I've changed that to using UUIDs. We also generate consents with UUIDs, and we call those reference ids. In an election, since an election could be on different entities, we use the reference id instead of the database id.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
rushtong: Good catch, will add that

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
rushtong: The deprecated props are deprecated from `DataAccessRequestData`. I think I should move the deprecated json props to that object instead to avoid confusion.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequestData.java
rushtong: Yes, this is where I think I may have introduced a bug in the UI. If the `darInfo` object is expecting `onegender` vs `oneGender` (same concern with some other fields), that field may never populate. It should be an easy fix since I refactored everything to look at the same `darInfo` obhect a while back.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -375,6 +379,87 @@ public DataAccessRequest cancelDataAccessRequest(String referenceId) {
...
+     * @param dataAccessRequest DataAccessRequest with populated DAR data
+     * @return List of created DARs.
+     */
+    public List<DataAccessRequest> createDataAccessRequest(User user, DataAccessRequest dataAccessRequest) {
+        Date now = new Date();
+        List<DataAccessRequest> newDARList = new ArrayList<>();
+        DataAccessRequestData darData = dataAccessRequest.getData();
+        darData.setPartialDarCode(null);
+        darData.setUserId(user.getDacUserId());
+        if (Objects.isNull(darData.getCreateDate())) {
```
JVThomas, line 399: Minor nitpick, it might be a better idea to save now.getTime() to a variable and then assign that as an argument to setCreateDate and setSortDate as opposed to passing in two separate calls of getTime()

JVThomas: Also another question: what does sortDate represent? The naming isn't really clear to me, from what I've read from the code it acts like an update timestamp, but I'm not sure

src/main/resources/assets/api-docs.yaml
```diff
@@ -2959,110 +3032,149 @@ definitions:
...
+      createDate:
         type: string
-        description: 'Describes the principal address.'
-      address2:
+        description: Creation Date
+      sortDate:
         type: string
-        description: 'Describes the secondary address. Optional.'
-      city:
+        description: Sort Date
```
JVThomas, line 3047: Another nitpick, I'd like to have a more expanded description for the dates in the docs. Jonathan and I didn't know what exactly some of the dates represented in our last retro meeting and I expect similar problems in the future. Having it laid out in the docs would be a good way to provide context

JVThomas: It doesn't have to be longer than one or two sentences

src/main/resources/assets/api-docs.yaml
```diff
@@ -2959,110 +3032,149 @@ definitions:
...
+        description: Describes if study is gender specific
+      gender:
+        type: string
+        description: F/M if oneGender is specified
       pediatric:
         type: boolean
-        description: 'Describes if study is restricted to a pediatric population.'
-      illegalbehave:
+        description: Describes if study is restricted to a pediatric population
+      illegalbehavior:
```
JVThomas: The variable was serialized earlier in camelCase terms, should that be the case here as well?

src/main/resources/assets/api-docs.yaml
```diff
@@ -2959,110 +3032,149 @@ definitions:
...
         type: boolean
-        description: 'If TRUE, triggers MANUAL REVIEW.'
-      sexualdiseases:
+        description: If TRUE, triggers MANUAL REVIEW
+      sexualDiseases:
         type: boolean
-        description: 'If TRUE, triggers MANUAL REVIEW.'
-      stigmatizediseases:
+        description: If TRUE, triggers MANUAL REVIEW
+      stigmatizeDiseases:
```
JVThomas: The variable was serialized as ```stigmatizedDiseases```, but it's written ```stigmatizeDiseases``` here (missing the 'd' for the past-tense), should that be the case here?

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
rushtong: I'm adding some comments around the sort date. For a preview:
```
  /**
   * Legacy property on DARs. Used to display the sort order for a DAR.
   * In practice, this also functions as the Update Date.
   * See also https://broadinstitute.atlassian.net/browse/DUOS-728
   */
  @JsonProperty public Date sortDate;
```

src/main/resources/assets/api-docs.yaml
rushtong: I'm also waffling on the format. I think I'll change them to longs so the UI can work with them more easily.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
JVThomas: I have another question. If we're updating a pre-existing record, shouldn't we fetch the original record, update the data attributes individually with the incoming payload (like Object.assign or _.assign in JS), and then save the updated record? Right now the endpoint just takes the incoming json data and assigns it to the record, so if the payload is just ```{projectTitle: <string>}```, it can potentially update a fully filled DAR to a DAR that only has a projectTitle assigned.

rushtong: With a `PUT` request, that's the intention. What you're describing here is `PATCH`.

JVThomas: Right, shouldn't this be a PATCH request?

rushtong: We don't currently support that method. We could support that, but it's not what the UI is expecting right now (it does a PUT). PATCH should be a separate story.

rushtong: Will file a story so we can look into this more but we need to be restrictive. Many fields on submitted DARs should not be updatable (i.e. datasetIds, etc.).

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
rushtong: Done.

src/main/resources/assets/api-docs.yaml
rushtong: Done, and changed dates in json to number.

JVThomas (dismissed):
Looks good to me!

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequestData.java
```diff
@@ -2,92 +2,119 @@
...
+    private List<DatasetDetailEntry> datasetDetail;
+
+    private Boolean anvilUse;
+    private Boolean cloudUse;
+    private String cloudProvider;
+    private String cloudProviderType;
+    private Boolean geneticStudiesOnly;
+    private Boolean irb;
+    private String irbDocumentLocation;
+    private String irbProtocolExpiration;
```
rushtong, line 113: Need a document name field too

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequestData.java
```diff
@@ -2,92 +2,119 @@
...
+    private Boolean publication;
+    private Boolean collaboration;
+    private String collaborationLetterLocation;
```
rushtong, line 116: Need a document name here too

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
rushtong: See https://broadinstitute.atlassian.net/browse/DUOS-734 for this story.

src/main/java/org/broadinstitute/consent/http/service/DataAccessParser.java
```diff
@@ -138,7 +139,9 @@ PDAcroForm fillDARForm(Document dar, Map<String, String> researcherProperties, U
                     field.setValue(getDefaultValue(dar.getString(DarConstants.RUS)));
                     break;
                 }
-                case DarConstants.NON_TECH_RUS: {
+                case DarConstants.NON_TECH_RUS:
+                    // Handle legacy "non_tech_rush"
+                case "non_tech_rush": {
```
JVThomas: This isn't a typo? The legacy variable name was ```non_tech_rush``` and not ```non_tech_rus```?

src/main/java/org/broadinstitute/consent/http/service/DataAccessParser.java
```diff
@@ -163,35 +168,49 @@ PDAcroForm fillDARForm(Document dar, Map<String, String> researcherProperties, U
                     field.setValue(getYesOrNoValue(dar.getBoolean(DarConstants.PEDIATRIC)));
                     break;
                 }
-                case DarConstants.ILLEGAL_BEHAVE: {
+                case DarConstants.ILLEGAL_BEHAVE:
+                    // Handle legacy "onegender"
```
JVThomas: Need to adjust the comment for illegalBehavior

rushtong: Yeah, I'm reviewing myself and just caught a couple of these!

src/main/java/org/broadinstitute/consent/http/service/DatabaseMatchingServiceAPI.java
```diff
@@ -69,17 +69,20 @@ public Match findSingleMatch(String consentId, String purposeId) {
...
     public Match findMatchForPurpose(String purposeId){
-        Document dar = dataAccessAPI.describeDataAccessRequestById(purposeId);
-        Consent consent = findRelatedConsents(purposeId);
         Match match = null;
-        if(dar != null){
+        Document dar = dataAccessAPI.describeDataAccessRequestById(purposeId);
+        if (Objects.nonNull(dar)) {
+            List<Integer> dataSetIdList = DarUtil.getIntegerList(dar, DarConstants.DATASET_ID);
+            Consent consent = findRelatedConsent(dataSetIdList);
+            if (Objects.nonNull(consent)) {
```
rushtong, line 77: Null checks necessary to catch a bug where this would fail and the DAR creation process was interrupted. Failures here are not an exception and should not stop DAR creation.

src/main/java/org/broadinstitute/consent/http/service/DataAccessParser.java
rushtong: This was a typo too, good catch.

JVThomas:
One quick comment, should the partial DARs have timestamps (create, sort, update)?

rushtong:
Partial DARs will have the same fields as full DARs since they are in reality, full DARs with just a flag for draft status or not. This PR wouldn't change that. The next PR will need to take this into account since we have a separate API for updating partial DARs.

PR #700, opened 2020-08-13
[DUOS-720][risk=no] Putting up a new test endpoint

lindojong:
As part of https://broadinstitute.atlassian.net/browse/DUOS-720, this PR addressed some preliminary setup in Consent to prepare for the new endpoint. Created a new DataSetService and a new method in DataSetResource that will eventually take in the form data from the UI to create a new dataset. For now, it simply returns a 200 OK response at `/api/dataset/test` and is included in the Swagger docs.

Future PRs for this ticket will include:
- actually creating an instance of a Dataset given valid input
- try/catch and error logging if something goes wrong
- adding and associating a set of DatasetProperties for the new Dataset
- renaming the endpoint and reducing permissions down to only Admin + Chairperson

In all cases:


src/main/java/org/broadinstitute/consent/http/service/DataSetService.java
```diff
@@ -0,0 +1,30 @@
+package org.broadinstitute.consent.http.service;
+
+import org.broadinstitute.consent.http.db.DataSetDAO;
+
+import javax.inject.Inject;
+import java.util.Date;
+
+
+public class DataSetService {
```
rushtong: Nitpick: Can you please name this `DatasetService`? The mixed casing of `DataSet` has always bugged me.

src/main/resources/assets/api-docs.yaml
```diff
@@ -1863,6 +1863,19 @@ paths:
...
+      description: Creates the Dataset from JSON (WIP)
+      consumes:
+        - application/json
+      tags:
+        - Datasets
+      responses:
+        200:
+          description: Successfully created Dataset
+          schema:
+            $ref: '#/definitions/Dataset'
```
rushtong, line 1884: This should be a 201 created response. 
Also, we'll need to document a 400 here for when the JSON is not valid, and we'll also need a 500 for any other error.

src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -57,16 +58,28 @@
...
     }
 
+    @POST
+    @Consumes("application/json")
+    @Produces("application/json")
+    @Path("/test")
+    @PermitAll
+    public Response createDataSet(@Context UriInfo info, DataSet dataSet) {
+//        Integer status = dataSetService.createDataSet(dataSet.getName(), null, dataSet.getObjectId(), dataSet.getActive(), dataSet.getAlias());
+        return Response.ok().build();
```
rushtong: We should return a 201 created here instead of a 200/ok

src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -57,16 +58,28 @@
...
+        this.dataSetService = dataSetService;
         this.userService = userService;
     }
 
+    @POST
+    @Consumes("application/json")
+    @Produces("application/json")
+    @Path("/test")
+    @PermitAll
+    public Response createDataSet(@Context UriInfo info, DataSet dataSet) {
```
rushtong: One thing I want all of our API endpoints to do is to take `@Auth AuthUser` as an argument. That makes the currently authenticated user available to the endpoint. I would argue that we don't need the `UriInfo` in general. If we return a full object that has all of the required information in it, then the client can do with that what they wish.

rushtong: Additionally, for this method, we should be passing in a String instead and then deserializing that string inside the method. What this does is rely on Jackson to deserialize whatever object is passed in as a `DataSet`. This pattern is very difficult to debug and I find Jackson to be a huge pain and general security risk. For that reason, it should be standard practice for us to manage deserialization ourselves.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -64,7 +68,7 @@ public void setUp() {
     private void initResource() {
         when(AbstractDataSetAPI.getInstance()).thenReturn(api);
         when(AbstractDataAccessRequestAPI.getInstance()).thenReturn(dataAccessRequestAPI);
-        resource = new DataSetResource(userService);
+        resource = new DataSetResource(dataSetService, userService);
     }
```
rushtong, line 83: For the test, I like to have a very simple test to ensure that the response we get is a 201. Something as simple as 
```
    @Mock
    UriInfo info;

    @Mock
    UriBuilder builder;

    @Test
    public void testCreateDataset() {
        when(builder.build()).thenReturn(URI.create("https://test.domain.org/some/path"));
        when(info.getRequestUriBuilder()).thenReturn(builder);
        initResource();
        Response response = resource.createDataSet(info, "");
        assertEquals(201, response.getStatus());
    }
```
should be enough for us to get a good pattern started so we can layer in other tests that look for failure conditions later on down the line.

rushtong (changes requested):
Please see comments inline. Most of this is generic style stuff for our code base. We'll all get the hang of it soon enough!

src/main/java/org/broadinstitute/consent/http/service/DataSetService.java
lindojong: Noted! To be honest I actually prefer Dataset as well. Would it make sense to refactor DataSet -> Dataset wherever touched moving forward?

rushtong: Yes, please!!! We love pure refactoring PRs too because they should never break any functionality.

src/main/resources/assets/api-docs.yaml
JVThomas: As an aside, MDN is generally a good source of info for web development. They have an organized list of http status codes here.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

I've always kept it bookmarked as a good reference

src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -57,16 +58,28 @@
...
         this.userService = userService;
     }
 
+    @POST
+    @Consumes("application/json")
+    @Produces("application/json")
+    @Path("/test")
+    @PermitAll
+    public Response createDataset(@Auth AuthUser user, String json) {
+        DataSet dataset = datasetService.createTestDataSet(json);
```
lindojong, line 79: In order to pass codacy, I had to actually "use" the service or make it a local variable. I opted to create a mostly empty dataset object (the string here is assigned as its objectId), and return that dataset in the response body (which we are aiming to do by the end with the actual dataset).

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -0,0 +1,35 @@
...
+    @Inject
+    public DatasetService(DataSetDAO dataSetDAO) {
+        this.dataSetDAO = dataSetDAO;
+    }
+
+    public DataSet createTestDataSet(String json) {
+        return new DataSet(json);
+    }
+
+    public Integer createDataset(String name, String objectId, Boolean active, Integer alias) {
```
lindojong, line 23: left this here so that the DatasetService class would use the DataSetDAO it contains (without flagging codacy)

src/main/resources/assets/api-docs.yaml
```diff
@@ -1863,6 +1863,33 @@ paths:
             type: array
             items:
               type: string
+  /api/dataset/test:
+    post:
+      summary: Creates the Dataset from JSON
+      description: Creates the Dataset from JSON (WIP)
+      consumes:
+        - application/json
+      parameters:
```
lindojong, line 1872: Added params so that you can test it when the server is running and see the mostly empty Dataset returned in a 201 response

src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
rushtong: Nice idea!

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
rushtong: 👍 For final implementation, we'll return a full dataset. This is good for this PR.

rushtong (approved):
Looks great!

PR #705, opened 2020-08-19
[DUOS-720][risk=no] Handle Dataset creation

lindojong:
Last time we created an endpoint that could POST back what you gave it. This time, that same endpoint is fleshed out a bit more and verifies the input JSON somewhat before creating and inserting a new Dataset (and properties) into the database. It checks for a pre-existing dataset with the same name and the existence of a properties field. At the end, the created Dataset and its corresponding properties is returned back in the response body.

Notes:
- The created dataset is given a null objectId and false active status.
- It is assumed that the input will contain any desired property fields and that their values are non-null. I assume that the properties will be given and formatted as an array of {"propertyKey":int, "propertyValue":"string"}. After the dataset is created successfully (and now has an id), the properties are updated with that corresponding datasetId and its createDate, then inserted into the db. 

Sample Input:
`{"name":"new dataset", "properties":[{"propertyKey":2,"propertyValue":"Data Type"},{"propertyKey":3,"propertyValue":"Species"},{"propertyKey":4,"propertyValue":"Phenotype/Indication"},{"propertyKey":5,"propertyValue":"# of participants"},{"propertyKey":6,"propertyValue":"Description"},{"propertyKey":7,"propertyValue":"dbGAP"},{"propertyKey":8,"propertyValue":"Data Depositor"},{"propertyKey":9,"propertyValue":"Principal Investigator(PI)"},{"propertyKey":10,"propertyValue":"Sample Collection ID"},{"propertyKey":11,"propertyValue":"Consent ID"}]}
`

Sample Output:
`{
  "dacId": null,
  "dataSetId": 466,
  "consentId": null,
  "translatedUseRestriction": null,
  "deletable": null,
  "properties": [
    {
      "propertyName": "Dataset Name",
      "propertyValue": "new dataset"
    },
    {
      "propertyName": "Sample Collection ID",
      "propertyValue": "Sample Collection ID"
    },
    {
      "propertyName": "Data Type",
      "propertyValue": "Data Type"
    },
    {
      "propertyName": "Species",
      "propertyValue": "Species"
    },
    {
      "propertyName": "Phenotype/Indication",
      "propertyValue": "Phenotype/Indication"
    },
    {
      "propertyName": "# of participants",
      "propertyValue": "# of participants"
    },
    {
      "propertyName": "Description",
      "propertyValue": "Description"
    },
    {
      "propertyName": "dbGAP",
      "propertyValue": "dbGAP"
    },
    {
      "propertyName": "Data Depositor",
      "propertyValue": "Data Depositor"
    },
    {
      "propertyName": "Principal Investigator(PI)",
      "propertyValue": "Principal Investigator(PI)"
    },
    {
      "propertyName": "Consent ID",
      "propertyValue": "Consent ID"
    }
  ],
  "active": false,
  "needsApproval": false,
  "isAssociatedToDataOwners": null,
  "updateAssociationToDataOwnerAllowed": null,
  "alias": "DUOS-000095",
  "objectId": null
}`

In all cases:


src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -57,16 +60,42 @@
...
+        }
+        if (ds.getName() == null) {
+            throw new BadRequestException("Dataset name is required");
+        }
+        if (ds.getProperties() == null) {
+            throw new BadRequestException("Dataset must contain required properties");
+        }
+        Integer nameId = datasetService.findDatasetByName(ds.getName());
+        if (nameId >= 0) {
+            throw new Exception("Dataset name: " + ds.getName() + " is already in use");
```
lindojong: I preferred checking nameId >= 0 rather than != null, but either works

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -0,0 +1,62 @@
...
+    public DatasetService(DataSetDAO dataSetDAO) {
+        this.dataSetDAO = dataSetDAO;
+    }
+
+    public DataSetDTO createDataset(String name, Set<DataSetProperty> properties) {
+        Date now = new Date();
+        int lastAlias = dataSetDAO.findLastAlias();
+        int alias = lastAlias + 1;
+
+        Integer id = dataSetDAO.insertDataset(name, now, null, false, alias);
```
lindojong: Do we want a dataset created this way to default to active or not?

src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -99,6 +99,13 @@
             "where d.dataSetId = :dataSetId order by d.dataSetId, k.displayOrder")
     Set<DataSetDTO> findDataSetWithPropertiesByDataSetId(@Bind("dataSetId") Integer dataSetId);
 
+  @UseRowMapper(DataSetPropertiesMapper.class)
+  @SqlQuery("select d.*, k.key, dp.propertyValue, ca.consentId, c.dac_id, c.translatedUseRestriction " +
+      "from dataset d left outer join datasetproperty dp on dp.dataSetId = d.dataSetId left outer join dictionary k on k.keyId = dp.propertyKey " +
+      "left outer join consentassociations ca on ca.dataSetId = d.dataSetId left outer join consents c on c.consentId = ca.consentId " +
+      "where d.dataSetId = :dataSetId order by d.dataSetId, k.displayOrder")
```
rushtong: Couple bits of feedback here.
* I like upper cased sql operators - it's easier to read
* Postgres doesn't require mixed case column names - they are all lowercased
* I also like white space to break up the sql to make it more legible
This should look something like this:
```
SELECT d.*, k.key, dp.propertyvalue, ca.consentid, c.dac_id, c.translateduserestriction
FROM dataset d 
LEFT OUTER JOIN datasetproperty dp on dp.datasetid = d.datasetid 
LEFT OUTER JOIN dictionary k on k.keyid = dp.propertykey
LEFT OUTER JOIN consentassociations ca on ca.datasetid = d.datasetid 
LEFT OUTER JOIN consents c on c.consentid = ca.consentid 
WHERE d.datasetid = :datasetId 
ORDER BY d.datasetid, k.displayorder
```

src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -99,6 +99,13 @@
...
+  Set<DataSetDTO> findDataSetWithPropertiesByDataSetIdWithOuterJoins(@Bind("dataSetId") Integer dataSetId);
```
rushtong: nit, casing:
```suggestion
  Set<DataSetDTO> findDataSetWithPropertiesByDataSetIdWithOuterJoins(@Bind("datasetId") Integer datasetId);
```

src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -57,16 +60,42 @@
...
         this.userService = userService;
     }
 
+    @POST
+    @Consumes("application/json")
+    @Produces("application/json")
+    @Path("/test")
+    @PermitAll
+    public Response createDataset(@Auth AuthUser user, String json) {
+        DataSet ds = new Gson().fromJson(json, DataSet.class);
```
rushtong: I think we need to do something a little different here. Since datasets have raw dataset properties, that means your incoming json string has to mimic that structure exactly. This leads to an unfriendly situation where the client needs to know what the property key integer values are. What I'd like to see instead is input that looks almost exactly like the output (minus the things that the system calculates):
```
{
  "active": true, 
  "consentId": "DEV-CG-4412", 
  "dacId": 4, 
  "deletable": null, 
  "isAssociatedToDataOwners": null, 
  "needsApproval": false, 
  "objectId": null, 
  "properties": [
    {
      "propertyName": "Dataset Name", 
      "propertyValue": "TOPMed CCAF_GRU_IRB_phs001189.v1.p1.c1"
    }, 
    {
      "propertyName": "Data Type", 
      "propertyValue": "Whole Genome"
    }, 
    {
      "propertyName": "Species", 
      "propertyValue": "human"
    }, 
    {
      "propertyName": "Phenotype/Indication", 
      "propertyValue": "Atrial Fibrilation"
    }, 
    {
      "propertyName": "# of participants", 
      "propertyValue": "357"
    }, 
    {
      "propertyName": "Description", 
      "propertyValue": "Cleveland Clinic Atrial Fibrillation Study"
    }, 
    {
      "propertyName": "dbGAP", 
      "propertyValue": "https://www.ncbi.nlm.nih.gov/projects/gap/cgi-bin/study.cgi?study_id=phs001189.v1.p1"
    }, 
    {
      "propertyName": "Data Depositor", 
      "propertyValue": "NHLBI"
    }, 
    {
      "propertyName": "Principal Investigator(PI)", 
      "propertyValue": "NHLBI"
    }
  ], 
  "translatedUseRestriction": "Samples are restricted for use under the following conditions:\nData is available for general research use. [GRU]\nCommercial use is not prohibited.\nData use for methods development research irrespective of the specified data use limitations is not prohibited.\nFuture use as a control set for any type of health/medical/biomedical study is not prohibited.\nOther restrictions: IRB Approval Required.", 
  "updateAssociationToDataOwnerAllowed": null
}
```
I think if you deserialize as a `DataSetDTO`, then you could backport that object to a dataset and then save it.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
rushtong: Default active or not will come from the passed in dataset. On the ticket, you'll see a checkbox for public availability ... that's what determines the value of active.

I think a better approach here is to pass in just the full `Dataset` object. That way you can look at for fields and properties that are required and throw an error if they aren't there. You'll find a list of what's required in the ticket. Here, you should only be adding the values that are new or calculated, like the alias, for example. 

Side note, and optional ... I'd also like to know who is doing this. We should be adding in create user to the mix as a top level field on dataset, but let's do that in a separate ticket.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -0,0 +1,64 @@
...
+        Set<DataSetDTO> set = dataSetDAO.findDataSetWithPropertiesByDataSetIdWithOuterJoins(id);
+
+        for (DataSetDTO ds : set) {
+            result = ds;
+        }
+        return result;
+    }
+
+    // return -1 if no ds found
+    public Integer findDatasetByName(String name) {
```
rushtong: This should throw a `NotFoundException` if it can't find it. Callers of this method should understand that and handle it appropriately.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public DataSetDTO createDataset(String name, Set<DataSetProperty> properties) {
+        Date now = new Date();
+        int lastAlias = dataSetDAO.findLastAlias();
+        int alias = lastAlias + 1;
+
+        Integer id = dataSetDAO.insertDataset(name, now, null, false, alias);
+
+        List<DataSetProperty> propertyList = this.processDatasetProperties(id, now, properties);
```
rushtong: nit: don't need `this`

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -0,0 +1,64 @@
...
+    public List<DataSetProperty> processDatasetProperties(Integer datasetId, Date now, Set<DataSetProperty> properties) {
+        List<DataSetProperty> result = new ArrayList<>(11);
+        Iterator<DataSetProperty> iterator = properties.iterator();
+        while (iterator.hasNext()) {
+            DataSetProperty dsp = iterator.next();
+            dsp.setDataSetId(datasetId);
+            dsp.setCreateDate(now);
+            result.add(dsp);
+        }
+        return result;
```
rushtong: This is a bit of a convoluted way of updating values inside of the set of properties. When you find yourself creating empty lists, iterators, and populating them, there's always a cleaner way of doing that functionally. Here is a lambda that does the same thing:
```
return properties.stream()
    .peek(
        p -> {
          p.setDataSetId(datasetId);
          p.setCreateDate(now);
        })
```
You could also use a forEach to do something similar. The nice thing about functional approaches is that we completely bypass all of the extra scaffolding objects required in a procedural approach.

src/main/resources/assets/api-docs.yaml
```diff
@@ -1863,6 +1863,33 @@ paths:
             type: array
             items:
               type: string
+  /api/dataset/test:
```
rushtong: Can we rename this to `/api/dataset/v2`?

src/main/resources/assets/api-docs.yaml
```diff
@@ -1863,6 +1863,33 @@ paths:
...
+    post:
+      summary: Creates the Dataset from JSON
+      description: Creates the Dataset from JSON (WIP)
+      consumes:
+        - application/json
+      parameters:
+        - name: json
+          in: body
+          description: Submitted dataset registration form
```
rushtong:
```suggestion
          description: Dataset in json format
```

rushtong (changes requested):
Great progress! See feedback inline.

src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
rushtong: One more bit of naming feedback. Can we name this something like,
* findDatasetByIdWithProps
* findDatasetDTOById
Or something like that? Naming is always hard, but I'd have a hard time understanding what this really means without examining the sql, which can really slow a developer down when they're looking at lots of these in different daos.

src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -99,6 +99,17 @@
...
+  @UseRowMapper(DataSetPropertiesMapper.class)
+  @SqlQuery("select d.*, k.key, dp.propertyValue, ca.consentId, c.dac_id, c.translatedUseRestriction " +
+      "FROM dataset d " +
+      "LEFT OUTER JOIN datasetproperty dp on dp.dataSetId = d.dataSetId " +
+      "LEFT OUTER JOIN dictionary k on k.keyId = dp.propertyKey " +
+      "LEFT OUTER JOIN consentassociations ca on ca.dataSetId = d.dataSetId " +
+      "LEFT OUTER JOIN consents c on c.consentId = ca.consentId " +
+      "WHERE d.dataSetId = :datasetId " +
+      "ORDER BY d.dataSetId, k.displayOrder")
+  Set<DataSetDTO> findDatasetDTOWithPropsByDatasetId(@Bind("datasetId") Integer datasetId);
```
rushtong: For postgres, we have all lowercase all column names. The mixed casing only works because jdbi3 does the right thing and converts the query appropriately. There was a huge number of headaches associated with the mixed cased names when this was all mysql, and I want to avoid that for the future.
* `propertyValue` -> `propertyvalue`
* `dataSetId` -> `datasetid`
* `consentId` -> `consentid`
* etc.

For variables, we should keep using camel case naming so it's more clear in the query that this is NOT a db field.

src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -99,6 +99,17 @@
             "where d.dataSetId = :dataSetId order by d.dataSetId, k.displayOrder")
     Set<DataSetDTO> findDataSetWithPropertiesByDataSetId(@Bind("dataSetId") Integer dataSetId);
 
+  @UseRowMapper(DataSetPropertiesMapper.class)
+  @SqlQuery("select d.*, k.key, dp.propertyValue, ca.consentId, c.dac_id, c.translatedUseRestriction " +
```
rushtong: Upper case select
```suggestion
  @SqlQuery("SELECT d.*, k.key, dp.propertyValue, ca.consentId, c.dac_id, c.translatedUseRestriction " +
```

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -0,0 +1,73 @@
...
+    Set<DataSetDTO> set = dataSetDAO.findDatasetDTOWithPropsByDatasetId(id);
+
+    for (DataSetDTO ds : set) {
+      result = ds;
+    }
+    return result;
+  }
+
+  // return -1 if no ds found
+  public Integer findDatasetByName(String name) {
```
rushtong: I think I mentioned this before. A service class method that says `find me a dataset by name` should return an actual dataset, not something other than a dataset. There are some options here:
1. Return the object, or null. This is not the best idea.
2. Return a `Optional<DataSet>` which makes it clear in the signature that you might get a dataset, but you might not. This is probably the most flexible/usable option.
3. Throw a `NotFoundException` and let the caller deal with the consequences since it should know what will happen because that exception should be in the signature.

The goal you're shooting for here is the idea of self-documenting code. The class name, method name, signature should be enough for a lazy developer to just know what to expect from reading it. This method should be able to stand on it's own, but to use it as is, I have to dig into the method to understand why I'm getting an ID instead of a dataset. I dug into the dao call you're using, and it's terrible. I have no idea why it does what it does, but that is a terrible pattern. The name doesn't tell you what it really does and you have to read the query to understand it. That slows down development overall. 

This should be the dao call you want:
```
    @SqlQuery("SELECT * FROM dataset WHERE lower(name) = lower(:name)")
    DataSet findDataSetByName(@Bind("name") String name);
```

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+    return result;
+  }
+
+  // assumes that you will receive non-null values for all properties in receiveOrder
+  public List<DataSetProperty> processDatasetProperties(Integer datasetId, Date now,
+      List<DataSetPropertyDTO> properties) {
+    List<Dictionary> dictionaries = dataSetDAO.getMappedFieldsOrderByReceiveOrder();
+    // removes Dataset Name property, which is generated and doesn't need to be manually inserted
+    dictionaries.remove(0);
```
rushtong: This could fail if receive order ever changes. Better to look for the actual property you want instead of guessing the position in the array.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -0,0 +1,73 @@
...
+
+    return properties.stream()
+        .map(
+            p -> {
+              Dictionary dictionary = dictionaries.stream()
+                  .filter(d -> d.getKey().equals(p.getPropertyName()))
+                  .findAny()
+                  .orElse(null);
+              return new DataSetProperty(datasetId, dictionary.getKeyId(), p.getPropertyValue(),
```
rushtong, line 69: `dictionary` can be null here. I think you don't want to do anything if you have a property name that you don't understand. Might be good to even log something if we find an invalid key.

rushtong (changes requested):
Making good progress! See comments inline.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
rushtong: This might be a safer query:
```
    @SqlQuery("SELECT * FROM dataset WHERE lower(name) = lower(:name) LIMIT 1")
    DataSet findDataSetByName(@Bind("name") String name);
```
Another alternative is to just retuen a list of datasets with the same name. Don't know if we have a unique constraint on that column. Lower casing the name also means we could find two datasets with similar names, i.e. `Dataset 1` vs `dATAsET 1`

src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -57,16 +62,43 @@
...
+        String name = ds.getPropertyValue("Dataset Name");
+        if (name == null) {
+            throw new BadRequestException("Dataset name is required");
+        }
+        DataSet datasetNameAlreadyUsed = datasetService.findDatasetByName(name);
+        if (Objects.nonNull(datasetNameAlreadyUsed)) {
+            throw new NotFoundException("Dataset name: " + name + " is already in use");
+        }
+        DataSetDTO dataset = datasetService.createDataset(ds, name);
+        return Response.ok().entity(dataset).status(201).build();
```
rushtong: Optional. If you call the created method, it will have the right status. Unfortunately, it also requires a URI to describe the location of the new dataset. That could be pretty simple though. Add in a `@Context UriInfo info` to the method signature and then change these lines:
```suggestion
        URI uri = info.getRequestUriBuilder().replacePath("api/dataset/{datasetId}").build(ds.getDataSetId());
        return Response.created().entity(dataset).build();
```
Caveat ... untested, but this should work out of the box.

src/main/java/org/broadinstitute/consent/http/service/DatabaseDataSetAPI.java
```diff
@@ -402,7 +402,7 @@ private boolean isValid(List<DataSet> dataSets, boolean overwrite) {
     private void processAssociation(List<DataSet> dataSetList) {
         dataSetList.forEach(dataSet -> {
             if (StringUtils.isNotEmpty(dataSet.getConsentName())) {
-                Integer datasetId = dsDAO.getDataSetByName(dataSet.getName());
+                Integer datasetId = dsDAO.getDatasetIdByName(dataSet.getName());
```
rushtong, line 405: Nice change here and in the DAO.

src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -99,6 +99,17 @@
...
+  @UseRowMapper(DataSetPropertiesMapper.class)
+  @SqlQuery("SELECT d.*, k.key, dp.propertyvalue, ca.consentid, c.dac_id, c.translateduserestriction " +
+      "FROM dataset d " +
+      "LEFT OUTER JOIN datasetproperty dp on dp.datasetid = d.datasetid " +
+      "LEFT OUTER JOIN dictionary k on k.keyid = dp.propertykey " +
+      "LEFT OUTER JOIN consentassociations ca on ca.datasetid = d.datasetid " +
+      "LEFT OUTER JOIN consents c on c.consentid = ca.consentid " +
+      "WHERE d.datasetid = :datasetId " +
+      "ORDER BY d.datasetid, k.displayorder")
+  Set<DataSetDTO> findDatasetDTOWithPropsByDatasetId(@Bind("datasetId") Integer datasetId);
```
rushtong: I'm confused about why we're returning a set here. We know there's only one, why not just this?
```suggestion
  DataSetDTO findDatasetDTOWithPropsByDatasetId(@Bind("datasetId") Integer datasetId);
```

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -0,0 +1,72 @@
...
+
+    List<DataSetProperty> propertyList = processDatasetProperties(id, now, dataset.getProperties());
+    dataSetDAO.insertDataSetsProperties(propertyList);
+
+    DataSetDTO result = new DataSetDTO();
+    Set<DataSetDTO> set = dataSetDAO.findDatasetDTOWithPropsByDatasetId(id);
+
+    for (DataSetDTO ds : set) {
+      result = ds;
+    }
```
rushtong: See my comment in the DAO class that should make this code moot. If you do need to do something like this, though, there's a more functional approach that I like. In this code, we have two intermediate objects, `result` and `set`, where `result` gets overwritten if we find something in the set. This structure has no intermediate objects:
```
    return dataSetDAO
        .findDatasetDTOWithPropsByDatasetId(id)
        .stream()
        .findFirst()
        .orElse(new DataSetDTO());
```

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -0,0 +1,72 @@
...
+    return properties.stream()
+        .filter(p -> !p.getPropertyName().equals("Dataset Name"))
+        .map(
+            p -> {
+              Dictionary dictionary = dictionaries.stream()
+                  .filter(d -> d.getKey().equals(p.getPropertyName()))
+                  .findAny()
+                  .orElse(null);
+              if (Objects.isNull(dictionary)) {
+                throw new NotFoundException("The given property name could not be recognized or associated with a key: " + p.getPropertyName());
```
rushtong: I don't think we want to throw an exception here, that would break the upload of valid properties as well. We could do a couple of things, but it would be more user friendly to validate the properties in the resource class so you can give the client a useful Bad Request exception if they try and send bad data through. At the service class level, we should just ignore anything we don't want to save and perhaps log that case.

src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -57,16 +62,43 @@
...
     }
 
+    @POST
+    @Consumes("application/json")
+    @Produces("application/json")
+    @Path("/v2")
+    @PermitAll
+    public Response createDataset(@Auth AuthUser user, String json) {
+        DataSetDTO ds = new Gson().fromJson(json, DataSetDTO.class);
+        if (ds == null) {
```
rushtong: Optional: I like `Objects.isNull` for these kinds of cases.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -0,0 +1,72 @@
...
+    return dataSetDAO.findDatasetByName(name);
+  }
+
+  public List<DataSetProperty> processDatasetProperties(Integer datasetId, Date now,
+      List<DataSetPropertyDTO> properties) {
+    List<Dictionary> dictionaries = dataSetDAO.getMappedFieldsOrderByReceiveOrder();
+
+    return properties.stream()
+        .filter(p -> !p.getPropertyName().equals("Dataset Name"))
+        .map(
```
rushtong, line 64: Right above this map, you could also filter to be sure the property name is in the list of dictionary keys. That would cut down on the required logic inside the map.

rushtong (changes requested):
Still making good progress! See comments inline.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
lindojong: Sure! Would that look more like a method that only does the checking against the dictionaries for the properties (kind of splitting off the validation here so it can be called before a dataset is created)?

rushtong: That sounds like a good approach.

src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
lindojong: So from what I can tell, the DataSetDTO is somewhat misleading of a name here. What is being returned by the DAO initially is a list of properties that DataSetPropertiesMapper then reformats into what we know as the DataSetDTO. This has to be a set so that we can grab all of the properties for a given dataset, rather than only the first property. It's pretty confusing but I'm not sure how to make it more obvious without separating the fetch of properties and dataset and tying them back together in the service instead? Would that be a better approach to using the mapper?

rushtong: If you look at `DataSetPropertiesMapper`, it creates a map of dataset id to DataSetDTO. It will only return a single DataSetDTO for a dataset id. If you queried on multiple different ids, then you'd get more than one back. The mapper will take all of the properties with the same dataset id, any number of them, and turn them into single DataSetDTO. Give my suggestion a shot and you should see all the resulting rows come back as a single object.

PR #706, opened 2020-08-20
[DUOS-662][risk=no] New V2 POST, PUT for Draft Data Access Requests

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-662
Also partially implements security risk issues found in https://broadinstitute.atlassian.net/browse/DUOS-442, specifically the `PUT` requests for dars and draft dars.

## Changes
* New v2 POST endpoint for draft DARs
* New v2 PUT endpoint for draft DARs
* Updated api docs
* Added external collaborators to DAR fields

### DIFF
--- a/src/main/resources/assets/api-docs.yaml
+++ b/src/main/resources/assets/api-docs.yaml
@@ -1191,6 +1191,10 @@ paths:
           description: Returns a Data Access Request
           schema:
             $ref: '#/definitions/DataAccessRequest'
+        403:
+          description: Updates are restricted to the create user.
+        500:
+          description: Internal Server Error.
   /api/dar/v2/draft:
     post:
       summary: Create Draft Data Access Request
@@ -1230,6 +1234,8 @@ paths:
       responses:
         201:
           description: Returns the upated draft Data Access Request.
+        403:
+          description: Updates are restricted to the create user.
         500:
           description: Internal Server Error.
 
--- a/src/main/resources/assets/api-docs.yaml
+++ b/src/main/resources/assets/api-docs.yaml
@@ -1191,6 +1191,25 @@ paths:
           description: Returns a Data Access Request
           schema:
             $ref: '#/definitions/DataAccessRequest'
+  /api/dar/v2/draft:
+    post:
+      summary: Create Draft Data Access Request
+      description: Create Draft Data Access Request
+      parameters:
+        - name: dar
+          in: body
+          description: The fields that represent a Data Access Request, json format.
+          required: true
+          schema:
+            $ref: '#/definitions/DataAccessRequest'
+      tags:
+        - Data Access Request
+      responses:
+        201:
+          description: Returns the created draft Data Access Request.
+        500:
+          description: Internal Server Error.
+
   /api/dar:
     post:
       summary: createdDataAccessRequest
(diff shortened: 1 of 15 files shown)


src/main/resources/assets/api-docs.yaml
```diff
@@ -3056,6 +3100,25 @@ definitions:
       updateStatus:
         type: boolean
         description: Indicates if a consent has been updated from orsp
+  Collaborator:
+    type: object
+    description: A Collaborator on a Data Access Request
+    properties:
+      approverStatus:
+        type: boolean
+        description: Collaborator approver/downloader status
```
JVThomas: Can you add a note about how approverStatus is relevant to lab staff and not to internal collaborators? The nuance behind approverStatus is not obvious from the name alone.

src/main/resources/assets/api-docs.yaml
```diff
@@ -3205,6 +3268,18 @@ definitions:
       sharingDistribution:
         type: boolean
         description: Study requires sharing distribution
+      internalCollaborators:
+        type: array
+        description: Array of ontology terms
```
JVThomas: You need to update the description here

src/main/resources/assets/api-docs.yaml
```diff
@@ -3205,6 +3268,18 @@ definitions:
...
+        items:
+          type: object
+          $ref: '#/definitions/Collaborator'
+      labCollaborators:
+        type: array
+        description: Array of ontology terms
```
JVThomas: Same issue with description as internalCollaborators

src/main/resources/assets/api-docs.yaml
rushtong: Yes, can do.

src/main/resources/assets/api-docs.yaml
rushtong: Good catch!

src/main/resources/assets/api-docs.yaml
rushtong: Good catch!

lindojong (commented):
Tried out the endpoint locally using the example input data model and noticed two small inconsistencies between doc data model and actual data model: datasetIds as Array of String -> Array of Integer, ontologies as Array of String -> Array of OntologyEntry. Maybe it's worth updating the model definition for DataAccessRequest inside api-docs?

rushtong:
Thank you @lindojong - it is worth updating the docs. Will do that now.

JVThomas (approved):
Dates look good. createDate is only defined on POST and is simply returned on the PUT request, and none of the dates have zeroed out times. Collaborators are saving properly as well.

rushtong:
I'm going to add one more check to the PUTs to ensure the same user is making the update, then call it a day for this PR.

JVThomas:
User validation looks like it's working. Separate issue: a user can send in a payload that changes the reference id value within the json, causing a mismatch between the column and the json attribute. Instead of simply accepting the user input, it'll be a good idea to just assign the value to the json manually server-side by reading the value from the db record itself.

rushtong:
Nice validation catch, will correct that.

rushtong:
@JVThomas Feedback applied, please have another look at your convenience!

JVThomas:
Another nitpick: The CREATE/POST endpoint is reading the reference_id from the json payload and assigning it as a value for both the json and db column. The endpoint should be generating the value itself and assigning that value to those attributes.

rushtong:
@JVThomas Another nice catch. Updated.

JVThomas (approved):
CREATE looks good, it ignores the user supplied json values for reference_id and instead applies its own generated value. PR as a whole looks good to me.

PR #707, opened 2020-08-20
[DUOS-660][risk=no] Fix for string constants and some typos

rushtong:
## Addresses
Some fallout from https://broadinstitute.atlassian.net/browse/DUOS-660

## Changes
* Use constants instead of raw string values which are no longer correct. 
* Fix some typos


lindojong (approved):
LGTM!

PR #718, opened 2020-08-28
[DUOS-720] DatasetService layer

lindojong:
Separates out service layer changes from https://github.com/DataBiosphere/consent/pull/705, with some changes to return a Dataset instead of a DatasetDTO. Also includes tests for the new DatasetService and a new sql query in the dao to return a dataset (rather than just its id) when searching by name.

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
@@ -72,4 +72,13 @@ public class DatasetService {
             .collect(Collectors.toList());
     }
 
+    public List<DataSetPropertyDTO> findInvalidProperties(List<DataSetPropertyDTO> properties) {
+        List<Dictionary> dictionaries = dataSetDAO.getMappedFieldsOrderByReceiveOrder();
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -1,12 +1,16 @@
 package org.broadinstitute.consent.http.service;
 
-import java.util.Set;
+import java.util.*;
```
rushtong: Avoid * imports

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -45,4 +59,25 @@ public DataSet getDatasetWithPropertiesById(Integer datasetId) {
         return dataset;
     }
 
+    public List<DataSetProperty> processDatasetProperties(Integer datasetId, Date createDate, List<DataSetPropertyDTO> properties) {
```
rushtong: Minor suggestion, not required. Drop create date as an argument and instead create a new date inside this method.

rushtong (approved):
Looks great! A couple minor suggestions inline.

PR #727, opened 2020-09-08
[DUOS-695][risk=no] Use the researcher email from the DAR user 

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-695

## Changes
Previously, we looked for the researcher's email in the dar data. Instead, first try ti get the real user from the dar's user id and in case we can't use that, fall back to looking at the dar data as we did previously.


lindojong (approved):
LGTM!

rushtong:
Thanks @lindojong!

PR #729, opened 2020-09-10
[DUOS-747][risk=no] Upgrade api docs

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-747

## Changes
* Updated the swagger resource to to the right string substitution for the latest version of swagger UI
* Significant update to the api-docs for the latest version
* Minor cleanup around some of the schemas

## Note to Reviewers
* Recommended to hide white space changes
* I did not test every API (although I did test many of them) and it is likely that there are some errors. I am making an executive decision that this is an acceptable amount of technical debt and we should be aware of this for future updates and/or new functionality.

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/SwaggerResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/SwaggerResource.java
@@ -140,7 +140,7 @@ public class SwaggerResource {
                 + "        realm: 'Broad Institute',\n"
                 + "        appName: 'Consent',\n"
                 + "        scopeSeparator: ' ',\n"
-                + "        scopes: 'openid profile email',\n"
+                + "        scopes: 'profile email',\n"
                 + "        additionalQueryStringParams: {}\n"
                 + "      })\n"
                 + "      window.ui = ui\n");
--- a/src/main/java/org/broadinstitute/consent/http/resources/SwaggerResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/SwaggerResource.java
@@ -1,9 +1,9 @@
 package org.broadinstitute.consent.http.resources;
 
-import org.apache.commons.lang3.StringUtils;
-import org.broadinstitute.consent.http.configurations.GoogleOAuth2Config;
-import org.parboiled.common.FileUtils;
-
+import java.io.InputStream;
+import java.net.URI;
+import java.util.Properties;
+import java.util.logging.Logger;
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/consent/http/resources/SwaggerResource.java
```diff
@@ -1,138 +1,144 @@
...
+        mediaType = MEDIA_TYPE_GIF;
+        break;
+      default:
+        mediaType = MediaType.TEXT_HTML;
+        break;
     }
+    return mediaType;
+  }
 
+  private String getIndex(String swaggerResource) {
```
lindojong, line 119: Are the \n line breaks necessary here?

lindojong (approved):
I think you did a great job of changing what was necessary to get the docs back up and running again, with all the syntax changes! I'm also enjoying the small QOL change with email + profile scopes being preselected in authorization.

src/main/java/org/broadinstitute/consent/http/resources/SwaggerResource.java
rushtong: Technically, no, they're not required. It was helpful in debugging the rendered javascript though. Getting the configs syntactically correct was a bit of a challenge.

src/main/java/org/broadinstitute/consent/http/resources/SwaggerResource.java
```diff
@@ -1,138 +1,144 @@
...
+    String content = FileUtils.readAllTextFromResource(swaggerResource + "index.html");
+    return content
+        .replace(
+            "url: \"https://petstore.swagger.io/v2/swagger.json\",",
```
JVThomas, line 123: More of a question than a comment: Is there no way to configure Swagger to set these attributes?

rushtong: The swagger ui web jar has a pre-constructed html page with an existing swagger configuration. In the implementation we have, we pull that index.html file out and rewrite it with custom values. This has been the approach we've taken since we started using the web jar. I do think this is the simplest way to get our custom configurations into that html page. We've been doing it this way for quite a few years though, so there certainly could be a better way. Here's how we do this in our scala artifacts: https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala#L226

PR #731, opened 2020-09-10
[DUOS-663][risk=high] DAR Document upload/download

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-663

## Changes
* New endpoints for irb and collaboration document uploads
* New GCS storage methods for interacting with contents as Blobs
* Updated Resource abstract class so it always returns errors in JSON format. This should fix a related error where endpoints that return binary content, and then fail, generate a secondary error trying to convert the error to binary content.
* New library to handle upload file name and size validation


src/main/java/org/broadinstitute/consent/http/resources/Resource.java
```diff
@@ -51,11 +52,11 @@ protected Response createExceptionResponse(Exception e) {
             if (handler != null) {
                 return handler.handle(e);
             } else {
-                return Response.serverError().entity(new Error(e.getMessage(), Response.Status.INTERNAL_SERVER_ERROR.getStatusCode())).build();
+                return Response.serverError().type(MediaType.APPLICATION_JSON).entity(new Error(e.getMessage(), Response.Status.INTERNAL_SERVER_ERROR.getStatusCode())).build();
```
rushtong, line 56: We need to force errors to JSON for the cases when a non-json endpoint fails. In those cases, Jersey tries to force the error into the non-json format and then fails, again.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
```diff
@@ -213,4 +325,46 @@ private void checkAuthorizedUpdateUser(User user, DataAccessRequest dar) {
...
+        if (Objects.nonNull(dar.getData().getCollaborationLetterLocation())) {
+          deleteDarDocument(dar, dar.getData().getCollaborationLetterLocation());
+        }
+        dar.getData().setCollaborationLetterLocation(blobId.getName());
+        dar.getData().setCollaborationLetterName(fileName);
+        break;
+      default:
+        break;
+    }
+    dataAccessRequestService.updateByReferenceIdVersion2(user, dar);
```
JVThomas: This function already calls findByReferenceId as it's return value so you don't need to call ```dataAccessRequestService.findByReferenceId()``` again

rushtong: Nice catch!

JVThomas:
Code logic looks fine, I just want to know what precautionary steps we need to take to ensure that malicious files are not uploaded to GCS (does it scan files for us, do we get explicit error responses from GCS for malicious files that we are expected to handle, etc.)

rushtong:
RE:
> Code logic looks fine, I just want to know what precautionary steps we need to take to ensure that malicious files are not uploaded to GCS (does it scan files for us, do we get explicit error responses from GCS for malicious files that we are expected to handle, etc.)

I'm not sure what constitutes a malicious file. Tagging @adrazhi for guidance on this question. Do we need to do anything in code outside of what we are doing now when storing files uploaded by a user to GCS?

adrazhi:
For malicious content, if we're talking about GCS buckets, it's tough. Google does not scan those - they only scan the GDrive/Gsuite content.

Since malware scanning is geared more towards consumer products. tldr; don't worry about malware scanning because we'll likely look at that from a higher level in Terra in general.

rushtong:
@JVThomas @zbedo 
I have added file name and size validation to this PR. Please have a second look, thank you!

src/main/java/org/broadinstitute/consent/http/resources/Resource.java
```diff
@@ -71,6 +72,18 @@ StreamingOutput createStreamingOutput(InputStream inputStream) {
         };
     }
 
+    void validateFileDetails(FormDataContentDisposition fileDetail) {
+        FileValidator validator = new FileValidator();
+        boolean validName = validator.isValidFileName("validating uploaded file name", fileDetail.getFileName(), true);
```
JVThomas, line 77: So does the validator ensure that the filename follows a uuid format? What constitutes a valid filename?

rushtong: No, it uses a library regex. You can dig into the code here: https://github.com/augustd/owasp-java-fileio/blob/b2f13dbc0002cc6c2ffe880f76211df90e7aa4d5/src/main/java/org/owasp/fileio/FileValidator.java#L35

rushtong: In this code, we're not trying to force the file into a UUID format. We only want to validate it for OWASP reasons, and that's it. For our purposes, we really don't care what the file name is. We store that separately as a raw string on the DAR and not as part of the actual file.

JVThomas (approved):

PR #736, opened 2020-09-16
[DUOS-720][risk=no] Add create/update date, user id to Dataset model

lindojong:
Addresses additional requirements: https://broadinstitute.atlassian.net/browse/DUOS-720

As part of pushing out the new POST v2 endpoint, we want to start capturing the create and update dates and user ids for the datasets created. This PR adds these 4 fields to the Dataset and DatasetDTO models both in code and postgres, and any necessary changes to calculate them from a POST request or return them in a response. We also change the format of the "date" to Timestamp, so that we can get time of day sent (instead of just the calendar date).

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
@@ -1,5 +1,6 @@
 package org.broadinstitute.consent.http.db;
 
+import java.sql.Timestamp;
 import org.apache.commons.lang3.tuple.Pair;
 import org.broadinstitute.consent.http.db.mapper.AssociationMapper;
 import org.broadinstitute.consent.http.db.mapper.AutocompleteMapper;
@@ -42,6 +43,10 @@ public interface DataSetDAO extends Transactional<DataSetDAO> {
     @SqlUpdate("insert into dataset (name, createDate, objectId, active, alias) values (:name, :createDate, :objectId, :active, :alias)")
     @GetGeneratedKeys
     Integer insertDataset(@Bind("name") String name, @Bind("createDate") Date createDate, @Bind("objectId") String objectId, @Bind("active") Boolean active, @Bind("alias") Integer alias);
+
+    @SqlUpdate("INSERT INTO dataset (name, createdate, create_user_id, update_date, update_user_id, objectId, active, alias) VALUES (:name, :createDate, :createUserId, :createDate, :createUserId, :objectId, :active, :alias)")
+    @GetGeneratedKeys
+    Integer insertDatasetV2(@Bind("name") String name, @Bind("createDate") Timestamp createDate, @Bind("createUserId") Integer createUserId, @Bind("objectId") String objectId, @Bind("active") Boolean active, @Bind("alias") Integer alias);
 
     @SqlQuery("select * from dataset where dataSetId = :dataSetId")
     DataSet findDataSetById(@Bind("dataSetId") Integer dataSetId);
--- a/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
@@ -99,6 +99,13 @@ public interface DataSetDAO extends Transactional<DataSetDAO> {
     Set<DataSetDTO> findDataSets();
 
     @UseRowMapper(DataSetPropertiesMapper.class)
+    @SqlQuery("SELECT d.*, k.key, dp.propertyvalue, ca.consentid, c.dac_id, c.translateduserestriction " +
+          "FROM dataset d LEFT OUTER JOIN datasetproperty dp on dp.datasetid = d.datasetid LEFT OUTER JOIN dictionary k on k.keyid = dp.propertykey " +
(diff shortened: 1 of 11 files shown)


lindojong:
The Timestamp dates are not very human readable in the json (ex. `1600278704315`). Should we be formatting it to something like `MM-DD-YYYY` before it gets returned?

src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -93,6 +98,13 @@
             "order by d.dataSetId, k.displayOrder")
     Set<DataSetDTO> findDataSets();
 
+    @UseRowMapper(DataSetPropertiesMapper.class)
+    @SqlQuery("SELECT d.*, k.key, dp.propertyvalue, ca.consentid, c.dac_id, c.translateduserestriction " +
+          "FROM dataset d LEFT OUTER JOIN datasetproperty dp on dp.datasetid = d.datasetid LEFT OUTER JOIN dictionary k on k.keyid = dp.propertykey " +
```
rushtong: optional nit: For legibility, I like join statements on their own line

src/main/java/org/broadinstitute/consent/http/db/mapper/DataSetMapper.java
```diff
@@ -13,7 +13,10 @@ public DataSet map(ResultSet r, StatementContext ctx) throws SQLException {
         r.getInt("dataSetId"),
         r.getString("objectId"),
         r.getString("name"),
-        r.getTimestamp("createDate"),
+        r.getTimestamp("createdate"),
+        r.getInt("create_user_id"),
+        r.getTimestamp("update_date"),
+        r.getInt("update_user_id"),
```
rushtong: The new columns, if not included in ALL queries that use this mapper, could cause NPEs. Look at the `hasColumn` method in `RowMapperHelper` which I use to first check before trying to `get` the value.

src/main/java/org/broadinstitute/consent/http/db/mapper/DataSetPropertiesMapper.java
```diff
@@ -35,6 +35,10 @@ public DataSetDTO map(ResultSet r, StatementContext ctx) throws SQLException {
       dataSetDTO.setDataSetId(dataSetId);
       dataSetDTO.setActive(r.getBoolean("active"));
       dataSetDTO.setTranslatedUseRestriction(r.getString("translatedUseRestriction"));
+      dataSetDTO.setCreateDate(r.getDate("createdate"));
+      dataSetDTO.setCreateUserId(r.getInt("create_user_id"));
+      dataSetDTO.setUpdateDate(r.getTimestamp("update_date"));
+      dataSetDTO.setUpdateUserId(r.getInt("update_user_id"));
```
rushtong: Similar comment to above.

src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -176,7 +178,7 @@ public Response describeDataSets(@Context HttpServletRequest request , @QueryPar
     @PermitAll
     public Response describeDataSet( @PathParam("datasetId") Integer datasetId){
         try {
-            return Response.ok(api.getDataSetDTO(datasetId), MediaType.APPLICATION_JSON).build();
+            return Response.ok(datasetService.getDatasetDTO(datasetId), MediaType.APPLICATION_JSON).build();
```
rushtong: As a general practice, I like to assign the object before using in the response - it just makes debugging a lot easier in the long run. Something like this:
```
DatasetDTO ds = datasetService.getDatasetDTO(datasetId);
return Response.ok(ds, MediaType.APPLICATION_JSON).build();
```

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -23,13 +24,15 @@ public DatasetService(DataSetDAO dataSetDAO) {
         this.dataSetDAO = dataSetDAO;
     }
 
-    public DataSet createDataset(DataSetDTO dataset, String name) {
-        Date now = new Date();
+    public DataSet createDataset(DataSetDTO dataset, String name, Integer userId) {
+        Date date = new Date();
+        long time = date.getTime();
+        Timestamp now = new Timestamp(time);
```
rushtong: This could be a one liner if you never need to access `date` or `time` again.
```
Timestamp now = new Timestamp(new Date().getTime());
```

src/main/resources/changesets/changelog-consent-54.0.xml
```diff
@@ -0,0 +1,34 @@
...
+    </rollback>
+  </changeSet>
+  <changeSet dbms="postgresql" author="lzhang" id="54.2">
+    <addColumn tableName="dataset">
+      <column name="update_user_id" type="bigint" />
+    </addColumn>
+    <rollback>
+      <dropColumn columnName="update_user_id" tableName="dataset" />
+    </rollback>
+  </changeSet>
```
rushtong, line 20: I would put these all in one changeset instead of three different ones.

rushtong (commented):
Looking good! Some minor change requests inline.

rushtong (approved):
Thank you for the changes!

PR #740, opened 2020-09-18
[DUOS-767][risk=no] New /me endpoint

rushtong:
## Addresses
A bonus addition to https://broadinstitute.atlassian.net/browse/DUOS-767

## Changes
New `/me` endpoint that should simplify the profile page load and reduce the number of APi calls required for it.

---- 

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
@@ -39,11 +39,27 @@ public class UserResource extends Resource {
 
     private final UserService userService;
     private final WhitelistService whitelistService;
+    private final Gson gson;
 
     @Inject
     public UserResource(UserService userService, WhitelistService whitelistService) {
         this.userService = userService;
         this.whitelistService = whitelistService;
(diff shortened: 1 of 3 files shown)


lindojong (approved):
LGTM! Will we always want to return a generic user object and manually add researcherProperties and whitelistEntries, or should those fields be added to the User class in the future?

rushtong:
>LGTM! Will we always want to return a generic user object and manually add researcherProperties and whitelistEntries, or should those fields be added to the User class in the future?

I debated in my head about what the right approach to take here. In general, these should be properties on the user and I think I'll end up writing a story for that. The goal here was to get something in place quickly that wasn't too disruptive to all usages of the `User` class. I don't like to have transient properties on objects that are only populated in certain circumstances as it adds developer confusion down the line when trying to use it. For example, user roles. There are calls where we don't get the roles for a user, leading to confusion when we use that call, and then try to read the roles in some downstream operation. So, for those reasons, I didn't add these as real properties to the user. We should do that in a future story.

Edit: https://broadinstitute.atlassian.net/browse/DUOS-772

PR #744, opened 2020-09-23
[DUOS-720][risk=no] wrong id used when returning created dataset

lindojong:
createDataset v2 errors out on a successful insert into the database because we try to return the dataset by searching for the id given in the json input (which shouldn't exist). Renamed the variables to be more clear about which dataset is which and to grab the correct id.

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
@@ -106,7 +106,7 @@ public class DataSetResource extends Resource {
         User dacUser = userService.findUserByEmail(user.getGoogleUser().getEmail());
         Integer userId = dacUser.getDacUserId();
         DataSet dataset = datasetService.createDataset(ds, name, userId);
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -105,9 +105,9 @@ public Response createDataset(@Auth AuthUser user, @Context UriInfo info, String
         }
         User dacUser = userService.findUserByEmail(user.getGoogleUser().getEmail());
         Integer userId = dacUser.getDacUserId();
-        DataSet dataset = datasetService.createDataset(ds, name, userId);
-        URI uri = info.getRequestUriBuilder().replacePath("api/dataset/{datasetId}").build(ds.getDataSetId());
-        return Response.created(uri).entity(dataset).build();
+        DataSet createdDataset = datasetService.createDataset(inputDataset, name, userId);
+        URI uri = info.getRequestUriBuilder().replacePath("api/dataset/{datasetId}").build(createdDataset.getDataSetId());
```
rushtong, line 109: Nice catch, I should have seen that before as I have made the same mistake.

rushtong (approved):
👍

PR #756, opened 2020-09-30
[DUOS-721][risk=no] API: Put update dataset

lindojong:
Addresses: https://broadinstitute.atlassian.net/browse/DUOS-721

In all cases:


src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -78,6 +79,9 @@
     @SqlBatch("delete from datasetproperty where dataSetId = :dataSetId")
     void deleteDataSetsProperties(@Bind("dataSetId") Collection<Integer> dataSetsIds);
 
+    @SqlUpdate("UPDATE datasetproperty SET propertyvalue = :propertyValue WHERE datasetid = :datasetId AND propertykey = :propertyKey")
```
lindojong, line 81: I considered checking by propertyId here as an extra safeguard but decided against it since we never use the propertyId anywhere else (resulting in a whole new set of changes to properly identify properties by id)

src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -90,6 +94,9 @@
     @SqlUpdate("update dataset set needs_approval = :needs_approval where dataSetId = :dataSetId")
     void updateDataSetNeedsApproval(@Bind("dataSetId") Integer dataSetId, @Bind("needs_approval") Boolean needs_approval);
 
+    @SqlUpdate("UPDATE dataset SET update_date = :updateDate, update_user_id = :updateUserId WHERE datasetid = :datasetId")
```
lindojong, line 99: This only updates the update date and update user id, which could be slightly misleading since I don't change any fields on the dataset itself here, but this indicates when changes occurred in the associated dataset properties (which only store a createDate).

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -57,6 +57,36 @@ public DataSet getDatasetWithPropertiesById(Integer datasetId) {
...
+                    .noneMatch(op -> op.getPropertyKey() == p.getPropertyKey()))
+              .collect(Collectors.toList());
+
+        List<DataSetProperty> propertiesToUpdate = jsonProperties.stream()
+              .filter(p -> oldProperties.stream()
+                    .noneMatch(op -> equalsProperty(op.getPropertyKey(), op.getPropertyValue(), p.getPropertyKey(), p.getPropertyValue())))
+              .collect(Collectors.toList());
+
+        if (propertiesToAdd.isEmpty() && propertiesToUpdate.isEmpty()) {
+            return null;
```
lindojong: Couldn't think of a better way to distinguish "not modified" datasets, but I wanted to make sure this was checked for so we don't update the dataset (with date, user id) if no properties were actually updated.

src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
rushtong: This is a fine approach, no need to look for the row id.

src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
rushtong: Maybe rename this `updateDatasetUpdateUserAndDate` to clarify what it really does.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
rushtong: A clearer way to handle this is to return and `Optional<Dataset>` instead of just a `Dataset`. That way, callers know that the dataset is not guaranteed and don't have to do null checking. If you just return Optional.empty here, the caller would know nothing happened.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -88,4 +118,7 @@ public DataSetDTO getDatasetDTO(Integer datasetId) {
             .collect(Collectors.toList());
     }
 
+    private boolean equalsProperty(Integer propOneKey, String propOneValue, Integer propTwoKey, String propTwoValue) {
+        return (propOneKey.equals(propTwoKey) && propOneValue.equalsIgnoreCase(propTwoValue));
+    }
```
rushtong, line 152: An alternative to this approach is to add a equals to DatasetProperty. I like apache commons `EqualsBuilder` for this. It would make sense to use dataset id, property key, and property value. Those three things should be unique or we have a different problem. See `User` for an example.

src/main/resources/assets/api-docs.yaml
```diff
@@ -2325,6 +2325,35 @@ paths:
...
+              $ref: '#/components/schemas/Dataset'
+      tags:
+        - Datasets
+      responses:
+        200:
+          description: Successfully updated Dataset
+        304:
+          description: Not modified
+        400:
+          description: External Error (invalid input)
```
rushtong: nit, language:
```suggestion
          description: Bad Request (invalid input)
```

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -139,6 +139,39 @@ public void testCreateDatasetErrors() {
...
+        when(datasetService.getDatasetByName("test")).thenReturn(preexistingDataset);
+        when(datasetService.updateDataset(any(), any(), any())).thenReturn(preexistingDataset);
+        when(authUser.getGoogleUser()).thenReturn(googleUser);
+        when(googleUser.getEmail()).thenReturn("email@email.com");
+        when(userService.findUserByEmail(any())).thenReturn(dacUser);
+        when(dacUser.getDacUserId()).thenReturn(1);
+        when(uriInfo.getRequestUriBuilder()).thenReturn(uriBuilder);
+        when(uriBuilder.replacePath(anyString())).thenReturn(uriBuilder);
+        initResource();
+        Response response = resource.updateDataset(authUser, uriInfo, "{\"properties\":[{\"propertyName\":\"Dataset Name\",\"propertyValue\":\"test\"}]}", 1);
```
rushtong: Instead of passing in a string, it would be cleaner to pass in a Gson.toJson string value constructed from an actual instance of DataSetDTO. If we do that, then any kind of refactoring around DataSetDTO would be picked up in this test method without having to find it by a test failure.

rushtong (commented):
Looking good! See feedback inline.

src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -110,6 +111,45 @@ public Response createDataset(@Auth AuthUser user, @Context UriInfo info, String
         return Response.created(uri).entity(createdDataset).build();
     }
 
+    @PUT
+    @Consumes("application/json")
+    @Produces("application/json")
+    @Path("/{datasetId}")
+    @PermitAll
+    public Response updateDataset(@Auth AuthUser user, @Context UriInfo info, String json, @PathParam("datasetId") Integer datasetId) {
```
rushtong: Super minor nit, I like to have all of the annotated arguments toward the front of the method, non-annotated to the end. I don't know if there's any style guide for this, but it's easier for me to read that way. I welcome other opinions.
```suggestion
    public Response updateDataset(@Auth AuthUser user, @Context UriInfo info, @PathParam("datasetId") Integer datasetId,  String json) {
```

src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -110,6 +111,45 @@ public Response createDataset(@Auth AuthUser user, @Context UriInfo info, String
...
+        DataSetDTO inputDataset = new Gson().fromJson(json, DataSetDTO.class);
+        if (Objects.isNull(inputDataset)) {
+            throw new BadRequestException("Dataset is required");
+        }
+        if (Objects.isNull(inputDataset.getProperties()) || inputDataset.getProperties().isEmpty()) {
+            throw new BadRequestException("Dataset must contain required properties");
```
rushtong, line 129: TOL: we should make sure the swagger doc has an appropriate description about which properties are required.

src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -110,6 +111,45 @@ public Response createDataset(@Auth AuthUser user, @Context UriInfo info, String
...
+        }
+        List<DataSetPropertyDTO> invalidProperties = datasetService.findInvalidProperties(inputDataset.getProperties());
+        if (invalidProperties.size() > 0) {
+            List<String> invalidKeys = invalidProperties.stream().map(p -> p.getPropertyName()).collect(
+                  Collectors.toList());
+            throw new BadRequestException("Dataset contains invalid properties that could not be recognized or associated with a key: " + invalidKeys.toString());
+        }
+        String name = inputDataset.getPropertyValue("Dataset Name");
```
rushtong: `"Dataset Name"` should be a static constant somewhere, probably on `DatasetService`. I also see this string in `DatasetService`.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -57,6 +58,36 @@ public DataSet getDatasetWithPropertiesById(Integer datasetId) {
         return dataset;
     }
 
+    public Optional<DataSet> updateDataset(DataSetDTO dataset, Integer datasetId, Integer userId) {
+        Timestamp now = new Timestamp(new Date().getTime());
+
+        DataSet old = getDatasetWithPropertiesById(datasetId);
+        Set<DataSetProperty> oldProperties = old.getProperties();
+
+        List<DataSetPropertyDTO> dtos = dataset.getProperties();
```
rushtong: Naming nit (naming is always hard): I think this should reflect more about what it is, like `updateDatasetPropertyDTOs` so that you can quickly distinguish that these are on the new dataset.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -57,6 +58,36 @@ public DataSet getDatasetWithPropertiesById(Integer datasetId) {
...
+        List<DataSetProperty> jsonProperties = processDatasetProperties(datasetId, dtos);
```
rushtong: same comment here, something like `updateDatasetProperties`

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -57,6 +58,36 @@ public DataSet getDatasetWithPropertiesById(Integer datasetId) {
...
+
+        List<DataSetProperty> propertiesToAdd = jsonProperties.stream()
+              .filter(p -> oldProperties.stream()
+                    .noneMatch(op -> op.getPropertyKey() == p.getPropertyKey()))
+              .collect(Collectors.toList());
+
+        List<DataSetProperty> propertiesToUpdate = jsonProperties.stream()
+              .filter(p -> oldProperties.stream()
+                    .noneMatch(op -> p.equals(op)))
+              .collect(Collectors.toList());
```
rushtong, line 81: Do you need to also look at properties to delete? If I `PUT` a dataset with fewer props than the one that's there, I would expect that the missing props would be removed. This is in contrast to a `PATCH` where you're only making updates to the specific things that were passed in.

src/main/resources/assets/api-docs.yaml
```diff
@@ -2325,6 +2325,35 @@ paths:
           description: The Dataset was deleted.
         500:
           description: Server error.
+  /api/dataset/{datasetId}:
+    put:
+      summary: Updates the Dataset from JSON
+      description: Updates the Dataset from JSON
```
rushtong, line 2417: Suggestion from previous comment, we should fully describe the required properties here in an expanded description.

src/test/java/org/broadinstitute/consent/http/db/DataSetDAOTest.java
```diff
@@ -158,6 +163,31 @@ public void testFindDatasetPropertiesByDatasetId() {
...
+        DataSetProperty newProperty = new DataSetProperty(d.getDataSetId(), 1, "Updated Value", new Date());
+        List<DataSetProperty> updatedProperties = new ArrayList<>();
+        updatedProperties.add(newProperty);
+        dataSetDAO.updateDatasetProperty(d.getDataSetId(), updatedProperties.get(0).getPropertyKey(), updatedProperties.get(0).getPropertyValue());
+        Set<DataSetProperty> returnedProperties = dataSetDAO.findDatasetPropertiesByDatasetId(d.getDataSetId());
+        DataSetProperty returnedProperty = returnedProperties.stream().collect(Collectors.toList()).get(0);
+        assertEquals(originalProperty.getPropertyKey(), returnedProperty.getPropertyKey());
+        assertEquals(originalProperty.getPropertyId(), returnedProperty.getPropertyId());
+        assertNotEquals(originalProperty.getPropertyValue(), returnedProperty.getPropertyValue());
+    }
```
rushtong, line 189: Some more test ideas: 
* Update with multiple changed props
* Update with one changed prop and one unchanged prop
* Update with multiple unchanged props
* If we do tackle removed props, test that too.

rushtong (changes requested):
Coming along nicely! I do think we need to tackle the removal of properties. See feedback inline.

src/test/java/org/broadinstitute/consent/http/db/DataSetDAOTest.java
lindojong: Since I chose to implement the SQL queries in the DAO as single property update/delete (rather than a batch set in one call), the DAO tests `testUpdateDatasetProperty` and `testDeleteDatasetPropertyByKey` accordingly only test a single property change at a time. Instead, I added tests for multiple simultaneously changed props in `DatasetServiceTest`. Although the DAO calls are mocked and we can't verify the returned dataset here, since the service layer checks for the presence of changed props, for the sake of the test it was possible to verify that the correct workflow was followed (resulting in a non-null dataset).

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -57,6 +61,46 @@ public DataSet getDatasetWithPropertiesById(Integer datasetId) {
...
+            return Optional.empty();
+        }
+
+        updateDatasetProperties(datasetId, propertiesToUpdate, propertiesToDelete, propertiesToAdd);
+        dataSetDAO.updateDatasetUpdateUserAndDate(datasetId, now, userId);
+        DataSet updatedDataset = getDatasetWithPropertiesById(datasetId);
+        return Optional.of(updatedDataset);
+    }
+
+    private void updateDatasetProperties(Integer datasetId, List<DataSetProperty> updateProperties, List<DataSetProperty> deleteProperties, List<DataSetProperty> addProperties) {
```
rushtong: nit: minor possible optimization here ... `DataSetProperty` has a dataset id field. If that were populated before arriving at this method, you wouldn't need to pass in the id at all, just read them from the property.

src/test/java/org/broadinstitute/consent/http/db/DataSetDAOTest.java
rushtong: Gotcha, thanks!

src/test/java/org/broadinstitute/consent/http/service/DatasetServiceTest.java
```diff
@@ -126,6 +127,20 @@ public void testFindInvalidProperties() {
...
+        initService();
+
+        List<DataSetPropertyDTO> input = getDatasetPropertiesDTO();
+        DataSetPropertyDTO duplicateProperty = input.get(0);
+        input.add(duplicateProperty);
+
+        List<DataSetPropertyDTO> properties = datasetService.findDuplicateProperties(input);
+
+        Assert.assertFalse(properties.isEmpty());
+        Assert.assertEquals(properties.get(0), duplicateProperty);
```
rushtong: minor nit: I like to use the static imports so it's just:
```suggestion
        assertEquals(properties.get(0), duplicateProperty);
```

rushtong (approved):
Thank you for the changes! Looks good with a couple minor suggestions inline. Feel free to address or not if you don't want to.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
lindojong: Nice catch!

PR #768, opened 2020-10-07
[DUOS-789][risk=no] Update pending case call for chairs

rushtong:
## Addresses
Partially addresses https://broadinstitute.atlassian.net/browse/DUOS-789
See also: https://github.com/DataBiosphere/duos-ui/pull/630

## Changes
This PR:
* Updates the database call to filter out closed elections in addition to canceled ones
* Deprecate the pending case endpoint that expects a user id and instead use one that pulls the user from the authenticated user. This is actually a security improvement as we don't want chairs/members querying for data visible to other users.
* Updates api docs

----

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
@@ -148,9 +148,9 @@ public interface ElectionDAO extends Transactional<ElectionDAO> {
             "     AND electionView.referenceId = e.referenceId " +
             "     AND LOWER(e.electionType) = LOWER(:type) " +
             "     AND e.finalAccessVote = :vote " +
-            "     AND LOWER(e.status) != 'canceled' " +
+            "     AND LOWER(e.status) not in ('canceled', 'closed') " +
             " ORDER BY createDate ASC")
-    List<Election> findLastElectionsByTypeAndFinalAccessVoteChairPerson(@Bind("type") String type, @Bind("vote") Boolean finalAccessVote);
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
```diff
@@ -148,9 +148,9 @@ void bulkCancelOpenElectionByReferenceIdAndType(@Bind("electionType") String ele
             "     AND electionView.referenceId = e.referenceId " +
             "     AND LOWER(e.electionType) = LOWER(:type) " +
             "     AND e.finalAccessVote = :vote " +
-            "     AND LOWER(e.status) != 'canceled' " +
+            "     AND LOWER(e.status) not in ('canceled', 'closed') " +
```
JVThomas, line 151: Can elections have a status = "Final"?

rushtong: Yes, and final elections are still vote-able. It is not intuitive, and I don't think I understood that until recently. Closed and canceled are the only election types that you cannot vote on. Final represents a legacy middle state where there were several votes on an election, including member and chair votes, but before it was really, really, really finally voted on. We don't use the final election status in the V2 voting process

### Legacy election status paths:
* Open -> Canceled
* Open -> Closed
* Open -> Final
* Open -> Final -> Canceled
* Open -> Final -> Closed
* Closed -> Open
* Canceled -> Open

### New election status paths:
* Open -> Canceled
* Open -> Closed
* Closed -> Open
* Canceled -> Open

JVThomas (approved):

PR #779, opened 2020-10-15
[DUOS-786][risk=no] Find DAC by Id throws a 500 when not found

lindojong:
Addresses: https://broadinstitute.atlassian.net/browse/DUOS-786

At the service level, find DAC by id was setting chairpersons and members without first checking if the dac itself existed. If a dac wasn't properly found, it would hit a NPE here before being flagged as null at the resource level (where the 404 NotFoundException is thrown).

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DacService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DacService.java
@@ -124,8 +124,10 @@ public class DacService {
         Dac dac = dacDAO.findById(dacId);
         List<User> chairs = dacDAO.findMembersByDacIdAndRoleId(dacId, UserRoles.CHAIRPERSON.getRoleId());
         List<User> members = dacDAO.findMembersByDacIdAndRoleId(dacId, UserRoles.MEMBER.getRoleId());
-        dac.setChairpersons(chairs);
-        dac.setMembers(members);
+        if (dac != null) {
+            dac.setChairpersons(chairs);
+            dac.setMembers(members);
+        }
         return dac;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/consent/http/service/DacService.java
```diff
@@ -124,8 +124,10 @@ public Dac findById(Integer dacId) {
         Dac dac = dacDAO.findById(dacId);
         List<User> chairs = dacDAO.findMembersByDacIdAndRoleId(dacId, UserRoles.CHAIRPERSON.getRoleId());
         List<User> members = dacDAO.findMembersByDacIdAndRoleId(dacId, UserRoles.MEMBER.getRoleId());
-        dac.setChairpersons(chairs);
-        dac.setMembers(members);
+        if (dac != null) {
+            dac.setChairpersons(chairs);
+            dac.setMembers(members);
+        }
         return dac;
```
rushtong: I would only return the dac if not null (and I like to use `Objects.nonNull` for this). Something like:
```suggestion
        if (Objects.nonNull(dac)) {
            dac.setChairpersons(chairs);
            dac.setMembers(members);
            return dac;
        }
        throw new NotFoundException("Could not find DAC with the provided id: " + dacId);
```

rushtong (changes requested):
Nice work! I think we can either throw a 404 here (which we do in other service level classes as well), or we can do it in the resource class, which we also have some precedent for.

rushtong (approved):
Thank you for the changes!

PR #785, opened 2020-10-19
[DUOS-806][risk=no] Move trivy check to a github action

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-806
See also: https://github.com/marketplace/actions/aqua-security-trivy

## Example
This is a failing run similar to what we found when running as a build step in jenkins. Now that we have this info at the PR stage, we can avoid downstream surprises.
```
docker.io/broadinstitute/consent:92adae45120db529ff8f729c0f3260540abe3f29 (debian 10.6)
=======================================================================================
Total: 1 (CRITICAL: 1)

+---------+------------------+----------+-------------------+---------------+--------------------------------+
| LIBRARY | VULNERABILITY ID | SEVERITY | INSTALLED VERSION | FIXED VERSION |             TITLE              |
+---------+------------------+----------+-------------------+---------------+--------------------------------+
| libbsd0 | CVE-2019-20367   | CRITICAL | 0.9.1-2           |               | nlist.c in libbsd before       |
|         |                  |          |                   |               | 0.10.0 has an out-of-bounds    |
|         |                  |          |                   |               | read during a comparison...    |
+---------+------------------+----------+-------------------+---------------+--------------------------------+
```

----

In all cases:


JVThomas (approved):

dinvlad:
Looks good, though I'd be a little cautious about making an out-of-bounds check here. Could you confirm if this would still prevent a finding in Trivy from deploying a vulnerable image to prod?

rushtong:
@dinvlad 
> Could you confirm if this would still prevent a finding in Trivy from deploying a vulnerable image to prod?

I don't think that's completely possible in either this approach or the build script-approach. There's always a possibility that a vulnerable image goes to prod. The time between PR merges and jenkins deployments are relatively negligible. Once the image is built, it is pushed to GCR and then potentially can be a prod image. The only way to address your concern is to scan the images in GCR on a regular basis. It will absolutely be the case that you'll find some prod images right now that are vulnerable. For example, both https://github.com/DataBiosphere/consent and https://github.com/DataBiosphere/consent-ontology use `openjdk8` in prod right now and we know it has a vulnerability.

dinvlad:
I see, makes sense! So the deployments to prod are triggered manually, and as part of the deployment process you'll consult Trivy findings from GH Action also? Just to make sure they're not purely informational and won't be out of sight :)

rushtong:
@dinvlad 
> Just to make sure they're not purely informational

We're keeping the `exit-code: '1'` configuration so that the PR will fail if trivy finds an issue. That will, per DSP SDLC policy, prevent devs from merging the PR. And in that case, it would be a standard practice to place the failing PR on hold until a separate PR with a security fix is put into place that addresses whatever trivy finds. Then, once that's fixed, the original failing PR can be rebased and life moves on normally.

dinvlad:
Got it, then LGTM!

dinvlad (approved):

PR #791, opened 2020-10-20
[DUOS-732][risk=no] Find Dacs By User

lindojong:
Addresses https://broadinstitute.atlassian.net/browse/DUOS-732

The updated Dataset Catalog page adds a DAC column naming the owning DAC for a given dataset. I thought it was unnecessary to use the available `findAll` that also returns the list of members inside the DAC just to populate this field, so I made a new endpoint instead to return a minimal list of dacs available to a user, restricted to only admins and chairpersons. 

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
@@ -204,6 +204,15 @@ public class DacResource extends Resource {
         return Response.ok().entity(users).build();
     }
 
+    @GET
+    @Path("partial")
+    @Produces("application/json")
+    @RolesAllowed({ADMIN, CHAIRPERSON})
+    public Response findDacsByUser(@Auth AuthUser authUser) {
+        List<Dac> dacs = dacService.findDacsByUser(authUser);
+        return Response.ok().entity(dacs).build();
+    }
+
     private User findDacUser(Integer userId) {
         User user = dacService.findUserById(userId);
         if (user == null) {
--- a/src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
@@ -3,6 +3,7 @@ package org.broadinstitute.consent.http.resources;
 import com.google.gson.Gson;
 import com.google.inject.Inject;
 import io.dropwizard.auth.Auth;
+import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/consent/http/service/DacService.java
```diff
@@ -121,6 +121,15 @@ public DacService(DacDAO dacDAO, UserDAO userDAO, DataSetDAO dataSetDAO,
         return dacToUserMap;
     }
 
+    public List<Dac> findDacsByUser(AuthUser authUser) {
+        if (isAuthUserAdmin(authUser)) {
+            List<Dac> allDacs = dacDAO.findAll();
+            return allDacs;
+        }
+        List<Dac> dacs = dacDAO.findDacsForEmail(authUser.getGoogleUser().getEmail());
```
rushtong: The auth user's name is their email address, so this also works:
```suggestion
        List<Dac> dacs = dacDAO.findDacsForEmail(authUser.getName());
```

rushtong (changes requested):
I like this idea in principle. I would request that we change how we're getting this information. Instead of making a new endpoint, update the current `findAll` resource method to take an optional query param (i.e. `Optional<Boolean>`), `withUsers` with a default value of true. That way existing API calls will work and new ones that don't need users can make a more lightweight API call. 

Some reference info:
* https://www.dropwizard.io/en/latest/getting-started.html#creating-a-resource-class

src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
```diff
@@ -40,8 +42,15 @@ public DacResource(DacService dacService) {
...
-        List<Dac> dacs = dacService.findAllDacsWithMembers();
+    public Response findAll(@Auth AuthUser authUser, @QueryParam("withUsers") Optional<Boolean> withUsers) {
+        List<Dac> dacs = new ArrayList<>();
+        final Boolean includeUsers = withUsers.isPresent() ? withUsers.get() : true;
+        if (includeUsers) {
+            dacs.addAll(dacService.findAllDacsWithMembers());
+        }
+        else {
+            dacs.addAll(dacService.findDacsByUser(authUser));
+        }
```
rushtong, line 53: Thank you for this change, I think it simplifies the overall work this class does.

rushtong (approved):
Thank you, this looks good!

PR #803, opened 2020-10-23
[DUOS-723][risk=no] Associate Consent to Dataset

lindojong:
Addresses: https://broadinstitute.atlassian.net/browse/DUOS-723
Foundational work to create a consent from a given dataset.

In all cases:


src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -1,29 +1,91 @@
...
+            String associationType = AssociationType.SAMPLESET.getValue();
+            consentDAO.insertConsentAssociation(consentId, associationType, dataset.getDacId());
+            return consentDAO.findConsentById(consentId);
+        } else {
+            throw new NullPointerException("Dataset is missing Data Use information. Consent could not be created.");
+        }
+
+    }
+
+    private boolean isConsentDataUseManualReview(DataUse dataUse) {
```
lindojong, line 77: I was having a bit of trouble understanding what to do with this, as it appears that many of the fields in DataUse won't actually have values, so it wouldn't be safe to arbitrarily access them. In v2, we can "fix" this by guaranteeing that all these fields are defaulted to false on the UI on the info coming in, but that would be a departure from how things are done today.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -2,32 +2,93 @@
...
+             * translated use restriction
+             */
+            consentDAO.insertConsent(consentId, manualReview, null, dataset.getDataUse().toString(),
+                  null, name, null, createDate, createDate, null,
+                  true, groupName, dataset.getDacId());
+            String associationType = AssociationType.SAMPLESET.getValue();
+            consentDAO.insertConsentAssociation(consentId, associationType, dataset.getDacId());
+            return consentDAO.findConsentById(consentId);
+        } else {
+            throw new NullPointerException("Dataset is missing Data Use information. Consent could not be created.");
```
rushtong: This should be an `IllegalArgumentException`

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
rushtong: I think that's fine. The code checks for null-ness so we shouldn't fail in the case you're describing. "Manual Review"-ness is purely a function of checking a variety of fields for any value at all.

rushtong (approved):
One minor codacy nit inline, but otherwise, 👍

rushtong (approved):

PR #810, opened 2020-11-03
[DUOS-820][risk=no] Add user validation to legacy DAR modification calls

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/DUOS-820

## Changes
* Validate that the user is the create user for the DAR or an admin in these methods:
  * PUT /api/dar/{id}
  * PUT /api/dar/partial
  * PUT /api/dar/cancel/{id}
  * DELETE /api/dar/{id}
  * DELETE /api/dar/partial/{id}  

## Reviewer Notes
The codacy checks that are failing are unrelated legacy code that I do not want to address as part of this PR. Please ignore the following:

> Avoid reassigning parameters such as 'dar'
> public Response updateDataAccessRequest(@Auth AuthUser authUser, Document dar, @PathParam("id") String id) {
 
> Avoid reassigning parameters such as 'dar'
> public Response updatePartialDataAccessRequest(@Auth AuthUser authUser, @Context UriInfo info, Document dar) {

----

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -1,6 +1,7 @@
 package org.broadinstitute.consent.http.resources;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.api.client.http.HttpStatusCodes;
 import com.google.inject.Inject;
 import freemarker.template.TemplateException;
 import io.dropwizard.auth.Auth;
@@ -14,12 +15,15 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import javax.annotation.security.PermitAll;
 import javax.annotation.security.RolesAllowed;
 import javax.mail.MessagingException;
 import javax.ws.rs.BadRequestException;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
+import javax.ws.rs.ForbiddenException;
 import javax.ws.rs.GET;
 import javax.ws.rs.NotFoundException;
 import javax.ws.rs.POST;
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -468,4 +475,35 @@ private User findUserByEmail(String email) {
...
+     * they can also have access to the DAR.
+     *
+     * @param allowableRoles List of roles that would allow the user to access the resource
+     * @param authUser The AuthUser
+     * @param referenceId The referenceId of the resource.
+     */
+    private void validateAuthedRoleUser(final List<UserRoles> allowableRoles, AuthUser authUser, String referenceId) {
+        DataAccessRequest dataAccessRequest = findDataAccessRequestById(referenceId);
+        User user = findUserByEmail(authUser.getName());
+        if (Objects.nonNull(dataAccessRequest.getUserId()) && dataAccessRequest.getUserId() > 0) {
```
JVThomas, line 502: Don't know much about how the Java back-end is set up, so I have some questions. Is userId the only way to validate the user before role check? Is there a session token or timed OAuth token saved in server memory or db that can be referenced?

rushtong: The `AuthUser authUser` is the OAuthed user that we get from the bearer token when the user makes an authenticated request, so that is, in effect, what you're asking in this question, I believe. That information is validated against our user table in `findUserByEmail(authUser.getName())` so we ensure that the person who logged in is really a user we know about.

JVThomas: And the AuthUser is determined at log in right? There's no way to alter that on each request?

rushtong: That's exactly right. If you look at the API methods, they are instantiated with `@Auth AuthUser authUser` which is populated by the framework. There are auth filters in place that examine the bearer token, query Google for the user, and then populate the `AuthUser`.

JVThomas: Ok and validateAuthedRoleUser call ensures that a request can't impersonate another user by making sure that the userId provided in the request matches with the AuthUser id on the server right?

rushtong: Correct. Here is that method in `Resource`: 
```
    /**
     * Validate that the current authenticated user can access this resource.
     * If the user has one of the provided roles, then access is allowed.
     * If not, then the authenticated user must have the same identity as the
     * `userId` parameter they are requesting information for.
     *
     * @param authedRoles   List of UserRoles enums
     * @param authedUser    The authenticated DACUser
     * @param userId        The id of the DACUser the authenticated user is requesting access to
     */
    void validateAuthedRoleUser(final List<UserRoles> authedRoles, final User authedUser, final Integer userId) {
        List<Integer> authedRoleIds = authedRoles.stream().
                map(UserRoles::getRoleId).
                collect(Collectors.toList());
        boolean authedUserHasRole = authedUser.getRoles().stream().
                anyMatch(userRole -> authedRoleIds.contains(userRole.getRoleId()));
        if (!authedUserHasRole && !authedUser.getDacUserId().equals(userId)) {
            throw new ForbiddenException("User does not have permission");
        }
    }
```

lindojong (approved):
I'm ok with these changes, they're consistent with how I understand user validation is done.

JVThomas (approved):

PR #814, opened 2020-11-03
[DUOS-485][risk=no] DAC Endpoints: Filter by user level

lindojong:
Addresses: https://broadinstitute.atlassian.net/browse/DUOS-485
Before, DAC findAll would find all DACs, irregardless of user role. We would rather find DACS specific to a user. 

I separated out the logic that adds member information and attached it to the DACS returned by findDacsByUser if flagged. I kept the original findAllDacsWithMembers, since the MetricsService uses it.

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
@@ -43,9 +43,8 @@ public class DacResource extends Resource {
     @Produces("application/json")
     @RolesAllowed({ADMIN, MEMBER, CHAIRPERSON})
     public Response findAll(@Auth AuthUser authUser, @QueryParam("withUsers") Optional<Boolean> withUsers) {
-        List<Dac> dacs = new ArrayList<>();
         final Boolean includeUsers = withUsers.isPresent() ? withUsers.get() : true;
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
```diff
@@ -45,12 +45,7 @@ public DacResource(DacService dacService) {
     public Response findAll(@Auth AuthUser authUser, @QueryParam("withUsers") Optional<Boolean> withUsers) {
         List<Dac> dacs = new ArrayList<>();
         final Boolean includeUsers = withUsers.isPresent() ? withUsers.get() : true;
-        if (includeUsers) {
-            dacs.addAll(dacService.findAllDacsWithMembers());
-        }
-        else {
-            dacs.addAll(dacService.findDacsByUser(authUser));
-        }
+        dacs.addAll(dacService.findDacsByUser(authUser, includeUsers));
```
rushtong: Minor optimization. Get rid of line 46 and just create the list on line 48:
```suggestion
        List<Dac> dacs = dacService.findDacsByUser(authUser, includeUsers);
```

rushtong (approved):
One minor comment inline, but looks great, thank you!

PR #834, opened 2020-11-17
DUOS-842 [risk=no] Timestamp and Partial DAR Code bugfixes

JVThomas:
Addresses [DUOS-842](https://broadinstitute.atlassian.net/browse/DUOS-842)

PR adds create date and sort date to document records on GET requests, PR also generates partialDarCode for new drafts (v2 resource file). Additions address the current issue on the UI where requested DARs do not have partialDarCodes assigned on generation and does not have create and sort dates added to document returns, resulting in blank fields and occasional component breakage due to missing attributes.

As mentioned on the ticket, the PR is a companion piece to DUOS-681 since that PR utilizes the new v2 endpoints (which is how the bug was initially discovered), so testing alongside that ticket is advised.

Branch is also based off of DUOS-832, which is another companion piece to DUOS-681

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
@@ -1,7 +1,7 @@
 package org.broadinstitute.consent.http.service;
 
 import static java.util.stream.Collectors.toList;
-
+import java.text.SimpleDateFormat;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.inject.Inject;
@@ -291,6 +291,8 @@ public class DataAccessRequestService {
             throw new IllegalArgumentException("User and DataAccessRequest are required");
         }
         Date now = new Date();
+        SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd");
+        dar.getData().setPartialDarCode(DataAccessRequestData.partialDarCodePrefix + sdf.format(now));
         dataAccessRequestDAO.insertVersion2(
             dar.getReferenceId(),
             user.getDacUserId(),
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -245,6 +245,8 @@ private Document createDocumentFromDar(DataAccessRequest d) {
         Document document = Document.parse(gson.toJson(d.getData()));
         document.put(DarConstants.DATA_ACCESS_REQUEST_ID, d.getId());
         document.put(DarConstants.REFERENCE_ID, d.getReferenceId());
+        document.put(DarConstants.CREATE_DATE, d.getCreateDate());
+        document.put(DarConstants.SORT_DATE, d.getCreateDate());
```
rushtong: This line should be: 
```suggestion
        document.put(DarConstants.SORT_DATE, d.getSortDate());
```

rushtong: I think we also need to make a couple similar updates to `public DataAccessRequest insertDraftDataAccessRequest(User user, DataAccessRequest dar)`. Can you add these lines just before the dao insert call?
```
        SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd");
        dar.getData().setCreateDate(now.getTime());
        dar.getData().setPartialDarCode("temp_DAR_" + sdf.format(now));
```

rushtong: While we're at it, we could abstract `"temp_DAR_"` into a public static string so both classes can refer to it.

rushtong (commented):
Some minor changes requested, but looks good, nice catch on this bug!

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
rushtong: Still need this one change:
```suggestion
        document.put(DarConstants.SORT_DATE, d.getSortDate());
```

rushtong (changes requested):
Thank you for the changes! Still one change to make, see inline.

rushtong (approved):
Thank you!

rushtong:
@JVThomas Have run `mvn clean package` locally on this branch and it runs fine. Please ignore the current maven action failure, it is a false negative. Merge when you are ready.

JVThomas:
Note: At time of merge there were issues obtaining maven packages (most likely external issue), resulting in the failure of the Coverage and Maven jobs. Code was able to perform locally, indicating that these failures are likely to be false positives.

PR #837, opened 2020-12-01
[DUOS-816] Create new consent during create dataset v2

lindojong:
When creating a new dataset through the v2 path, we want to also create a new consent and consent association from scratch for that dataset. This PR takes the previously written code for creating a consent from a dataset and integrates it to the v2 pathway, as well as updating the consent definition for DataUse to include new fields added in consent-ontology (`secondaryOther` and `publicationMoratorium`).

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
@@ -42,6 +42,7 @@ import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.consent.http.models.AuthUser;
+import org.broadinstitute.consent.http.models.Consent;
 import org.broadinstitute.consent.http.models.DataSet;
 import org.broadinstitute.consent.http.models.Dictionary;
 import org.broadinstitute.consent.http.models.User;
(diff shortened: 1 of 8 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -112,9 +113,14 @@ public Response createDataset(@Auth AuthUser user, @Context UriInfo info, String
         }
         User dacUser = userService.findUserByEmail(user.getGoogleUser().getEmail());
         Integer userId = dacUser.getDacUserId();
-        DataSet createdDataset = datasetService.createDataset(inputDataset, name, userId);
-        URI uri = info.getRequestUriBuilder().replacePath("api/dataset/{datasetId}").build(createdDataset.getDataSetId());
-        return Response.created(uri).entity(createdDataset).build();
+        DataSetDTO createdDataset = datasetService.createDataset(inputDataset, name, userId);
+        Consent createdConsent = datasetService.createConsentForDataset(createdDataset);
```
rushtong: This can throw an `IllegalArgumentException` so we should handle that case here.

src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -112,9 +113,14 @@ public Response createDataset(@Auth AuthUser user, @Context UriInfo info, String
...
+        if (Objects.isNull(createdConsent)) {
+            throw new ClientErrorException("Consent was not able to be created for newly created dataset with id: " + createdDataset.getDataSetId(), Status.CONFLICT);
```
rushtong: Are we sure this is a conflict? Seems like this would be more of a server error.

src/main/resources/assets/api-docs.yaml
```diff
@@ -4125,6 +4123,9 @@ components:
         objectId:
           type: string
           description: The sample collection ID, if defined
+        dataUse:
+          type: object
```
rushtong, line 4127: Outside the scope of this PR - it would be nice to have a model for `DataUse`

rushtong (changes requested):
Couple minor suggestions, but looking good!

src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
lindojong: Sure! In the event that a Consent cannot be created, should we be rolling back the new Dataset as well?

rushtong: Excellent point, yes, the whole operation should fail in that case. If there's anything we can do to get the reason behind the failure out to the UI, that would be helpful.

src/main/resources/assets/api-docs.yaml
lindojong: Filed a ticket for this here: https://broadinstitute.atlassian.net/browse/DUOS-855

rushtong (approved):
Thank you for the changes!

PR #862, opened 2020-12-14
[DUOS-694][risk=low] Update findDatasets query behavior

lindojong:
Addresses: https://broadinstitute.atlassian.net/browse/DUOS-694

- Consolidated query for `findDatasets` general use, for researchers, and for autocomplete string into one. 
- Added additional service level processing to filter and parse autocomplete output into previous expected format. 
- Removing the filtering by election status rendered the general use case and the for researcher case to be the same, and routed those through a new `DatasetService.getAllActiveDatasets`.
- Wrote tests for any new methods added.
- Removed the old `AutocompleteMapper` specific for formatting autocomplete DAO output (since it no longer exists), and any references to it.

In all cases:


src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -100,10 +99,14 @@
...
-            "from dataset d inner join datasetproperty dp on dp.dataSetId = d.dataSetId and d.name is not null inner join dictionary k on k.keyId = dp.propertyKey " +
-            "inner join consentassociations ca on ca.dataSetId = d.dataSetId inner join consents c on c.consentId = ca.consentId " +
-            "order by d.dataSetId, k.displayOrder")
+    @SqlQuery("SELECT d.*, k.key, dp.propertyvalue, ca.consentid, c.dac_id, c.translateduserestriction, c.datause " +
+          "FROM dataset d " +
+          "INNER JOIN datasetproperty dp on dp.datasetid = d.datasetid " +
+          "INNER JOIN dictionary k on k.keyid = dp.propertykey " +
+          "INNER JOIN consentassociations ca on ca.datasetid = d.datasetid " +
+          "INNER JOIN consents c on c.consentid = ca.consentid " +
+          "WHERE d.active = true and d.name IS NOT NULL " +
```
rushtong: nit: capitalize `on` keywords for consistency

src/main/java/org/broadinstitute/consent/http/db/mapper/AutocompleteMapper.java
```diff
@@ -1,19 +0,0 @@
-package org.broadinstitute.consent.http.db.mapper;
-
-import java.sql.ResultSet;
-import java.sql.SQLException;
-import java.util.HashMap;
-import java.util.Map;
-import org.jdbi.v3.core.mapper.RowMapper;
-import org.jdbi.v3.core.statement.StatementContext;
-
-public class AutocompleteMapper implements RowMapper<Map<String, String>> {
```
rushtong, line 10: Great work pruning unused code!

src/main/java/org/broadinstitute/consent/http/service/DatabaseDataSetAPI.java
```diff
@@ -172,13 +172,11 @@ public ParseResult overwrite(File dataSetFile, Integer userId) {
...
-
+    @Deprecated
     @Override
     public Collection<DataSetDTO> describeDataSets(Integer dacUserId) {
-        Collection<DataSetDTO> dataSetDTOList;
-        if (userIs(UserRoles.RESEARCHER.getRoleName(), dacUserId)) {
-            dataSetDTOList = dsDAO.findDataSetsForResearcher();
-        } else {
+        Collection<DataSetDTO> dataSetDTOList = dsDAO.findDataSets();
+        if (!userIs(UserRoles.RESEARCHER.getRoleName(), dacUserId)) {
```
lindojong: Have not completely removed this old method to ask some follow up questions... In the old behavior, we would check for whether or not the given user is a researcher or admin. (This would short circuit on researcher even if a user was both). If not a researcher, and yes an admin, the dataset list retrieved would be self-selected for certain election properties, upon which properties like `updateAssociationToDataOwnerAllowed`, `isAssociatedToDataOwners`, and `deletable` would be set. These properties are only set here, and are otherwise always null. Should we still be assigning these values based on the same criteria despite wanting to fetch all datasets irregardless of election status?

src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -351,7 +351,8 @@ public Response disableDataSet(@PathParam("datasetObjectId") Integer dataSetId,
     @Produces("application/json")
     @PermitAll
     public Response datasetAutocomplete(@PathParam("partial") String partial){
-        List<Map<String, String>> j = api.autoCompleteDataSets(partial);
+        List<Map<String, String>> j = datasetService.autoCompleteDatasets(partial);
```
rushtong: nit: naming. I know this was the original naming pattern, but `j` doesn't connote anything at all. `terms` seems like a reasonable name, but I'm open to anything more indicative of what it really is.

src/main/java/org/broadinstitute/consent/http/service/DatabaseDataSetAPI.java
lindojong: These fields affect whether or not certain buttons on the dataset catalog are disabled

src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
rushtong: The `d.active = true` condition is not valid in two cases
1. Admins can see all datasets
2. Chairs can see disabled datasets that are associated to their DACs

src/main/java/org/broadinstitute/consent/http/service/DatabaseDataSetAPI.java
rushtong: Great questions.
In the case of admins/chairs we want to get all datasets regardless of active/inactive. We need this because an admin/chair can enable a currently disabled dataset. These are all admin-level fields so a user with admin permissions on that dataset can execute the available functions
* Admins: Can perform admin functions on all datasets
* Chairs: Can perform admin functions only on datasets that are associated to their DACs. Other datasets should show a null prop so the UI can eventually do the right thing. We might need a story for this since I don't think we've ever made any catalog updates wrt a Chair view.
* Everyone else: Can see only active datasets

rushtong (commented):
See comments inline. Code is looking good!

src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
```diff
@@ -226,11 +226,11 @@ public Response createDataSet(
     @GET
     @Produces("application/json")
     @PermitAll
-    public Response describeDataSets(@Context HttpServletRequest request , @QueryParam("dacUserId") Integer dacUserId){
+    public Response describeDataSets(@Context HttpServletRequest request, @QueryParam("dacUserId") Integer dacUserId){
         if (StringUtils.isEmpty(request.getParameter("dacUserId"))) {
             return Response.status(Response.Status.NOT_FOUND).build();
         } else {
-            Collection<DataSetDTO> dataSetList = api.describeDataSets(dacUserId);
+            Collection<DataSetDTO> dataSetList = datasetService.describeDatasets(dacUserId);
```
lindojong, line 233: Went ahead and replaced the old describeDatasets as well, since it should have the same filtering logic (without a partial string to filter by)

rushtong: The dataset catalog calls this endpoint. According to the underlying logic in the service class:
```
Admins => see all datasets, active and inactive
Chairs => see all active datasets and inactive ones that are part of the user's DAC
Others => see all active datasets
```
This is correct, as far as it stands. The issue I see now is in the UI. It will need to be able to display the dataset correctly. For example, I, as both a researcher and an admin on dev, should be able to see disabled datasets, but should not be able to select them for creating a DAR. Let's follow up to ensure we have a story that covers this.

src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -100,11 +99,38 @@
...
+    Set<DataSetDTO> findDatasetsByUser(@Bind("dacUserId") Integer dacUserId);
+
+    @UseRowMapper(DataSetPropertiesMapper.class)
+    @SqlQuery("SELECT d.*, k.key, dp.propertyvalue, ca.consentid, c.dac_id, c.translateduserestriction, c.datause " +
+          "FROM dataset d " +
+          "INNER JOIN datasetproperty dp ON dp.datasetid = d.datasetid " +
+          "INNER JOIN dictionary k ON k.keyid = dp.propertykey " +
+          "INNER JOIN consentassociations ca ON ca.datasetid = d.datasetid " +
+          "INNER JOIN consents c ON c.consentid = ca.consentid " +
+          "WHERE d.name IS NOT NULL and d.active = true " +
```
rushtong: nit: casing
```suggestion
          "WHERE d.name IS NOT NULL AND d.active = true " +
```

src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -100,11 +99,38 @@
...
-    Set<DataSetDTO> findDataSets();
+    @SqlQuery("SELECT d.*, k.key, dp.propertyvalue, ca.consentid, c.dac_id, c.translateduserestriction, c.datause " +
+          "FROM dataset d " +
+          "INNER JOIN datasetproperty dp ON dp.datasetid = d.datasetid " +
+          "INNER JOIN dictionary k ON k.keyid = dp.propertykey " +
+          "INNER JOIN consentassociations ca ON ca.datasetid = d.datasetid " +
+          "INNER JOIN consents c ON c.consentid = ca.consentid " +
+          "INNER JOIN user_role ur ON ur.dac_id = c.dac_id " +
+          "INNER JOIN dacuser u ON ur.user_id = u.dacUserId " +
+          "WHERE u.dacUserId = :dacUserId and d.name IS NOT NULL " +
```
rushtong: nit casing
```suggestion
          "WHERE u.dacuserid = :dacUserId and d.name IS NOT NULL " +
```

src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -100,11 +99,38 @@
...
-            "order by d.dataSetId, k.displayOrder")
-    Set<DataSetDTO> findDataSets();
+    @SqlQuery("SELECT d.*, k.key, dp.propertyvalue, ca.consentid, c.dac_id, c.translateduserestriction, c.datause " +
+          "FROM dataset d " +
+          "INNER JOIN datasetproperty dp ON dp.datasetid = d.datasetid " +
+          "INNER JOIN dictionary k ON k.keyid = dp.propertykey " +
+          "INNER JOIN consentassociations ca ON ca.datasetid = d.datasetid " +
+          "INNER JOIN consents c ON c.consentid = ca.consentid " +
+          "INNER JOIN user_role ur ON ur.dac_id = c.dac_id " +
+          "INNER JOIN dacuser u ON ur.user_id = u.dacUserId " +
```
rushtong: nit: casing
```suggestion
          "INNER JOIN dacuser u ON ur.user_id = u.dacuserid " +
```

src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
rushtong: Follow up. The UI code is definitely impacted by this. [Here is the current code](https://github.com/DataBiosphere/duos-ui/blob/develop/src/pages/DatasetCatalog.js#L426) that doesn't have any logic around the validity of the dataset selection.

rushtong (commented):
This looks really good, thank you @lindojong!

Minor nits inline, and we need to follow up with the UI so invalid datasets don't have checkboxes allowing them to be selected for a DAR. Also, the dataset autocomplete will need to filter out invalid datasets so they aren't added to a DAR via the DAR form.

src/main/java/org/broadinstitute/consent/http/resources/DataSetResource.java
lindojong: Ok! Just to clarify, we want to disable the selection checkbox if a dataset is inactive/disabled?

lindojong: Screen Shot 2021-01-05 at 2 36 35 PM
Also, we might want to flip the enable/disable icons... kind of confusing to click cancel to enable, now that I think about it.

rushtong: Yes, exactly. I know that selection is also used for downloading the dataset, but we'll need to figure out something else to deal with that.

lindojong: Ticket created here: https://broadworkbench.atlassian.net/browse/DUOS-901

rushtong (approved):
Thank you for the changes!

PR #877, opened 2021-01-07
[DUOS-894][risk=no] Minor java 11 fixes

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-894

See also: https://github.com/DataBiosphere/consent-ontology/pull/372

## Changes
* Minor pom fixes for java 11 support
* Needed to add ignores for any test that uses `PowerMock` so it would run under java 11
  * `PowerMock` will need to go away at some point once we finish migrating away from the singleton service holder pattern
* Needed to update all of the actions that do builds as they were running java 8
  * Added caching to help speed up the builds ... they are horribly slow when pulling down maven resources
  * Removed unnecessary maven installs since it is already on the `ubuntu-latest` image
  * Update each one to set java 11 as the default
  * Standardize all actions on `ubuntu-latest`
  * Added a `mvn -v` for helpful debugging/run info

----

In all cases:

### DIFF
--- a/.github/workflows/maven.yaml
+++ b/.github/workflows/maven.yaml
@@ -10,10 +10,6 @@ jobs:
       - uses: actions/setup-java@v1
         with:
           java-version: 11
-      - name: Install
-        run: |
-          sudo apt update
-          sudo apt install maven
       - name: Package
         env:
           MAVEN_OPTS: -Dmaven.test.skip=true -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
--- a/.github/workflows/maven.yaml
+++ b/.github/workflows/maven.yaml
@@ -10,6 +10,12 @@ jobs:
       - uses: actions/setup-java@v1
         with:
           java-version: 11
+      - uses: actions/cache@v1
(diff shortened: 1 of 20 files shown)


raejohanek (commented):
This looks good to me although I didn't test it locally

.github/workflows/maven.yaml
```diff
@@ -7,12 +7,18 @@ jobs:
...
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
       - name: Package
         env:
           MAVEN_OPTS: -Dmaven.test.skip=true -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
         run: |
+          mvn -v
```
JVThomas, line 23: Is the version output for logging purposes?

rushtong: Yes, exactly. I can see in the logs what the maven and java versions are this way and can cross reference any changes with previous builds/runs.

JVThomas (approved):

PR #887, opened 2021-01-13
[DUOS-865][risk=no] Bug fix for Find Dataset By Name

lindojong:
Follow up to: https://broadworkbench.atlassian.net/browse/DUOS-865

### Scope of Changes
---
- We wanted to match dataset names irrespective of case, but it was implemented only removing case from one half. This change compares both names as lowercase.

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
@@ -196,7 +196,7 @@ public interface DataSetDAO extends Transactional<DataSetDAO> {
     @SqlQuery("SELECT dataSetId FROM dataset WHERE name = :name")
     Integer getDatasetIdByName(@Bind("name") String name);
 
-    @SqlQuery("SELECT * FROM dataset WHERE LOWER(name) LIKE LOWER(:name)")
+    @SqlQuery("SELECT * FROM dataset WHERE LOWER(name) LIKE CONCAT('%', LOWER(:name), '%')")
     DataSet getDatasetByName(@Bind("name") String name);
 
     @SqlQuery("select *  from dataset where name in (<names>) ")
--- a/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
@@ -196,7 +196,7 @@ public interface DataSetDAO extends Transactional<DataSetDAO> {
     @SqlQuery("SELECT dataSetId FROM dataset WHERE name = :name")
     Integer getDatasetIdByName(@Bind("name") String name);
 
-    @SqlQuery("SELECT * FROM dataset WHERE name LIKE LOWER(:name)")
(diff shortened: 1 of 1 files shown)


rushtong (dismissed):
This works as it stands, but could be better. Like clauses require `%` around the ~search~ query term to work as expected. I should have caught this first time around.

lindojong:
@rushtong I could attach `%`s around the query term in the service layer before sending it to the DAO if that's preferable?

rushtong:
@lindojong I think the DAO is where it makes sense. Have a look at this example from `DacDAO`:
```
    @SqlQuery("select du.*, r.roleId, r.name, ur.user_role_id, ur.user_id, ur.role_id, ur.dac_id from dacuser du " +
              " inner join user_role ur on ur.user_id = du.dacUserId " +
              " inner join roles r on r.roleId = ur.role_id " +
              " where lower(du.displayName) like concat('%', lower(:term), '%') " +
              " or lower(du.email) like concat('%', lower(:term), '%') " +
              " or lower(du.additional_email) like concat('%', lower(:term), '%') ")
```

rushtong (commented):
Still ruminating on this. If I search for "Smith" ... this will return "Smith" as well as "Smithson" or "Hamsmith" ... etc. In retrospect, I think a like query is the wrong approach here.

rushtong:
As an example, this returns 48 rows on dev:
```
SELECT * FROM dataset WHERE LOWER(name) LIKE CONCAT('%', LOWER('TOP'), '%');
```

lindojong:
If we're ensuring both sides are being compared in lowercase, would it suffice to just use equals? `SELECT * FROM dataset WHERE LOWER(name) = LOWER(:name);` Alternatively, the only bonus to fuzzy search that I can think of would be checking for stray special characters/punctuation, such as `Dataset Name  ` vs `Dataset name!` vs `Dataset Name...`, etc. Do we want to flag those as being the same name?

rushtong:
To address the requirements of the story, I think your first suggestion is the thing we should do:

> If we're ensuring both sides are being compared in lowercase, would it suffice to just use equals? SELECT * FROM dataset WHERE LOWER(name) = LOWER(:name);

src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
```diff
@@ -196,7 +196,7 @@
     @SqlQuery("SELECT dataSetId FROM dataset WHERE name = :name")
     Integer getDatasetIdByName(@Bind("name") String name);
 
-    @SqlQuery("SELECT * FROM dataset WHERE LOWER(name) LIKE CONCAT('%', LOWER(:name), '%')")
+    @SqlQuery("SELECT * FROM dataset WHERE LOWER(name) = LOWER(:name);")
```
rushtong: We don't need the semi-colon, but I don't think it hurts either. Best to remove it for consistency.

rushtong (approved):
One minor comment inline, 👍🏽  once that's addressed.

PR #892, opened 2021-01-20
[DUOS-940][risk=low] Conditional election final vote query fix

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-940

## Changes
The get-election-by-id query was pulling the final vote from the chairperson vote. That is technically incorrect as that only applies to RP and DUL elections. DAR elections have a final vote based on a vote type of final. Data Owner elections have a final vote based on the data owner vote type.

----

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
@@ -95,10 +95,10 @@ public interface ElectionDAO extends Transactional<ElectionDAO> {
           + "FROM election e "
           + "LEFT JOIN vote v ON v.electionid = e.electionid AND "
           + "    CASE "
-          + "        WHEN LOWER(e.electiontype) = 'dataaccess' THEN LOWER(v.type) = 'final' "
-          + "        WHEN LOWER(e.electiontype) = 'dataset' THEN LOWER(v.type) = 'data_owner' "
-          + "        ELSE LOWER(v.type) = 'chairperson' "
-          + "    END "
+          + "        WHEN LOWER(e.electiontype) = 'dataaccess' THEN 'final' "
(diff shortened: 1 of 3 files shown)


raejohanek (approved):

src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
```diff
@@ -87,10 +87,19 @@ void bulkCancelOpenElectionByReferenceIdAndType(@Bind("electionType") String ele
...
+    @SqlQuery("SELECT distinct "
+          + "    e.electionid, e.datasetid, v.vote finalvote, e.status, e.createdate, "
+          + "    e.referenceid, v.rationale finalrationale, v.createdate finalvotedate, "
+          + "    e.lastupdate, e.finalaccessvote, e.electiontype,  e.datauseletter, e.dulname, "
+          + "    e.archived, e.version "
+          + "FROM election e "
+          + "LEFT JOIN vote v ON v.electionid = e.electionid AND "
+          + "    CASE "
+          + "        WHEN LOWER(e.electiontype) = 'dataaccess' THEN 'final' "
+          + "        WHEN LOWER(e.electiontype) = 'dataset' THEN 'data_owner' "
```
JVThomas, line 99: Nothing wrong with the code, I'm just curious about the context of the data_owner vote. When I think of the voting process, the things that come to mind are the chair vote, the member votes, and the final vote, but not the data_owner vote (but it's being used here as a backup check on final).

rushtong: The data owner vote is probably something none of us has actually seen. It's a separate console that shows up just for data owners when there is a DO election that involves one of their datasets. In that case, the final vote belongs to them.

PR #902, opened 2021-01-26
DUOS-942[risk=no]Represent ToT as an integer of days

raejohanek:
## Scope:
- change turnaround time from a string to an integer representing days
- wrap conversion from long to integer in a try catch fo the unlikely occasion that number of days will exceed integer bounds
- NOTE: All changes outside of DarMetricsDecision and DacMetricsDecision have already been approved and merged as a part of DUOS-731, the new changes on this PR are all in the very last commit

## Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-942

In all cases:


src/main/java/org/broadinstitute/consent/http/models/DacDecisionMetrics.java
```diff
@@ -181,15 +181,19 @@ private void setAverageTurnaroundTimeMillis(Double averageTurnaroundTimeMillis)
...
-      this.averageTurnaroundTime =
-          DurationFormatUtils.formatDurationWords(
-              this.getAverageTurnaroundTimeMillis().longValue(), true, true);
+      //this will only ever catch an exception if there is no final date, or an unreasonable amount of time between them
+      //in this case, the upward bound of Integer is displayed
+      try {
+        this.averageTurnaroundTime = Math.toIntExact(TimeUnit.MILLISECONDS.toDays(this.averageTurnaroundTimeMillis.longValue()));
+      } catch (ArithmeticException e) {
+        this.averageTurnaroundTime = 2147483647;
+      }
```
rushtong: Since this is copied almost exactly with the one in the other class, this is a great time for a refactor. In the interface, add this method:
```
    default Integer convertMillisToDays(Long tot) {
        //this will only ever catch an exception if there is no final date, or an unreasonable amount of time between them
        //in this case, the upward bound of Integer is displayed
        try {
            return Math.toIntExact(TimeUnit.MILLISECONDS.toDays(tot));
        } catch (ArithmeticException e) {
            return 2147483647;
        }
    }
```
Then, in the Dar and Dac decision metrics classes, just call that method:
```
this.averageTurnaroundTime = convertMillisToDays(this.averageTurnaroundTimeMillis.longValue());
```

It also gives you a chance to look at the method more critically. The comments, when extracted out into this method, are confusing to me. We should be able to convert any long to a number of days, no?

rushtong (changes requested):
Looking good. One opportunity for refactoring inline.

src/main/java/org/broadinstitute/consent/http/models/DacDecisionMetrics.java
raejohanek: okay good idea!! and yes we should be able to convert any long to an integer but since the method Math.toIntExact throws an exception if the long is too large to be converted to an integer I thought it would be safer to wrap it just in case

src/main/java/org/broadinstitute/consent/http/models/DecisionMetrics.java
```diff
@@ -1,7 +1,19 @@
...
 
+import java.util.concurrent.TimeUnit;
+
 public interface DecisionMetrics {
 
     String toString(String joiner);
 
+    default Integer convertMillisToDays(Long tot) {
+        //this will only ever catch an exception if the number of days
+        //exceeds 2147483647 and thus can't be converted to an integer
```
rushtong, line 11: Much clearer, thank you!

src/test/java/org/broadinstitute/consent/http/resources/MetricsResourceTest.java
```diff
@@ -30,7 +30,8 @@ public void setUp() {
     MockitoAnnotations.initMocks(this);
   }
 
-  private void initResource() {
+  @Before
```
rushtong: We actually don't want this to run before every test invocation. It is intended to be called only by tests after they have done all of the initialization they need. The net result of adding a `@Before` here is that we initialize the resource twice for every test instead of only once.

rushtong (approved):
One issue inline, otherwise looking good 👍🏽  ! Feel free to merge after reverting `MetricsResourceTest`

src/test/java/org/broadinstitute/consent/http/resources/MetricsResourceTest.java
```diff
@@ -30,7 +30,7 @@ public void setUp() {
     MockitoAnnotations.initMocks(this);
   }
 
-  private void initResource() {
+  public void initResource() {
```
rushtong, line 33: This can remain private

PR #913, opened 2021-01-28
DUOS-944[risk=no]Bug fix the incorrect % fields

raejohanek:
Scope:
-set empty column for revealAlgorithmDecision percentages instead of 0
-fix math on setPercentAgreementAlgorithm
-fix math on setPercentSRPAccurate
-fix math on setPercentReviewed

Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-944

In all cases:


src/main/java/org/broadinstitute/consent/http/models/DacDecisionMetrics.java
```diff
@@ -202,16 +187,39 @@ public Integer getPercentAgreementAlgorithm() {
...
+    List<DarDecisionMetrics> srpMetricsDenominator =
+      completedDarMetrics.stream()
+        .filter(m -> Objects.nonNull(m.getSrpDecision()))
+        .collect(Collectors.toList());
+
+    List<DarDecisionMetrics> srpMetricsNumerator =
+      srpMetricsDenominator.stream()
+        .filter(m -> m.getSrpDecision().equalsIgnoreCase("yes"))
+        .collect(Collectors.toList());
+    int percentSrp = (int) (((double) srpMetricsNumerator.size() / (double) srpMetricsDenominator.size()) * 100);
```
rushtong: One suggestion for refactoring. Since we're calculating the percentages multiple times (3 times in this PR), it is probably a good idea to extract that out to a private method that takes two `int`s and does the same thing in terms of casting to doubles (nice catch with `long`->`double`, by the way) and then returning the value in `int` form. That way we're not leaving it up to each method to figure out the right implementation via copy/paste.

rushtong (approved):
One suggestion for refactoring, but otherwise this looks good! 👍🏽

PR #914, opened 2021-01-29
DUOS-946[risk=no]Add count of unique users on approved dars by dac

raejohanek:
ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-946

SCOPE:
-add properties to DAR tracking internal collaborators and lab collaborators (researchers and PIs are not currently being returned on DAR API)
-add count of users to darmetrics
-calculate average users on approved dars in dacmetrics using data from darmetrics

In all cases:


src/main/java/org/broadinstitute/consent/http/db/mapper/DataAccessRequestMapper.java
```diff
@@ -27,6 +28,8 @@ public DataAccessRequest map(ResultSet resultSet, StatementContext statementCont
         dar.setSortDate(resultSet.getTimestamp("sort_date"));
         dar.setSubmissionDate(resultSet.getTimestamp("submission_date"));
         dar.setUpdateDate(resultSet.getTimestamp("update_date"));
+        dar.setCountCollaborators(resultSet.getArray("internalCollaborators"));
+        dar.setCountCollaborators(resultSet.getArray("labCollaborators"));
```
rushtong: This class loops over the results of a query and tries to extract the query columns from the results array. In these lines, you're expecting that query results will contain columns for `internalCollaborators` and `labCollaborators`, but those values don't exist in the database as columns on the data access request table, which is the root of the failures. These values have to be extracted from the `data` column but this is probably not the right place to look for those values.

A second bit of feedback here - you're calling the same method (`setCountCollaborators`) with two different things, so if this did work, it would be overwriting the first with the second.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
```diff
@@ -42,6 +46,8 @@
 
   @JsonProperty public Timestamp updateDate;
 
+  @JsonProperty public ArrayList<String> countCollaborators = new ArrayList<>();
+
```
rushtong: I don't think we need this method. You can get the collaborators by looking at the `DataAccessRequestData` object instead.

rushtong (changes requested):
See comments inline

src/main/java/org/broadinstitute/consent/http/models/DacDecisionMetrics.java
```diff
@@ -149,6 +159,20 @@ private void setDatasetCount(List<DataSetDTO> datasets) {
     }
   }
 
+  private void setAverageCountUniqueUser(List<DarDecisionMetrics> metrics) {
+    Integer totalUsers = 0;
+    Integer totalDars = 0;
+    for(DarDecisionMetrics m : metrics) {
+      totalDars++;
+      totalUsers = totalUsers + m.getCountUniqueUsers();
+    }
```
rushtong: `totalDars` looks like `metrics.size()`. I don't think you need to increment it in the loop.

rushtong: Without the unnecessary increment, what this loop is doing is just summing up a bunch of integers:
```
    for(DarDecisionMetrics m : metrics) {
        totalUsers = totalUsers + m.getCountUniqueUsers();
    }
```
This can be simplified with a stream -> map -> sum like so:
```
    int totalUsers = metrics.stream()
        .map(DarDecisionMetrics::getCountUniqueUsers)
        .mapToInt(Integer::intValue)
        .sum();
```
The nice thing about this approach is that it is functional. There is no mutation of variables this way which reduces complexity.

src/main/java/org/broadinstitute/consent/http/models/DacDecisionMetrics.java
```diff
@@ -149,6 +159,20 @@ private void setDatasetCount(List<DataSetDTO> datasets) {
...
+    if (totalDars > 0) {
+      this.averageCountUniqueUser = Math.toIntExact(Math.round((double) totalUsers / (double) totalDars));
+    }
```
rushtong: Since we're averaging, there is a simpler way to average elements of a list with streams, which would remove the need for any of the above code:
```
    OptionalDouble averageUniqueUsers = metrics.stream()
        .map(DarDecisionMetrics::getCountUniqueUsers)
        .mapToInt(Integer::intValue)
        .average();
    if (averageUniqueUsers.isPresent()) {
      this.averageCountUniqueUser = (int) averageUniqueUsers.getAsDouble();
    }
```
If you want to take it even further and remove the intermediate variable, you can inline the whole thing:
```
    this.averageCountUniqueUser = (int) metrics.stream()
        .map(DarDecisionMetrics::getCountUniqueUsers)
        .mapToInt(Integer::intValue)
        .average()
        .orElse(0);
```
which is slightly different in that it will always set a `0` value if there are no unique users.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
```diff
@@ -114,7 +116,28 @@ public void setUpdateDate(Timestamp updateDate) {
     this.updateDate = updateDate;
   }
 
-  /**
+  public Integer getCountCollaborators() {
+    ArrayList<String> group = new ArrayList<>();
+    if (Objects.nonNull(getUserId())) {
+      group.add(getUserId().toString());
```
rushtong: I don't understand this. `getUserId` returns the user id of the researcher who created the DAR, not a collaborator. I don't think you want to count this value.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
```diff
@@ -114,7 +116,28 @@ public void setUpdateDate(Timestamp updateDate) {
...
+    }
+
+    List<Collaborator> collabs = data.getInternalCollaborators();
+
+    if (Objects.nonNull(collabs)) {
+      collabs.addAll(data.getLabCollaborators());
```
rushtong: What we should be doing here is concatenating the two lists into a single one instead of creating holding variables. With a concatenated list, you can stream over them and test for uniqueness in a more legible manner.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
```diff
@@ -114,7 +116,28 @@ public void setUpdateDate(Timestamp updateDate) {
...
+
+      for (Collaborator c: collabs) {
+        String userID = c.getUuid();
+        if (!group.contains(userID)) {
+          group.add(userID);
+        }
```
rushtong: This will likely not do what it seems to be doing. The UUID is only used for the UI. There could be two `j.smith@someplace.edu` collaborators with different UUIDs, so this would consider the to be two different people. What you should probably be looking at is a lower-cased email address instead. Even this isn't ideal since a single person can be referenced by multiple emails by different researchers, but it's the best thing we have with collaborators.

rushtong (changes requested):
See comments inline, thank you!

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
raejohanek: I was talking to Johnathan and he said we are looking to include the researcher who created the DAR in the count of unique users

rushtong: If that's the case, then we definitely need to collect all the emails as a list and test for unique counts on that. The naming is also confusing here since it's no longer `Collaborators` but any associated user to the DAR. We should name this to be more reflective of what it really is.

raejohanek: Okay, good idea, where could we find the email for the researcher?

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
```diff
@@ -114,7 +117,30 @@ public void setUpdateDate(Timestamp updateDate) {
...
+      if (Objects.nonNull(data.getAcademicEmail())) {
+        emails.add(data.getAcademicEmail());
+      }
+      if (Objects.nonNull(data.getPiEmail())) {
+        emails.add(data.getPiEmail());
+      }
+      emails = emails.stream().map(String::toLowerCase).distinct().collect(Collectors.toList());
+    }
+    return Objects.nonNull(emails) ? emails.size() : 0;
+  }
```
rushtong: This method really should be in the metrics class. In general, we don't need to know this value for a DAR. The only reason for it is for metrics, so let's keep that code close to where it is used.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
```diff
@@ -114,7 +117,30 @@ public void setUpdateDate(Timestamp updateDate) {
...
+        collabs.addAll(data.getLabCollaborators());
+      }
+      emails = collabs.stream().map(collaborator -> Objects.requireNonNull(collaborator).getEmail()).collect(Collectors.toList());
+
+      if (Objects.nonNull(data.getAcademicEmail())) {
+        emails.add(data.getAcademicEmail());
+      }
+      if (Objects.nonNull(data.getPiEmail())) {
+        emails.add(data.getPiEmail());
+      }
```
rushtong: We don't want to get `User` information from the data access request, we want to get it from the `User`'s `ResearcherProperty` values. In the ticket, I suggested that we pull this feature out into a separate ticket so we can address the design in a better way. Let's revert this code and make sure we're doing the right thing in the metrics class, which we're not doing now. For example, we need the `User` which we don't have in the constructor right now.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
```diff
@@ -114,7 +117,30 @@ public void setUpdateDate(Timestamp updateDate) {
...
+    List<String> emails = null;
+    if (Objects.nonNull(data)) {
+      ArrayList<Collaborator> collabs = new ArrayList<>();
+      if (Objects.nonNull(data.getInternalCollaborators())) {
+        collabs.addAll(data.getInternalCollaborators());
+      }
+      if (Objects.nonNull(data.getLabCollaborators())) {
+        collabs.addAll(data.getLabCollaborators());
+      }
+      emails = collabs.stream().map(collaborator -> Objects.requireNonNull(collaborator).getEmail()).collect(Collectors.toList());
```
rushtong: This can be streamlined and a little more legible with some lambda cleanup:
```
      emails = collabs
          .stream()
          .filter(Objects::nonNull)
          .map(Collaborator::getEmail)
          .filter(Objects::nonNull)
```

rushtong (changes requested):
Definitely making progress! See comments inline.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
```diff
@@ -114,7 +114,7 @@ public void setUpdateDate(Timestamp updateDate) {
     this.updateDate = updateDate;
   }
 
-  /**
+    /**
```
rushtong: suggested that you revert this

src/main/java/org/broadinstitute/consent/http/models/DarDecisionMetrics.java
```diff
@@ -113,6 +120,28 @@ private void setDatasetId(DataSet dataset) {
     }
   }
 
+  private void setCountUniqueUser(DataAccessRequest dar) {
+    List<String> emails = null;
+    if (Objects.nonNull(dar.data)) {
```
rushtong: We should always call the getter and not pull the field directly.
```suggestion
    if (Objects.nonNull(dar.getData())) {
```

src/main/java/org/broadinstitute/consent/http/models/DarDecisionMetrics.java
```diff
@@ -113,6 +120,28 @@ private void setDatasetId(DataSet dataset) {
...
+        collabs.addAll(dar.data.getLabCollaborators());
+      }
+      emails = collabs
+        .stream()
+        .filter(Objects::nonNull)
+        .map(Collaborator::getEmail)
+        .filter(Objects::nonNull)
+        .collect(Collectors.toList());
+    }
+    this.countUniqueUser = Objects.nonNull(emails) ? emails.size() : 0;
```
rushtong: There are at least two extraneous holder variables that can be extracted out. `emails` and `collabs` are both unnecessary:
```
    if (Objects.nonNull(dar.getData())) {
      this.countUniqueUser = (int) Streams
          .concat(
              dar.getData().getLabCollaborators().stream(),
              dar.getData().getInternalCollaborators().stream())
          .filter(Objects::nonNull)
          .map(Collaborator::getEmail)
          .filter(Objects::nonNull)
          .map(String::toLowerCase)
          .distinct()
          .count();
    } else {
      this.countUniqueUser = 0;
    }
```

That can also be re-written to be a one-liner ternary:
```
    this.countUniqueUser =
        (Objects.nonNull(dar.getData())) ? 
            (int) Streams
              .concat(
                  dar.getData().getLabCollaborators().stream(),
                  dar.getData().getInternalCollaborators().stream())
              .filter(Objects::nonNull)
              .map(Collaborator::getEmail)
              .filter(Objects::nonNull)
              .map(String::toLowerCase)
              .distinct()
              .count()
            : 0;
```

rushtong (changes requested):
Looking good, thanks for the changes so far! One more suggestion inline.

rushtong (approved):
👍🏽  Thanks for the changes!

PR #915, opened 2021-02-01
DUOS-941[risk=no] Fix SRPs that all show mis-matching results

raejohanek:
Scope:
-ensure election is correct type
-ensure dac decision has already been made

Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-941

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/DarDecisionMetrics.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/DarDecisionMetrics.java
@@ -230,7 +230,7 @@ public class DarDecisionMetrics implements DecisionMetrics {
    * @param election The election
    */
   private void setSrpDecision(Election election) {
-    if (Objects.nonNull(election)) {
+    if (Objects.nonNull(dacDecision) && Objects.nonNull(election) && election.getElectionType().equals("RP")) {
       Boolean rpVote =
         Objects.nonNull(election.getFinalVote())
           ? election.getFinalVote()
--- a/src/main/java/org/broadinstitute/consent/http/models/DarDecisionMetrics.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/DarDecisionMetrics.java
@@ -230,7 +230,7 @@ public class DarDecisionMetrics implements DecisionMetrics {
    * @param election The election
    */
   private void setSrpDecision(Election election) {
-    if (Objects.nonNull(dacDecision) && Objects.nonNull(election) && election.getElectionType().equals("RP")) {
+    if (Objects.nonNull(dacDecision) && Objects.nonNull(election) && election.getElectionType().equals(ElectionType.RP.getValue())) {
       Boolean rpVote =
         Objects.nonNull(election.getFinalVote())
           ? election.getFinalVote()
--- a/src/main/java/org/broadinstitute/consent/http/models/DarDecisionMetrics.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/DarDecisionMetrics.java
@@ -230,7 +230,7 @@ public class DarDecisionMetrics implements DecisionMetrics {
    * @param election The election
    */
   private void setSrpDecision(Election election) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/consent/http/models/DarDecisionMetrics.java
```diff
@@ -230,7 +230,7 @@ public String getSrpDecision() {
    * @param election The election
    */
   private void setSrpDecision(Election election) {
-    if (Objects.nonNull(election)) {
+    if (Objects.nonNull(dacDecision) && Objects.nonNull(election) && election.getElectionType().equals("RP")) {
```
rushtong: Try to avoid raw strings in `.equals()`. We have an enum for the election types:
```suggestion
    if (Objects.nonNull(dacDecision) && Objects.nonNull(election) && election.getElectionType().equals(ElectionType.RP.getValue())) {
```

rushtong: I'm more concerned here with the fact that this method needs to make the check at all. In the setter, we are setting it with a `rpElection` so we should already know that this is an RP election and that it's non-null. Look at the constructor: 
```
DarDecisionMetrics(
    DataAccessRequest dar,
    Dac dac,
    DataSet dataset,
    Election accessElection,
    Election rpElection,
    Match match)
```
where we are clearly expecting an RP election. If we reach this point in the code, something went really wrong in the constructor. This happens in `MetricsService` where we are also clearly trying to pass in an RP election. Looking more closely there, I see that we find it like this:
```
          Optional<Election> rpElection =
            elections.stream()
              .filter(
                e ->
                  e.getReferenceId()
                    .equalsIgnoreCase(dataAccessRequest.getReferenceId()))
              .filter(e -> e.getElectionType().equalsIgnoreCase(ElectionType.RP.getValue()))
              .findFirst();
```
So, the enum check that is in your method shouldn't be necessary. We should never get an incorrect election type here.

rushtong: Lastly, in your testing, were the mismatches simply an artifact of either a null `dacDecision` or null `election`?

rushtong (changes requested):
I think we need to dig a little further to solve this bug.

src/main/java/org/broadinstitute/consent/http/models/DarDecisionMetrics.java
raejohanek: So, I think a lot of there from nulls where the decisions were not populating at all, but also, a lot of them return No, when they should be Yes, which is not from a null. Either the value if finalVote or finalAccessVote is false when it should be true. I see that the finalAccessVote is passed in to an Election in the constructor, but I am not so sure where the finalVote value gets assigned

rushtong: > Either the value if finalVote or finalAccessVote is false when it should be true.

I think I know what this might be. It's likely that we're not getting the right final vote/final access vote in the election query. We recently updated the get-election-by-reference-id DAO call to do the right thing, but I'm guessing the underlying election DAO query here is not doing the same thing. Basically, the final vote for an election depends on the type of the election. Let's drill down to the underlying query and see what that is.

raejohanek: I was thinking of adding a check that said if either one of those values (finalVote or finalAccessVote) are true, set the SRPdecision to be true. Bc the issue could be that finalVote isn't null, but always false, and since it takes precedence over finalAccessVote in the code currently, they would always be returning No.

raejohanek: Okay, good idea!!!! Thank you

> I think I know what this might be. It's likely that we're not getting the right final vote/final access vote in the election query. We recently updated the get-election-by-reference-id DAO call to do the right thing, but I'm guessing the underlying election DAO query here is not doing the same thing. Basically, the final vote for an election depends on the type of the election. Let's drill down to the underlying query and see what that is.

rushtong (approved):
👍🏽

PR #929, opened 2021-02-11
DUOS-974[risk=no]Remove admin console "Add Datasets" screens

raejohanek:
## Scope
-remove unused endpoint from Consent

## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-974

In all cases:


src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -171,56 +171,6 @@ public Response updateDataset(@Auth AuthUser authUser, @Context UriInfo info, @P
...
-                if (inputFile != null) {
-                    inputFile.delete();
-                }
-            }
-        }
-
-        errors.add("The file type is not the expected one. Please download the Dataset Spreadsheet Model from the 'Add Datasets' window.");
-        return Response.status(Response.Status.BAD_REQUEST).entity(errors).build();
-    }
-
```
rushtong, line 223: After removing this method, there are quite a few unused imports left over, please optimize those.

rushtong: `DataSetAPI.overwrite` is now unused ... that and the implementation should be removed as well.

rushtong: `DataSetAPI.create` is now unused, please remove that and it's implementation method.

rushtong: I don't see this API documented in `api-docs.yaml`, so we're good there. Usually, we need to remove the doc there as well.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -201,74 +201,6 @@ public void testUpdateDatasetNoContent() {
...
-        result.setDatasets(Collections.emptyList());
-        result.setErrors(Collections.emptyList());
-        when(api.overwrite(any(), any())).thenReturn(result);
-        when(api.create(any(), any())).thenReturn(result);
-
-        initResource();
-        Response response = resource.createDataSet(is, fileDataBodyPart, 1, false);
-        assertEquals(200, response.getStatus());
-    }
-
```
rushtong, line 271: As a result of removing these tests, there is a good amount of unused imports and unused static strings. Please remove those as well.

rushtong: Unfortunately, even though `DataAccessRequestAPI dataAccessRequestAPI` isn't used in the test, it needs to be mocked for the resource class, so we need to keep that.

rushtong (changes requested):
There are some unused code paths that need to be traced through and removed in addition to the resource methods. Please remove those and make sure to optimize imports. Thank you!

src/main/java/org/broadinstitute/consent/http/service/DatabaseDataSetAPI.java
```diff
@@ -241,237 +123,4 @@ public DataSet updateNeedsReviewDataSets(Integer dataSetId, Boolean needsApprova
...
-            auditProperties.add(new DataSetAuditProperty(
-                    property.getPropertyId(),
-                    property.getDataSetId(),
-                    property.getPropertyKey(),
-                    property.getPropertyValue(),
-                    property.getCreateDate(),
-                    dataSetAuditId));
-        });
-        return auditProperties;
-    }
```
rushtong, line 476: Wow, that's a lot of deleted code, nice work!
Also, `UserRoleDAO userRoleDAO` is unused after this.

As more fallout, there are now unused queries in `DataSetDAO`
* findDataSetByObjectId
* updateAllByObjectId
* searchDataSetsByObjectIdList
* searchDataSetsIdsByObjectIdList
* getDataSetsForNameList
* getAssociationsForObjectIdList
* getConsentAssociationByObjectId
* searchDataSetsByNameList
* searchByNameIdList
* getDataSetsWithValidNameForObjectIdList

rushtong: Additionally, `DataSetAuditDAO.insertDataSetAuditProperties` is now unused.

rushtong (changes requested):
Much better! There is still fallout in DAO calls. We can remove them as well if they're unused. One more thing to be aware of is that we have tests for DAO calls, so sometimes the *only* usage is a test usage, so we need to be on the lookout for that and remove those as well.

rushtong (approved):
Very nice work, thanks for digging into all of the affected paths.

PR #934, opened 2021-02-16
DUOS-715[risk=no]Tech Debt: Delete help table

raejohanek:
Scope:
add changeset 60 to drop the help_report table 
dropping the help_report table removes the help_report_report_id_seq sequence as well
add the changeset to the master file

Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-715

In all cases:

### DIFF
--- a/migration/postgres/README.md
+++ b/migration/postgres/README.md
@@ -57,7 +57,6 @@ consent.accesselection_consentelection          0        168     1.2 kB         
                consent.researchpurpose          0          0                     0.738s
                      consent.user_role          0        100     1.9 kB          0.797s
                           consent.vote          0       2327   145.6 kB          0.884s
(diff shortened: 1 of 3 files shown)


migration/postgres/README.md
```diff
@@ -247,9 +245,6 @@ select count(*) from user_role
 count(*)
 107
 
-select count(*) from help_report
-count(*)
-13
```
rushtong: Please revert these changes to this file, it's an historical doc that we should keep around in case we need double check our validation.

rushtong (changes requested):
One suggestion inline, but otherwise looks good!

rushtong (approved):
Looks great, thanks for the changes!

PR #937, opened 2021-02-17
[DUOS-936][risk=low] Run automation tests off of fresh db

rfricke-asymmetrik:
**Addresses**
https://broadworkbench.atlassian.net/browse/DUOS-936

**Changes**

- Make Docker Compose file to run a fresh postgres db and consent api
- Load test data into postgres db using custom liquibase changeset
- Update jenkins scripts to compose api and db containers and run automation container against them

In all cases:

### DIFF
--- a/automation/consent-automation-config.yml
+++ b/automation/consent-automation-config.yml
@@ -7,8 +7,8 @@ server:
     port: 8000
 database:
   driverClass: org.postgresql.Driver
-  user: consent
-  password: consent
+  user: automation
+  password: automation
   url: jdbc:postgresql://consent-postgres:5432/consent
   initialSize: 20
   minSize: 20
--- a/automation/consent-automation-config.yml
+++ b/automation/consent-automation-config.yml
@@ -1,0 +1,52 @@
+server:
+  type: simple
+  applicationContextPath: /
+  adminContextPath: /admin
+  connector:
+    type: http
+    port: 8000
+database:
+  driverClass: org.postgresql.Driver
+  user: consent
+  password: consent
+  url: jdbc:postgresql://consent-postgres:5432/consent
+  initialSize: 20
+  minSize: 20
+  maxSize: 128
+  validationQuery: SELECT 1
+googleStore:
+  password: password
+  bucket: bucket
+  endpoint: endpoint
+# Logging settings
+logging:
+  # default logging level
+  level: WARN
+  # logger specific levels
+  loggers:
+    "org.broadinstitute.consent.http.cloudstore.GCSService": OFF
+services:
+    ontologyURL: https://consent-ontology.dsde-dev.broadinstitute.org/
+    localURL: http://localhost:8000/#/
+mailConfiguration:
+  activateEmailNotifications: false
+  googleAccount: test@gmail.com
+  sendGridApiKey: test-key
+freeMarkerConfiguration:
+  templateDirectory: /freemarker
+  defaultEncoding: UTF-8
+googleAuthentication:
+  clientId: clientId
+basicAuthentication:
+  users:
+    - user: testuser
+      password: testpassword
+    - user: testuser2
+      password: testpassword2
+storeOntology:
+  bucketSubdirectory: bucket
+  configurationFileName: configuration
+elasticSearch:
+  servers:
+    - localhost
+  indexName: iName
--- a/docker-compose-automation.yml
+++ b/docker-compose-automation.yml
@@ -1,0 +1,22 @@
+version: '3.8'
+services:
+  consent-postgres:
+    image: postgres:11.6-alpine
+    environment:
+      - POSTGRES_USER=consent
+      - POSTGRES_PASSWORD=consent
+      - POSTGRES_DB=consent
+    ports:
+      - 5432:5432
+  consent-api:
+    image: consent
+    depends_on: 
+      - consent-postgres
+    build:
+      context: ./
+      dockerfile: Dockerfile-automation
+    ports:
+      - 8000:8000
+    environment:
+      - CONSENT_CHANGELOG_FILE=test-data.xml
+    
--- a/docker-compose-automation.yml
+++ b/docker-compose-automation.yml
@@ -3,8 +3,8 @@ services:
   consent-postgres:
     image: postgres:11.6-alpine
     environment:
-      - POSTGRES_USER=consent
-      - POSTGRES_PASSWORD=consent
+      - POSTGRES_USER=automation
+      - POSTGRES_PASSWORD=automation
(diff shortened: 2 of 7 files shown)


docker-compose-automation.yml
```diff
@@ -0,0 +1,22 @@
+version: '3.8'
+services:
+  consent-postgres:
+    image: postgres:11.6-alpine
+    environment:
+      - POSTGRES_USER=consent
+      - POSTGRES_PASSWORD=consent
```
rushtong: Can we change this to something less realistic-sounding? Something like `automation` or `testing`?

src/main/resources/test-data.xml
```diff
@@ -0,0 +1,201 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+    <include file="changelog-master.xml" relativeToChangelogFile="true"/>
```
rushtong, line 5: Took me a bit to get how we were pulling in all of the existing changelogs as well. Nice approach.

src/main/resources/test-data.xml
```diff
@@ -0,0 +1,201 @@
...
+
+    <!--insert dac 1-->
+    <changeSet author="rfricke" id="test">
+        <validCheckSum>ANY</validCheckSum>
```
rushtong: I don't think you'll need any of these lines since we're always working off of a fresh database. These are required more for the case when you need to update an existing changeset file that's already been run and exists in the changelog table. That table stores a checksum of the changeset and usually throws an error if the checksums are different. This drops that constraint and ignores modified changeset files.

automation/consent-automation-config.yml
```diff
@@ -0,0 +1,52 @@
...
+  password: password
+  bucket: bucket
+  endpoint: endpoint
+# Logging settings
+logging:
+  # default logging level
+  level: WARN
+  # logger specific levels
+  loggers:
+    "org.broadinstitute.consent.http.cloudstore.GCSService": OFF
```
rushtong, line 27: Might be worth setting this to debug and removing the custom loggers. We can dial it back if the logs get too crazy.

automation/consent-automation-config.yml
```diff
@@ -0,0 +1,52 @@
...
+services:
+    ontologyURL: https://consent-ontology.dsde-dev.broadinstitute.org/
+    localURL: http://localhost:8000/#/
+mailConfiguration:
+  activateEmailNotifications: false
```
rushtong, line 32: Nice

automation/consent-automation-config.yml
```diff
@@ -0,0 +1,52 @@
...
+  googleAccount: test@gmail.com
+  sendGridApiKey: test-key
+freeMarkerConfiguration:
+  templateDirectory: /freemarker
+  defaultEncoding: UTF-8
+googleAuthentication:
+  clientId: clientId
```
rushtong, line 39: This is only used for the swagger UI so it's fine to have a dummy value here.

automation/consent-automation-config.yml
```diff
@@ -0,0 +1,52 @@
...
+      password: testpassword
+    - user: testuser2
+      password: testpassword2
+storeOntology:
+  bucketSubdirectory: bucket
+  configurationFileName: configuration
+elasticSearch:
+  servers:
+    - localhost
+  indexName: iName
```
rushtong, line 52: This will cause the status endpoint that checks on ES to report as down. An alternative is to also spin up an instance of ES with an empty index. That can be done as part of an additional story, however, and not required for this PR.

automation/consent-automation-config.yml
```diff
@@ -0,0 +1,52 @@
...
+  clientId: clientId
+basicAuthentication:
+  users:
+    - user: testuser
+      password: testpassword
+    - user: testuser2
+      password: testpassword2
+storeOntology:
+  bucketSubdirectory: bucket
+  configurationFileName: configuration
```
rushtong, line 48: This will cause the status endpoint to fail on GCS and also fail any functions that call GCS. Not necessary to fix for this PR, but we do eventually need a solution so we can do full testing.

rushtong: As a side note, we can pass in configs on the command line that override what exists in the config file. I did this for our [consent k8s implementation](https://github.com/broadinstitute/terra-helm/blob/master/charts/consent/templates/deployment.yaml#L82). See the values prefixed with `-Ddw.*` ... each of those references a config.

src/main/java/org/broadinstitute/consent/http/ConsentApplication.java
```diff
@@ -368,8 +368,17 @@ private void initializeLiquibase(ConsentConfiguration config) throws LiquibaseEx
...
         );
         Database database = DatabaseFactory.getInstance().findCorrectDatabaseImplementation(new JdbcConnection(connection));
-        Liquibase liquibase = new Liquibase("changelog-master.xml", new ClassLoaderResourceAccessor(), database);
+        Liquibase liquibase = new Liquibase(liquibaseFile(), new ClassLoaderResourceAccessor(), database);
         liquibase.update(new Contexts(), new LabelExpression());
     }
 
+    private String liquibaseFile() {
+        String changeLogFile = System.getenv("CONSENT_CHANGELOG_FILE");
+        if (changeLogFile == null || changeLogFile.trim().isEmpty()) {
```
rushtong: minor nit, I prefer `Objects.nonNull()` for null checks
```suggestion
        if (Objects.nonNull(changeLogFile) || changeLogFile.trim().isEmpty()) {
```

src/main/java/org/broadinstitute/consent/http/ConsentApplication.java
```diff
@@ -368,8 +368,17 @@ private void initializeLiquibase(ConsentConfiguration config) throws LiquibaseEx
...
+            changeLogFile = "changelog-master.xml";
+        }
+        LOGGER.info(changeLogFile);
```
rushtong: Can we expand this a little? Something like
```suggestion
        LOGGER.info("Initializing db with: " + changeLogFile);
```

rushtong (approved):
Great work here. See minor suggestions inline. Once addressed, 👍🏽

rushtong (approved):
Thanks for the changes!

PR #941, opened 2021-02-19
DUOS-653[risk=no]Update SC ID to Dataset ID in Data Owner approval email

raejohanek:
Scope:
-change SC ID to Dataset ID

Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-653

In all cases:


src/main/resources/freemarker/owner-dar-approved.html
```diff
@@ -95,13 +95,13 @@
...
                     <tr>
-                        <th align="left" style="font-family: 'Roboto', sans-serif; font-style: italic; font-size: 15px; color: #00609f; border-bottom: 1px solid #cccccc; padding: 0 0 5px 5px; text-align: left;">SC ID</th>
+                        <th align="left" style="font-family: 'Roboto', sans-serif; font-style: italic; font-size: 15px; color: #00609f; border-bottom: 1px solid #cccccc; padding: 0 0 5px 5px; text-align: left;">Dataset ID</th>
                         <th align="left" style="font-family: 'Roboto', sans-serif; font-style: italic; font-size: 15px; color: #00609f; border-bottom: 1px solid #cccccc; padding: 0 0 5px 5px; text-align: left;">Dataset Name</th>
                     </tr>
                     <#list dsl>
                         <#items as ds>
                             <tr style="font-family: 'Roboto', sans-serif;">
-                                <td style="font-family: 'Roboto', sans-serif; padding: 5px; border-bottom: 1px solid #cccccc;">${ds.objectId}</td>
+                                <td style="font-family: 'Roboto', sans-serif; padding: 5px; border-bottom: 1px solid #cccccc;">${ds.dataSetId}</td>
```
rushtong: This is not the same dataset id that the story is asking for. The ID Jonathan is looking for is a manipulated version of the `alias` property. Look at `DatasetUtil.parseAlias()` for how that value is generated.

rushtong (changes requested):
One minor change, but looking good!

src/main/java/org/broadinstitute/consent/http/models/DataSet.java
```diff
@@ -168,4 +174,10 @@ public Integer getAlias() {
     public void setAlias(Integer alias) {
         this.alias = alias;
     }
+
+    public String getAliasParsed() {return aliasParsed; }
```
rushtong: You could simplify the class a little by doing the parsing here in the getter and avoid the setter altogether:
```suggestion
    public String getAliasParsed() {
        return DatasetUtil.parseAlias(alias); 
    }
```

rushtong: Another issue we're working around is the fact that the UI refers to this value as "Dataset ID" instead of alias. Since we already have a `datasetId`, can we call this instead a `datasetIdentifier` (I'm open to other suggestions as well, i.e. `uniqueDatasetIdentifier` or anything similar that conveys the UI presentation)? I'd like to get to a place where the naming is more intuitive between this class and what we use in the display.

src/main/java/org/broadinstitute/consent/http/util/DatasetUtil.java
```diff
@@ -1,31 +1,14 @@
...
-        else if(aliasUid < 100000) {
-            alias = PREFIX + "0" + aliasUid;
-        }
-        else {
-            alias = PREFIX + aliasUid;
-        }
+        String alias = PREFIX + StringUtils.leftPad(aliasUid.toString(), 6, "0");
         return alias;
-
     }
```
rushtong: Fantastic simplification. This also leaves me with the impression that this class is completely useless. The dataset class itself should just return this value from the current `getAliasParsed()` method.

rushtong: I think refactoring this class away might be beyond the scope of the PR, but if you're willing to take that one, please do. Otherwise, we should file a ticket for the refactoring.

src/main/java/org/broadinstitute/consent/http/db/mapper/DataSetMapper.java
```diff
@@ -33,6 +34,7 @@ public DataSet map(ResultSet r, StatementContext ctx) throws SQLException {
       }
       dataset.setActive(r.getBoolean("active"));
       dataset.setAlias(r.getInt("alias"));
+      dataset.setAliasParsed(DatasetUtil.parseAlias(r.getInt("alias")));
```
rushtong: With my other suggestions, you wouldn't need any setter here.

rushtong (commented):
See suggestions inline. This is the right approach (thank you!). I think we can clean it up a little more.

src/main/java/org/broadinstitute/consent/http/util/DatasetUtil.java
raejohanek: Sounds good I'll make a new ticket and PR, and remove the class there

rushtong (approved):
👍🏽

PR #945, opened 2021-02-22
DUOS-1046[risk=no]Remove DatasetUtils class

raejohanek:
Scope:
Delete datasetUtil
Delete datasetUtilTest
Refactor usages of parseAlias()

Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-1046

In all cases:


src/main/java/org/broadinstitute/consent/http/models/dto/DataSetDTO.java
```diff
@@ -155,7 +156,7 @@ public Integer getDataSetId() {
     }
 
     public void setAlias(Integer alias) {
-        this.alias = DatasetUtil.parseAlias(alias);
+        this.alias = "DUOS-" + StringUtils.leftPad(alias.toString(), 6, "0");
     }
```
rushtong: We should be referencing the Dataset class here if we need it. One way to do that is to make it a public static method there so we don't have to copy the guts of that method in different places.

src/main/java/org/broadinstitute/consent/http/models/dto/DataOwnerCase.java
```diff
@@ -93,7 +93,7 @@ public String getAlias() {
     }
 
     public void setAlias(Integer alias) {
-        this.alias = DatasetUtil.parseAlias(alias);
+        this.alias = "DUOS-" + StringUtils.leftPad(alias.toString(), 6, "0");
 
     }
```
rushtong: See other comment, we should be getting this from the Dataset, not copying the content of the method.

src/main/java/org/broadinstitute/consent/http/service/DataAccessReportsParser.java
```diff
@@ -103,7 +102,7 @@ private void addDARLine(FileWriter darWriter, Document dar, String customContent
         List<Integer> dataSetIds = DarUtil.getIntegerList(dar, DarConstants.DATASET_ID);
         List<String> dataSetUUIds = new ArrayList<>();
         for(Integer id : dataSetIds) {
-            dataSetUUIds.add(DatasetUtil.parseAlias(id));
+            dataSetUUIds.add("DUOS-" + StringUtils.leftPad(id.toString(), 6, "0"));
```
rushtong: similar feedback here.

src/main/java/org/broadinstitute/consent/http/models/DataSet.java
```diff
@@ -168,4 +171,7 @@ public Integer getAlias() {
     public void setAlias(Integer alias) {
         this.alias = alias;
     }
+
+    public String getDatasetIdentifier() {return PREFIX + StringUtils.leftPad(alias.toString(), 6, "0"); }
```
rushtong: One way to make this more generic so other classes can use it is to move it to a public static method, then call that method from inside here to get what you want. For example, 
```
public String getDatasetIdentifier() {
    return parseAliasToIdentifier(this.getAlias()); 
}

public static String parseAliasToIdentifier(Integer alias) {
    return PREFIX + StringUtils.leftPad(alias.toString(), 6, "0"); 
}
```
That way, other classes that don't have a full `Dataset` object can use this method the same way `Dataset` does.

src/main/resources/freemarker/owner-dar-approved.html
```diff
@@ -95,13 +95,13 @@
...
                     <tr>
-                        <th align="left" style="font-family: 'Roboto', sans-serif; font-style: italic; font-size: 15px; color: #00609f; border-bottom: 1px solid #cccccc; padding: 0 0 5px 5px; text-align: left;">SC ID</th>
+                        <th align="left" style="font-family: 'Roboto', sans-serif; font-style: italic; font-size: 15px; color: #00609f; border-bottom: 1px solid #cccccc; padding: 0 0 5px 5px; text-align: left;">Dataset ID</th>
                         <th align="left" style="font-family: 'Roboto', sans-serif; font-style: italic; font-size: 15px; color: #00609f; border-bottom: 1px solid #cccccc; padding: 0 0 5px 5px; text-align: left;">Dataset Name</th>
                     </tr>
                     <#list dsl>
                         <#items as ds>
                             <tr style="font-family: 'Roboto', sans-serif;">
-                                <td style="font-family: 'Roboto', sans-serif; padding: 5px; border-bottom: 1px solid #cccccc;">${ds.objectId}</td>
+                                <td style="font-family: 'Roboto', sans-serif; padding: 5px; border-bottom: 1px solid #cccccc;">${ds.getDatasetIdentifier()}</td>
```
rushtong: Changes to this class are outside the scope of the PR and should be in your other PR. Wait until this one is merged and then you can simplify your other PR.

rushtong (changes requested):
See comments inline.

src/main/java/org/broadinstitute/consent/http/models/dto/DataSetDTO.java
```diff
@@ -155,8 +156,7 @@ public Integer getDataSetId() {
...
 
     public void setAlias(Integer alias) {
-        this.alias = DatasetUtil.parseAlias(alias);
-    }
+        this.alias = DataSet.parseAliasToIdentifier(alias); }
```
rushtong: minor nit: add newline between semi-colon and closing brace.

rushtong (approved):
One minor suggestion inline, but otherwise looks great 👍🏽

PR #951, opened 2021-02-26
[DUOS-1057][risk=?] Refactor DAR Use Restriction Logic

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1057
See also: https://github.com/DataBiosphere/duos-ui/pull/910

## Changes
* rm unnecessary `hasUseRestriction` API method
* rm all references to restriction fields
* refactor how we generate the UseRestriction for a DAR when generating a match

---

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
@@ -46,7 +46,7 @@ import org.broadinstitute.consent.http.util.DarConstants;
 import org.broadinstitute.consent.http.util.DarUtil;
 import org.bson.Document;
 
-@Path("{api : (api/)?}dataRequest/{requestId}/vote")
+@Path("api/dataRequest/{requestId}/vote")
 public class DataRequestVoteResource extends Resource {
 
     private final DACUserAPI dacUserAPI;
@@ -105,8 +105,10 @@ public class DataRequestVoteResource extends Resource {
             Vote vote = api.firstVoteUpdate(rec, id);
             Document access = accessRequestAPI.describeDataAccessRequestById(requestId);
             List<Integer> dataSets = DarUtil.getIntegerList(access, DarConstants.DATASET_ID);
-            List<Vote> votes = vote.getType().equals(VoteType.FINAL.getValue()) ? api.describeVoteByTypeAndElectionId(VoteType.AGREEMENT.getValue(), vote.getElectionId()) :  api.describeVoteByTypeAndElectionId(VoteType.FINAL.getValue(), vote.getElectionId());
-            if(vote.getVote() != null && votes.get(0).getVote() != null){
+            List<Vote> votes = vote.getType().equals(VoteType.FINAL.getValue()) ?
+                api.describeVoteByTypeAndElectionId(VoteType.AGREEMENT.getValue(), vote.getElectionId()) :
+                api.describeVoteByTypeAndElectionId(VoteType.FINAL.getValue(), vote.getElectionId());
+            if (vote.getVote() != null && votes.get(0).getVote() != null) {
                 electionAPI.updateFinalAccessVoteDataRequestElection(rec.getElectionId());
             }
             createDataOwnerElection(requestId, vote, access, dataSets);
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
@@ -105,12 +105,8 @@ public class DataRequestVoteResource extends Resource {
             Vote vote = api.firstVoteUpdate(rec, id);
             Document access = accessRequestAPI.describeDataAccessRequestById(requestId);
             List<Integer> dataSets = DarUtil.getIntegerList(access, DarConstants.DATASET_ID);
-            if(access.containsKey(DarConstants.RESTRICTION)){
-                List<Vote> votes = vote.getType().equals(VoteType.FINAL.getValue()) ? api.describeVoteByTypeAndElectionId(VoteType.AGREEMENT.getValue(), vote.getElectionId()) :  api.describeVoteByTypeAndElectionId(VoteType.FINAL.getValue(), vote.getElectionId());
-                if(vote.getVote() != null && votes.get(0).getVote() != null){
-                    electionAPI.updateFinalAccessVoteDataRequestElection(rec.getElectionId());
-                }
-            }else {
+            List<Vote> votes = vote.getType().equals(VoteType.FINAL.getValue()) ? api.describeVoteByTypeAndElectionId(VoteType.AGREEMENT.getValue(), vote.getElectionId()) :  api.describeVoteByTypeAndElectionId(VoteType.FINAL.getValue(), vote.getElectionId());
+            if(vote.getVote() != null && votes.get(0).getVote() != null){
                 electionAPI.updateFinalAccessVoteDataRequestElection(rec.getElectionId());
             }
             createDataOwnerElection(requestId, vote, access, dataSets);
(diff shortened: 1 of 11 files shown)


src/main/java/org/broadinstitute/consent/http/service/DatabaseMatchingServiceAPI.java
```diff
@@ -163,12 +162,15 @@ private Consent findRelatedConsent(List<Integer> dataSetIdList) {
...
     }
 
-    private RequestMatchingObject createRequestObject(Consent consent, Document dar) throws Exception {
-        String restriction = new Gson().toJson(dar.get(DarConstants.RESTRICTION, Map.class));
-        return new RequestMatchingObject(consent.getUseRestriction(), UseRestriction.parse(restriction));
+    private RequestMatchingObject createRequestObject(Consent consent, DataAccessRequest dar) {
+        DataUse dataUse = useRestrictionConverter.parseDataUsePurpose(dar);
+        UseRestriction darUseRestriction = useRestrictionConverter.parseUseRestriction(dataUse);
+        return new RequestMatchingObject(consent.getUseRestriction(), darUseRestriction);
     }
```
rushtong, line 175: This is the meat of the PR. Previously, we were storing the `UseRestriction` on the data access request as a custom field. We stopped capturing that field when we migrated from mongo to postgres. Now, we first calculate the data use, then query Ontology for the translation to a Use Restriction. We can then generate a match object with the Use Restrictions from both the DAR and the Consent.

src/main/java/org/broadinstitute/consent/http/service/DatabaseMatchingServiceAPI.java
```diff
@@ -118,11 +121,7 @@ private Match singleEntitiesMatch(Consent consent, Document dar) throws Exceptio
             logger.error("Data Access Request is null");
             throw new IllegalArgumentException("Data Access Request cannot be null");
         }
-        if(!dar.containsKey(DarConstants.RESTRICTION)){
-            logger.error("Error finding single matchData Access Request: "+ dar.getString(DarConstants.DAR_CODE) + " does not have a proper Use Restriction.");
-            throw new Exception("Data Access Request: " + dar.getString(DarConstants.DAR_CODE) + " cannot be matched. Missing Use Restriction field.");
-        }
-        Match match = createMatch(consent.getConsentId(), dar.getString(DarConstants.REFERENCE_ID), false, false);
+        Match match = createMatch(consent.getConsentId(), dar.getReferenceId(), false, false);
```
rushtong, line 124: Simplified to remove `DarConstants.RESTRICTION`

src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
```diff
@@ -105,12 +105,10 @@ public Response updateFinalAccessVote(Vote rec,
...
-                if(vote.getVote() != null && votes.get(0).getVote() != null){
-                    electionAPI.updateFinalAccessVoteDataRequestElection(rec.getElectionId());
-                }
-            }else {
+            List<Vote> votes = vote.getType().equals(VoteType.FINAL.getValue()) ?
+                api.describeVoteByTypeAndElectionId(VoteType.AGREEMENT.getValue(), vote.getElectionId()) :
+                api.describeVoteByTypeAndElectionId(VoteType.FINAL.getValue(), vote.getElectionId());
+            if (vote.getVote() != null && votes.get(0).getVote() != null) {
                 electionAPI.updateFinalAccessVoteDataRequestElection(rec.getElectionId());
             }
```
rushtong: Simplified to remove `DarConstants.RESTRICTION`

src/main/java/org/broadinstitute/consent/http/service/DatabaseMatchProcessAPI.java
```diff
@@ -41,10 +40,10 @@ public void processMatchesForConsent(String consentId) {
...
         removeMatchesForPurpose(purposeId);
-        Document rp = dataAccessRequestService.getDataAccessRequestByReferenceIdAsDocument(purposeId);
-        if (rp != null && rp.get(DarConstants.RESTRICTION) != null) {
-            Match match = matchingServiceAPI.findMatchForPurpose(purposeId);
-            saveMatch(Arrays.asList(match));
+        DataAccessRequest dar = dataAccessRequestService.findByReferenceId(purposeId);
+        if (Objects.nonNull(dar)) {
+            Match match = matchingServiceAPI.findMatchForPurpose(dar.getReferenceId());
+            saveMatch(Collections.singletonList(match));
         }
```
rushtong, line 47: Simplified to remove `DarConstants.RESTRICTION`
Also removes a dependency on `Document` which we need to do as general cleanup.

src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
```diff
@@ -105,12 +105,10 @@ public Response updateFinalAccessVote(Vote rec,
...
             Document access = accessRequestAPI.describeDataAccessRequestById(requestId);
             List<Integer> dataSets = DarUtil.getIntegerList(access, DarConstants.DATASET_ID);
-            if(access.containsKey(DarConstants.RESTRICTION)){
-                List<Vote> votes = vote.getType().equals(VoteType.FINAL.getValue()) ? api.describeVoteByTypeAndElectionId(VoteType.AGREEMENT.getValue(), vote.getElectionId()) :  api.describeVoteByTypeAndElectionId(VoteType.FINAL.getValue(), vote.getElectionId());
-                if(vote.getVote() != null && votes.get(0).getVote() != null){
-                    electionAPI.updateFinalAccessVoteDataRequestElection(rec.getElectionId());
-                }
-            }else {
+            List<Vote> votes = vote.getType().equals(VoteType.FINAL.getValue()) ?
+                api.describeVoteByTypeAndElectionId(VoteType.AGREEMENT.getValue(), vote.getElectionId()) :
```
raejohanek: Maybe I am interpreting this wrong, but if the vote type is Final then why pass in the Agreement vote type on the next line, instead of passing in Final?

lindojong (approved):
LGTM!

src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
rushtong: My intention here was to leave the original logic in place and do as little refactoring as possible that might cause a regression. However, after reviewing what this code is doing, the end result is the same regardless of what votes we query for, either final or agreement. For example, in both cases, we end up calling `electionAPI.updateFinalAccessVoteDataRequestElection(rec.getElectionId());` This code can be simplified even further, nice call 👍🏽

rushtong: I've rewritten the logic here to accomplish the same thing, but hopefully in a more code-legible manner. Please have a second look.

src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
```diff
@@ -105,12 +107,17 @@ public Response updateFinalAccessVote(Vote rec,
...
+            List<Vote> votes = voteService
+                  .findVotesByElectionIds(Collections.singletonList(rec.getElectionId()))
+                  .stream()
+                  .filter(v ->
+                      v.getType().equalsIgnoreCase(VoteType.FINAL.getValue()) ||
+                      v.getType().equalsIgnoreCase(VoteType.AGREEMENT.getValue()))
+                  .filter(v -> Objects.nonNull(v.getVote()))
+                  .collect(Collectors.toList());
+            // If we have any votes with a populated vote, we can now update the election.
+            if (!votes.isEmpty()) {
```
rushtong, line 120: Previously, this call was using a confusing ternary that didn't make clear the intention. The goal here was to find out if there were any existing FINAL or AGREEMENT votes that had been voted on (i.e., have a non-null vote value). Such a value would mean that a chairperson had recorded the final vote for a DAR election. If any such vote existed, then it was appropriate to update the election's final access vote value.

raejohanek (approved):

PR #960, opened 2021-03-08
DUOS-976[risk=no]Service class plumbing for institute

raejohanek:
## Scope:
major changes:
Create institutionMapper
Add Institution support to DAOTestHelper
Create tests of InstitutionDAO
Create tests of InstitutionService

minor changes:
Register Mapper in InstitutionDAO Class
Simplify methods in Institution Model class
Override equals in Institution Model class
Adjust methods in InstitutionService class to include the date rather than create it

## Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-976

In all cases:


src/main/java/org/broadinstitute/consent/http/models/Institution.java
```diff
@@ -36,24 +41,36 @@ public Institution(Integer id, String name, String itDirectorName, String itDire
     this.updateUser = updateUser;
   }
 
-  public void setName(String name, Integer updateUser) {
```
raejohanek, line 39: remove the update user parameter from all set methods and add set methods for update user, create user, create date, and update date

src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
```diff
@@ -17,27 +17,42 @@ public InstitutionService(InstitutionDAO institutionDAO) {
   }
 
   public Integer createInstitution(String name, String itDirectorName, String itDirectorEmail, Integer createUser) {
-    return this.institutionDAO.insertInstitution(name, itDirectorName, itDirectorEmail, createUser, new Date());
+    Date date = new Date();
+    return institutionDAO.insertInstitution(name, itDirectorName, itDirectorEmail, createUser, date);
+  }
+
+  public Integer createInstitution(String name, String itDirectorName, String itDirectorEmail, Integer createUser, Date date) {
```
raejohanek: Address errors about incorrect number of matchers in service class by making versions of the create and update methods that include all parameters including date

src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
```diff
@@ -17,27 +17,42 @@ public InstitutionService(InstitutionDAO institutionDAO) {
...
+    return institutionDAO.insertInstitution(name, itDirectorName, itDirectorEmail, createUser, date);
   }
 
   public void updateInstitutionById(Integer id,
                                     String institutionName,
                                     String itDirectorName,
                                     String itDirectorEmail,
                                     Integer updateUser) {
-    this.institutionDAO.updateInstitutionById(
```
raejohanek, line 28: removing unnecessary 'this' in this file

src/main/java/org/broadinstitute/consent/http/models/Institution.java
raejohanek: this follows the structure of other model classes and relies on the user to update user and update date when they make a change instead of doing it automatically

src/main/java/org/broadinstitute/consent/http/db/mapper/InstitutionMapper.java
```diff
@@ -0,0 +1,37 @@
...
+  @Override
+  public Institution map(ResultSet resultSet, StatementContext statementContext) throws SQLException {
+    Institution institution;
+
+    if (institutionMap.containsKey(resultSet.getInt("institution_id"))) {
+      institution = institutionMap.get(resultSet.getInt("institution_id"));
+    } else {
+      institution = new Institution();
+      institution.setId(resultSet.getInt("institution_id"));
+    }
```
rushtong: You could pull the `institution_id` call out to an int above the if clause. That would clean up the copy/past of `resultSet.getInt("institution_id"))`
```suggestion
    int institutionId = resultSet.getInt("institution_id"));
    if (institutionMap.containsKey(institutionId) {
      institution = institutionMap.get(institutionId);
    } else {
      institution = new Institution();
      institution.setId(institutionId);
    }
```

src/main/java/org/broadinstitute/consent/http/db/mapper/InstitutionMapper.java
```diff
@@ -0,0 +1,37 @@
...
+    institution.setName(resultSet.getString("institution_name"));
```
rushtong: I don't think it's necessary for this PR, but you could add some safety for each column by checking if it exists (see helper method in `RowMapperHelper`). Currently, we query for all fields. If, at some point in time, we have a query for IDs or Names, etc., then this mapper will throw an exception on the fields that aren't in the query.

src/main/java/org/broadinstitute/consent/http/models/Institution.java
```diff
@@ -85,4 +102,11 @@ public Date getUpdateDate() {
   public Integer getUpdateUser() {
     return updateUser;
   }
+
+  @Override
+  public boolean equals(Object institution) {
+    if (institution == this) return true;
+    if (institution == null || institution.getClass() != getClass()) return false;
+    return (((Institution)institution).getId() == getId());
```
rushtong: For consistency, we should use `EqualsBuilder`. If we need to add something else to determine equality, it will be trivial to add an append line.
```suggestion
    Institution other = (Institution) obj;
    return new EqualsBuilder()
          .append(id, other.getId())
          .isEquals();
```

src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
```diff
@@ -17,27 +17,42 @@ public InstitutionService(InstitutionDAO institutionDAO) {
   }
 
   public Integer createInstitution(String name, String itDirectorName, String itDirectorEmail, Integer createUser) {
-    return this.institutionDAO.insertInstitution(name, itDirectorName, itDirectorEmail, createUser, new Date());
+    Date date = new Date();
+    return institutionDAO.insertInstitution(name, itDirectorName, itDirectorEmail, createUser, date);
+  }
```
rushtong: As a general principle, methods should return a full object, not an ID, unless that's really the intent. This should return a full institution by finding by id.

src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
```diff
@@ -17,27 +17,42 @@ public InstitutionService(InstitutionDAO institutionDAO) {
...
 
+  public void updateInstitutionById(Integer id,
+                                    String institutionName,
+                                    String itDirectorName,
+                                    String itDirectorEmail,
+                                    Integer updateUser,
+                                    Date date) {
+institutionDAO.updateInstitutionById(
+id, institutionName, itDirectorName, itDirectorEmail, updateUser, date);
+}
```
rushtong: nit: spacing is off here.

src/test/java/org/broadinstitute/consent/http/db/InstitutionDAOTest.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.List;
+
+public class InstitutionDAOTest extends DAOTestHelper {
+
+  @Test
+  public void testInsertInstitution() {
+    Institution institution = createInstitution();
+    List<Institution> all = institutionDAO.findAllInstitutions();
+    Assert.assertTrue(all.contains(institution));
+  }
```
rushtong, line 19: nit: newline between methods

src/test/java/org/broadinstitute/consent/http/db/InstitutionDAOTest.java
```diff
@@ -0,0 +1,53 @@
...
+  @Test
+  public void testUpdateInstitutionById() {
+    Integer userId = createUser().getDacUserId();
+    String newValue = "New Value";
+    Institution institution = createInstitution();
+    institutionDAO.updateInstitutionById(institution.getId(), newValue, newValue, newValue, userId, new Date());
+    Institution updated = institutionDAO.findInstitutionById(institution.getId());
+    Assert.assertEquals(updated.getName(), newValue);
```
rushtong: Import the static `assertEquals` so you don't have to prefix this on every assertion.

src/test/java/org/broadinstitute/consent/http/db/InstitutionDAOTest.java
```diff
@@ -0,0 +1,53 @@
...
+    Assert.assertEquals(updated.getItDirectorName(), newValue);
+    Assert.assertEquals(updated.getItDirectorEmail(), newValue);
+  }
+
+  @Test
+  public void testDeleteInstitutionById() {
+    Institution institution = createInstitution();
+    Integer id = institution.getId();
+    institutionDAO.deleteInstitutionById(id);
+    Assert.assertFalse(institutionDAO.findAllInstitutions().contains(institution));
```
rushtong: Thinking out loud ... you could also find by id and assert a null object.

src/test/java/org/broadinstitute/consent/http/db/InstitutionDAOTest.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  @Test
+  public void testFindInstitutionById() {
+    Institution institution = createInstitution();
+    Integer id = institution.getId();
+    Assert.assertEquals(institutionDAO.findInstitutionById(id).getName(), institution.getName());
```
rushtong: It would be safer to get the full institution, then check all properties on that institution to make sure all of them are populated correctly. This one just tells you that the name was mapped correctly.

src/test/java/org/broadinstitute/consent/http/db/InstitutionDAOTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+  @Test
+  public void testFindAllInstitutions() {
+    List<Institution> instituteList = institutionDAO.findAllInstitutions();
+    Assert.assertEquals(0, instituteList.size());
+    createInstitution();
+    List<Institution> instituteListUpdated = institutionDAO.findAllInstitutions();
+    Assert.assertEquals(1, instituteListUpdated.size());
+  }
+}
```
rushtong: nit: add a newline

src/test/java/org/broadinstitute/consent/http/service/InstitutionServiceTest.java
```diff
@@ -0,0 +1,93 @@
...
+  private void initService() {
+    service = new InstitutionService(institutionDAO);
+  }
+
+  @Test
+  public void testCreateInstitution() {
+    initService();
+    when(institutionDAO.insertInstitution(anyString(), anyString(), anyString(), anyInt(), eq(new Date()))).thenReturn(getInstitutions().get(0).getId());
+    Integer instituteId = service.createInstitution(anyString(), anyString(), anyString(), anyInt(), eq(new Date()));
+    Assert.assertSame(getInstitutions().get(0).getId(), instituteId);
```
rushtong: similar feedback to the dao test, import the static `assertSame` to avoid the duplication of `Assert.`

src/test/java/org/broadinstitute/consent/http/service/InstitutionServiceTest.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  @Test
+  public void testUpdateInstitutionById() {
+    initService();
+    Integer id = institutionDAO.insertInstitution(anyString(), anyString(), anyString(), anyInt(), eq(new Date()));
```
rushtong: `institutionDAO.insertInstitution` should be wrapped in a mock method like `when` or `doNothing` here, not called directly. A mocked object like this will not actually perform any actions. If you put the debugger on this and step into it, the id that comes back is `0` which actually comes from `org.powermock.api.mockito.repackaged.MethodInterceptorFilter.intercept()`, not `InstitutionDAO`.

rushtong: Since the update is a void method, you probably want to wrap that call in a `doNothing` (see https://www.baeldung.com/mockito-void-methods). There are really two cases that you care about ... the update does something right, or it does something wrong. If it does something right, there's no error. If it does something wrong, there's some kind of exception thrown (probably a postgres exception) so those are really the two cases that you want to handle in your tests, a positive case and a negative case.

src/test/java/org/broadinstitute/consent/http/service/InstitutionServiceTest.java
```diff
@@ -0,0 +1,93 @@
...
+        service.deleteInstitutionById(1);
+    } catch (Exception e) {
+        Assert.fail("Delete should not fail");
+    }
+  }
+
+  @Test
+  public void testFindInstitutionById() {
+    initService();
+    Integer id = institutionDAO.insertInstitution(anyString(), anyString(), anyString(), anyInt(), eq(new Date()));
```
rushtong: By calling the dao class like this, you're trying to do a unit test on the dao, not the service. We need to mock the insert call and then see if the service class can handle the different kinds of responses, either a real ID or an exception.

src/test/java/org/broadinstitute/consent/http/service/InstitutionServiceTest.java
```diff
@@ -0,0 +1,93 @@
...
+    Assert.assertEquals(institutionDAO.findInstitutionById(id), service.findInstitutionById(id));
+  }
+
+  @Test
+  public void testFindAllInstitutions() {
+    initService();
+    when(institutionDAO.findAllInstitutions()).thenReturn(Collections.emptyList());
```
rushtong, line 88: This is correct mocking behavior!

src/test/java/org/broadinstitute/consent/http/service/InstitutionServiceTest.java
```diff
@@ -0,0 +1,93 @@
...
+   */
+  private List<Institution> getInstitutions() {
+    return IntStream.range(0, 4).
+      mapToObj(i -> {
+        Institution institute = new Institution();
+        institute.setId(i);
+        return institute;
+      }).collect(Collectors.toList());
+  }
+}
```
rushtong: nit: add newline to the end of the file.

rushtong (changes requested):
See comments inline, good work so far!

src/test/java/org/broadinstitute/consent/http/service/InstitutionServiceTest.java
```diff
@@ -0,0 +1,110 @@
...
+  public void testUpdateInstitutionByIdFail() {
+    initService();
+    doThrow(new RuntimeException("Update method should pass on error from DAO")).when(institutionDAO).updateInstitutionById(anyInt(), anyString(), anyString(), anyString(), anyInt(), eq(new Date()));
+  
+    try {
+      service.updateInstitutionById(anyInt(), eq("New Name"), anyString(), anyString(), anyInt(), eq(new Date()));
+    } catch (Exception e) {
+      assertEquals(e.getMessage(), "Update method should pass on error from DAO");
+    }
+  }
```
rushtong: Instead of wrapping the call in a try, you can add an `expected` to the test annotation. In testing scenarios, prefer checked vs runtime exceptions
```suggestion
  @Test(expected = Throwable.class)
  public void testUpdateInstitutionByIdFail() {
    initService();
    doThrow(new Exception("Update method should pass on error from DAO")).when(institutionDAO).updateInstitutionById(anyInt(), anyString(), anyString(), anyString(), anyInt(), eq(new Date()));
    service.updateInstitutionById(anyInt(), eq("New Name"), anyString(), anyString(), anyInt(), eq(new Date()));
  }
```

rushtong (approved):
One optional suggestion inline, but don't let that hold up this PR, merge when you're ready.

rushtong (approved):
👍🏽

PR #965, opened 2021-03-10
[DUOS-1103][risk=no] Clean up final vote submission logic

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1103

## Changes
Terminology: 
* `finalVote`: The value of the `Vote` object whose type is `FINAL`. It is a derived value attached to election value objects, but stored in the vote table. This is one of the (several) votes the chairperson submits when clicking "Vote" in the chairperson's voting UI.
* `finalAccessVote`: A value stored on the election object itself that should reflect the `finalVote` value. Due to a bug, these are not in sync and are confusing when they disagree. In a future story, we'll investigate consolidating these fields to reduce confusion.

Updated the flow around submitting a final vote. When submitting a final access vote, move some of the vote checking logic into the service class method and out of the resource class method. Original code was always incorrectly setting the `finalAccessVote` to a true value. We now update that value based on the actual final vote that is submitted.

* Deconstruct json string into a Vote in the resource class manually with Gson instead of through Jackson
* minor method renaming to make it more clear what we're doing to votes:
  * `firstVoteUpdate` -> `updateVoteById`
  * `updateFinalAccessVote ` -> `submitFinalAccessVote`
  * `updateFinalAccessVoteDataRequestElection` -> `submitFinalAccessVoteDataRequestElection`

----

In all cases:


raejohanek (approved):
refactoring and bug fix look good to me!

PR #968, opened 2021-03-11
DUOS-1104[risk=no]Cancel Election does not update all of the election's votes correctly

JVThomas:
Addresses [DUOS-1104](https://broadworkbench.atlassian.net/browse/DUOS-1104)

PR corrects vote assignments on votes with `type == 'FINAL'` via `updateFinalVote` method within `DatabaseElectionAPI`. Instead of finding a single Final vote, the method will now query for a list of final votes and perform attribute updates based on the election argument. 

**Note:** Method feels like it should belong in a Vote or Election service, and it's only referenced in a single method so moving shouldn't be difficult. I'll write the test up as a follow-up to the feedback.

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DatabaseElectionAPI.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DatabaseElectionAPI.java
@@ -611,11 +611,21 @@ public class DatabaseElectionAPI extends AbstractElectionAPI {
 
     private void updateFinalVote(Election rec, Integer electionId) {
         if (rec.getFinalVote() != null) {
-            Vote vote = voteDAO.findVoteByElectionIdAndType(electionId, VoteType.CHAIRPERSON.getValue());
-            vote.setVote(rec.getFinalVote());
-            vote.setCreateDate(rec.getFinalVoteDate());
-            vote.setRationale(rec.getFinalRationale());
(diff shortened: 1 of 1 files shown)


JVThomas:
Upon further discussion, it seems like the target function, `updateFinalVote` does not belong within an Election update path and should be moved elsewhere. Ideal action is to remove the method and its function call within updateElectionById.

src/main/java/org/broadinstitute/consent/http/service/DatabaseElectionAPI.java
```diff
@@ -611,11 +611,21 @@ private void updateAccessElection(Integer electionId, String type, String status
...
+            for (Vote vote : finalVotes) {
+                vote.setVote(rec.getFinalVote());
+                vote.setUpdateDate(finalVoteDate);
+                vote.setRationale(rec.getFinalRationale());
+                voteDAO.updateVote(
+                    vote.getVote(), vote.getRationale(), vote.getUpdateDate(), vote.getVoteId(),
+                    vote.getIsReminderSent(), vote.getElectionId(), vote.getCreateDate(),
+                    vote.getHasConcerns()
+                );
+            }
```
rushtong: In the context of where this is called from, `updateElectionById`, I don't think we should be modifying any votes at all. Vote modification should only ever be done when creating an election (and creating votes for that election), or when updating a vote's value directly through the Vote APIs. My recommendation here is to remove this call altogether. It is named one thing (update final vote) and then queries for a Chairperson vote, making this even more suspect.

rushtong (changes requested):
My suggestion is to remove this method altogether and not update any vote based on an election's status change.

rushtong (approved):
👍🏽

PR #973, opened 2021-03-15
[DUOS-919][risk=low] consent api to service

rfricke-asymmetrik:
**Addresses**
https://broadworkbench.atlassian.net/browse/DUOS-919

**Changes**
- Remove ConsentAPI objects
- Move ConsentAPI methods to ConsentService
- Adds Unit tests for new ConsentService methods

In all cases:


src/main/java/org/broadinstitute/consent/http/service/DatabaseMatchingServiceAPI.java
```diff
@@ -39,12 +39,12 @@
...
-        ServicesConfiguration config, DatasetService datasetService, UseRestrictionConverter useRestrictionConverter) {
-        MatchAPIHolder.setInstance(new DatabaseMatchingServiceAPI(client, dataAccessRequestDAO, config, AbstractConsentAPI.getInstance(), datasetService, useRestrictionConverter));
+        ServicesConfiguration config, ConsentService consentService, DatasetService datasetService, UseRestrictionConverter useRestrictionConverter) {
+        MatchAPIHolder.setInstance(new DatabaseMatchingServiceAPI(client, dataAccessRequestDAO, config, consentService, datasetService, useRestrictionConverter));
     }
 
-    DatabaseMatchingServiceAPI(Client client, DataAccessRequestDAO dataAccessRequestDAO, ServicesConfiguration config, ConsentAPI consentAPI, DatasetService datasetService, UseRestrictionConverter useRestrictionConverter){
-        this.consentAPI = consentAPI;
+    DatabaseMatchingServiceAPI(Client client, DataAccessRequestDAO dataAccessRequestDAO, ServicesConfiguration config, ConsentService consentService, DatasetService datasetService, UseRestrictionConverter useRestrictionConverter){
+        this.consentService = consentService;
```
rushtong: This is a case where we want to try and minimize pulling one service into another. I see that there are only two usages here that we care about, maybe we can instead pull in the consent DAO and make those db calls directly.

src/main/java/org/broadinstitute/consent/http/service/ConsentService.java
```diff
@@ -100,6 +227,47 @@ public void updateConsentDac(String consentId, Integer dacId) {
...
+        if (election != null) {
+            consent.setLastElectionStatus(election.getStatus());
+            consent.setLastElectionArchived(election.getArchived());
+        }
+        return consent;
+    }
+
+    public Consent getConsentFromDatasetID(Integer datasetId) {
+        return consentDAO.findConsentFromDatasetID(datasetId);
+    }
```
rushtong, line 269: Seems like the only usage of this is in `AbstractMatchingServiceAPI`. Can we instead just make the dao call in that class and drop this method completely from the service?

rushtong (changes requested):
This looks great. Two suggestions inline that should help reduce the potential for circular service references.

rushtong (approved):
Thank you for the changes, this is great!

PR #982, opened 2021-03-23
[DUOS-920][risk=low] Replace match APIs with service

rfricke-asymmetrik:
**Addresses**
https://broadworkbench.atlassian.net/browse/DUOS-920

**Changes**

- Remove AbstractMatchAPI, AbstractMatchProcessAPI, AbstractMatchingServiceAPI
- Remove Interfaces
- Add methods to new MatchService

In all cases:


src/main/java/org/broadinstitute/consent/http/service/MatchService.java
```diff
@@ -32,35 +24,136 @@
...
 
-    DatabaseMatchingServiceAPI(Client client, DataAccessRequestDAO dataAccessRequestDAO, ServicesConfiguration config, ConsentDAO consentDAO, ElectionDAO electionDAO, DatasetService datasetService, UseRestrictionConverter useRestrictionConverter){
+    @Inject
+    public MatchService(Client client, ServicesConfiguration config, ConsentDAO consentDAO, MatchDAO matchDAO,
+                        ElectionDAO electionDAO, DataAccessRequestDAO dataAccessRequestDAO, DataSetDAO dataSetDAO,
+                        UseRestrictionConverter useRestrictionConverter) {
+        this.matchDAO = matchDAO;
         this.consentDAO = consentDAO;
-        this.dataAccessRequestDAO = dataAccessRequestDAO;
+        this.accessAPI = AbstractDataAccessRequestAPI.getInstance();
```
rushtong: The only need for accessAPI here is for the method:
```
    private void validatePurpose(String purposeId) {
        if (accessAPI.describeDataAccessRequestById(purposeId) == null) {
            throw new IllegalArgumentException("Purpose for the specified id does not exist");
        }
    }
```
However, that can easily be replaced with a call from `dataAccessRequestDAO` which is preferable to using the singleton here.

src/main/java/org/broadinstitute/consent/http/service/MatchService.java
```diff
@@ -95,11 +187,10 @@ public Match findMatchForPurpose(String purposeId){
         return match;
     }
 
-    @Override
     public List<Match> findMatchesForConsent(String consentId) {
         List<Match> matches = new ArrayList<>();
         Consent consent = findConsent(consentId);
-        List<DataSet> dataSets = datasetService.getDataSetsForConsent(consentId);
+        List<DataSet> dataSets = dataSetDAO.getDataSetsForConsent(consentId);
```
rushtong, line 191: Nice update 👍🏽

rushtong (changes requested):
One suggestion inline, but looks very good - thank you!

rushtong (approved):
Thank you for the changes!

PR #984, opened 2021-03-23
DUOS-1119[risk=no]Consent: Unicode vulnerability

raejohanek:
SCOPE/EXPLANATION OF CHANGES:
Please read before review : UPDATED

for update method:
- DACUserResource.update creates userMap by calling constructUserMapFromJson and passes the userMap and userId to DatabaseDACUserAPI.updateDACUserId (constructUserMap calls User(json) which calls setEmail)
-  DatabaseDACUserAPI.updateDACUserId calls userDAO.updateUser(updatedUser.getEmail(), updatedUser.getDisplayName(), id, updatedUser.getAdditionalEmail());
- So by wrapping setEmail we are prevented the update user method from posting unicode to the database

for create method:
- DACUserResource.createDACUser calls UserService.createUser(new User(json)) (User(json) calls setEmail)
- UserService.createUser calls userDAO.insertUser(user.getEmail(), user.getDisplayName(), new Date());
- So by wrapping setEmail we are preventing the create user method from posting Unicode to the database

general:
- UNDONE CHANGES: wrapping the getEmail method is now unnecessary since prior to that call setEmail is called and makes the conversion

- add new dependency to pom.xml file

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1119

In all cases:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -258,6 +258,13 @@
   <dependencies>
 
     <dependency>
+      <groupId>net.gcardone.junidecode</groupId>
+      <artifactId>junidecode</artifactId>
+      <version>0.4.1</version>
+    </dependency>
+
+
+    <dependency>
       <groupId>org.owasp</groupId>
       <artifactId>java-file-io</artifactId>
       <version>1.0.0</version>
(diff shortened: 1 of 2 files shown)


rushtong:
> so by wrapping getEmail (not setEmail since that is not used in this process), we are preventing the API from posting Unicode to the database

We do call `setEmail()` when deconstructing a user from a json string (see `public User(String json)`) which is what we do in `constructUserMapFromJson(String json)`. The approach of just  doing this in the getter will not be sufficient to protect us from inserting unicode.

rushtong (changes requested):
See main comment, we need a more comprehensive approach as outlined in the ticket.

rushtong (approved):
This works as advertised, thank you! It also brings up another flaw in the system in that when a user updates their own email address, it breaks the `/me` call because the auth-ed user now has a different email than what we are storing. I will file a ticket to remove email edit-ability from this API and prevent users from breaking their own login info.

PR #985, opened 2021-03-23
DUOS-977[risk=medium] Implemented Institution Resource

JVThomas:
Addresses [DUOS-977](https://broadworkbench.atlassian.net/browse/DUOS-977)

- PR adds InstitutionResource, InstitutionUtil (utility class), and associated tests. PR also slightly refactors InstitutionService and its associated tests.
- Requesting thorough review as code is adding new API endpoints.
- Would appreciate any feedback with regards to Java coding conventions

In all cases:


src/main/resources/assets/api-docs.yaml
```diff
@@ -2895,6 +2895,69 @@ paths:
...
+          description: Returns all institutions. If no results are found, an empty array is returned. 
+            Admin will see all details while non-Admins will only see a subset.
+        500:
+          description: Internal Server Error
+    post:
+      summary: Insitution CREATE resource
+      description: Creates new institution record based on json payload
+      responses:
+        200:
+          description: Creates and returns newly created record.
```
rushtong: This should be a 201 created response

src/main/resources/assets/api-docs.yaml
```diff
@@ -2895,6 +2895,69 @@ paths:
...
+        - name: id
+          in: path
+          description: Institution ID
+          required: true
+          schema:
+            type: String
+      responses:
+        200:
+          description: Returns target institution.
+            Admin will see all details while non-Admins will only see a subset.
```
rushtong, line 2910: I think we need a 404 here too.

src/main/java/org/broadinstitute/consent/http/ConsentApplication.java
```diff
@@ -251,6 +255,9 @@ public void run(ConsentConfiguration config, Environment env) {
         System.setProperty("sun.net.http.allowRestrictedHeaders", "true");
         configureCors(env);
 
+        //Response Formatters
+        final InstitutionUtil institutionUtil = injector.getProvider(InstitutionUtil.class).get();
```
rushtong: Does this need to be here as an injected instance? It could just be a final instance created on the resource itself and not necessarily be passed into the constructor.

src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
```diff
@@ -14,36 +14,41 @@
...
-    Date date = new Date();
-    Integer id = institutionDAO.insertInstitution(name, itDirectorName, itDirectorEmail, createUser, date);
-    return this.findInstitutionById(id);
-  }
+  };
 
-  public Institution createInstitution(String name, String itDirectorName, String itDirectorEmail, Integer createUser, Date date) {
-    Integer id = institutionDAO.insertInstitution(name, itDirectorName, itDirectorEmail, createUser, date);
-    return this.findInstitutionById(id);
+  //NOTE: if NAME column has the UNIQUE attribute in Postgres, can we trust that to throw errors in Java?
```
rushtong: yes, we can.

src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
```diff
@@ -14,36 +14,41 @@
...
+  public Institution createInstitution(Institution institution, Integer userId) {
+      //NOTE: should an error be thrown if id is NOT null? Data may be valid, but id present indicates non-normal request
```
rushtong: We have enough knowledge at this point to make an informed decision about what to do. We should just ignore the id value if it's there.

rushtong: The thing you do want to do here, since we've changed the method signature, is to make sure all of the required fields are non-null so we aren't inserting nulls.

src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
```diff
@@ -57,4 +62,11 @@ public Institution findInstitutionById(Integer id) {
   public List<Institution> findAllInstitutions() {
     return institutionDAO.findAllInstitutions();
   }
+
+  private void checkForEmptyName(Institution institution) {
+    String name = institution.getName();
+    if(name == null || name.isBlank()) {
```
rushtong: Prefer `Objects.nonNull()` to `== null`

src/main/java/org/broadinstitute/consent/http/resources/InstitutionResource.java
```diff
@@ -0,0 +1,117 @@
...
+      return createExceptionResponse(e);
+    }
+  }
+
+  @PUT
+  @Consumes("application/json")
+  @Produces("application/json")
+  @Path("/{id}")
+  @RolesAllowed(ADMIN)
+  public Response updateInstitution(@Auth AuthUser authUser, @PathParam("id") String paramId, String institution) {
```
rushtong: `paramId` can be an integer here and you won't have to parse it in the method

src/main/java/org/broadinstitute/consent/http/resources/InstitutionResource.java
```diff
@@ -0,0 +1,117 @@
...
+    try{
+      User user = userService.findUserByEmail(authUser.getName());
+      Integer id = Integer.parseInt(paramId);
+      Institution payload = new Gson().fromJson(institution, Institution.class);
+      Institution updatedInstitution = institutionService.updateInstitutionById(payload, id, user.getDacUserId());
```
rushtong, line 98: Before trying to update, we should check that it's a real one first and throw a NFE it there is no institution with the provided ID.

src/main/java/org/broadinstitute/consent/http/resources/InstitutionResource.java
```diff
@@ -0,0 +1,117 @@
...
+      String jsonResponse = new Gson().toJson(updatedInstitution);
+      return Response.ok().entity(jsonResponse).build();
```
rushtong: You don't need to convert it to a string first - just putting the updated institution into `entity(updatedInstitution)` will work.

src/main/java/org/broadinstitute/consent/http/resources/InstitutionResource.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  @DELETE
+  @Produces("application/json")
+  @Path("/{id}")
+  @RolesAllowed(ADMIN)
+  public Response deleteInstitution(@Auth AuthUser authUser, @PathParam("id") String paramId) {
+    try {
+      Integer id = Integer.parseInt(paramId);
+      institutionService.deleteInstitutionById(id);
```
rushtong, line 111: Similar feedback here: use `Integer` in the signature and make sure there really is an institution with the id before trying to delete it.

src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
```diff
@@ -14,36 +14,41 @@
...
-                                    Integer updateUser,
-                                    Date date) {
-    institutionDAO.updateInstitutionById(
-      id, institutionName, itDirectorName, itDirectorEmail, updateUser, date);
+      institutionPayload.getId(), 
+      institutionPayload.getName(), 
+      institutionPayload.getItDirectorName(), 
+      institutionPayload.getItDirectorEmail(), 
+      institutionPayload.getUpdateUser(),
+      institutionPayload.getUpdateDate());
```
rushtong: Similar feedback here, since we've changed signature, we should do some validation on required fields.
Also, we should be setting the update date here as a new date in the service class and not getting it from the object.

src/test/java/org/broadinstitute/consent/http/resources/InstitutionResourceTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+public class InstitutionResourceTest {
+  private final AuthUser authUser = new AuthUser("test@test.com");
+  private final List<UserRole> adminRoles = Collections.singletonList(new UserRole(UserRoles.ADMIN.getRoleId(), UserRoles.ADMIN.getRoleName()));
+  private final List<UserRole> researcherRoles = Collections.singletonList(new UserRole(UserRoles.RESEARCHER.getRoleId(), UserRoles.RESEARCHER.getRoleName()));
+  private final User adminUser = new User(1, authUser.getName(), "Display Name", new Date(), adminRoles, authUser.getName());
+  private final User researcherUser = new User(1, authUser.getName(), "Display Name", new Date(), researcherRoles, authUser.getName());
+
+  @Mock InstitutionService institutionService;
+  @Mock UserService userService;
```
rushtong: These can be private too

src/test/java/org/broadinstitute/consent/http/resources/InstitutionResourceTest.java
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  private Institution initInsitutionModel() {
+    Institution mockInstitution = new Institution();
+    mockInstitution.setName("Test Name");
+    return mockInstitution;
+  }
+
+  private void initResource() {
+    resource = new InstitutionResource(userService, institutionService, institutionUtil);
```
rushtong: Having this in two places (here and `setUp`) is a bit confusing. As it stands now, resource will be initialized twice for any method that calls `initResource`. It would be cleaner to only have it here, and call `initResource` for every test method.

src/test/java/org/broadinstitute/consent/http/resources/InstitutionResourceTest.java
```diff
@@ -0,0 +1,160 @@
...
+  public void setUp() {
+    MockitoAnnotations.initMocks(this);
+    resource = new InstitutionResource(userService, institutionService, institutionUtil);
+  }
+
+  private Institution initInsitutionModel() {
+    Institution mockInstitution = new Institution();
+    mockInstitution.setName("Test Name");
+    return mockInstitution;
+  }
```
rushtong: This is only used in `institutionSetup` ... I think it would be more legible to just fold this code into that method instead of having two that do two parts of the same job.

src/test/java/org/broadinstitute/consent/http/resources/InstitutionResourceTest.java
```diff
@@ -0,0 +1,160 @@
...
+  @Mock InstitutionService institutionService;
+  @Mock UserService userService;
+  @Mock private UriInfo info;
+  @Mock private UriBuilder builder;
+  @Mock private User mockUser;
+  @Spy InstitutionUtil institutionUtil = new InstitutionUtil();
+
+  private InstitutionResource resource;
+
+  private Institution institutionSetup() {
```
rushtong: method naming nit: This is just creating an institution with mock values, not really setting anything up which would connote configuring services or resources.

src/test/java/org/broadinstitute/consent/http/resources/InstitutionResourceTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+  private void initResource() {
+    resource = new InstitutionResource(userService, institutionService, institutionUtil);
+  }
+
+  @Test
+  public void testGetInstitutionsForAdmin() {
+    List<Institution> institutions = new ArrayList<Institution>();
+    Institution mockInstitution = institutionSetup();
+    institutions.add(mockInstitution);
```
rushtong: Condense:
```
List<Institution> institutions = Collections.singletonList(institutionSetup());
```

rushtong: Similar feedback in other test methods

src/test/java/org/broadinstitute/consent/http/resources/InstitutionResourceTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+  @Test
+  public void testDeleteInstitution() { 
+    when(userService.findUserByEmail(anyString())).thenReturn(adminUser);
+    when(institutionService.updateInstitutionById(any(), anyInt(), anyInt())).thenReturn(null);
+    String paramId = "1";
+    Response response = resource.deleteInstitution(authUser, paramId);
+    assertEquals(200, response.getStatus());
+  }  
+}
```
rushtong, line 230: Other tests to consider are the 404's and making sure that response codes are appropriate.

src/test/java/org/broadinstitute/consent/http/resources/InstitutionResourceTest.java
```diff
@@ -0,0 +1,160 @@
...
+  @Test
+  public void getInstitutionNonAdmin() {
+    Institution mockInstitution = institutionSetup();
+    when(userService.findUserByEmail(anyString())).thenReturn(researcherUser);
+    when(institutionService.findInstitutionById(anyInt())).thenReturn(mockInstitution);
+    String id = "1";
+    Response researcherResponse = resource.getInstitution(authUser, id);
+    String json = researcherResponse.getEntity().toString();
+    assertEquals(200, researcherResponse.getStatus());
+    assertNotNull(json);
```
rushtong, line 119: For the non-admin responses, it would be nice to know that the exclusion strategy worked. You can coerce the string back into an object with Gson and check the fields

rushtong (changes requested):
Really good work here, love the exclusion strategy approach. See comments inline.

src/test/java/org/broadinstitute/consent/http/resources/InstitutionResourceTest.java
JVThomas: The InstitutionUtil tests covers this since that's where the ```ExclusionStrategy``` and the ```GsonBuilder``` are defined. This way I don't have to repeat this test across all of the ```@PermitAll``` routes. (Though for the non Admin response I just check the json directly since comparing the string is faster to type out than deserializing the ```Institution``` object and then checking each field).

src/main/java/org/broadinstitute/consent/http/resources/InstitutionResource.java
```diff
@@ -0,0 +1,111 @@
...
+  @GET
+  @Produces("application/json")
+  @PermitAll
+  public Response getInstitutions(@Auth AuthUser authUser) {
+    try{
+      User user = userService.findUserByEmail(authUser.getName());
+      Boolean isAdmin = institutionUtil.checkIfAdmin(user);
+      Gson gson = institutionUtil.getGsonBuilder(isAdmin);
+      List<Institution> institutions = institutionService.findAllInstitutions();
+      return Response.ok().entity(gson.toJson(institutions)).build();
```
rushtong, line 52: TOL: Normally, we'd just return the list as the entity. In this case, we **need** to run it through Gson so it handles the exclusion strategy. I think this could use a comment to explain why we're doing it this way. Same feedback for the other `PermitAll` method(s).

src/main/resources/assets/api-docs.yaml
```diff
@@ -2872,6 +2872,82 @@ paths:
...
+    post:
+      summary: Institution CREATE resource
+      description: Creates new institution record based on json payload
+      responses:
+        201:
+          description: Creates and returns newly created record.
+        500:
+          description: Internal Server Error
+        400:
+          description: "Bad request"
```
rushtong: nit: order by code, so 400 goes before 500. Same feedback throughout.

src/main/resources/assets/api-docs.yaml
```diff
@@ -2872,6 +2872,82 @@ paths:
...
+          required: True
+          schema: 
+            type: string
+      responses:
+        204: 
+          description: Deletes target institution (no return entity)
+        500: 
+          description: Internal Server Error
+        404:
+          description: "Institution not found"
```
rushtong: nit: rm the quotes.

src/test/java/org/broadinstitute/consent/http/service/InstitutionServiceTest.java
```diff
@@ -38,40 +39,126 @@ private void initService() {
...
-  public void testUpdateInstitutionById() {
+  public void testCreateInstitutionBlankName() {
+    Institution mockInstitution = initMockModel();
+    mockInstitution.setName("");
+    initService();
+    try{
+      service.createInstitution(mockInstitution, 1);
+      Assert.fail("Institution CREATE should not succeed with an empty name");
+    } catch(Exception e) {
+      assertEquals("Institution name cannot be null or empty", e.getMessage());
```
rushtong: nit: In general testing, I find that testing message string equality is unnecessary. I prefer to test exception types instead. It's more convenient to add an expect clause in the test annotation so that you know a certain call should throw a certain exception and then you can also simplify the code by removing the try/catch.
Similar feedback in other try/catch/assert blocks.

src/test/java/org/broadinstitute/consent/http/service/InstitutionServiceTest.java
```diff
@@ -38,40 +39,126 @@ private void initService() {
...
+      Assert.fail("Institution CREATE should not succeed with an empty name");
     } catch (Exception e) {
-      Assert.fail("Update should not fail");
+      assertEquals("Institution name cannot be null or empty", e.getMessage());
     }
   }
 
-  @Test(expected = Throwable.class)
-  public void testUpdateInstitutionByIdFail() {
+  //NOTE: would like to add a test for name duplication, but exception will differ depending on whether name column has the UNIQUE keyword applied
```
rushtong: See `changelog-consent-61.0.xml` ... we do not have a unique constraint on that column (yet). Feel free to add one or file a new ticket to add it as an update along with new tests.

src/test/java/org/broadinstitute/consent/http/service/InstitutionServiceTest.java
```diff
@@ -38,40 +39,126 @@ private void initService() {
...
-    doThrow(new Exception("Update method should pass on error from DAO")).when(institutionDAO).updateInstitutionById(anyInt(), anyString(), anyString(), anyString(), anyInt(), eq(new Date()));
-    service.updateInstitutionById(anyInt(), eq("New Name"), anyString(), anyString(), anyInt(), eq(new Date()));
+    try{
+      service.updateInstitutionById(mockInstitution, 1, 1);
+      Assert.fail("UPDATE should not succeed");
+    } catch(Exception e) {
+      assertEquals(error.getMessage(), e.getMessage());
+    }
+    
+  }
```
rushtong: Here's how I would rework this to be more generic
```
  @Test(expected = NotFoundException.class)
  public void testUpdateInstitutionByIdNotFound() {
    Institution mockInstitution = initMockModel();
    when(institutionDAO.findInstitutionById(anyInt())).thenThrow(new NotFoundException());
    initService();
    service.updateInstitutionById(mockInstitution, 1, 1);
  }
```

rushtong (changes requested):
Very minor suggestions inline, looking very good!

rushtong (approved):
Thank you for the changes!

raejohanek (approved):
works well, i love your tests!

PR #988, opened 2021-03-25
DUOS-1131[risk=no]Remove email updates from user APIs

raejohanek:
SCOPE:
the email should not be updatable
remove email field from update call
merge before DUOS-1131 in UI

Addresses:
https://broadworkbench.atlassian.net/browse/DUOS-1131
With DUOS-1131 in UI

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
@@ -86,7 +86,7 @@ public interface UserDAO extends Transactional<UserDAO> {
                           @Bind("displayName") String displayName,
                           @Bind("createDate") Date createDate);
 
-    @SqlUpdate("update dacuser set displayName=:displayName, additional_email=:additionalEmail where dacUserId=:id")
+    @SqlUpdate("UPDATE dacuser SET displayname=:displayName, additional_email=:additionalEmail WHERE dacuserid=:id")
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
```diff
@@ -86,9 +86,8 @@ Integer insertUser(@Bind("email") String email,
                           @Bind("displayName") String displayName,
                           @Bind("createDate") Date createDate);
 
-    @SqlUpdate("update dacuser set email=:email, displayName=:displayName, additional_email=:additionalEmail where dacUserId=:id")
-    void updateUser(@Bind("email") String email,
-                       @Bind("displayName") String displayName,
+    @SqlUpdate("update dacuser set displayName=:displayName, additional_email=:additionalEmail where dacUserId=:id")
```
rushtong: sql nit: We need to standardize the sql when we are touching it. Capitalize keywords and lower case column names:
```suggestion
    @SqlUpdate("UPDATE dacuser SET displayname=:displayName, additional_email=:additionalEmail WHERE dacuserid=:id")
```

rushtong (changes requested):
We also need to update the api-docs.yaml so it's clear that email cannot be updated with the update APIs. One minor comment inline, but looks good!

rushtong (approved):
Thank you for the changes! Assuming tests pass, 👍🏽

PR #991, opened 2021-03-29
[DUOS-925][risk=low] election api to service

rfricke-asymmetrik:
**Addresses**
https://broadworkbench.atlassian.net/browse/DUOS-925

**Changes**

- Remove ElectionAPI and interface
- Replace with ElectionService
- Add ElectionServiceTest and add tests for methods moved into ElectionService

In all cases:


src/test/java/org/broadinstitute/consent/http/resources/ElectionResourceTest.java
```diff
@@ -32,7 +31,6 @@
 @SuppressWarnings("FieldCanBeLocal")
 @RunWith(PowerMockRunner.class)
 @PowerMockIgnore("jdk.internal.reflect.*")
-@PrepareForTest({AbstractElectionAPI.class})
```
rushtong: I think we can get rid of all of these if we aren't using `AbstractElectionAPI` any more

rushtong (approved):
One minor cleanup suggestion inline, but looks great 👍🏽

PR #993, opened 2021-03-30
DUOS-1135[risk=no]Bug on Admin Manage Access and New Chair Console when creating Election

raejohanek:
SCOPE:
-remove breaking subquery for version in ElectionDAO.getElectionWithFinalVoteByReferenceIdAndType

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1135

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
@@ -79,7 +79,7 @@ public interface ElectionDAO extends Transactional<ElectionDAO> {
     @SqlQuery("SELECT e.electionId,  e.datasetId, v.vote finalVote, e.status, e.createDate, e.referenceId, v.rationale finalRationale, v.createDate finalVoteDate, "
             + " e.lastUpdate, e.finalAccessVote, e.electionType, e.dataUseLetter, e.dulName, e.archived, e.version FROM election e"
             + " INNER JOIN vote v on v.electionId = e.electionId and LOWER(v.type) = 'chairperson' "
-            + " WHERE e.referenceId = :referenceId AND lower(e.electionType) = LOWER(:type) LIMIT 1")
+            + " WHERE e.referenceId = :referenceId AND lower(e.electionType) = LOWER(:type) ORDER BY createDate desc LIMIT 1")
     Election getElectionWithFinalVoteByReferenceIdAndType(@Bind("referenceId") String referenceId, @Bind("type") String type);
 
     @SqlQuery("SELECT distinct "
--- a/src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
@@ -76,10 +76,10 @@ public interface ElectionDAO extends Transactional<ElectionDAO> {
             + " where e.referenceId = :referenceId and lower(e.status) = 'open' and lower(e.electionType) = lower(:type)")
     Election getOpenElectionWithFinalVoteByReferenceIdAndType(@Bind("referenceId") String referenceId, @Bind("type") String type);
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
```diff
@@ -76,10 +76,10 @@ void bulkCancelOpenElectionByReferenceIdAndType(@Bind("electionType") String ele
...
 
-    @SqlQuery("select e.electionId,  e.datasetId, v.vote finalVote, e.status, e.createDate, e.referenceId, v.rationale finalRationale, v.createDate finalVoteDate, "
-            + "e.lastUpdate, e.finalAccessVote, e.electionType, e.dataUseLetter, e.dulName, e.archived, e.version  from election e"
-            + " inner join vote v on v.electionId = e.electionId and lower(v.type) = 'chairperson' "
-            + " where e.referenceId = :referenceId and lower(e.electionType) = lower(:type) and e.version = (select MAX(version) from election where referenceId = :referenceId)")
+    @SqlQuery("SELECT e.electionId,  e.datasetId, v.vote finalVote, e.status, e.createDate, e.referenceId, v.rationale finalRationale, v.createDate finalVoteDate, "
+            + " e.lastUpdate, e.finalAccessVote, e.electionType, e.dataUseLetter, e.dulName, e.archived, e.version FROM election e"
+            + " INNER JOIN vote v on v.electionId = e.electionId and LOWER(v.type) = 'chairperson' "
+            + " WHERE e.referenceId = :referenceId AND lower(e.electionType) = LOWER(:type) LIMIT 1")
```
rushtong: We do still need the subquery, but the `version` version of these queries are wrong (and we need to update them where we see them). Instead of using `version`, we should instead use `createdate`. Version was not very well implemented and has led to quite a number of bugs in the past.

rushtong (changes requested):
Nice find ... see comment inline.

src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
raejohanek: so I know my solution is not a subquery but I believe sorting by descending create date and then only returning the first one has the same effect as sub querying for the max create date

raejohanek: So, when I check this solution using postico it works but when I test it out through the UI electionDAO.getElectionWithFinalVoteByReferenceIdAndType(requestId, ElectionType.DATA_ACCESS.getValue()) still returns null even though the same exact query returns the election in postico, I'm not sure if this is an issue with the way I'm testing it or if there is something else going on, @JVThomas  ? (this call is made from line 223 DatabaseElectionAPI.describeDataRequestElection)

rushtong: Using a limit 1 with a descending create date should be just as functional as the subquery approach.

rushtong: Just a note that limit+createdate+sort only works for single entity queries. When we need to pull a list of entities, we will need to use a subquery approach to ensuring we're getting the most recent election.

raejohanek: I don't understand what you mean because as i understand it this is always querying for a single entity. The reason I am using this approach is because with the subquery  for createdate it was the same issue as the subquery for version, it wasn't returning anything, but with the sort and limit 1 it returns the most recent

rushtong: THIS query always only returns a single election. In most other queries, we need to query for N elections, and we usually have to subquery to ensure we're getting the latest election for whatever criteria we're looking for, usually a reference id.

raejohanek: Ah I see!

rushtong (approved):
👍🏽

PR #1012, opened 2021-04-09
[DUOS-1178][risk=no] Update dataset queries

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1178

Update queries to outer joins so that datasets with some missing properties can be displayed.

----

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
@@ -1,18 +1,11 @@
 package org.broadinstitute.consent.http.db;
 
-import java.sql.Timestamp;
-import java.util.Collection;
-import java.util.Date;
-import java.util.List;
-import java.util.Set;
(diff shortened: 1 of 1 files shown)


JVThomas:
The update works for me locally. Just making sure, the left joins are tied to properties that don't have to exist on a dataset made via the form?

raejohanek (approved):
tested out locally and it works well for me!

rushtong:
@JVThomas 
> The update works for me locally. Just making sure, the left joins are tied to properties that don't have to exist on a dataset made via the form?

No, that's not what this is trying to do. The inner joins were preventing the DS from being visible if they were missing values. This prevents us from deleting them via the UI, because we can't see them. It also prevents any ability to fix them in the edit screen. 

That being said, we do have a restriction on the data use properties. Those are non-editable as a policy decision. Instead, the user must delete (which this PR helps us do, but not all the way, see DUOS-1179) and then recreate them.

PR #1013, opened 2021-04-09
[DUOS-1130][risk=no] Fix dataset delete by id/user

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1130

Clean up the delete by id/user code.

----

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataSetDAO.java
@@ -63,6 +63,9 @@ public interface DataSetDAO extends Transactional<DataSetDAO> {
     @SqlBatch("delete from datasetproperty where dataSetId = :dataSetId")
     void deleteDataSetsProperties(@Bind("dataSetId") Collection<Integer> dataSetsIds);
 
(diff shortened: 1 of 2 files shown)


raejohanek (changes requested):
when I test this out I still get the 500 failed to rollback error, is that bc this needs the 1178 PR too?

rushtong:
That can't happen, there's no rollback code in this PR. Remember to do a clean maven build.

JVThomas (approved):
Can confirm successful deletion of dataset and dataset properties with Swagger and Postico (local deletion).

JVThomas:
Would also like to add that the ```deleteDataset``` function is called on the exception handler for ```createDataset```. Could the old implementation with failed CREATE requests be the reason why impartial datasets exist in the database?

rushtong:
> Would also like to add that the deleteDataset function is called on the exception handler for createDataset. Could the old implementation be the reason why impartial datasets persisted?

It could be ... I haven't gotten to testing that specific error yet. Need to get some of these other bugs fixed before I can replicate the issue in https://broadworkbench.atlassian.net/browse/DUOS-1177

raejohanek (approved):
tested locally- deletes dataset, dataset properties, and adds delete to dataset audit

PR #1018, opened 2021-04-10
[DUOS-1180][risk=no] Update handling of delete dataset

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1180

## Changes
This PR is a significant overhaul of how we delete a dataset.  Prior to recent updates, we were handling transactions poorly. Recent updates were quick fixes to get past several bugs. This update handles deletion properly in a jdbi3-friendly transaction.

* Some class and variable renaming
* Moved the queries associated with deleting a dataset to the same DAO so we can wrap the whole process in a single, rollback-able transaction. More information on this here: https://jdbi.org/#_transactions
* Cleaned up unused methods and classes related to moving those delete queries
* Added `DELETE` to the standard auditable actions
* Removed the concept of "logical delete" of a dataset, this was not implemented well and has no purpose as we audit all deletions.

----

In all cases:


src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
```diff
@@ -66,6 +67,16 @@
...
 
+    @SqlUpdate("INSERT INTO dataset_audit (datasetid, changeaction, modifiedbyuser, modificationdate, objectid, name, active) VALUES (:dataSetId, :action, :user, :date, :objectId, :name, :active )")
+    @GetGeneratedKeys
+    Integer insertDataSetAudit(@BindBean DataSetAudit dataSets);
+
+    @SqlUpdate("DELETE FROM dataset_user_association WHERE datasetid = :datasetId")
+    void deleteUserAssociationsByDatasetId(@Bind("datasetId") Integer datasetId);
+
+    @SqlUpdate("DELETE FROM consentassociations WHERE datasetid = :datasetId")
+    void deleteConsentAssociationsByDataSetId(@Bind("datasetId") Integer datasetId);
```
rushtong, line 78: These queries were originally in other DAOs. These were moved to here so they can all be wrapped in a single transaction block.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -312,27 +307,34 @@ public DataSetDTO getDatasetDTO(Integer datasetId) {
...
+                    return true;
+                });
+            } catch (Exception e) {
+                logger.error(e.getMessage());
+                throw e;
             }
-        } catch (Exception e) {
-            throw new IllegalStateException(e.getMessage());
         }
     }
```
rushtong, line 339: This is the meat of the PR. Previous delete statements were spread across separate DAOs which prevented us from wrapping them in a single transaction. This ensures that all are committed, or none are.

src/main/java/org/broadinstitute/consent/http/db/ConsentDAO.java
```diff
@@ -130,9 +130,6 @@ void deleteAllAssociationsForType(@Bind("consentId") String consentId,
     @SqlUpdate("delete from consentassociations where consentId = :consentId")
     void deleteAllAssociationsForConsent(@Bind("consentId") String consentId);
 
-    @SqlUpdate("delete from consentassociations where dataSetId = :dataSetId")
-    void deleteAssociationsByDataSetId(@Bind("dataSetId") Integer dataSetId);
-
```
rushtong, line 135: moved to `DatasetDAO`

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -310,29 +305,29 @@ public DataSetDTO getDatasetDTO(Integer datasetId) {
...
+                        h.deleteDatasetById(datasetId);
+                    } catch (Exception e) {
+                        h.rollback();
+                        throw e;
+                    }
+                });
+            } catch (Exception e) {
+                logger.error(e.getMessage());
+                throw e;
             }
```
rushtong, line 330: This is the meat of the PR. Wrapping the individual DAO calls in this transaction allows for them to all succeed, or all fail.

raejohanek (commented):
tested out locally, I have a couple questions, after I delete a dataset with a successful response, and then try to find that dataset by ID instead of returning a 404 it returns a successful response with this body: {
  "dacId": null,
  "dataSetId": null,
  "consentId": null,
  "translatedUseRestriction": null,
  "deletable": null,
  "properties": null,
  "active": null,
  "needsApproval": null,
  "isAssociatedToDataOwners": null,
  "updateAssociationToDataOwnerAllowed": null,
  "alias": null,
  "objectId": null,
  "createDate": null,
  "createUserId": null,
  "updateDate": null,
  "updateUserId": null,
  "dataUse": null
}
Is this the expected behavior?
Additionally, when I try to delete a dataset that is missing information such as dataset 4 which does not have a name, delete gives a 500 error because it is trying to add the delete dataset call to the dataset_audit table which has a non-null constraint on the name

rushtong:
@raejohanek 

> after I delete a dataset with a successful response, and then try to find that dataset by ID instead of returning a 404 it returns a successful response with this body: {
> "dacId": null,
> "dataSetId": null,
> "consentId": null,
> "translatedUseRestriction": null,
> "deletable": null,
> "properties": null,
> "active": null,
> "needsApproval": null,
> "isAssociatedToDataOwners": null,
> "updateAssociationToDataOwnerAllowed": null,
> "alias": null,
> "objectId": null,
> "createDate": null,
> "createUserId": null,
> "updateDate": null,
> "updateUserId": null,
> "dataUse": null
> }
> Is this the expected behavior?

This looks like an error in the find by id code, it's returning the null object instead of throwing a 404. This looks like a separate bug ... will file a ticket for that.

rushtong:
@raejohanek 

> when I try to delete a dataset that is missing information such as dataset 4 which does not have a name, delete gives a 500 error because it is trying to add the delete dataset call to the dataset_audit table which has a non-null constraint on the name

This PR is trying to fix that exact problem, you're sure you have the right code pulled and built? See https://github.com/DataBiosphere/consent/pull/1018/files#diff-b48387687c627ee6703cc3122b9105cbc0197a3a1ec985a9f0bd0e5050f65e7eR312 for where we check on the name to ensure that we have something to enter into that column.

rushtong:
RE: null dataset, confirmed that this is an existing bug on dev:
```
curl -X 'GET' 'https://consent.dsde-dev.broadinstitute.org/api/dataset/20000' 
```
yields:
```
{
  "dacId": null,
  "dataSetId": null,
  "consentId": null,
  "translatedUseRestriction": null,
  "deletable": null,
  "properties": null,
  "active": null,
  "needsApproval": null,
  "isAssociatedToDataOwners": null,
  "updateAssociationToDataOwnerAllowed": null,
  "alias": null,
  "objectId": null,
  "createDate": null,
  "createUserId": null,
  "updateDate": null,
  "updateUserId": null,
  "dataUse": null
}
```

raejohanek:
> @raejohanek
> 
> > when I try to delete a dataset that is missing information such as dataset 4 which does not have a name, delete gives a 500 error because it is trying to add the delete dataset call to the dataset_audit table which has a non-null constraint on the name
> 
> This PR is trying to fix that exact problem, you're sure you have the right code pulled and built? See https://github.com/DataBiosphere/consent/pull/1018/files#diff-b48387687c627ee6703cc3122b9105cbc0197a3a1ec985a9f0bd0e5050f65e7eR312 for where we check on the name to ensure that we have something to enter into that column.

yeah, I see the solution in the code and that makes sense, I'm just not seeing it in action for some reason. I am on branch 1180 with all the latest changes I ran docker-compose down, then mvn clean compile and then docker-compose up and I am on the local swagger testing these out, any suggestions?

raejohanek (approved):
tested locally, delete works as expected

PR #1028, opened 2021-04-13
DUOS-1187[risk=no] Updated Researcher Manage Endpoint (removed userId as queryParam)

JVThomas:
Addresses [DUOS-1187](https://broadworkbench.atlassian.net/browse/DUOS-1187)

PR removes userId query param in favor of relying on current user stored in session as source of userId, a change that [has been implemented on the front-end](https://github.com/DataBiosphere/duos-ui/pull/991/files). PR also implements the use of ```try...catch``` and ```createExceptionResponse``` for error response handling.

Problem with initial code stemmed from the fact that the statement ```userId != null``` would evaluate to false, even though inspection of ```userId``` shows it to be ```null```. This incorrect evaluation would affect the filter conditionals in ```DataAccessRequestService.describeDataAccessRequestManage```

In all cases:


rushtong (approved):
Nice fix 👍🏽

PR #1031, opened 2021-04-15
DUOS-1192[risk=no]Cancelled DARs visible in Manage Screens

raejohanek:
SCOPE:
- filter out dars that have a cancelled status (any dar that has a status that is not null has a canceled status)

ADDRESS:
https://broadworkbench.atlassian.net/browse/DUOS-1192?

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
@@ -362,9 +362,10 @@ public class DataAccessRequestService {
         }
         DataAccessRequestData darData = dar.getData();
         List<Election> elections = electionDAO.findElectionsByReferenceId(referenceId);
-        List<Election> openElections = elections.stream().filter(e -> e.getStatus() == "Open").collect(Collectors.toCollection(ArrayList::new));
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -144,9 +144,13 @@ public Integer getTotalUnReviewedDars(AuthUser authUser) {
     public List<DataAccessRequestManage> describeDataAccessRequestManageV2(AuthUser authUser) {
         List<DataAccessRequest> allDars = findAllDataAccessRequests();
         List<DataAccessRequest> filteredAccessList = dacService.filterDataAccessRequestsByDac(allDars, authUser);
-        filteredAccessList.sort(sortTimeComparator());
-        if (CollectionUtils.isNotEmpty(filteredAccessList)) {
-            return createAccessRequestManageV2(filteredAccessList);
+        List<DataAccessRequest> openDarList = filteredAccessList.stream().filter(dar -> {
+            DataAccessRequestData data = dar.getData();
+            return (Objects.isNull(data) ? true : (Objects.isNull(data.getStatus())));
```
rushtong: This filters on a null status, which isn't exactly what we're looking for. What we want is to filter out anything with a status of "Canceled". The semantics are different in that there could be a status we want to include and this would lead to some head-scratching as to why certain statuses were not showing up.

rushtong (changes requested):
One comment inline, otherwise, looks good.

rushtong (approved):
👍🏽

PR #1042, opened 2021-04-23
DUOS-1227[risk=no] Filter nulls on Dataset Properties for AutoComplete

JVThomas:
Addresses [DUOS-1227](https://broadworkbench.atlassian.net/browse/DUOS-1227)

Problematic datasets (the ones that are incomplete or partially saved) may have null values for certain properties. This null value can cause the ```toLowerCase()``` function call in ```DatasetService.autoCompleteDatasets``` (around line 390) to error out. The PR adds an additional null check on those properties within the filter function.

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
@@ -407,16 +407,19 @@ public class DatasetService {
 
     private boolean filterDatasetOnProperties(DatasetDTO dataset, String term) {
         //datasets need to have consentId, null check to prevent NPE
-        if(Objects.isNull(dataset.getConsentId())) {
-            return false;
-        }
+        Boolean consentIdMatch;
         String consentId = dataset.getConsentId();
+        if(Objects.isNull(consentId)) {
+            consentIdMatch = false;
+        } else {
+            consentIdMatch = consentId.toLowerCase().contains(term);
+        }
         return dataset.getProperties()
         .stream()
         .filter(p -> Objects.nonNull(p.getPropertyValue()))
         .anyMatch(p -> {
-            return (consentId.toLowerCase().contains(term) 
-            || p.getPropertyValue().toLowerCase().contains(term));
+            return consentIdMatch
+            || p.getPropertyValue().toLowerCase().contains(term);
         });
     }
 
--- a/src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
@@ -388,7 +388,13 @@ public class DatasetService {
                     ds.getConsentId().toLowerCase().contains(lowercasePartial) ||
                     ds.getProperties().stream()
                           .anyMatch(
-                                p -> p.getPropertyValue().toLowerCase().contains(lowercasePartial))
+                                p -> {
+                                    String value = p.getPropertyValue();
+                                    if(Objects.isNull(value)) {
+                                        value = "";
+                                    }
+                                    return value.toLowerCase().contains(lowercasePartial);
+                                })
               )).collect(Collectors.toSet());
         return filteredDatasetsContainingPartial.stream().map(ds ->
               {
--- a/src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
@@ -406,6 +406,10 @@ public class DatasetService {
     }
 
     private boolean filterDatasetOnProperties(DatasetDTO dataset, String term) {
+        //datasets need to have consentId, null check to prevent NPE
+        if(Objects.isNull(dataset.getConsentId())) {
+            return false;
+        }
         String consentId = dataset.getConsentId();
         return dataset.getProperties()
         .stream()
--- a/src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
@@ -407,13 +407,8 @@ public class DatasetService {
 
     private boolean filterDatasetOnProperties(DatasetDTO dataset, String term) {
         //datasets need to have consentId, null check to prevent NPE
-        Boolean consentIdMatch;
         String consentId = dataset.getConsentId();
-        if(Objects.isNull(consentId)) {
-            consentIdMatch = false;
-        } else {
-            consentIdMatch = consentId.toLowerCase().contains(term);
-        }
+        Boolean consentIdMatch = Objects.isNull(consentId) ? false : consentId.toLowerCase().contains(term);
         return dataset.getProperties()
             .stream()
             .filter(p -> Objects.nonNull(p.getPropertyValue()))
--- a/src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
@@ -415,12 +415,12 @@ public class DatasetService {
             consentIdMatch = consentId.toLowerCase().contains(term);
         }
         return dataset.getProperties()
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -388,7 +388,13 @@ public void deleteDataset(Integer datasetId, Integer userId) throws Exception {
...
                     ds.getProperties().stream()
                           .anyMatch(
-                                p -> p.getPropertyValue().toLowerCase().contains(lowercasePartial))
+                                p -> {
+                                    String value = p.getPropertyValue();
+                                    if(Objects.isNull(value)) {
+                                        value = "";
+                                    }
+                                    return value.toLowerCase().contains(lowercasePartial);
+                                })
```
rushtong: A simpler way to do this is just to filter out the nulls before trying to examine them, that way you know you won't hit the error. In this solution, the empty string should match any value provided which looks like it might be giving you too many positive results.

This should work:
```
                    ds.getProperties().stream()
                            .filter(Objects::nonNull)
                            .filter(p -> Objects.nonNull(p.getPropertyValue()))
                            .anyMatch(p -> p.getPropertyValue().toLowerCase().contains(lowercasePartial))
```
Additionally, just above this, we have the same problem on line 386. I think it might be better to do the filtering before either of these checks happen, probably right after line 383.

rushtong (changes requested):
See comments inline, I think the problem goes a little deeper than this block.

rushtong (approved):
Changes look good 👍🏽

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -406,6 +406,10 @@ public void deleteDataset(Integer datasetId, Integer userId) throws Exception {
     }
 
     private boolean filterDatasetOnProperties(DatasetDTO dataset, String term) {
+        //datasets need to have consentId, null check to prevent NPE
+        if(Objects.isNull(dataset.getConsentId())) {
+            return false;
+        }
         String consentId = dataset.getConsentId();
         return dataset.getProperties()
         .stream()
```
rushtong: If you return false on 411, then you'll never do any checking of the properties. I think you want to just ignore it if it isn't there, not return early.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -411,6 +405,24 @@ public void deleteDataset(Integer datasetId, Integer userId) throws Exception {
...
+        } else {
+            consentIdMatch = consentId.toLowerCase().contains(term);
+        }
+        return dataset.getProperties()
+        .stream()
+        .filter(p -> Objects.nonNull(p.getPropertyValue()))
+        .anyMatch(p -> {
+            return consentIdMatch
+            || p.getPropertyValue().toLowerCase().contains(term);
+        });
```
rushtong: Minor nit: indentation. Not sure what your IDE settings are, but it should indent 4 or 8 spaces. If your IDE let's you use [Google's Java Style](https://medium.com/swlh/configuring-google-style-guide-for-java-for-intellij-c727af4ef248), see if you can select that and it should format your code automatically.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -411,6 +405,24 @@ public void deleteDataset(Integer datasetId, Integer userId) throws Exception {
...
+        } else {
+            consentIdMatch = consentId.toLowerCase().contains(term);
+        }
+        return dataset.getProperties()
+            .stream()
+            .filter(p -> Objects.nonNull(p.getPropertyValue()))
+            .anyMatch(p -> {
+                return consentIdMatch
+                || p.getPropertyValue().toLowerCase().contains(term);
+            });
```
rushtong: Whenever you find yourself creating intermediate holding variables, there is usually a more fluid and less memory-consuming approach. My refactoring approach to this would be to first try and remove the need for `consentId`:

```
        consentIdMatch = Objects.isNull(dataset.getConsentId()) ? 
                false :
                dataset.getConsentId().contains(term);
```

That's a literal translation of 411-416. In general, I like to check the positive case first, then let the negative case be second, so I'd rewrite that as:

```
        consentIdMatch = Objects.nonNull(dataset.getConsentId()) ?
                dataset.getConsentId().contains(term) :
                false;
```
Groovy has a cool operator for this called the [Elvis Operator](https://groovy-lang.org/operators.html#_elvis_operator) which simplifies that even more, but we're in java land

Now that we have a single assignment, we can instead use a primitive, and 7 loc reduces down to 3

```
        boolean consentIdMatch = Objects.nonNull(dataset.getConsentId()) ?
                dataset.getConsentId().contains(term) :
                false;
```

Once I have that, Intelli-j suggests that it can be simplified even more, which it can:

```
        boolean consentIdMatch = Objects.nonNull(dataset.getConsentId()) && 
                dataset.getConsentId().contains(term);
```

Now, we have a single boolean and the properties to look at. By putting the `consentIdMatch` inside the `anyMatch`, we're forcing the code to loop over the properties. In the true case, it will return after the first property, but we don't need to since we know it's a match. We only need to look at properties in the false case. I would take that out of the inside loop and instead move it before the property loop so the loop gets short-circuited:

```
        boolean consentIdMatch = Objects.nonNull(dataset.getConsentId()) &&
                dataset.getConsentId().contains(term);
        return consentIdMatch || dataset.getProperties()
                .stream()
                .filter(p -> Objects.nonNull(p.getPropertyValue()))
                .anyMatch(p -> p.getPropertyValue().toLowerCase().contains(term));
```

I think that is as concise as I can imagine it being.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -380,16 +381,9 @@ public void deleteDataset(Integer datasetId, Integer userId) throws Exception {
...
-                          p -> p.getPropertyName().equalsIgnoreCase("Principal Investigator(PI)"))
-                    &&
-                    ds.getConsentId().toLowerCase().contains(lowercasePartial) ||
-                    ds.getProperties().stream()
-                          .anyMatch(
-                                p -> p.getPropertyValue().toLowerCase().contains(lowercasePartial))
-              )).collect(Collectors.toSet());
+        Set<DatasetDTO> filteredDatasetsContainingPartial = datasets.stream()
+        .filter(ds -> filterDatasetOnProperties(ds, lowercasePartial))
+        .collect(Collectors.toSet());
```
rushtong, line 386: nit: indentation

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -411,6 +405,19 @@ public void deleteDataset(Integer datasetId, Integer userId) throws Exception {
...
+        //datasets need to have consentId, null check to prevent NPE
+        String consentId = dataset.getConsentId();
+        Boolean consentIdMatch = Objects.nonNull(consentId) && consentId.toLowerCase().contains(term);
+        return dataset.getProperties()
+            .stream()
+            .filter(p -> Objects.nonNull(p.getPropertyValue()))
+            .anyMatch(p -> {
+                return consentIdMatch
+                || p.getPropertyValue().toLowerCase().contains(term);
+            });
```
rushtong: Minor optimization can be made by pulling `consentIdMatch ||` out into the front of the overall return statement:
```suggestion
        return consentIdMatch || dataset.getProperties()
            .stream()
            .filter(p -> Objects.nonNull(p.getPropertyValue()))
            .anyMatch(p -> {
                return p.getPropertyValue().toLowerCase().contains(term);
            });
```

rushtong (approved):
Minor feedback inline that shouldn't hold up this PR - feel free to merge without making any changes.

PR #1044, opened 2021-04-26
[DUOS-1226][risk=no] Update Roles Allowed Query

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1226

## Changes
Minor update to the underlying query that gets a user's role names. Previously, with a case sensitive search, if the email address the user used to authenticate with is different from what was initially saved in our user table, then an authenticated user would experience 403 errors when hitting any `@RolesAllowed` endpoint. When that same authenticated user hit a `@PermitAll` endpoint, the APIs would respond correctly. This was because the role names query is a different code path from the `AuthUser` query. 

This PR: 
* Standardizes how we match on email
* Streamlines the role name query since we don't need full `UserRole`s for what we're doing, we only need the names
* Adds a test to ensure case doesn't matter when querying for roles

----

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/UserRoleDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/UserRoleDAO.java
@@ -27,7 +27,7 @@ public interface UserRoleDAO extends Transactional<UserRoleDAO> {
             "  INNER JOIN user_role ur ON ur.role_id = r.roleid " +
             "  INNER JOIN dacuser u ON u.dacuserid = ur.user_id " +
             "  WHERE LOWER(u.email) = LOWER(:email)")
-    List<String> findRoleNamessByUserEmail(@Bind("email") String email);
+    List<String> findRoleNamesByUserEmail(@Bind("email") String email);
 
     @UseRowMapper(DatabaseRoleMapper.class)
     @SqlQuery("select * from roles")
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/consent/http/db/UserRoleDAO.java
```diff
@@ -22,9 +22,12 @@
...
 
-    @SqlQuery("select * from roles r inner join user_role ur on ur.role_id = r.roleId  " +
-              "inner join dacuser u on u.dacUserId = ur.user_id where u.email = :email")
-    List<UserRole> findRolesByUserEmail(@Bind("email") String email);
+    @SqlQuery("SELECT DISTINCT name " +
+            "  FROM roles r " +
+            "  INNER JOIN user_role ur ON ur.role_id = r.roleid " +
+            "  INNER JOIN dacuser u ON u.dacuserid = ur.user_id " +
+            "  WHERE LOWER(u.email) = LOWER(:email)")
+    List<String> findRoleNamesByUserEmail(@Bind("email") String email);
```
rushtong, line 30: This is the meat of the PR

JVThomas (approved):
Can reproduce the bug locally by changing the casing on the saved user name table. PR now allows the user to log in with a case-insensitive email value.

Emails are usually handled in a case-insensitive manner, so it makes sense for us to do the same.

PR #1049, opened 2021-04-28
DUOS-1198[risk=no] DAR Cancel Not Allowed if Open/Approved/Denied elections exist

JVThomas:
Addresses [DUOS-1198](https://broadworkbench.atlassian.net/browse/DUOS-1198)

PR prevents researchers from canceling a submitted DAR if an election has been created for that DAR. New criteria is checked for in ```DataAccessRequestService.cancelDataAccessRequest```. Tests have also been updated to account for this new condition. 

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -2,9 +2,7 @@ package org.broadinstitute.consent.http.resources;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.inject.Inject;
-import freemarker.template.TemplateException;
 import io.dropwizard.auth.Auth;
-import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
@@ -15,7 +13,6 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.annotation.security.PermitAll;
 import javax.annotation.security.RolesAllowed;
-import javax.mail.MessagingException;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.ForbiddenException;
 import javax.ws.rs.GET;
@@ -34,7 +31,6 @@ import org.broadinstitute.consent.http.models.DataAccessRequest;
 import org.broadinstitute.consent.http.models.DataAccessRequestManage;
 import org.broadinstitute.consent.http.models.User;
 import org.broadinstitute.consent.http.models.darsummary.DARModalDetailsDTO;
-import org.broadinstitute.consent.http.models.dto.Error;
 import org.broadinstitute.consent.http.service.ConsentService;
 import org.broadinstitute.consent.http.service.DataAccessRequestService;
 import org.broadinstitute.consent.http.service.ElectionService;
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -217,14 +213,8 @@ public Response describeDraftManageDataAccessRequests(@Auth AuthUser authUser) {
     public Response cancelDataAccessRequest(@Auth AuthUser authUser, @PathParam("referenceId") String referenceId) {
         validateAuthedRoleUser(Collections.emptyList(), authUser, referenceId);
         try {
-            List<User> usersToNotify = dataAccessRequestService.getUserEmailAndCancelElection(referenceId);
```
rushtong, line 220: This method is no longer used by any application code, just in tests. We can remove it from the service class. There are likely other unused methods as a result of removing this functionality.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -217,14 +213,8 @@ public Response describeDraftManageDataAccessRequests(@Auth AuthUser authUser) {
...
             DataAccessRequest dar = dataAccessRequestService.cancelDataAccessRequest(referenceId);
-            if (CollectionUtils.isNotEmpty(usersToNotify)) {
-                emailNotifierService.sendCancelDARRequestMessage(usersToNotify, dar.getData().getDarCode());
```
rushtong, line 223: This method is no longer in use, we can remove it from the email service.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -363,6 +364,11 @@ public DataAccessRequest cancelDataAccessRequest(String referenceId) {
         if (Objects.isNull(dar)) {
             throw new NotFoundException("Unable to find Data Access Request with the provided id: " + referenceId);
         }
+        List<Election> elections = electionDAO.findElectionsByReferenceId(referenceId);
+        if (!elections.isEmpty()) {
+            //Is this the right exception to use?
+            throw new IllegalArgumentException("Cancelling this DAR is not allowed");
```
rushtong: I think `UnsupportedOperationException` is more appropriate here. The reference id is technically a valid argument, but we are choosing not to perform the requested operation due to business policy. You could even make an argument that this is a 409 conflict exception, but I think that's a bit of a stretch.

rushtong (changes requested):
See comments inline. Good work, but there is some additional cleanup that can be done.

raejohanek (approved):
Code looks good! nice clean up!

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -363,6 +362,11 @@ public DataAccessRequest cancelDataAccessRequest(String referenceId) {
         if (Objects.isNull(dar)) {
             throw new NotFoundException("Unable to find Data Access Request with the provided id: " + referenceId);
         }
+        List<Election> elections = electionDAO.findElectionsByReferenceId(referenceId);
+        if (!elections.isEmpty()) {
+            //Is this the right exception to use?
```
rushtong: nit, outdated comment

rushtong (approved):
One minor comment inline, then 👍🏽

PR #1050, opened 2021-04-28
DUOS-1215[risk=no]Add missing final vote info to Review_Results page

raejohanek:
SCOPE:
Currently, the UI calls getAccessElectionReviewByReferenceId to render the votes on the access review and review result page, with the boolean flag true meaning isfinalaccess. Due to this, the call in consent was filtering for only votes with FINAL type. These are all chair votes. This means that we are not receiving the votes for the members. 
This PR removes that boolean flag.
By switching to use this DAO call that has no boolean flag, we receive the DAC votes and the FINAL votes. This update is needed for this PR so that we can display both member and chair votes distinctly on the review page. Currently, the pages say "How members voted" but only show chair votes, so this will allow the page to be more accurate. 

DO NOT MERGE WITHOUT DUOS-1215 on UI

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1215

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/VoteDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/VoteDAO.java
@@ -28,8 +28,10 @@ public interface VoteDAO extends Transactional<VoteDAO> {
     @UseRowMapper(ElectionReviewVoteMapper.class)
     List<ElectionReviewVote> findElectionReviewVotesByElectionId(@Bind("electionId") Integer electionId);
 
-    @SqlQuery("select v.*, u.email, u.displayName from vote v inner join election on election.electionId = v.electionId inner join dacuser u on u.dacUserId = v.dacUserId "
-            + "where election.electionId = :electionId")
+    @SqlQuery("SELECT v.*, u.email, u.displayName FROM vote v " 
+            + " INNER JOIN election ON election.electionId = v.electionId "
+            + " INNER JOIN dacuser u ON u.dacUserId = v.dacUserId "
+            + " WHERE election.electionId = :electionId")
     @UseRowMapper(ElectionReviewVoteMapper.class)
     List<ElectionReviewVote> findAllElectionReviewVotesByElectionId(@Bind("electionId") Integer electionId);
 
--- a/src/main/java/org/broadinstitute/consent/http/db/VoteDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/VoteDAO.java
@@ -26,6 +26,11 @@ public interface VoteDAO extends Transactional<VoteDAO> {
     @SqlQuery("select v.*, u.email, u.displayName from vote v inner join election on election.electionId = v.electionId inner join dacuser u on u.dacUserId = v.dacUserId "
             + "where election.electionId = :electionId")
     @UseRowMapper(ElectionReviewVoteMapper.class)
+    List<ElectionReviewVote> findAllElectionReviewVotesByElectionId(@Bind("electionId") Integer electionId);
+
+    @SqlQuery("select v.*, u.email, u.displayName from vote v inner join election on election.electionId = v.electionId inner join dacuser u on u.dacUserId = v.dacUserId "
+    + "where election.electionId = :electionId and lower(v.type) != 'chairperson'")
+    @UseRowMapper(ElectionReviewVoteMapper.class)
     List<ElectionReviewVote> findElectionReviewVotesByElectionId(@Bind("electionId") Integer electionId);
 
     @SqlQuery("select v.*, u.email, u.displayName from vote v inner join election on election.electionId = v.electionId inner join dacuser u on u.dacUserId = v.dacUserId "
--- a/src/main/java/org/broadinstitute/consent/http/db/VoteDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/VoteDAO.java
@@ -24,7 +24,7 @@ public interface VoteDAO extends Transactional<VoteDAO> {
     List<Vote> findVotesByReferenceId(@Bind("referenceId") String referenceId);
 
     @SqlQuery("select v.*, u.email, u.displayName from vote v inner join election on election.electionId = v.electionId inner join dacuser u on u.dacUserId = v.dacUserId "
-            + "where election.electionId = :electionId and lower(v.type) != 'chairperson'")
+            + "where election.electionId = :electionId")
     @UseRowMapper(ElectionReviewVoteMapper.class)
     List<ElectionReviewVote> findElectionReviewVotesByElectionId(@Bind("electionId") Integer electionId);
 
--- a/src/main/java/org/broadinstitute/consent/http/db/VoteDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/VoteDAO.java
@@ -24,14 +24,14 @@ public interface VoteDAO extends Transactional<VoteDAO> {
     List<Vote> findVotesByReferenceId(@Bind("referenceId") String referenceId);
 
     @SqlQuery("select v.*, u.email, u.displayName from vote v inner join election on election.electionId = v.electionId inner join dacuser u on u.dacUserId = v.dacUserId "
-            + "where election.electionId = :electionId")
-    @UseRowMapper(ElectionReviewVoteMapper.class)
-    List<ElectionReviewVote> findAllElectionReviewVotesByElectionId(@Bind("electionId") Integer electionId);
-
-    @SqlQuery("select v.*, u.email, u.displayName from vote v inner join election on election.electionId = v.electionId inner join dacuser u on u.dacUserId = v.dacUserId "
     + "where election.electionId = :electionId and lower(v.type) != 'chairperson'")
     @UseRowMapper(ElectionReviewVoteMapper.class)
     List<ElectionReviewVote> findElectionReviewVotesByElectionId(@Bind("electionId") Integer electionId);
+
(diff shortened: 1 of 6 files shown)


rushtong:
The underlying premise of this PR is not accurate. The UI does query for all votes on an election. I think the right approach here is to update the UI to make use of the votes it is already receiving. See screenshot for an example where you can see that we're getting votes of all types for the election.

Screen Shot 2021-04-29 at 6 35 03 AM

raejohanek:
> The underlying premise of this PR is not accurate. The UI does query for all votes on an election. I think the right approach here is to update the UI to make use of the votes it is already receiving. See screenshot for an example where you can see that we're getting votes of all types for the election.
> 
> Screen Shot 2021-04-29 at 6 35 03 AM

To see the votes I am talking about go to to review results page and look at the accessElectionReview that is saved in state, this contains the votes that are being displayed in the vote summaries and this includes only the chairperson type FINAL votes, meaning all member votes are being ignored, and never get displayed. This PR attempts to correct that functionality by receiving DAC (member) and FINAL (chair) votes. If we want to get away from using accessElectionReview altogether, like you were saying in stand up, then that is a larger issue outside the scope of this PR. But on that topic, the only information the electionreview objects are distinctly used for are the reviewvotes which contain user names. So, if we want to eliminate this object altogether, then that is the one thing we will need to implement elsewhere.

src/main/java/org/broadinstitute/consent/http/resources/ElectionReviewResource.java
```diff
@@ -66,7 +66,7 @@ public ElectionReview getElectionReviewByElectionId(@PathParam("electionId") Int
     @Path("access/{electionId}")
     @Produces("application/json")
     @RolesAllowed({ADMIN, MEMBER, CHAIRPERSON, ALUMNI})
-    public ElectionReview getAccessElectionReviewByReferenceId(@PathParam("electionId") Integer electionId, @QueryParam("isFinalAccess") Boolean isFinalAccess) {
+    public ElectionReview getAccessElectionReviewByReferenceId(@PathParam("electionId") Integer electionId) {
```
rushtong: This changes the api signature which would require a change to the api-docs as well. This method is called from the UI and provides the `isFinalAccess` as well so this PR will require a companion on the UI side.

src/main/java/org/broadinstitute/consent/http/resources/ElectionReviewResource.java
```diff
@@ -75,7 +75,7 @@ public ElectionReview getAccessElectionReviewByReferenceId(@PathParam("electionI
             dataSetId.addAll(dar.getData().getDatasetIds());
         }
         Consent consent = consentService.getConsentFromDatasetID(dataSetId.get(0));
-        ElectionReview accessElectionReview = service.describeElectionReviewByElectionId(electionId, isFinalAccess);
+        ElectionReview accessElectionReview = service.describeElectionReviewByElectionId(electionId, null);
```
rushtong: I think using null for an argument as a generic anti-pattern. We should strive to avoid it if at all possible. That said, it isn't always possible and sometimes we need to handle legacy code this way.

src/main/java/org/broadinstitute/consent/http/service/ReviewResultsService.java
```diff
@@ -48,7 +48,10 @@ public ElectionReview describeElectionReviewByElectionId(Integer electionId, Boo
             review = new ElectionReview();
             review.setElection(election);
             Consent consent = consentDAO.findConsentById(review.getElection().getReferenceId());
-            List<ElectionReviewVote> rVotes = (isFinalAccess == null || isFinalAccess == false) ? voteDAO.findElectionReviewVotesByElectionId(electionId, VoteType.DAC.getValue()) :  voteDAO.findElectionReviewVotesByElectionId(electionId, VoteType.FINAL.getValue());
+            List<ElectionReviewVote> rVotes = isFinalAccess == null ? 
+              voteDAO.findElectionReviewVotesByElectionId(electionId)
+              : isFinalAccess == false ? voteDAO.findElectionReviewVotesByElectionId(electionId, VoteType.DAC.getValue())
+                : voteDAO.findElectionReviewVotesByElectionId(electionId, VoteType.FINAL.getValue());
```
rushtong: The nested ternaries here add a significant degree of complexity that's hard to follow. I'd like to see this broken down so the logic can be read more quickly and cleanly. 

What I think makes this more complex is that we now have 3 operational states of `isFinalAccess`, null, true, or false, and each has different semantics. Using null this way is definitely an anti-pattern. A better approach would be to create an overloaded method with the arguments you want, and then calling that one from the resource class, it might look like this ... 

```
    public ElectionReview describeElectionReviewByElectionId(Integer electionId) {
        ElectionReview review = null;
        Election election = electionDAO.findElectionWithFinalVoteById(electionId);
        if(election != null){
            review = new ElectionReview();
            review.setElection(election);
            Consent consent = consentDAO.findConsentById(review.getElection().getReferenceId());
            List<ElectionReviewVote> rVotes = voteDAO.findElectionReviewVotesByElectionId(electionId);
            review.setReviewVote(rVotes);
            review.setConsent(consent);
        }
        return review;
    }
```

rushtong (changes requested):
See comments inline. Let's make sure this is well tested against the UI.

src/main/java/org/broadinstitute/consent/http/service/ReviewResultsService.java
raejohanek: yeah good idea I agree and if we want to make this same change for the rp election, so that we are receiving both member and chair votes, then we could get rid of the isfinalaccess boolean flag altogether and only have this version

rushtong: That also makes a lot of sense. I'm a little worried that would introduce unexpected regressions, but it's probably worth the effort to do that now rather than later.

rushtong (approved):
Pending a successful test run, 👍🏽. Try re-running the failing job, I believe it's a flaky test that might have been made more flaky with a recent library update.

raejohanek:
> Pending a successful test run, 👍🏽. Try re-running the failing job, I believe it's a flaky test that might have been made more flaky with a recent library update.

After testing this through the UI, I made a change to the voteDAO, I removed the filter on vote type != 'chairperson'. On the access election this filter does not cause any problems because every chairperson has a chairperson vote, and a final vote. But on the rp election, the chairs don't have a final vote. So this filter is causing the "How committee chairs voted:" section for question 2 not to populate. If we don't want to remove this filter for some reason (I don't know why it is there in the first place) we could simply replace it with type != 'final'

rushtong:
@raejohanek I'm not following this conversation well, I think we need to talk through what's going on.

> After testing this through the UI, I made a change to the voteDAO, I removed the filter on vote type != 'chairperson'. On the access election this filter does not cause any problems because every chairperson has a chairperson vote, and a final vote. But on the rp election, the chairs don't have a final vote. So this filter is causing the "How committee chairs voted:" section for question 2 not to populate. If we don't want to remove this filter for some reason (I don't know why it is there in the first place) we could simply replace it with type != 'final'

src/main/java/org/broadinstitute/consent/http/db/VoteDAO.java
```diff
@@ -28,6 +28,11 @@
     @UseRowMapper(ElectionReviewVoteMapper.class)
     List<ElectionReviewVote> findElectionReviewVotesByElectionId(@Bind("electionId") Integer electionId);
 
+    @SqlQuery("select v.*, u.email, u.displayName from vote v inner join election on election.electionId = v.electionId inner join dacuser u on u.dacUserId = v.dacUserId "
+            + "where election.electionId = :electionId")
+    @UseRowMapper(ElectionReviewVoteMapper.class)
+    List<ElectionReviewVote> findAllElectionReviewVotesByElectionId(@Bind("electionId") Integer electionId);
+
```
raejohanek, line 37: made new DAO call instead of modifying

src/main/java/org/broadinstitute/consent/http/resources/ElectionReviewResource.java
rushtong: Here and below, to make this more backwards compatible, leave the query param in place so that the method signature doesn't change, but instead just ignore the query param. That way, UI code can change separately from this and either call this endpoint with or without the query param. In a follow-up PR, we can remove the unused query param. That maintains backward compatibility and allows for the back end and the front end to be released independently.

rushtong (changes requested):
One requested change to keep the original method signatures in place. After that, 👍🏽

src/main/resources/assets/api-docs.yaml
```diff
@@ -2819,7 +2819,7 @@ paths:
                 $ref: '#/components/schemas/ElectionReview'
   /api/electionReview/rp/{electionId}:
     get:
-      summary: describeElectionById
+      summary: getRPElectionReviewByReferenceId
```
rushtong: This should be a plain text shortened version of the description, not the method name. The method-name pattern was used by Belatrix, but we should update these to be more descriptive so the swagger UI is easier to read. Don't hold up the PR on this, so feel free to merge if you do update these text summaries.

src/main/resources/assets/api-docs.yaml
```diff
@@ -2773,7 +2773,7 @@ paths:
                 $ref: '#/components/schemas/ElectionReview'
   /api/electionReview/access/{electionId}:
     get:
-      summary: describeElectionById
+      summary: getAccessElectionReviewByReferenceId
```
rushtong: Same description comment here as below.

rushtong (approved):
Two minor suggestions inline, but then 👍🏽

PR #1058, opened 2021-04-29
DUOS-1155[risk=no]Add API: get votes by Election IDs, use instead of get votes by reference ID

raejohanek:
SCOPE:
DO NOT MERGE WITHOUT PR DUOS-1155 in DUOS-UI

Add getVotesByElectionIds to vote resource (the plumbing in service and DAO is already there to support it) remove describeAllVotes which will be unused because of this change and the UI change
This will prevent the UI from querying for hundreds on unused votes just to filter out the ones for the correct electionIds

Additionally, a minor change to createElection, not functional change, but update response to accurately reflect what the method does. This change will be incredibly beneficial now that the query on the UI will be using electionIds instead of a dar referenceId. Because once the elections are created we will easily be able to query for their votes. If we did not add the RP election to the response then we would not be able to easily query for its votes after the election is created.

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1155 

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataRequestElectionResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataRequestElectionResource.java
@@ -3,7 +3,6 @@ package org.broadinstitute.consent.http.resources;
 import com.google.inject.Inject;
 import java.io.File;
 import java.net.URI;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collectors;
@@ -60,14 +59,13 @@ public class DataRequestElectionResource extends Resource {
                                               @PathParam("requestId") String requestId) {
         URI uri;
         Election accessElection = null;
-        Election rpElection = null;
         try {
             DataAccessRequest dar = darService.findByReferenceId(requestId);
             boolean manualReview = dar.requiresManualReview();
             accessElection = electionService.createElection(rec, requestId, ElectionType.DATA_ACCESS);
             List<Vote> votes = voteService.createVotes(accessElection, ElectionType.DATA_ACCESS, manualReview);
             //create RP election
-            rpElection = electionService.createElection(rec, requestId, ElectionType.RP);
+            Election rpElection = electionService.createElection(rec, requestId, ElectionType.RP);
             voteService.createVotes(rpElection, ElectionType.RP, false);
             List<Vote> darVotes = votes.stream().
                     filter(vote -> vote.getType().equals(VoteType.DAC.getValue())).
@@ -84,8 +82,7 @@ public class DataRequestElectionResource extends Resource {
             }
             return createExceptionResponse(e);
         }
-        List<Election> elections = Arrays.asList(accessElection, rpElection);
-        return Response.created(uri).entity(elections).build();
+        return Response.created(uri).entity(accessElection).build();
     }
 
 
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataRequestElectionResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataRequestElectionResource.java
@@ -3,6 +3,7 @@ package org.broadinstitute.consent.http.resources;
 import com.google.inject.Inject;
 import java.io.File;
 import java.net.URI;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collectors;
@@ -59,13 +60,14 @@ public class DataRequestElectionResource extends Resource {
                                               @PathParam("requestId") String requestId) {
         URI uri;
         Election accessElection = null;
+        Election rpElection = null;
         try {
             DataAccessRequest dar = darService.findByReferenceId(requestId);
             boolean manualReview = dar.requiresManualReview();
             accessElection = electionService.createElection(rec, requestId, ElectionType.DATA_ACCESS);
             List<Vote> votes = voteService.createVotes(accessElection, ElectionType.DATA_ACCESS, manualReview);
             //create RP election
-            Election rpElection = electionService.createElection(rec, requestId, ElectionType.RP);
+            rpElection = electionService.createElection(rec, requestId, ElectionType.RP);
             voteService.createVotes(rpElection, ElectionType.RP, false);
             List<Vote> darVotes = votes.stream().
                     filter(vote -> vote.getType().equals(VoteType.DAC.getValue())).
@@ -82,7 +84,8 @@ public class DataRequestElectionResource extends Resource {
             }
             return createExceptionResponse(e);
         }
-        return Response.created(uri).entity(accessElection).build();
+        List<Election> elections = Arrays.asList(accessElection, rpElection);
+        return Response.created(uri).entity(elections).build();
     }
 
 
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
@@ -164,8 +164,8 @@ public class DataRequestVoteResource extends Resource {
     @GET
     @Produces("application/json")
     @PermitAll
-    public List<Vote> describeAllVotes(@PathParam("requestId") String requestId) {
-        return voteService.describeVotes(requestId);
+    public List<Vote> describeVotesOnElections(@PathParam("electionIds") List<Integer> electionIds) {
+        return voteService.findVotesByElectionIds(electionIds);
     }
 
     @DELETE
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
@@ -164,7 +164,7 @@ public class DataRequestVoteResource extends Resource {
     @GET
     @Produces("application/json")
     @PermitAll
-    public List<Vote> describeVotesOnElections(@PathParam("electionIds") List<Integer> electionIds) {
+    public List<Vote> describeVotesOnElections(@PathParam("requestId") String requestId, List<Integer> electionIds) {
         return voteService.findVotesByElectionIds(electionIds);
     }
 
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
@@ -164,8 +164,8 @@ public class DataRequestVoteResource extends Resource {
     @GET
     @Produces("application/json")
     @PermitAll
-    public List<Vote> describeVotesOnElections(@PathParam("requestId") String requestId, List<Integer> electionIds) {
-        return voteService.findVotesByElectionIds(electionIds);
+    public List<Vote> describeAllVotes(@PathParam("requestId") String requestId) {
+        return voteService.describeVotes(requestId);
     }
 
     @DELETE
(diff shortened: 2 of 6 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
```diff
@@ -164,8 +164,8 @@ public Response describeDataOwnerVote(@PathParam("requestId") String requestId,
     @GET
     @Produces("application/json")
     @PermitAll
-    public List<Vote> describeAllVotes(@PathParam("requestId") String requestId) {
-        return voteService.describeVotes(requestId);
+    public List<Vote> describeVotesOnElections(@PathParam("electionIds") List<Integer> electionIds) {
+        return voteService.findVotesByElectionIds(electionIds);
```
raejohanek, line 168: Should electionIds be a path param?

src/main/java/org/broadinstitute/consent/http/resources/DataRequestElectionResource.java
```diff
@@ -82,7 +84,8 @@ public Response createDataRequestElection(@Context UriInfo info, Election rec,
             }
             return createExceptionResponse(e);
         }
-        return Response.created(uri).entity(accessElection).build();
+        List<Election> elections = Arrays.asList(accessElection, rpElection);
+        return Response.created(uri).entity(elections).build();
```
rushtong: If we make the logical jump to returning all elections created due to the creation of an access election, then there is also the possibility of a data owner election. Those are created in one of the service methods if the dataset has a data owner and is marked as requiring data owner approval.

rushtong: I disagree with this approach conceptually. An endpoint should be focused on the response based on what it is doing. If we need an API to get ALL elections from a created dar election, that should be a separate endpoint. Please revert this logic, we need to have a larger design discussion before making changes like this.

src/main/resources/assets/api-docs.yaml
```diff
@@ -1868,7 +1868,7 @@ paths:
           description: The election or the DAR couldn't be found.
     post:
       summary: createDataRequestElection
-      description: Creates an Election for the DAR identified by the ID.
+      description: Creates an access and RP Election for the DAR identified by the ID.
```
rushtong: The original language is preferable in my mind. On a technical level, we are creating a DAR election. In some cases, we also create an RP election, but not always. In some cases, we also create a Data Owner election, but not always. Describing that internal complexity here would require a good deal more description than we have.

src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
```diff
@@ -164,8 +164,8 @@ public Response describeDataOwnerVote(@PathParam("requestId") String requestId,
     @GET
     @Produces("application/json")
     @PermitAll
-    public List<Vote> describeAllVotes(@PathParam("requestId") String requestId) {
-        return voteService.describeVotes(requestId);
+    public List<Vote> describeVotesOnElections(@PathParam("requestId") String requestId, List<Integer> electionIds) {
+        return voteService.findVotesByElectionIds(electionIds);
     }
```
rushtong: This should be a new API, not a modification of an existing one. We need to maintain backwards compatibility and support outdated/deprecated APIs for at least one release cycle before removing them.

rushtong (changes requested):
See comments inline. This PR should be limited in scope to the get-votes-for-electionIds part of the functionality.

src/main/java/org/broadinstitute/consent/http/resources/DataRequestElectionResource.java
raejohanek: I think this response is more clearly based on what it is doing than returning only the access election which gives the false impression that was the only election created. But this method produces both, access and rp. Also, this endpoint is only used in one place where it would be beneficial to see the results of the two created elections (and not the data owner election, which may or may not exist). See my 1555 draft PR on the UI. If we want to talk about the design, I would be happy to talk today, also I would like some help trying to get my 1555 PR on the UI to connect and run with the new API, if you have time.


> An endpoint should be focused on the response based on what it is doing.

src/main/java/org/broadinstitute/consent/http/service/VoteService.java
```diff
@@ -53,15 +53,26 @@ public VoteService(UserDAO userDAO, DatasetAssociationDAO dataSetAssociationDAO,
     }
 
     /**
-     * Find all votes for an election id.
+     * Find all votes for election ids.
      *
-     * @param electionIds The election ids for the election.
-     * @return Collection of votes for the given reference id
+     * @param electionIds The election ids for the elections.
+     * @return Collection of votes for the given reference ids
```
raejohanek, line 59: documentation was innacurate

src/main/resources/assets/api-docs.yaml
```diff
@@ -2062,7 +2062,7 @@ paths:
       parameters:
         - name: requestId
           in: path
-          description: referenceId of the Election of the requested Votes
+          description: referenceId of dar the Elections are on
           required: true
```
raejohanek, line 2066: documentation was inaccurate

src/main/resources/assets/api-docs.yaml
```diff
@@ -2071,7 +2071,7 @@ paths:
         - DataRequest
       responses:
         200:
-          description: Returns the DataRequest.
+          description: Returns the list of votes.
           content:
```
raejohanek, line 2075: documentation was inaccurate

src/test/java/org/broadinstitute/consent/http/resources/ElectionResourceTest.java
```diff
@@ -125,6 +125,20 @@ public void testIsDataSetElectionOpenError() {
...
+        Assert.assertEquals(OK, response.getStatus());
+    }
+
+    @Test
+    public void testDescribeVotesOnElectionError() {
+        when(voteService.findVotesByElectionId(randomInt())).thenThrow(new NotFoundException());
+        electionResource = new ElectionResource(voteService, electionService);
+        Response response = electionResource.describeVotesOnElection(randomInt());
+        Assert.assertEquals(NOT_FOUND, response.getStatus());
+    }
```
raejohanek, line 143: currently testDescribeVotesOnElectionError does not work, as it still returns a 200 response status, I'm not really sure why, thoughts?

rushtong: Yes ... the issue is with the random int ...
```
        when(voteService.findVotesByElectionId(randomInt())).thenThrow(new NotFoundException());
        ...
        Response response = electionResource.describeVotesOnElection(randomInt());
```
Note that `randomInt()` returns an integer. It is almost guaranteed that the two times you are using it are returning different integers, so the when condition is never triggered. Try changing the when to
```
        when(voteService.findVotesByElectionId(any())).thenThrow(new NotFoundException());
```

raejohanek: ah makes sense thank you!

src/main/java/org/broadinstitute/consent/http/resources/ElectionResource.java
```diff
@@ -115,4 +116,16 @@ public Response isDataSetElectionOpen(@Context UriInfo info) {
         }
     }
 
+    @GET
+    @Produces("application/json")
+    @PermitAll
+    @Path("/{electionId}/votes")
+    public Response describeVotesOnElection(@PathParam("electionId") Integer electionId) {
```
rushtong: Minor nit: We should add an auth user as a param even though we're not using it in any logic. 
A slightly bigger nit: I think we need to restrict this to admins, chairs, and members right now. This shouldn't be used in any researcher views yet, and when SOs become a thing, we can add them to the list.

src/main/java/org/broadinstitute/consent/http/service/VoteService.java
```diff
@@ -53,15 +53,26 @@ public VoteService(UserDAO userDAO, DatasetAssociationDAO dataSetAssociationDAO,
...
      */
     public List<Vote> findVotesByElectionIds(List<Integer> electionIds) {
         return voteDAO.findVotesByElectionIds(electionIds);
     }
 
+       /**
+     * Find all votes for an election id.
```
rushtong: nit, indentation
```suggestion
    /**
     * Find all votes for an election id.
```

src/main/resources/assets/api-docs.yaml
```diff
@@ -2652,6 +2652,28 @@ paths:
...
+        - Election
+      responses:
+        200:
+          description: List of votes
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Election'
+        400:
+          description: Could not find Election with the specified Id
```
rushtong: 400 is bad request, 404 is not found

rushtong: We should also add 500 for the server error case.

rushtong (changes requested):
I like it, this is clean and concise. See minor comments inline.

src/main/java/org/broadinstitute/consent/http/resources/ElectionResource.java
```diff
@@ -115,4 +118,16 @@ public Response isDataSetElectionOpen(@Context UriInfo info) {
...
+    @GET
+    @Produces("application/json")
+    @RolesAllowed({ADMIN, CHAIRPERSON, MEMBER})
+    @Path("/{electionId}/votes")
+    public Response describeVotesOnElection(@Auth AuthUser authUser, @PathParam("electionId") Integer electionId) {
+        try {
+            return Response.ok().entity(voteService.findVotesByElectionId(electionId)).build();
+        } catch (Exception e) {
+            return createExceptionResponse(e);
+        }
```
rushtong, line 130: When I test this with a non-existent election id, I get back an empty list, not a 404 like we describe in the api docs. There are a couple of ways to solve this problem. There are two cases here, the case where there are no votes for an election (generally an error case, but one that is theoretically possible), and the case where there is no election for the provided id. This should be handled in the service class, so will comment more on that line.

src/main/java/org/broadinstitute/consent/http/service/VoteService.java
```diff
@@ -53,15 +53,26 @@ public VoteService(UserDAO userDAO, DatasetAssociationDAO dataSetAssociationDAO,
...
     }
 
+    /**
+     * Find all votes for an election id.
+     *
+     * @param electionId The election id for the election.
+     * @return Collection of votes on the election specified by the election id
+     */
+    public List<Vote> findVotesByElectionId(Integer electionId) {
+        return voteDAO.findVotesByElectionId(electionId);
```
rushtong, line 76: There are two cases to capture here:
* An existing election: we should do exactly what we're doing here and return the results, whatever they may be, empty list or not.
* No election for the id: we should throw a `NotFoundException`

src/test/java/org/broadinstitute/consent/http/resources/ElectionResourceTest.java
```diff
@@ -125,6 +128,20 @@ public void testIsDataSetElectionOpenError() {
...
+        Assert.assertEquals(OK, response.getStatus());
+    }
+
+    @Test
+    public void testDescribeVotesOnElectionError() {
+        when(voteService.findVotesByElectionId(any())).thenThrow(new NotFoundException());
+        electionResource = new ElectionResource(voteService, electionService);
+        Response response = electionResource.describeVotesOnElection(authUser, any());
+        Assert.assertEquals(NOT_FOUND, response.getStatus());
+    }
```
rushtong, line 143: The reason this test works is because we're telling `voteService.findVotesByElectionId` to throw an exception that it really doesn't throw in the current implementation.

rushtong (changes requested):
Making good progress, see comments inline.

rushtong (approved):
Thank you for the changes, 👍🏽

PR #1061, opened 2021-05-04
DUOS-1245[risk=no] Added exception handler for UnableToExecuteStatementException

JVThomas:
Addresses[DUOS-1245](https://broadworkbench.atlassian.net/browse/DUOS-1245)

PR adds custom exception helper method to process JDBI's ```UnableToExecuteStatementException```, which is a wrapper exception that can cover different types of ```PSQLExceptions```. For now the helper method only checks on Unique constraint violations (vendor code ```23505```), but more codes can be added down the line (Scope of the PR is to properly cover the exception that occurs when a Institution POST carries a name value that already exists in the database).

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/Resource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/Resource.java
@@ -127,22 +127,19 @@ abstract public class Resource {
 
     //Helper method to process generic JBDI Postgres exceptions for responses
     private static Response unableToExecuteExceptionHandler(Exception e) {
-        //default response definition
-        Response response = Response.status(Response.Status.INTERNAL_SERVER_ERROR)
-            .type(MediaType.APPLICATION_JSON)
-            .entity(new Error("Database Error", Response.Status.INTERNAL_SERVER_ERROR.getStatusCode()))
-            .build();
-        
+        //default status definition
+        Integer status = Response.Status.INTERNAL_SERVER_ERROR.getStatusCode();
+       
         if(e.getCause() instanceof PSQLException) {
             //NOTE: implement more Postgres vendor code handlers as we encounter them
             if(((PSQLException)e.getCause()).getSQLState().equals("23505")) {
-                response = Response.status(Response.Status.CONFLICT)
-                    .type(MediaType.APPLICATION_JSON)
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/consent/http/resources/Resource.java
```diff
@@ -42,6 +46,13 @@
     public final static String MEMBER = "Member";
     public final static String RESEARCHER = "Researcher";
 
+    // NOTE: implement more Postgres vendor codes as we encounter them
+    private static Map<String, Integer> vendorCodeStatusMap;
+    static {
+        vendorCodeStatusMap = new HashMap<>();
+        vendorCodeStatusMap.put("23505", Response.Status.CONFLICT.getStatusCode());
```
rushtong: It looks like `23505` is `PSQLState.UNIQUE_VIOLATION` ... it would be cleaner to map that enum value here instead of relying on the string always being what is defined here.
```suggestion
        vendorCodeStatusMap.put(PSQLState.UNIQUE_VIOLATION.getState(), Response.Status.CONFLICT.getStatusCode());
```

src/main/java/org/broadinstitute/consent/http/resources/Resource.java
```diff
@@ -42,6 +46,13 @@
     public final static String MEMBER = "Member";
     public final static String RESEARCHER = "Researcher";
 
+    // NOTE: implement more Postgres vendor codes as we encounter them
+    private static Map<String, Integer> vendorCodeStatusMap;
```
rushtong: can be final

rushtong: Also, can be initialized a little more simply with java 9 stuff ... 
```
    private static final Map<String, Integer> vendorCodeStatusMap = Map.ofEntries(
        new AbstractMap.SimpleEntry<>("23505", Response.Status.CONFLICT.getStatusCode())
    );
```

rushtong (changes requested):
Very nice idea to start mapping sql errors to status codes. See comments inline.

rushtong (approved):
Very nice!

PR #1062, opened 2021-05-04
[DUOS-981][risk=low] - Add Institution ID to Update Profile

rfricke-asymmetrik:
**Addresses**
https://broadworkbench.atlassian.net/browse/DUOS-981
**Changes**
This is the backend portion of the ticket.
- Adds Institution Id support to update user profile
- Adds Institution Id to the describes for Users
- Updates unit tests

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/UserService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/UserService.java
@@ -13,6 +13,7 @@ import javax.ws.rs.BadRequestException;
 import javax.ws.rs.NotFoundException;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.broadinstitute.consent.http.db.InstitutionDAO;
 import org.broadinstitute.consent.http.db.UserPropertyDAO;
 import org.broadinstitute.consent.http.db.UserDAO;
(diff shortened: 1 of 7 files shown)


src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -137,8 +137,12 @@ public User updateDACUserById(Map<String, User> dac, Integer id) throws IllegalA
         if (StringUtils.isEmpty(updatedUser.getDisplayName())) {
             updatedUser.setDisplayName(existingUser.getDisplayName());
         }
+
+        if (Objects.isNull(updatedUser.getInstitutionId())) {
+            updatedUser.setInstitutionId(existingUser.getInstitutionId());
+        }
         try {
-            userDAO.updateUser(updatedUser.getDisplayName(), id, updatedUser.getAdditionalEmail());
+            userDAO.updateUser(updatedUser.getDisplayName(), id, updatedUser.getAdditionalEmail(), updatedUser.getInstitutionId());
```
rushtong, line 153: I think having a sanity check that the institution id is a valid one would be helpful somewhere in this method. We can return quickly with a BadRequestException if someone tries to enter an invalid institution as part of the user object.

raejohanek (approved):
this looks good to me, institution id is included everywhere it needs to be

PR #1064, opened 2021-05-05
DUOS-1246[risk=no]List Institutes should also return create user information

raejohanek:
SCOPE:
add dacuser information to InstitutionDAO call for createuser and updateuser
update institutionmapper to set createuser and updateuser information
add create user and update user fields to institution
renamed the id fields in institution to reflect that they are ids

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1246

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/InstitutionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/InstitutionDAO.java
@@ -6,11 +6,11 @@ import org.broadinstitute.consent.http.models.Institution;
 import org.broadinstitute.consent.http.models.User;
 import org.jdbi.v3.sqlobject.config.RegisterBeanMapper;
 import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
-import org.jdbi.v3.sqlobject.statement.UseRowReducer;
 import org.jdbi.v3.sqlobject.customizer.Bind;
 import org.jdbi.v3.sqlobject.statement.GetGeneratedKeys;
 import org.jdbi.v3.sqlobject.statement.SqlQuery;
 import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+import org.jdbi.v3.sqlobject.statement.UseRowReducer;
 import org.jdbi.v3.sqlobject.transaction.Transactional;
 
 import java.util.Date;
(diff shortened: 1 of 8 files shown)


src/main/java/org/broadinstitute/consent/http/db/InstitutionDAO.java
```diff
@@ -46,6 +46,12 @@ void updateInstitutionById(@Bind("institutionId") Integer institutionId,
...
 
-  @SqlQuery("SELECT * FROM institution")
+  @SqlQuery("SELECT i.*, u.*, u2.dacuserid AS updateUserId, u2.email AS updateUserEmail, " +
+  " u2.displayName AS updateUserName, u2.createdate AS updateUserCreateDate, " +
+  " u2.additional_email as updateUserAdditionalEmail, u2.email_preference as updateUserEmailPreference, " +
+  " u2.status as updateUserStatus, u2.rationale as updateUserRationale " +
+  " FROM institution i" +
+  " LEFT JOIN dacuser u ON u.dacuserid = i.create_user" +
+  " LEFT JOIN dacuser u2 ON u2.dacuserid = i.update_user")
   List<Institution> findAllInstitutions();
```
raejohanek, line 71: aliased the update user's properties so the mapper could understand

src/main/java/org/broadinstitute/consent/http/db/mapper/InstitutionMapper.java
```diff
@@ -36,18 +38,97 @@ public Institution map(ResultSet resultSet, StatementContext statementContext) t
...
+    if (hasColumn(resultSet, "status")) {
+      createUser.setStatus(getCreateStatus(resultSet));
+    }
+    if (hasColumn(resultSet, "rationale")) {
+      createUser.setRationale(resultSet.getString("rationale"));
+    }
+    //user model does not currently have an institutionId field or methods
+    // if (hasColumn(resultSet, "institute")) {
+    //   user.setInstitutionId(resultSet.getInt("update_user"));
+    // }
```
raejohanek: Currently the dacuser table has institution_id as a column, but the model does not have a field for this property. If this is something we want to implement I will file a ticket otherwise I will just delete these commented lines.

src/main/java/org/broadinstitute/consent/http/db/InstitutionDAO.java
rushtong: Using the institution mapper is probably the more complicated approach. When we have joins, we should try to use RowReducers instead since they manage a lot of the conditional logic for you.

rushtong: Look at how we're mapping user roles and users in `UserDAO` for some guidance. One thing we could be doing better in that code is to use column prefixes to disambiguate the joins - see https://jdbi.org/#_registerbeanmapper for docs on that.

src/main/java/org/broadinstitute/consent/http/db/mapper/InstitutionMapper.java
rushtong: prefer to remove the comments until we need to implement it.

src/main/java/org/broadinstitute/consent/http/db/InstitutionDAO.java
raejohanek: Okay, I'm not exactly sure how to do this, but I'll commit my best attempt and maybe we can talk about it tomorrow

rushtong: I'd be happy to help too. Give it a shot and we can work together tomorrow on it.

raejohanek: I based this class off of UserWithRolesReducer, when I test it out on swagger this is the error I get: "Mapping bean class org.broadinstitute.consent.http.models.User didn't find any matching columns in result set"
I also just commented out all the work I did in the Institution mapping class instead of deleting it because I wasn't sure if we would need any of it in implementing the reducer

src/main/java/org/broadinstitute/consent/http/db/mapper/InstitutionWithUsersReducer.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    User update_user = new User();
+    update_user.setDacUserId(rowView.getColumn("u2_dacuserid", Integer.class));
+    update_user.setEmail(rowView.getColumn("u2_email", String.class));
+    update_user.setDisplayName(rowView.getColumn("u2_displayname", String.class));
+    update_user.setCreateDate(rowView.getColumn("u2_createdate", Timestamp.class));
+    update_user.setAdditionalEmail(rowView.getColumn("u2_additional_email", String.class));
+    update_user.setEmailPreference(rowView.getColumn("u2_email_preference", Boolean.class));
+    update_user.setRationale(rowView.getColumn("u2_rationale", String.class));
```
rushtong, line 47: As we discussed, I think this is the only available solution to the case of trying to map the same class twice in a query. Nice work.

rushtong (approved):
👍🏽 Nice work!

PR #1065, opened 2021-05-07
DUOS-1062[risk=no]LC DB, DAO Classes

raejohanek:
SCOPE:
- Create table in changeset 67 and add to changelog
- Create DAO class and methods
- Create LibraryCardMapper and register in the DAO class
- Create LibraryCard Model class 
- Add support for Library Cards to DAO Test Helper
- Create LibraryCardDAOTest class

In all cases:


src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
```diff
@@ -0,0 +1,60 @@
...
+    " update_user_id = :updateUser, " +
+    " update_date = :updateDate " +
+    " WHERE id = :libraryCardId")
+  void updateLibraryCardById(@Bind("libraryCardId") Integer libraryCardId,
+                         @Bind("userId") Integer userId, 
+                         @Bind("institutionId") Integer institutionId,
+                         @Bind("eraCommonsId") String eraCommonsId,
+                         @Bind("userName") String userName,
+                         @Bind("userEmail") String userEmail,
+                         @Bind("updateUser") Integer updateUser,
```
rushtong: minor nit: should name this argument `updateUserId` to be consistent

src/main/java/org/broadinstitute/consent/http/db/mapper/LibraryCardMapper.java
```diff
@@ -0,0 +1,59 @@
...
+import org.broadinstitute.consent.http.models.LibraryCard;
+import org.jdbi.v3.core.mapper.RowMapper;
+import org.jdbi.v3.core.statement.StatementContext;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class LibraryCardMapper implements RowMapper<LibraryCard>, RowMapperHelper {
```
rushtong: This could be a lot simpler if we used `LinkedHashMapRowReducer` and used `@RegisterBeanMapper(value = LibraryCard.class)` in the DAO class. We wouldn't need to do any of the `hasColumn` checks and we could let jdbi figure out all of the field names with `rowView.getRow(LibraryCard.class)`

src/main/java/org/broadinstitute/consent/http/models/LibraryCard.java
```diff
@@ -0,0 +1,140 @@
...
+    private Date createDate;
+    private Integer createUser;
+    private Date updateDate;
+    private Integer updateUser;
+
+  public LibraryCard() {
+    this.createDate = new Date();
+  }
+
+  public LibraryCard(Integer id, Integer userId, Integer institutionId, String eraCommonsId, String name, String email, Date createDate, Integer createUser) {
```
rushtong: This looks unused

src/main/java/org/broadinstitute/consent/http/models/LibraryCard.java
```diff
@@ -0,0 +1,140 @@
...
+    this.userId = userId;
+    this.institutionId = institutionId;
+    this.eraCommonsId = eraCommonsId;
+    this.name = name;
+    this.email = email;
+    this.createDate = createDate;
+    this.createUser = createUser;
+  }
+
+  public LibraryCard(Integer id, Integer userId, Integer institutionId, String eraCommonsId, String name, String email, Date createDate, Integer createUser, Date updateDate, Integer updateUser) {
```
rushtong: This looks unused

src/main/resources/changesets/changelog-consent-67.0.xml
```diff
@@ -0,0 +1,59 @@
...
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="67.0" author="raejohanek">
+        <createTable tableName="library_card">
+            <column name="id" type="bigint" autoIncrement="true">
+                <constraints primaryKey="true" nullable="false"/>
+            </column>
+            <column name="user_id" type="bigint">
+            </column>
```
rushtong: minor nit: here and throughout, self-close tags when they don't have any inner content:
```suggestion
            <column name="user_id" type="bigint"/>
```

src/main/resources/changesets/changelog-consent-67.0.xml
```diff
@@ -0,0 +1,59 @@
...
+        <addForeignKeyConstraint baseColumnNames="update_user_id" baseTableName="library_card" constraintName="fkLibraryCardUpdateUser"
+                                 deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION"
+                                 referencedColumnNames="dacuserid" referencedTableName="dacuser"
+                                 referencesUniqueColumn="false"/>
+        <rollback>
+            <dropTable tableName="library_card"/>
+            <dropForeignKeyConstraint baseTableName="library_card" constraintName="fkLibraryCardUser"/>
+            <dropForeignKeyConstraint baseTableName="library_card" constraintName="fkLibraryCardInstitution"/>
+            <dropForeignKeyConstraint baseTableName="library_card" constraintName="fkLibraryCardCreateUser"/>
+            <dropForeignKeyConstraint baseTableName="library_card" constraintName="fkLibraryCardUpdateUser"/>
```
rushtong: Have not tested this, but I think you can just add `cascadeConstraints="true"` to `dropTable` instead of explicitly adding all of the constraints one at a time.

src/test/java/org/broadinstitute/consent/http/db/LibraryCardDAOTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+  @Test
+  public void testInsertLibraryCardNegative() {
+    Integer userId = createUser().getDacUserId();
+    Integer institutionId = createInstitution().getId();
+    String stringValue = "value";
+    try { 
+        Integer id = libraryCardDAO.insertLibraryCard(0, institutionId, stringValue, stringValue, stringValue, userId, new Date());
+    } catch (Exception e) {
+        assertEquals("23503", ((PSQLException)e.getCause()).getSQLState());
```
rushtong: We should use the constant values instead of string values here if we're expecting this error. I think this one is `PSQLState.FOREIGN_KEY_VIOLATION`. Similar comment in the other assertions.

src/test/java/org/broadinstitute/consent/http/db/LibraryCardDAOTest.java
```diff
@@ -0,0 +1,113 @@
...
+    assertTrue(all.contains(card));
+  }
+
+  @Test
+  public void testInsertLibraryCardNegative() {
+    Integer userId = createUser().getDacUserId();
+    Integer institutionId = createInstitution().getId();
+    String stringValue = "value";
+    try { 
+        Integer id = libraryCardDAO.insertLibraryCard(0, institutionId, stringValue, stringValue, stringValue, userId, new Date());
```
rushtong: Here and below, the id is unused. Just make a call to the insert and let the exception happen.

src/test/java/org/broadinstitute/consent/http/db/LibraryCardDAOTest.java
```diff
@@ -0,0 +1,113 @@
...
+    String newValue = "New Value";
+    try {
+      libraryCardDAO.updateLibraryCardById(0, userId, institutionId, newValue, newValue, newValue, userId, new Date());
+    } catch (Exception e) {
+        assertEquals("23505", ((PSQLException)e.getCause()).getSQLState());
+    }
+  }
+
+  @Test
+  public void testDeleteInstitutionById() {
```
rushtong: nit: We're deleting library cards, this looks like a copy/paste thing.

src/test/java/org/broadinstitute/consent/http/db/LibraryCardDAOTest.java
```diff
@@ -0,0 +1,113 @@
...
+    LibraryCard card = createLibraryCard();
+    Integer id = card.getId();
+    libraryCardDAO.deleteLibraryCardById(id);
+    assertNull(libraryCardDAO.findLibraryCardById(id));
+  }
+
+  @Test
+  public void testDeleteInstitutionByIdNegative() {
```
rushtong: nit: method naming

src/test/java/org/broadinstitute/consent/http/db/LibraryCardDAOTest.java
```diff
@@ -0,0 +1,113 @@
...
+    try{
+      libraryCardDAO.deleteLibraryCardById(RandomUtils.nextInt(1, 1000));
+    } catch (Exception e) {
+      assertEquals("23505", ((PSQLException)e.getCause()).getSQLState());
+    }
+  }
+
+  @Test
+  public void testFindInstitutionById() {
```
rushtong: nit: method naming

src/test/java/org/broadinstitute/consent/http/db/LibraryCardDAOTest.java
```diff
@@ -0,0 +1,113 @@
...
+    LibraryCard cardFromDAO = libraryCardDAO.findLibraryCardById(id);
+    assertEquals(cardFromDAO.getUserId(), card.getUserId());
+    assertEquals(cardFromDAO.getName(), card.getName());
+    assertEquals(cardFromDAO.getEmail(), card.getEmail());
+    assertEquals(cardFromDAO.getCreateUser(), card.getCreateUser());
+    assertEquals(cardFromDAO.getCreateDate(), card.getCreateDate());
+  }
+
+  @Test
+  public void testFindInstitutionByIdNegative() {
```
rushtong: nit: method naming

rushtong (changes requested):
Thank you for this, good work! See comments inline for where I think we can make some improvements.

src/test/java/org/broadinstitute/consent/http/db/LibraryCardDAOTest.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+    try { 
+        libraryCardDAO.insertLibraryCard(userId, institutionId, stringValue, stringValue, stringValue, 0, new Date());
+    } catch (Exception e) {
+        assertEquals(PSQLState.FOREIGN_KEY_VIOLATION.getState(), ((PSQLException)e.getCause()).getSQLState());
+    }
+  }
+
+  @Test
+  public void testUpdateLibraryCardById() {
```
raejohanek, line 50: this test currently fails with the commented out lines uncommented. It is only updating some information. Specifically, this is what the test returns as the updated library card:
{"id":2,"userId":5,"institutionId":4,"eraCommonsId":"New Value","createDate":"May 10, 2021, 3:18:40 PM","updateDate":"May 10, 2021, 3:18:40 PM"}
meaning that name, email, and update user are completely ignored, even when the name is set (on line 55) prior to the update the library card returned does not include the old or new value instead it is null. Ideas as to why this is happening now and wasn't happening before? the only functional changes I made was using the row reducer instead of the row mapper

rushtong: This failing test is telling us something, so it's doing its job. Name is not a field in the database, but `user_name` is. To make this work, the db column and POJO field names need to align. Same thing with `user_email` and `update_user_id` db columns - their names are not aligned with the class field names.

raejohanek: Ah that makes sense thank you! I'll make that change

raejohanek: They all pass now!

rushtong (approved):
Nice work!

PR #1069, opened 2021-05-12
DUOS-1151[risk=no]Enable Admin-only Removal/Delete of a Data Access Committee

raejohanek:
SCOPE:
This PR adds basic dataset objects to a dac in the findall() call in the DacDAO. Now, the findall method searches for all datasets that belong to that dac and returns them in a list on the dac. This change is backwards compatible because it only adds a 'datasets:' field to each dac in the response leaving the rest unchanged. This DAO call is used in several APIs: DacResource.findAll(), MetricsResource.getMetricsData() (DAC and DAR), and DataAccessRequestsResource.describeManageDataAccessRequests() (not V2). All of the places that these APIs are used on the front end are unaffected by these changes. Additionally, by adding datasets to the dac this will enable us to have a delete dac button on the admin manage dac page that does not need to make its own api call to check if there are datasets on that dac that need to be reassigned prior to deletion.

- Added datasets to the findall DAO call
- Added DatasetName to DatasetDTO model so the name can be set directly on the object (just like it is saved directly in the table not on an adjacent table like properties). By having the name on the object itself we no longer need to get properties.
- Added a list of DatasetDTO to the Dac model as a JsonProperty
- Created DacWithDatasetsReducer to compile the list of datasets for each dac, as well as set aliased fields, and set the data use restrictions on each dataset
- Modified test to include an associated dataset

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1151 

In all cases:


rushtong (approved):
Looks good! 👍🏽

PR #1071, opened 2021-05-13
[DUOS-1258][risk=no] Use a test container instead of a locally managed mock server

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1258

## Changes
* Migrated from spinning up a local server to using a test container to do the same work. Test containers have been very stable in CI but local servers have not. 
* Update tests to use the container.
* Minor refactoring for easier maintenance.

----
Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] I've followed [the instructions](https://github.com/DataBiosphere/consent/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [ ] I've updated the RC_XXX release ticket with any manual steps required to release this change
- [ ] I've updated the [FISMA documentation](https://github.com/DataBiosphere/consent/blob/develop/CONTRIBUTING.md#fisma-documentation-changes) if I've made any security-related changes, including auth, encryption, or auditing
- [ ] I've updated Swagger to reflect any API changes

In all cases:

- [ ] Get two thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; you can delete your branch after this **unless it's for a hotfix**. In that case, don't delete it!
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #1072, opened 2021-05-13
DUOS-1229[risk=no]Restore Vote Statistics Download functions

raejohanek:
ADDDRESES: 
https://broadworkbench.atlassian.net/browse/DUOS-1229 

SCOPE:
- ConsentCasesResource.getConsentSummaryDetailFile was deleted in another PR. This PR adds that deleted code back, that includes all of the helper methods, constants, and tests associated.
- Add back getConsentSummaryDetailFile in ConsentCasesResource which calls two summaryService methods, and its tests
- Add back describeConsentSummaryDetail() and describeDataAccessRequestSummaryDetail() in SummaryService, and its helper methods and its tests. This method calls several DAO methods listed below.
- Add back getAssociationsForDataSetIdList in DatasetDAO
- Add back findLastElectionsWithFinalVoteByReferenceIdsTypeAndStatus, verifyOpenElectionsForReferenceId, findAccessRPbyElectionAccessId in ElectionDAO
- Add back findMatchesPurposeId in MatchDAO
- Add back findMaxNumberOfDACMembers in VoteDAO, and its test
- Add ConsentCasesResource.getConsentSummaryDetailFile to api-docs

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
@@ -1,11 +1,13 @@
 package org.broadinstitute.consent.http.db;
 
 import org.apache.commons.lang3.tuple.Pair;
+import org.broadinstitute.consent.http.db.mapper.AccessRPMapper;
 import org.broadinstitute.consent.http.db.mapper.DacMapper;
 import org.broadinstitute.consent.http.db.mapper.DateMapper;
 import org.broadinstitute.consent.http.db.mapper.ImmutablePairOfIntsMapper;
 import org.broadinstitute.consent.http.db.mapper.SimpleElectionMapper;
 import org.broadinstitute.consent.http.db.mapper.ElectionMapper;
+import org.broadinstitute.consent.http.models.AccessRP;
 import org.broadinstitute.consent.http.models.Dac;
 import org.broadinstitute.consent.http.models.Election;
 import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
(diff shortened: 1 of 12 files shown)


rushtong:
FYI, these testing errors are flaky. They should be resolved by https://github.com/DataBiosphere/consent/pull/1071
```
Error:  Errors: 
Error:    UseRestrictionConverterTest.testUseRestrictionConverterConnection:84 » Processing
Error:    UseRestrictionValidatorTest.testValidateUseRestriction_Success » SocketConnection
```

src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
```diff
@@ -195,6 +197,21 @@ void bulkCancelOpenElectionByReferenceIdAndType(@Bind("electionType") String ele
     @UseRowMapper(SimpleElectionMapper.class)
     List<Election> findLastElectionsByReferenceIdsAndType(@BindList("referenceIds") List<String> referenceIds, @Bind("type") String type);
 
+    @SqlQuery("select distinct e.electionId,  e.datasetId, v.vote finalVote, e.status, e.createDate, e.referenceId, v.rationale finalRationale, " +
+            "v.createDate finalVoteDate, e.lastUpdate, e.finalAccessVote, e.electionType, e.dataUseLetter, e.dulName, e.archived, e.version  from election e " +
+            "inner join (select referenceId, MAX(createDate) maxDate from election e where lower(e.status) = lower(:status) group by referenceId) " +
```
JVThomas, line 202: I think this query might have the same issue as the MetricsDAO where it tried to find the last elections for `Access` and `RP` by date. The problem was that for certain referenceIds the `RP` election was made after the `Access` election  (around a 200ms difference), resulting in the query returning only the final `RP` votes.

Just to be safe, try to find an election on your local db where the RP and Access Elections have different timestamps but the same referenceId and see if this query works correctly (you can use the `SQL Query` feature in Postico to test the query out).

JVThomas: Wait, forget what I said, there's a type parameter, so this query is okay

rushtong: This is the original functionality, so even if it's not 100% correct, it will restore what we had prior to removing it. If this doesn't work as expected, I think we need to tackle that as a separate task.

rushtong (approved):
This restores the original functionality as requested 👍🏽

PR #1076, opened 2021-05-17
DUOS-1262/DUOS-1261[risk=no]API: Validate Library Card before updating Vote

raejohanek:
SCOPE:
In resource class, pass additional reference id param to service class.
In service class, retrieve the dar using the reference id, get the user id on dar, retrieve all Library cards for that user, if the list of library cards is empty then throw a not found exception.
In test class, mock the libraryCardDAO.findLibraryCardByUserId call for the submitFinalAccessVote test, and add a test for submitFinalAccessVote when the researcher does not have a library card.

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
@@ -102,8 +102,8 @@ public class DataRequestVoteResource extends Resource {
         String json) {
         try {
             Vote voteRecord = new Gson().fromJson(json, Vote.class);
+            electionService.submitFinalAccessVoteDataRequestElection(voteRecord.getElectionId(), referenceId);
             Vote updatedVote = voteService.updateVoteById(voteRecord, id);
-            electionService.submitFinalAccessVoteDataRequestElection(updatedVote.getElectionId());
             DataAccessRequest dar = dataAccessRequestService.findByReferenceId(referenceId);
             createDataOwnerElection(updatedVote, dar);
             return Response.ok(updatedVote).build();
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
@@ -102,7 +102,7 @@ public class DataRequestVoteResource extends Resource {
         String json) {
         try {
             Vote voteRecord = new Gson().fromJson(json, Vote.class);
-            electionService.submitFinalAccessVoteDataRequestElection(voteRecord.getElectionId(), referenceId);
+            electionService.submitFinalAccessVoteDataRequestElection(voteRecord.getElectionId());
             Vote updatedVote = voteService.updateVoteById(voteRecord, id);
             DataAccessRequest dar = dataAccessRequestService.findByReferenceId(referenceId);
             createDataOwnerElection(updatedVote, dar);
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
```diff
@@ -102,8 +102,8 @@ public Response submitFinalAccessVote(
         String json) {
         try {
             Vote voteRecord = new Gson().fromJson(json, Vote.class);
+            electionService.submitFinalAccessVoteDataRequestElection(voteRecord.getElectionId(), referenceId);
             Vote updatedVote = voteService.updateVoteById(voteRecord, id);
-            electionService.submitFinalAccessVoteDataRequestElection(updatedVote.getElectionId());
```
raejohanek, line 106: I moved this line before updating the vote and just get the election Id from the original vote instead of teh updated one. This way the vote won't be updated in the case the service method throws an error.

src/main/java/org/broadinstitute/consent/http/service/ElectionService.java
```diff
@@ -184,11 +188,20 @@ public Election updateElectionById(Election rec, Integer electionId) {
         return electionDAO.findElectionWithFinalVoteById(electionId);
     }
 
-    public Election submitFinalAccessVoteDataRequestElection(Integer electionId) throws Exception {
+    public Election submitFinalAccessVoteDataRequestElection(Integer electionId, String darId) throws Exception {
         Election election = electionDAO.findElectionWithFinalVoteById(electionId);
```
rushtong: The election has a reference id that is the DAR id, so you don't need to pass that in as an argument.

src/main/java/org/broadinstitute/consent/http/service/ElectionService.java
```diff
@@ -184,11 +188,20 @@ public Election updateElectionById(Election rec, Integer electionId) {
...
         }
+        DataAccessRequest dar = dataAccessRequestService.findByReferenceId(darId);
+        if (dar == null) {
+            throw new NotFoundException("Data Access Request for specified id does not exist");
+        } 
+        Integer userId = dar.getUserId();
+        List<LibraryCard> libraryCards = Objects.nonNull(userId) ? libraryCardDAO.findLibraryCardsByUserId(userId) : new ArrayList<>();
+        if (libraryCards.isEmpty()) {
+            throw new NotFoundException("No Library cards exist for the researcher on this dar");
+        }
```
rushtong: This introduces two new failure conditions that we should describe in `api-docs.yaml`. That only covers a missing election so we should add the cases that the DAR or the Library Card might not be found either.

rushtong (changes requested):
Nice work! See comments inline ... just a couple cleanup tasks.

src/main/java/org/broadinstitute/consent/http/service/ElectionService.java
```diff
@@ -189,6 +193,16 @@ public Election submitFinalAccessVoteDataRequestElection(Integer electionId) thr
         if (election == null) {
             throw new NotFoundException("Election for specified id does not exist");
         }
+        String darId = election.getReferenceId();
+        DataAccessRequest dar = Objects.nonNull(darId) ? dataAccessRequestService.findByReferenceId(darId) : null;
+        if (dar == null) {
+            throw new NotFoundException("Data Access Request for specified id does not exist");
+        } 
+        Integer userId = dar.getUserId();
+        List<LibraryCard> libraryCards = Objects.nonNull(userId) ? libraryCardDAO.findLibraryCardsByUserId(userId) : new ArrayList<>();
```
rushtong: It's more idiomatic in Java to use Collections.emptyList() as opposed to a new ArrayList<>.

rushtong (approved):
Looks good. One minor suggestion inline, but don't let that hold up the PR.

PR #1077, opened 2021-05-18
DUOS-1064[risk=medium] Implemented Library Card Resource

JVThomas:
Addresses [DUOS-1064](https://broadworkbench.atlassian.net/browse/DUOS-1064)

PR adds LC Resource endpoints alongside tests and updates to ```api-docs.yaml```. 

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/LibraryCardResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/LibraryCardResource.java
@@ -4,7 +4,6 @@ import com.google.gson.Gson;
 import com.google.inject.Inject;
 import io.dropwizard.auth.Auth;
 import java.util.List;
-import javax.annotation.security.PermitAll;
 import javax.annotation.security.RolesAllowed;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
@@ -18,7 +17,7 @@ import javax.ws.rs.core.Response;
 import org.broadinstitute.consent.http.models.AuthUser;
 import org.broadinstitute.consent.http.models.LibraryCard;
 import org.broadinstitute.consent.http.models.User;
-// import org.broadinstitute.consent.http.service.LibraryCardService;
+import org.broadinstitute.consent.http.service.LibraryCardService;
 import org.broadinstitute.consent.http.service.UserService;
 
 @Path("api/libraryCards")
@@ -40,8 +39,7 @@ public class LibraryCardResource extends Resource{
   @RolesAllowed(ADMIN)
   public Response getLibraryCards(@Auth AuthUser authUser) {
     try{
-      User user = userService.findUserByEmail(authUser.getName()); //NOTE: may not need this? Check tech doc to make sure
-      List<LibraryCard> libraryCards = libraryCardService.getLibraryCards();
+      List<LibraryCard> libraryCards = libraryCardService.findAllLibraryCards();
       return Response.ok().entity(libraryCards).build();
     } catch(Exception e) {
       return createExceptionResponse(e);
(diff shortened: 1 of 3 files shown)


src/main/resources/assets/api-docs.yaml
```diff
@@ -2973,6 +2973,133 @@ paths:
...
+          description: User not authorized
+        403:
+          description: User does not have permission to access endpoint
+        404:
+          description: Library Card not found
+        500:
+          description: Internal Server Error
+    put:
+      description: Library Card PUT resource
+      tags:
```
raejohanek, line 3052: The update and create methods can also error with the following descriptions: "Invalid Institution Id" "Invalid User Id" "Institution ID is a required parameter" "User ID is a required parameter"

src/main/java/org/broadinstitute/consent/http/resources/LibraryCardResource.java
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  @DELETE
+  @Produces("application/json")
+  @Path("/{id")
+  @RolesAllowed(ADMIN)
+  public Response deleteLibraryCard(@Auth AuthUser authUser, @PathParam("id") Integer id) {
+    try {
+      libraryCardService.deleteLibraryCardById(id);
+      return Response.status(204).build();
```
rushtong, line 114: In almost all of our other DELETE endpoints, we return a 200. The only exception is the institution delete, which I must not have looked at closely the first time around. I needed to go back to my [history books](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE) to sort out the right responses here: 

> If a DELETE method is successfully applied, there are several response status codes possible:
>  
> A 202 (Accepted) status code if the action will likely succeed but has not yet been enacted.
> A 204 (No Content) status code if the action has been enacted and no further information is to be supplied.
> A 200 (OK) status code if the action has been enacted and the response message includes a representation describing the status.

So this is correct. I think we need another tech-debt ticket to address how we're handling `DELETE`'s and align their response codes appropriately.

raejohanek (approved):
Looks good, all tests pass, and documentation is thorough!

src/main/java/org/broadinstitute/consent/http/resources/LibraryCardResource.java
```diff
@@ -0,0 +1,118 @@
...
+  @POST
+  @Consumes("application/json")
+  @Produces("application/json")
+  @RolesAllowed(ADMIN)
+  public Response createLibraryCard(@Auth AuthUser authUser, String libraryCard) {
+    try{
+      User user = userService.findUserByEmail(authUser.getName());
+      LibraryCard payload = new Gson().fromJson(libraryCard, LibraryCard.class);
+      LibraryCard newLibraryCard = libraryCardService.createLibraryCard(payload, user.getDacUserId());
+      return Response.ok().entity(newLibraryCard).build();
```
rushtong: This should be a 201 created

src/main/resources/assets/api-docs.yaml
```diff
@@ -2973,6 +2973,164 @@ paths:
...
+        409:
+          description: Library Card payload conflicts with an existing record
+        500:
+          description: Internal server error
+    delete:
+      summary: Library Card DELETE resource
+      description: Delete target Library Card based on id parameter
+      tags:
+        - Library Card
+        - Admin
```
rushtong, line 3090: This is a nice addition, but begs the question of whether we should add this tag to all of the other admin-endpoints. Also, it could be expanded to Researcher, DAC Chair/Member, Data Custodian, etc. Let's have a team discussion about whether we want to go down this path. This should not hold up the PR.

src/test/java/org/broadinstitute/consent/http/resources/LibraryCardResourceTest.java
```diff
@@ -0,0 +1,218 @@
...
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+import org.postgresql.util.PSQLException;
+import org.postgresql.util.PSQLState;
+
+public class LibraryCardResourceTest {
+  private final AuthUser authUser = new AuthUser("test@test.com");
+  private final List<UserRole> adminRoles = Collections.singletonList(
+    new UserRole(UserRoles.ADMIN.getRoleId(), UserRoles.ADMIN.getRoleName())
+  );
```
rushtong, line 39: We do this so much, it might be nice to extract them into constants in a base test class.

rushtong (changes requested):
I only have one real requested change (the 201 issue), but otherwise, this looks great! Tests are complete, granular, and make sense.

rushtong (approved):
Thank you for the changes 👍🏽

PR #1085, opened 2021-05-25
DUOS-1275[risk=no] Correct nuance behind LibraryCard validation on Votes

JVThomas:
Addresses [DUOS-1275](https://broadworkbench.atlassian.net/browse/DUOS-1275)

PR updates Library Card conditional check on votes to allow Data Access Chair Rejections, RP Chair Approvals, and RP Chair Rejections to be processed without error as those selections do not require a Library Card to be present. PR also includes additional tests on the updated conditional.

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/ElectionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/ElectionService.java
@@ -188,7 +188,7 @@ public class ElectionService {
         return electionDAO.findElectionWithFinalVoteById(electionId);
     }
 
-    public Election submitFinalAccessVoteDataRequestElection(Integer electionId) throws Exception {
+    public Election submitFinalAccessVoteDataRequestElection(Integer electionId, Boolean voteValue) throws Exception {
         Election election = electionDAO.findElectionWithFinalVoteById(electionId);
         if (election == null) {
             throw new NotFoundException("Election for specified id does not exist");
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/consent/http/service/ElectionService.java
```diff
@@ -200,15 +200,20 @@ public Election submitFinalAccessVoteDataRequestElection(Integer electionId) thr
...
-            throw new NotFoundException("No Library cards exist for the researcher on this DAR");
-        }
         List<Vote> finalVotes = voteDAO.findFinalVotesByElectionId(electionId);
         // The first final vote to be submitted is what determines the approval/denial of the election
         boolean isApproved = finalVotes.stream().
                 filter(Objects::nonNull).
                 filter(v -> Objects.nonNull(v.getVote())).
                 anyMatch(Vote::getVote);
+        //Users cannot submit a DataAccess approval if the researcher does not have a library card
+        //However Chairs can still approve/disapprove RP elections and/or reject DataAccess elections even if the researcher does not have a Library Card
```
raejohanek: this, submitFinalAccessVoteDataRequestElection, is always going to be on an access election and not an rp election there is a separate method for those, so I think that check is unnecessary, good catch on the negative vote tho

JVThomas: Ah, good point, I should only be checking on the negative vote here

JVThomas: Wait, I don't think that's true. Both RP and DataAccess votes, when submitted, always go to `api/dataRequest/${requestId}/vote/${vote.voteId}/final`, and that endpoint calls submitFinalAccessVoteDataRequestElection() always. 

I'm outlining the steps for my own reference (and to make sure I'm not off on anything)

Here's the RP submission as chair on `DacVotePanel.js`
Screen Shot 2021-05-25 at 1 47 05 PM

Here's `submitVote`...
Screen Shot 2021-05-25 at 1 49 36 PM

Since both RP and DataAccess have `type = FINAL`, it ends up calling `updateFinalAccessDarVote`
Screen Shot 2021-05-25 at 1 51 19 PM

The above code ends up here in `DataRequestVoteResource.js`
Screen Shot 2021-05-25 at 1 43 53 PM

raejohanek: So, on the UI the RP votes do not have FINAL type and so they will never reach that endpoint.  Here is an example of what the votes on the access review page look like for a chairperson on an open election. There are four votes for a chairperson for the access election, in this case 1442, and two votes for a chairperson for the rp election, 1443.
Screen Shot 2021-05-26 at 7 38 33 AM
Screen Shot 2021-05-26 at 7 38 58 AM

raejohanek (approved):
These changes look great, thanks for catching this and looking into it

src/main/java/org/broadinstitute/consent/http/service/ElectionService.java
```diff
@@ -200,15 +200,19 @@ public Election submitFinalAccessVoteDataRequestElection(Integer electionId) thr
...
         boolean isApproved = finalVotes.stream().
                 filter(Objects::nonNull).
                 filter(v -> Objects.nonNull(v.getVote())).
                 anyMatch(Vote::getVote);
+        //Users cannot submit a DataAccess approval if the researcher does not have a library card
+        //However Chairs can still reject DataAccess elections even if the researcher does not have a Library Card
+        //voteValue, which represents the vote from the payload, is referenced for this validation step
+        if (libraryCards.isEmpty() && (Boolean.TRUE.equals(voteValue))) {
+            throw new NotFoundException("No Library cards exist for the researcher on this DAR");
+        }
```
rushtong, line 215: Nice

rushtong (approved):
👍🏽

PR #1086, opened 2021-05-25
DUOS-1272[risk=low]Dataset Metrics: API

raejohanek:
SCOPE:
- MODEL: Create Dataset Metrics model with a dataset, list of objects representing dars, and list of elections
- API: Create MetricsResource.getDatastetMetrics API
- SERVICE: Create generateDatasetMetrics service method to gather needed information using new and pre-existing DAO calls, and helper method findPI
- DAO: Create findAllDataAccessRequestsForDatasetMetrics(datasetId) which returns only the needed dar data for only the dars that have the given datasetId in their datasetIds list
- DOC: Add new api information to api-docs
- TEST: Add tests for getDatastetMetrics in MetricsResourceTest
- TEST: Add test tests for generateDatasetMetrics MetricsServiceTest
- Modify service class constructor to include electionDAO, userPropertyDAO, and dataAccessRequestDAO since they are all used by generateDatasetMetrics 

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1272

In all cases:


src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
```diff
@@ -160,4 +172,33 @@ public String getHeaderRow(Type type) {
...
+
+    //get datasetDTO with properties and data use restrictions
+    Set<DatasetDTO> datasets = dataSetDAO.findDatasetDTOWithPropertiesByDatasetId(datasetId);
+    if (datasets == null || datasets.isEmpty()) {
+       throw new NotFoundException("Dataset with specified ID does not exist.");
+    }
+
+    //find all dars, then filter by datasetID because filtering cannot occur in DAO call because
+    //all dar data is saved in one column and not distinguishable until handled by the mapper
+    List<DataAccessRequest> dars = dataAccessRequestDAO.findAllDataAccessRequests();
```
JVThomas: I can see the response getting pretty large if a dataset is used in a lot of DARs (the DARs are pretty big as a data struct and the aforementioned scenario becomes more likely as time passes). It might be a good idea to simplify the DAR data to only include properties needed for the metrics page.

There's probably a couple different ways to approach this. Off the top of my head you could make a DAO that returns only some of the DAR properties, or you can just fetchAll as normal and just pick out the data you need. Feel free to come up with a method that best tackles this scenario.

raejohanek: Okay, good idea, I think I'll look at the pre-existing dar DAO calls to see if there's anything that I could use to get the info we need according to the mockups, and if not I'll make a new DAO call

rushtong: You can select/query a [specific field in the json](https://github.com/DataBiosphere/consent/blob/develop/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java#L48) like so: `(data #>> '{}')::jsonb->>'sortDate'`. See also the postgres docs for some additional examples: https://www.postgresql.org/docs/11/functions-json.html

raejohanek: Thanks Greg, I also found this to be very helpful: https://www.compose.com/articles/faster-operations-with-the-jsonb-data-type-in-postgresql/ 
I just made a new DAO query getting only the essential fields, and added an 'if' section to the mapper to account for the varied structure of these dars. Here is an example of what one looks like now! 
"dars": [
    {
      "id": 352,
      "referenceId": "938e4a2b-8435-485c-b8a8-c12236116c9f",
      "data": {
        "investigator": "Gregory Rushton",
        "projectTitle": "Collaborator Testing",
        "nonTechRus": "Collaborator Testing",
        "darCode": "DAR-261"
      },
      "draft": false,
      "userId": 3359,
      "createDate": 1597776900498,
      "sortDate": 1601994565814,
      "submissionDate": 1601994565814,
      "updateDate": 1601994565814
    }
]

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -36,6 +36,26 @@
...
+  /**
+   * Find all non-draft DataAccessRequests for the given datasetId including only the dar data needed for the metrics page
+   *
+   * @return List<DataAccessRequest>
+   */
+  @RegisterRowMapper(DataAccessRequestMapper.class)
+  @SqlQuery(
+      "SELECT id, reference_id, draft, user_id, create_date, sort_date, submission_date, update_date, "
+      + " ((data #>> '{}')::jsonb->>'projectTitle') AS project_title, "
+      + " ((data #>> '{}')::jsonb->>'investigator') AS investigator, "
```
rushtong: I think this should come from the user's profile in researcher properties. The fact that it exists in the dar json is legacy, deprecated behavior.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -36,6 +36,26 @@
...
+      + " ((data #>> '{}')::jsonb->>'nonTechRus') AS nonTechRus, "
+      + " ((data #>> '{}')::jsonb->>'non_tech_rus') AS non_tech_rus, "
+      + " ((data #>> '{}')::jsonb->>'darCode') AS darCode, "
+      + " ((data #>> '{}')::jsonb->>'dar_code') AS dar_code "
```
rushtong: I'm concerned about this approach. In our current reconstruction of the jsonb data, we run it through regexes to ensure it is convertible to a `DataAccessRequestData` object. All of the logic is centralized in that class. I think we should stick with that approach instead of trying to pull these values out individually. Selecting for the camel cased and snake cased names will becomes unmanageable over time.

src/main/java/org/broadinstitute/consent/http/db/mapper/DataAccessRequestMapper.java
```diff
@@ -42,6 +44,27 @@ public DataAccessRequest map(ResultSet resultSet, StatementContext statementCont
...
+            String nonTechRus = resultSet.getString("nonTechRus");
+            if (Objects.isNull(nonTechRus)) {
+                nonTechRus = resultSet.getString("non_tech_rus");
+            }
+            data.setDarCode(darCode);
+            data.setNonTechRus(nonTechRus);
+            data.setProjectTitle(resultSet.getString("project_title"));
+            data.setInvestigator(resultSet.getString("investigator"));
+            dar.setData(data);
         }
```
rushtong: I think we should stick with the original approach here.

src/main/java/org/broadinstitute/consent/http/resources/MetricsResource.java
```diff
@@ -1,16 +1,19 @@
...
 
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
-@Path("/metrics")
+@Path("api/metrics")
```
rushtong: We need to keep this at the `/metrics` level. Adding `api` to this makes it authentication-required which we can't do for public metrics calls.

rushtong: Additionally, changing this here will break all the existing metrics calls.

src/main/resources/assets/api-docs.yaml
```diff
@@ -3585,6 +3585,29 @@ paths:
                 type: string
         500:
           description: Internal Server Error
+  /api/metrics/dataset/{datasetId}:
```
rushtong: As noted above, please make this `/metrics/...`

rushtong (dismissed):
Have not tested this locally, but it is looking good so far. Please see comments inline, we need to make a couple of adjustments to the query and the endpoint path.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
raejohanek: Okay, so then, do we want to retrieve the whole dar object? I was doing this in response to Justin's feedback, so the response wouldn't be large and bulky. If we just want to get the entire dar, then before returning it do we just construct a new object with only the info we need/remove the information we don't need?

src/main/java/org/broadinstitute/consent/http/resources/MetricsResource.java
raejohanek: Okay, so I added this because on the front end, when we call the api in the ajax method we get the api url from the config file, and the api url in the config file has /api at the end, so this does not work with that url. It returns 404 path not found.  How should I get around this issue? in consent, I could remove the class path @Path("/metrics") and update each of the pre-existing apis' path to include /metrics at the front and then update the new datasetmetrics api to include api/metrics at the front so that just that one is accessible and the others don't break, or I could hardcode the api url with no '/api' at the end in the ajax method.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
rushtong: The response is a `DatasetMetrics` object that doesn't necessarily need to have a list of full DARs, it could be a different object altogether with just the info that you want. The issue with this code as it stands now is that it sets up two different ways of parsing the json which introduces room for bugs and inconsistencies. If instead you used the current mapping for DAR info and then grab only the fields you want for a specific metric, it would be more efficient in the response.

src/main/java/org/broadinstitute/consent/http/resources/MetricsResource.java
rushtong: The problem you're describing is a UI problem that we need to solve there, not here. This code should follow the same pattern we use for all metrics calls.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
raejohanek: Okay that makes sense thanks!

src/main/java/org/broadinstitute/consent/http/resources/MetricsResource.java
raejohanek: Okay, the other metric apis do not have ajax methods and are not called directly from the UI. So, I'm not sure what pattern to follow, I could add another api url (without /api at the end) to the config file and retrieve that one for this specific ajax method

rushtong: That's correct. The UI makes some assumptions that aren't true any longer, so we need to update our UI code to handle this new case. The built in assumption being that all ajax calls are prefixed with `api` and that's just not true any longer. For this case, we'll need to do some config refactoring in the UI.

raejohanek: Okay, thanks!

src/main/java/org/broadinstitute/consent/http/models/DatasetMetrics.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class DatasetMetrics {
+
+  private DatasetDTO dataset;
+  private List<DataAccessRequest> dars;
+  private List<Election> elections;
+
+  public DatasetDTO getDataset() {
+    return dataset;
+  }
```
rushtong: One way to lighten the response is to do more work here in the value class than in the display. For example, the UI doesn't really care about the full election or each complete DAR, it only needs some summary info and minimal calculations. Add each element individually, then when you are using the constructor, pass in a list of DARs and a list of Elections and populate the summary fields from those lists. Some suggestions:

* count of dars
* count of approved elections
* count of denied elections
* list of value objects, each containing fields for: 
  * DAR ID
  * Project Title
  * Last Submission (by the Researcher)
  * Principal Investigator
  * Non-Technical Summary

`Principal Investigator` is really one of the DAR's researcher's properties, so I would suggest getting the DAR's researcher in addition to the DAR. The fact that it exists in the DAR's json is actually a problem that we need to not perpetuate.

raejohanek: I considered this approach initially, but remembered out conversation about how storing calculated fields can get out of hand. So, instead I decided to just do those calculations where I need them in the front end. But, I actually just finished the approach you mentioned earlier, I created a much smaller Object and stored a list of those in Datasetmetrics model instead of a list of dars. I see your comment about the investigator now and I will make that change

raejohanek: So, I'm just working on trying to get the PI. I'm looking at the user, that doesn't have PI on it, I'm looking at user properties, and that doesn't have the PI either that just has 'isPI' and 'hasPi' but it doesn't have a 'PI' property so for all the cases that the user is not the PI I don't know who it is. So, where can I find this information besides the dar?

rushtong: Look at the class `UserFields`. That has the names of all possible fields in user properties. Not all users have a PI, some of them **_ARE_** the PI. So, to get that value, we need to look at `isThePI`, and if false, then `piName` and `piEmail`. You can see all of these fields in action on your profile page.

raejohanek: Okay thank you!

raejohanek: @rushtong I'm wondering if either of these are a preferred approach (storing the elections vs storing the count of approved/denied)?  Something I am thinking about, although it is not included in the mockups, I wonder if we would want to know which dars have been approved/denied. We display the total number of dars, list of dar titles, the number approved, the number denied, and the dar non technical summaries. Would it be helpful to whoever is reading those summaries to know whether or not that dar was approved?

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -36,6 +36,20 @@
...
+  /**
+   * Find all non-draft DataAccessRequests for the given datasetId including only the dar data needed for the metrics page
+   *
+   * @return List<DataAccessRequest>
+   */
+  @SqlQuery(
+      "SELECT id, reference_id, draft, user_id, create_date, sort_date, submission_date, update_date, "
+          + "(data #>> '{}')::jsonb AS data FROM data_access_request "
+          + " WHERE not (data #>> '{}')::jsonb ??| array['partial_dar_code', 'partialDarCode'] "
+          +  "AND draft = false"
```
rushtong: I think we can move to querying only on draft status here and in the other non-draft query. We started with the `partial_dar_code` check because we didn't have draft at the time. Draft was added, and we never removed the other check.

src/main/java/org/broadinstitute/consent/http/models/DatasetMetrics.java
```diff
@@ -0,0 +1,37 @@
...
+
+
+import java.util.List;
+
+import org.broadinstitute.consent.http.models.dto.DatasetDTO;
+
+public class DatasetMetrics {
+
+  private DatasetDTO dataset;
+  private List<Object> dars;
```
rushtong: Prefer a defined object over generic `Object`. We should create a real class to handle what this is intended to be. It could be an inner class here for example.

src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
```diff
@@ -160,4 +176,54 @@ public String getHeaderRow(Type type) {
...
+    //find dars with the given datasetId in their list of datasetIds, datasetId is a String so it can be converted to jsonb in query
+    //convert all dars into smaller objects that only contain the information needed
+    List<DataAccessRequest> dars = dataAccessRequestDAO.findAllDataAccessRequestsForDatasetMetrics(Integer.toString(datasetId));
+    List<Object> darInfo = dars.stream().map(dar -> new Object() {
+      public Timestamp updateDate = dar.getUpdateDate();
+      public String projectTitle = dar.data.getProjectTitle();
+      public String darCode = dar.data.getDarCode();
+      public String nonTechRus = dar.data.getNonTechRus();
+      public String investigator = findPI(userPropertyDAO.findResearcherPropertiesByUser(dar.userId), dar.data.getResearcher()); 
+    }).collect(Collectors.toList());
```
rushtong: Similar to previous comment, prefer a real object that captures the idea of what we want. Something like `DarMetricSummary` could work. That object should take a DAR in the constructor so it can populate itself with the same info you're using in this map call.

src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
```diff
@@ -160,4 +176,54 @@ public String getHeaderRow(Type type) {
...
+
+    //find dars with the given datasetId in their list of datasetIds, datasetId is a String so it can be converted to jsonb in query
+    //convert all dars into smaller objects that only contain the information needed
+    List<DataAccessRequest> dars = dataAccessRequestDAO.findAllDataAccessRequestsForDatasetMetrics(Integer.toString(datasetId));
+    List<Object> darInfo = dars.stream().map(dar -> new Object() {
+      public Timestamp updateDate = dar.getUpdateDate();
+      public String projectTitle = dar.data.getProjectTitle();
+      public String darCode = dar.data.getDarCode();
+      public String nonTechRus = dar.data.getNonTechRus();
+      public String investigator = findPI(userPropertyDAO.findResearcherPropertiesByUser(dar.userId), dar.data.getResearcher());
```
rushtong: The value of `dar.data.getResearcher()` should be ignored ... that is a legacy string that we should instead always be getting from `userPropertyDAO.findResearcherPropertiesByUser(dar.userId)`

rushtong: A more stable way to make this call is to pass in only the user id. Then, inside `findPI`, query for the full user object to get the props and make a determination there as to what the right string value should be.

src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
```diff
@@ -160,4 +176,54 @@ public String getHeaderRow(Type type) {
...
+    } else {
+      metrics.setElections(Collections.emptyList());
+    }
+    metrics.setDataset(datasets.iterator().next());
+    metrics.setDars(darInfo);
+    return metrics;
+
+  }
+
+  public String findPI(List<UserProperty> props, String researcher) {
```
rushtong: See comment above. I think this would be more legible if it took a user id, queried for the user+props, and then with that information, figure out the PI name string to return.

rushtong (changes requested):
Looking good! Please see comments inline.

src/main/java/org/broadinstitute/consent/http/models/DatasetMetrics.java
rushtong: Let's touch base with @solideoglori on this question: 
> Would it be helpful to whoever is reading those summaries to know whether or not that dar was approved?

raejohanek: @solideoglori gave his thumb for this, said it would be helpful to know which were approved and denied.

src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
```diff
@@ -160,4 +194,54 @@ public String getHeaderRow(Type type) {
...
+    Set<DatasetDTO> datasets = dataSetDAO.findDatasetDTOWithPropertiesByDatasetId(datasetId);
+    if (datasets == null || datasets.isEmpty()) {
+       throw new NotFoundException("Dataset with specified ID does not exist.");
+    }
+
+    //find dars with the given datasetId in their list of datasetIds, datasetId is a String so it can be converted to jsonb in query
+    //convert all dars into smaller objects that only contain the information needed
+    List<DataAccessRequest> dars = dataAccessRequestDAO.findAllDataAccessRequestsForDatasetMetrics(Integer.toString(datasetId));
+    List<DarMetricsSummary> darInfo = dars.stream().map(dar -> 
+      new DarMetricsSummary(dar.getUpdateDate(), dar.data.getProjectTitle(), dar.data.getDarCode(), dar.data.getNonTechRus(), findPI(dar.userId), dar.getReferenceId()))
```
rushtong: Use getters for fields. This would be a lot cleaner if the constructor took a Dar instead of individual fields. In the constructor, you could add null checks for safety, etc.

src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
```diff
@@ -160,4 +194,54 @@ public String getHeaderRow(Type type) {
...
+    return metrics;
+
+  }
+
+  public String findPI(Integer userId) {
+    List<UserProperty> props = userPropertyDAO.findResearcherPropertiesByUser(userId);
+
+    Optional<UserProperty> isResearcher = props.stream().filter(prop -> prop.getPropertyKey().equals("isThePI") && prop.getPropertyValue().toLowerCase().equals("true")).findFirst();
+    if (isResearcher.isPresent()) {
+      Optional<UserProperty> userName = props.stream().filter(prop -> prop.getPropertyKey().equals("profileName")).findFirst();
```
rushtong: We have a display name right on the user, so you don't need to go through the properties to find that value.

rushtong (changes requested):
Looking better! Please see comments inline.

src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
raejohanek: I don't think I have the user object. I looked for a DAO call that got a user and its properties by its userId but I don't see any like that in either the userDAO or the userPropertyDAO, so instead, I only get the user properties using the userpropertyDAO call

rushtong: No, you don't have the user object, but you do have the user id and can query on it. We could add a dao call to get the user with properties as well. We really should be looking directly at the user since the `profileName` is a brittle value. We need a story to deprecate some of these. Looking at `UserFields.PROFILE_NAME`, we never actually use that to **_set_** a property value, we only use it to **_get_** a property value.

raejohanek: okay makes sense! I will add a new DAO call

rushtong: We do use the `profileName` value in the UI code though - we should clean up these usages to all be using the same field.

raejohanek: Okay, I will file a ticket for that!

rushtong: For this PR, let me rephrase my request that we make the constructor easier to use. With respect to the user profile name, we can address that separately in a different PR.

src/main/java/org/broadinstitute/consent/http/models/UserProperty.java
```diff
@@ -3,8 +3,6 @@
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 
-import java.util.Date;
```
raejohanek, line 6: unused import

src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
```diff
@@ -160,4 +204,56 @@ public String getHeaderRow(Type type) {
...
+      }
+
+      Optional<UserProperty> piName = user.getProperties().stream().filter(prop -> prop.getPropertyKey().equals("piName")).findFirst();
+      if (piName.isPresent()) {
+        return piName.get().getPropertyValue();
+      }
+  }
+    
+    return "- -";
+  }
```
rushtong: nit: indentation looks off here.

rushtong (approved):
Nice work, thank you for the changes! Minor comment inline, feel free to merge if you adjust any whitespace.

PR #1090, opened 2021-05-27
DUOS-1274[risk=no] Refactored searchFilteredElectionList in MetricsService for correct % SRP Approved

JVThomas:
Addresses [DUOS-1274](https://broadworkbench.atlassian.net/browse/DUOS-1274)

PR corrects null check in searchFilteredElectionList and date references in associated MetricsDAO query.

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/MetricsDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/MetricsDAO.java
@@ -27,7 +27,7 @@ public interface MetricsDAO extends Transactional<MetricsDAO> {
 
   @SqlQuery(
     "SELECT * from ( "
-      + "SELECT e.*, v.vote finalvote, v.rationale finalrationale, v.createdate finalvotedate, MAX(e.electionid) "
+      + "SELECT e.*, v.vote finalvote, v.rationale finalrationale, v.updatedate finalvotedate, MAX(e.electionid) "
       + "OVER (PARTITION BY e.referenceid, e.electiontype) AS latest "
       + "FROM election e "
       + "LEFT JOIN vote v ON e.electionid = v.electionid AND "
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/consent/http/db/MetricsDAO.java
```diff
@@ -27,7 +27,7 @@
 
   @SqlQuery(
     "SELECT * from ( "
-      + "SELECT e.*, v.vote finalvote, v.rationale finalrationale, v.createdate finalvotedate, MAX(e.electionid) "
+      + "SELECT e.*, v.vote finalvote, v.rationale finalrationale, v.updatedate finalvotedate, MAX(e.electionid) "
```
rushtong: Looking at `VoteDAO`, the only time we set a create date is on the update method. On dev, there are many more records with null update dates than create dates:
```
consent=> select count(voteid) from vote;
 count 
-------
 16971
(1 row)

consent=> select count(voteid) from vote where createdate is null;
 count 
-------
 14182
(1 row)

consent=> select count(voteid) from vote where updatedate is null;
 count 
-------
 16489
(1 row)
```

In ElectionDAO, we use the createDate for the final vote date exclusively so I don't think this is the correct approach.

rushtong (changes requested):
I'm not confident this will solve the problem - have you run this code against staging? Let's get that set up for you if you haven't.

rushtong:
Re-reviewing, I was incorrectly looking at your LC branch. Looking at staging, this does indeed show more accurate numbers.

Screen Shot 2021-05-27 at 3 20 27 PM

rushtong (approved):
This does show more accurate numbers. It also introduces room for improving the vote population code which is lacking.

raejohanek (approved):
I'm not exactly show how to test this out since it was an issue on staging, but the code looks good and when I check it out locally everything works well!

src/main/java/org/broadinstitute/consent/http/db/MetricsDAO.java
```diff
@@ -27,7 +27,12 @@
 
   @SqlQuery(
     "SELECT * from ( "
-      + "SELECT e.*, v.vote finalvote, v.rationale finalrationale, v.createdate finalvotedate, MAX(e.electionid) "
+      + "SELECT e.*, v.vote finalvote, "
+      + "CASE "
+        + "WHEN v.updatedate IS NULL THEN v.createdate "
+        + "ELSE v.updatedate "
+      + "END as finalvotedate, "
```
rushtong, line 34: Very, very nice solution!

rushtong (approved):
Nice work!

PR #1093, opened 2021-06-01
[DUOS-1282][risk=low] Post LC Fix

rfricke-asymmetrik:
**Addresses**
https://broadworkbench.atlassian.net/browse/DUOS-1282
**Changes**
- Change Library Card resource to look up library card user by email and ignore the userId field from the payload
- Update tests to accommodate

In all cases:

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/LibraryCardResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/LibraryCardResource.java
@@ -85,10 +85,6 @@ public class LibraryCardResource extends Resource{
       LibraryCard payload = new Gson().fromJson(libraryCard, LibraryCard.class);
       User lcUser = userService.findUserByEmail(payload.getUserEmail());
 
-      if (Objects.isNull(lcUser)) {
-        throw new BadRequestException("User email does not belong to a valid user.");
-      }
-
       payload.setUserId(lcUser.getDacUserId());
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/consent/http/resources/LibraryCardResource.java
```diff
@@ -81,6 +83,13 @@ public Response createLibraryCard(@Auth AuthUser authUser, String libraryCard) {
     try{
       User user = userService.findUserByEmail(authUser.getName());
       LibraryCard payload = new Gson().fromJson(libraryCard, LibraryCard.class);
+      User lcUser = userService.findUserByEmail(payload.getUserEmail());
+
+      if (Objects.isNull(lcUser)) {
+        throw new BadRequestException("User email does not belong to a valid user.");
+      }
```
rushtong: The service call will throw a NFE if the user is null, so this check should never happen.

rushtong (changes requested):
Looks good. See comment inline about the exception case that will not be caught.

src/main/java/org/broadinstitute/consent/http/resources/LibraryCardResource.java
rfricke-asymmetrik: Removed the unecessary thrown exception.

rushtong (approved):
Thank you for the changes 👍🏽

PR #1103, opened 2021-06-08
DUOS-1267[risk=no]Backpopulate missing Matches in match_entity table

raejohanek:
SCOPE:
All missing matches on dev, staging, and prod have already been reprocessed. This PR just aims to add the script used for reprocessing to consent.
ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1267

- Verify all tests go green


migration/matchPopulation/reprocessMatches.groovy
```diff
@@ -0,0 +1,54 @@
...
+ *     https://local.broadinstitute.org:27443
+ */
+
+reprocessMatches(args[0], args[1])
+
+static void reprocessMatches(String authToken, String uriHost) {
+    Logger logger = Logger.getLogger("ReprocessMatches")
+
+    //list of purposeIds to reprocess matches for, comma separated list exported from postico
+    def purposeIds = ['fb885dcf-8639-44a7-97c4-0383a4ebd75c', '0caad95b-37c1-47fd-9e47-ac2b96fd70dd', 'd9a17040-cd35-4983-a9a1-fc080a691bb0']
```
rushtong: Can we pull these in as a file, one id per line? [File handling in groovy is pretty simple](https://stackoverflow.com/questions/7729302/how-to-read-a-file-in-groovy-into-a-string).

rushtong (changes requested):
One suggestion inline, but looks good!

migration/matchPopulation/purposeIds
```diff
@@ -0,0 +1,3 @@
+'fb885dcf-8639-44a7-97c4-0383a4ebd75c'
+'0caad95b-37c1-47fd-9e47-ac2b96fd70dd'
+'d9a17040-cd35-4983-a9a1-fc080a691bb0'
```
rushtong: I think maybe my earlier comments were not clear. We don't need to check in any id files - it should be something provided by the user of the script, not part of the code base. In other words, calling the script should look something like this:

```
groovy reprocessMatches.groovy \ 
    `gcloud auth print-access-token` \
    https://local.broadinstitute.org:27443 \
    <file path goes here as an argument>
```

rushtong (changes requested):
One more bit of feedback, thank you!

migration/matchPopulation/purposeIds
rushtong: This file should be removed. The ids are going to be specific to the environment.

rushtong (approved):
Thank you for the changes!

PR #1107, opened 2021-06-09
DUOS-1311[risk=no]Replace references to deprecated user properties with fields on User

raejohanek:
SCOPE:
- Everywhere institution is populated from user properties replace with institution name by getting the institution from the institution id on the user:
-- in ResearcherService.hasUpdatedFields and  ResearcherService.getResearcherPropertiesForDAR. Also, remove field references in  ResearcherServiceTest, DataAccessRequestServiceTest, and WhitelistServiceTest. DataAccessRequestService.createApprovedDARDocument and createDataSetApprovedUsersDocument. Also, remove field references in DataAccessRequestServiceTest.
- Everywhere a users name is populated from user properties replace with the display name on the user:
-- DataAccessRequestService.createApprovedDARDocument and createDataSetApprovedUsersDocument. Also, remove field references in DataAccessRequestServiceTest.

ADDRESSES: 
https://broadworkbench.atlassian.net/browse/DUOS-1311
----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/UserPropertyDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/UserPropertyDAO.java
@@ -44,14 +44,13 @@ public interface UserPropertyDAO extends Transactional<UserPropertyDAO> {
     void deletePropertyByUser(@BindList("propertyKeyList") List<String> propertyKeyList, @Bind("userId") Integer userId);
 
     @SqlQuery(value = "SELECT * FROM user_property WHERE " +
-            "(propertykey = '" + INSTITUTION + "' AND propertyvalue != :institutionName) OR " +
             "(propertykey = '" + ARE_YOU_PRINCIPAL_INVESTIGATOR + "' AND  propertyvalue != :isThePI) OR " +
             "(propertykey = '" + DO_YOU_HAVE_PI + "' AND  propertyvalue != :havePI) OR " +
             "(propertykey = '" + ERA_COMMONS_ID + "' AND  propertyvalue != :eRACommonsID) OR " +
             "(propertykey = '" + PUBMED_ID + "' AND  propertyvalue != :pubmedID) OR " +
             "(propertykey = '" + SCIENTIFIC_URL + "' AND  propertyvalue != :scientificURL) " +
             " AND userid = :userId")
-    List<UserProperty> findResearcherProperties(@Bind("userId") Integer userId, @Bind("institutionName") String institutionName,@Bind("isThePI") String isThePI,
+    List<UserProperty> findResearcherProperties(@Bind("userId") Integer userId, @Bind("isThePI") String isThePI,
                                                       @Bind("havePI") String havePI, @Bind("eRACommonsID") String eRACommonsID, @Bind("pubmedID") String pubmedID, @Bind("scientificURL") String scientificURL);
 
     @SqlQuery("SELECT propertyvalue FROM user_property WHERE userid = :userId and propertykey = :propertyKey")
--- a/src/main/java/org/broadinstitute/consent/http/service/DataAccessReportsParser.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DataAccessReportsParser.java
@@ -8,6 +8,7 @@ import org.broadinstitute.consent.http.util.DarConstants;
 import org.broadinstitute.consent.http.util.DarUtil;
 import org.bson.Document;
 import java.io.*;
+import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Date;
@@ -61,7 +62,7 @@ public class DataAccessReportsParser {
     public void addApprovedDARLine(FileWriter darWriter, Election election, Document dar, String profileName, String institution, String consentName, String translatedUseRestriction) throws IOException {        String rusSummary = StringUtils.isNotEmpty( dar.getString(DarConstants.NON_TECH_RUS)) ?  dar.getString(DarConstants.NON_TECH_RUS).replace("\n", " ") : "";
         String content1 =  profileName + DEFAULT_SEPARATOR + institution + DEFAULT_SEPARATOR;
         String content2 = rusSummary + DEFAULT_SEPARATOR +
-                formatTimeToDate(new Date(dar.getLong(DarConstants.SORT_DATE)).getTime()) + DEFAULT_SEPARATOR +
+                formatTimeToDate(((Timestamp) dar.get(DarConstants.SORT_DATE)).getTime()) + DEFAULT_SEPARATOR +
                 formatTimeToDate(election.getFinalVoteDate().getTime()) + DEFAULT_SEPARATOR +
                 "--";
         addDARLine(darWriter, dar, content1, content2, consentName, translatedUseRestriction);
@@ -97,13 +98,17 @@ public class DataAccessReportsParser {
     private void addDARLine(FileWriter darWriter, Document dar, String customContent1, String customContent2, String consentName, String translatedUseRestriction) throws IOException {
         List<Document> dataSetDetails = dar.get(DarConstants.DATASET_DETAIL, ArrayList.class);
         List<String> datasetNames = new ArrayList<>();
-        for(Document detail : dataSetDetails) {
-            datasetNames.add(StringUtils.defaultString(detail.getString("name")));
+        if (dataSetDetails != null) {
+            for (Document detail : dataSetDetails) {
+                datasetNames.add(StringUtils.defaultString(detail.getString("name")));
+            }
         }
         List<Integer> dataSetIds = DarUtil.getIntegerList(dar, DarConstants.DATASET_ID);
         List<String> dataSetUUIds = new ArrayList<>();
-        for(Integer id : dataSetIds) {
-            dataSetUUIds.add(DataSet.parseAliasToIdentifier(id));
+        if (dataSetIds != null) {
+            for (Integer id : dataSetIds) {
+                dataSetUUIds.add(DataSet.parseAliasToIdentifier(id));
+            }
         }
         String sDUL = StringUtils.isNotEmpty(translatedUseRestriction) ?  translatedUseRestriction.replace("\n", " ") : "";
         String translatedRestriction = StringUtils.isNotEmpty(dar.getString(DarConstants.TRANSLATED_RESTRICTION)) ? dar.getString(DarConstants.TRANSLATED_RESTRICTION).replace("<br>", " ") :  "";
--- a/src/main/java/org/broadinstitute/consent/http/service/DataAccessReportsParser.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DataAccessReportsParser.java
@@ -7,12 +7,14 @@ import org.broadinstitute.consent.http.models.Election;
 import org.broadinstitute.consent.http.util.DarConstants;
 import org.broadinstitute.consent.http.util.DarUtil;
 import org.bson.Document;
-import java.io.*;
+import java.io.FileWriter;
+import java.io.IOException;
 import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.List;
+import java.util.Objects;
 
 public class DataAccessReportsParser {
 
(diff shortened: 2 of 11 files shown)


src/main/java/org/broadinstitute/consent/http/db/UserPropertyDAO.java
```diff
@@ -44,14 +44,13 @@
...
 
     @SqlQuery(value = "SELECT * FROM user_property WHERE " +
-            "(propertykey = '" + INSTITUTION + "' AND propertyvalue != :institutionName) OR " +
             "(propertykey = '" + ARE_YOU_PRINCIPAL_INVESTIGATOR + "' AND  propertyvalue != :isThePI) OR " +
             "(propertykey = '" + DO_YOU_HAVE_PI + "' AND  propertyvalue != :havePI) OR " +
             "(propertykey = '" + ERA_COMMONS_ID + "' AND  propertyvalue != :eRACommonsID) OR " +
             "(propertykey = '" + PUBMED_ID + "' AND  propertyvalue != :pubmedID) OR " +
             "(propertykey = '" + SCIENTIFIC_URL + "' AND  propertyvalue != :scientificURL) " +
             " AND userid = :userId")
-    List<UserProperty> findResearcherProperties(@Bind("userId") Integer userId, @Bind("institutionName") String institutionName,@Bind("isThePI") String isThePI,
```
raejohanek, line 54: only used in researcherservice.hasupdatedfields , and field is not needed for that method

src/main/java/org/broadinstitute/consent/http/service/ResearcherService.java
```diff
@@ -185,7 +190,6 @@ private void validateRequiredFields(Map<String, String> properties) {
     private Boolean hasUpdatedFields(Integer userId, Map<String, String> researcherPropertiesMap, Boolean isUpdatedProfileCompleted) {
         Boolean hasUpdatedFields = false;
         if (isUpdatedProfileCompleted) {
-            String institutionName = researcherPropertiesMap.getOrDefault(UserFields.INSTITUTION.getValue(), "");
```
raejohanek, line 188: remove instead of replace because we are no longer updating this field

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -610,13 +612,14 @@ public File createApprovedDARDocument() throws IOException {
...
                 try {
-                    if (dar != null) {
+                    if (dar != null && user != null) {
                         Integer datasetId = dataAccessRequest.getData().getDatasetIds().get(0);
                         String consentId = dataSetDAO.getAssociatedConsentIdByDataSetId(datasetId);
                         Consent consent = consentDAO.findConsentById(consentId);
-                        String profileName = userPropertyDAO.findPropertyValueByPK(dataAccessRequest.getUserId(), DarConstants.PROFILE_NAME);
-                        String institution = userPropertyDAO.findPropertyValueByPK(dataAccessRequest.getUserId(), DarConstants.INSTITUTION);
+                        String profileName = user.getDisplayName();
+                        String institution = institutionDAO.findInstitutionById(user.getInstitutionId()).getName();
```
rushtong: This could use a null check on the institution. If we don't have one, log a warning and return the empty string.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -661,14 +664,13 @@ public File createDataSetApprovedUsersDocument(Integer dataSetId) throws IOExcep
...
                 Date approvalDate = electionDAO.findApprovalAccessElectionDate(referenceId);
                 if (approvalDate != null) {
                     String email = userPropertyDAO
                             .findPropertyValueByPK(dataAccessRequest.getUserId(), DarConstants.ACADEMIC_BUSINESS_EMAIL);
-                    String name = userPropertyDAO
-                            .findPropertyValueByPK(dataAccessRequest.getUserId(), DarConstants.PROFILE_NAME);
-                    String institution = userPropertyDAO
-                            .findPropertyValueByPK(dataAccessRequest.getUserId(), DarConstants.INSTITUTION);
+                    String name = (user == null) ? "" : user.getDisplayName();
+                    String institution = (user == null || user.getInstitutionId() == null) ? "" : institutionDAO.findInstitutionById(user.getInstitutionId()).getName();
```
rushtong: Nice. 

This method is a great area for refactoring, as an FYI. Anywhere we see `Document`, we need to replace that with `DataAccessRequest` which is what a document really is. `Document` is how the DARs were stored when they were mongo objects and are just generic object holders. Moving to `DataAccessRequest` gives us a lot more control over the fields. 

If you dig into this method a little more, you'll see that we do this work horribly inefficiently:

1. get ALL DARs (see describeDataAccessByDataSetId)
2. convert DARs to documents (see describeDataAccessByDataSetId)
3. filter documents on those having dataset ids with the list (see describeDataAccessByDataSetId)
4. re-query for the DAR (see line 666)

src/main/java/org/broadinstitute/consent/http/service/ResearcherService.java
```diff
@@ -101,8 +105,9 @@ public void deleteResearcherSpecificProperties(List<UserProperty> properties) {
 
     private Map<String, String> getResearcherPropertiesForDAR(Map<String, String> properties, Integer userId) {
         Map<String, String> rpForDAR = new HashMap<>();
-        rpForDAR.put(UserFields.INVESTIGATOR.getValue(), properties.getOrDefault(UserFields.PI_NAME.getValue(), userDAO.findUserById(userId).getDisplayName()));
-        rpForDAR.put(UserFields.INSTITUTION.getValue(), properties.getOrDefault(UserFields.INSTITUTION.getValue(), null));
+        User user =  userDAO.findUserById(userId);
+        rpForDAR.put(UserFields.INVESTIGATOR.getValue(), properties.getOrDefault(UserFields.PI_NAME.getValue(), user.getDisplayName()));
+        rpForDAR.put(DarConstants.INSTITUTION, institutionDAO.findInstitutionById(user.getInstitutionId()).getName());
```
rushtong: Null check would be good here.

src/main/java/org/broadinstitute/consent/http/service/ResearcherService.java
```diff
@@ -115,7 +120,7 @@ public void deleteResearcherSpecificProperties(List<UserProperty> properties) {
         rpForDAR.put(UserFields.DIVISION.getValue(), properties.getOrDefault(UserFields.DIVISION.getValue(), null));
         rpForDAR.put(UserFields.ERA_COMMONS_ID.getValue(), properties.getOrDefault(UserFields.ERA_COMMONS_ID.getValue(), null));
         rpForDAR.put(UserFields.PUBMED_ID.getValue(), properties.getOrDefault(UserFields.PUBMED_ID.getValue(), null));
-        rpForDAR.put(UserFields.PROFILE_NAME.getValue(), properties.getOrDefault(UserFields.PROFILE_NAME.getValue(), null));
+        rpForDAR.put(DarConstants.PROFILE_NAME, user.getDisplayName());
```
rushtong: It is common to find duplicated code like `PROFILE_NAME`. In this case, I think it would be preferable to use `UserFields` instead of `DarConstants`. `DarConstants` should eventually be deprecated altogether and `UserFields` seems like a more logical home for how we define the user's profile name. I think in an ideal world, we'd have a json schema to define these keys.

rushtong (changes requested):
Caveat, I haven't tested this locally, but the code looks good. Please see comments inline.

rushtong:
We need to do some more testing on this before it's ready for review. One of the code paths this affects is `/api/dataRequest/approved` which is currently undocumented. Add this to the swagger doc to get that endpoint, and one right above it, to be visible in swagger:
```
  /api/dataRequest/approved:
    get:
      summary: Get a list of approved DARs
      description: Get a list of approved DARs
      tags:
        - Data Access Request
      responses:
        200:
          description: A list of approved DARs
          content:
            application/octet-stream:
              schema:
                type: string
                format: binary
  /api/dataRequest/reviewed:
    get:
      summary: Get a list of reviewed DARs
      description: Get a list of reviewed DARs
      tags:
        - Data Access Request
      responses:
        200:
          description: A list of reviewed DARs
          content:
            application/octet-stream:
              schema:
                type: string
                format: binary
```

Next, hitting that endpoint is triggering a good number of errors like this:
```
consent  | WARN  [2021-06-11 12:57:40,728] org.broadinstitute.consent.http.service.DataAccessRequestService: No institution found for creator of this Data Access Request
consent  | ERROR [2021-06-11 12:57:40,729] org.broadinstitute.consent.http.service.DataAccessRequestService: Exception generating Approved DAR Document
consent  | ! java.lang.ClassCastException: class java.sql.Timestamp cannot be cast to class java.lang.Long (java.sql.Timestamp is in module java.sql of loader 'platform'; java.lang.Long is in module java.base of loader 'bootstrap')
consent  | ! at org.bson.Document.getLong(Document.java:266)
consent  | ! at org.broadinstitute.consent.http.service.DataAccessReportsParser.addApprovedDARLine(DataAccessReportsParser.java:64)
consent  | ! at org.broadinstitute.consent.http.service.DataAccessRequestService.createApprovedDARDocument(DataAccessRequestService.java:626)
consent  | ! at org.broadinstitute.consent.http.resources.DataRequestReportsResource.downloadApprovedDARs(DataRequestReportsResource.java:41)
``` The error shown here is fixed by removing Document.getLong and instead using Document.get, to get the Timestamp and then use .getTime() to get the long, I also added null checks for other errors coming from the parser

**Review by rushtong — CHANGES_REQUESTED** [review]
See comments on the PR, we need some more testing for all affected endpoints.

**Comment by raejohanek** [comment]
> 
> Next, hitting that endpoint is triggering a good number of errors like this:
> 
> ```
> consent  | WARN  [2021-06-11 12:57:40,728] org.broadinstitute.consent.http.service.DataAccessRequestService: No institution found for creator of this Data Access Request
> consent  | ERROR [2021-06-11 12:57:40,729] org.broadinstitute.consent.http.service.DataAccessRequestService: Exception generating Approved DAR Document
> consent  | ! java.lang.ClassCastException: class java.sql.Timestamp cannot be cast to class java.lang.Long (java.sql.Timestamp is in module java.sql of loader 'platform'; java.lang.Long is in module java.base of loader 'bootstrap')
> consent  | ! at org.bson.Document.getLong(Document.java:266)
> consent  | ! at org.broadinstitute.consent.http.service.DataAccessReportsParser.addApprovedDARLine(DataAccessReportsParser.java:64)
> consent  | ! at org.broadinstitute.consent.http.service.DataAccessRequestService.createApprovedDARDocument(DataAccessRequestService.java:626)
> consent  | ! at org.broadinstitute.consent.http.resources.DataRequestReportsResource.downloadApprovedDARs(DataRequestReportsResource.java:41)

> ``` 
The error shown here is fixed by removing Document.getLong and instead using Document.get, to get the Timestamp and then use .getTime() to get the long, I also added null checks for other errors coming from the parser

**Review by JVThomas — APPROVED** [review]
Refactor and additional null checks look good and the two endpoints added to swagger are working locally for me.

File: src/main/java/org/broadinstitute/consent/http/service/DataAccessReportsParser.java [hunk 7]
```diff
@@ -8,6 +8,7 @@
 import org.broadinstitute.consent.http.util.DarUtil;
 import org.bson.Document;
 import java.io.*;
```
rushtong: Not your code, but we should avoid star imports. If you set up your IDE correctly, then when you optimize imports, these will go away.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -610,13 +612,17 @@ public File createApprovedDARDocument() throws IOException {
...
                         Integer datasetId = dataAccessRequest.getData().getDatasetIds().get(0);
                         String consentId = dataSetDAO.getAssociatedConsentIdByDataSetId(datasetId);
                         Consent consent = consentDAO.findConsentById(consentId);
-                        String profileName = userPropertyDAO.findPropertyValueByPK(dataAccessRequest.getUserId(), DarConstants.PROFILE_NAME);
-                        String institution = userPropertyDAO.findPropertyValueByPK(dataAccessRequest.getUserId(), DarConstants.INSTITUTION);
+                        String profileName = user.getDisplayName();
+                        if (user.getInstitutionId() == null) {
+                          logger.warn("No institution found for creator of this Data Access Request");
+                        }
+                        String institution = (user.getInstitutionId() == null) ? "" : institutionDAO.findInstitutionById(user.getInstitutionId()).getName();
```
rushtong: Prefer `Objects.isNull()` over direct null checks.
```suggestion
                        String institution = Objects.isNull(user.getInstitutionId()) ? "" : institutionDAO.findInstitutionById(user.getInstitutionId()).getName();
```

src/main/java/org/broadinstitute/consent/http/service/ResearcherService.java
```diff
@@ -101,8 +105,9 @@ public void deleteResearcherSpecificProperties(List<UserProperty> properties) {
 
     private Map<String, String> getResearcherPropertiesForDAR(Map<String, String> properties, Integer userId) {
         Map<String, String> rpForDAR = new HashMap<>();
-        rpForDAR.put(UserFields.INVESTIGATOR.getValue(), properties.getOrDefault(UserFields.PI_NAME.getValue(), userDAO.findUserById(userId).getDisplayName()));
-        rpForDAR.put(UserFields.INSTITUTION.getValue(), properties.getOrDefault(UserFields.INSTITUTION.getValue(), null));
+        User user =  userDAO.findUserById(userId);
+        rpForDAR.put(UserFields.INVESTIGATOR.getValue(), properties.getOrDefault(UserFields.PI_NAME.getValue(), (user == null) ? "" : user.getDisplayName()));
+        rpForDAR.put(UserFields.INSTITUTION.getValue(), (user == null || user.getInstitutionId() == null) ? "" : institutionDAO.findInstitutionById(user.getInstitutionId()).getName());
```
rushtong: Here and in other places, prefer null checks with `Objects.isNull()`

rushtong (approved):
Code looks and functions well, I tested all of the endpoints affected. Please see the minor feedback inline around null checks, but don't let that hold up the PR, feel free to address if you wish.

PR #1109, opened 2021-06-10
[DUOS-1298][risk=no] Vote updates

rushtong:
## Addresses
Partial fixes for https://broadworkbench.atlassian.net/browse/DUOS-1298
Will need to manually fix the identified error cases via the admin API once this is merged.

## Changes
* Added logging when submitting a final access vote so we can better follow up on future errors
* Update the vote API so we can manually fix any errors that do happen
* Fix api docs so we can use swagger for vote updates

----

- Verify all tests go green


raejohanek (approved):
update vote looks good, and logging looks good!

rushtong:
There is another update I'd like to make to the voting endpoints to restrict on auth user but those are not related to the primary ticket.

PR #1113, opened 2021-06-11
DUOS-1310[risk=no] Fix DAC Pilot data not loading on metrics

JVThomas:
Addresses [DUOS-1310](https://broadworkbench.atlassian.net/browse/DUOS-1310)

PR updates DAC related queries and election filter function to 
- Correctly populate DAC Pilot data
- Correctly calculate `# of DARs Received` which in turn corrects `% of DARs Reviewed`

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
```diff
@@ -341,13 +341,21 @@ void updateElectionById(@Bind("electionId") Integer electionId,
...
         + "INNER JOIN election consentElection ON a.consent_election_id = consentElection.electionid "
         + "INNER JOIN consents c ON consentElection.referenceId = c.consentid "
         + "INNER JOIN dac d on d.dac_id = c.dac_id "
         + "WHERE e.electionId IN (<electionIds>) "
+        + "UNION "
+        + "SELECT d.*, e.electionid " + "FROM dac d " 
+        + "INNER JOIN consents "
+        + "ON d.dac_id = consents.dac_id " + "INNER JOIN consentassociations ca "
+        + "ON ca.consentid = consents.consentid " + "INNER JOIN dataset data "
+        + "ON data.datasetid = ca.datasetid " + "INNER JOIN election e "
```
rushtong: minor spacing nit: Either remove the string concatenation when on the same line, or break them into separate lines if you want them like this.

src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
```diff
@@ -217,6 +217,10 @@ public String getHeaderRow(Type type) {
       .stream()
       .filter(e -> Objects.nonNull(e.getFinalVote()))
       .findFirst();
+
+    if(election.isEmpty()) {
+      election = electionList.stream().findFirst();
+    }
```
rushtong: This took me a minute to sort through in my head. I think that what we're looking for in `electionList` is either the first one with a final vote or any other election at all, regardless of final vote status. I think we can clean up the lambda to do both things in one pass. If we first sort on final vote (treating nulls as lower in sort order) instead of filtering, then `findFirst` works in both of our cases. I _think_ this might work (have not tested, and there are further optimizations to be done here):

```
    Optional<Election> election = electionList
      .stream()
      .sorted(Comparator.nullsLast(Comparator.comparing(Election::getFinalVote)))
      .findFirst();
```

rushtong (approved):
One minor suggestion to try if you want, but not required for approval 👍🏽

src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
JVThomas: I changed the `sorted` call a bit since NullPointerExceptions can still be thrown within `Comparator.comparing`

```
Optional<Election> election = electionList
      .stream()
      .sorted(Comparator.comparing(Election::getFinalVote, Comparator.nullsLast(Comparator.naturalOrder())))
      .findFirst();
```

I removed the outer `nullsLast` call since the initial DAO call and the subsequent filters ensure that the electionList parameter will not have null elements or be null itself (it'll just be an empty list)

raejohanek (approved):
updated query looks good!

rushtong (approved):
Nice!

PR #1114, opened 2021-06-11
DUOS-1070[risk=no]eRA Commons Auth Hook

raejohanek:
SCOPE:
- get the eraCommonsId from the user properties
- update all library cards that belong to that user with their eraCommonsId
ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1070

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/NihAccountResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/NihAccountResource.java
@@ -4,9 +4,12 @@ import com.google.inject.Inject;
 import io.dropwizard.auth.Auth;
 import org.broadinstitute.consent.http.models.AuthUser;
 import org.broadinstitute.consent.http.models.User;
+import org.broadinstitute.consent.http.models.UserProperty;
 import org.broadinstitute.consent.http.models.NIHUserAccount;
 import org.broadinstitute.consent.http.service.NihService;
 import org.broadinstitute.consent.http.service.UserService;
+
+import java.util.List;
 
 import javax.annotation.security.RolesAllowed;
 import javax.ws.rs.DELETE;
@@ -32,8 +35,9 @@ public class NihAccountResource extends Resource {
     @RolesAllowed(RESEARCHER)
     public Response registerResearcher(NIHUserAccount nihAccount, @Auth AuthUser user) {
         try {
-            userService.findUserByEmail(user.getName());
-            return Response.ok(nihService.authenticateNih(nihAccount, user)).build();
+            User u = userService.findUserByEmail(user.getName());
+            List<UserProperty> authUserProps = nihService.authenticateNih(nihAccount, user, u.getDacUserId());
+            return Response.ok(authUserProps).build();
         } catch (Exception e){
             return createExceptionResponse(e);
         }
--- a/src/main/java/org/broadinstitute/consent/http/resources/NihAccountResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/NihAccountResource.java
@@ -21,8 +21,8 @@ import javax.ws.rs.core.Response;
 @Path("api/nih")
 public class NihAccountResource extends Resource {
 
-    private NihService nihService;
-    private UserService userService;
+    private final NihService nihService;
+    private final UserService userService;
 
     @Inject
     public NihAccountResource(NihService nihService, UserService userService) {
@@ -33,10 +33,10 @@ public class NihAccountResource extends Resource {
     @POST
     @Produces("application/json")
     @RolesAllowed(RESEARCHER)
-    public Response registerResearcher(NIHUserAccount nihAccount, @Auth AuthUser user) {
+    public Response registerResearcher(NIHUserAccount nihAccount, @Auth AuthUser authUser) {
         try {
-            User u = userService.findUserByEmail(user.getName());
-            List<UserProperty> authUserProps = nihService.authenticateNih(nihAccount, user, u.getDacUserId());
+            User user = userService.findUserByEmail(authUser.getName());
+            List<UserProperty> authUserProps = nihService.authenticateNih(nihAccount, authUser, user.getDacUserId());
             return Response.ok(authUserProps).build();
         } catch (Exception e){
             return createExceptionResponse(e);
--- a/src/main/java/org/broadinstitute/consent/http/service/NihService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/NihService.java
@@ -2,6 +2,7 @@ package org.broadinstitute.consent.http.service;
 
 import com.google.inject.Inject;
 import org.apache.commons.lang3.StringUtils;
+import org.broadinstitute.consent.http.db.LibraryCardDAO;
 import org.broadinstitute.consent.http.enumeration.UserFields;
 import org.broadinstitute.consent.http.models.AuthUser;
 import org.broadinstitute.consent.http.models.NIHUserAccount;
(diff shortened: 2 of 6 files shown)


JVThomas (changes requested):
There's a compilation error when running `mvn clean compile` because the constructor arguments in ConsentModule haven't been updated.

```
Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project consent: Compilation failure
[ERROR] /Users/jthomas/consent/src/main/java/org/broadinstitute/consent/http/ConsentModule.java:[497,16] constructor NihService in class org.broadinstitute.consent.http.service.NihService cannot be applied to given types;
[ERROR]   required: org.broadinstitute.consent.http.service.ResearcherService,org.broadinstitute.consent.http.db.LibraryCardDAO
[ERROR]   found:    org.broadinstitute.consent.http.service.ResearcherService
[ERROR]   reason: actual and formal argument lists differ in length

**Comment by raejohanek** [comment]
> There's a compilation error when running `mvn clean compile` because the constructor arguments in ConsentModule haven't been updated.

good catch, thanks!

File: src/main/java/org/broadinstitute/consent/http/service/NihService.java [hunk 1]
```diff
@@ -12,23 +13,31 @@
...
-    public List<UserProperty> authenticateNih(NIHUserAccount nihAccount, AuthUser user) throws BadRequestException {
+    public List<UserProperty> authenticateNih(NIHUserAccount nihAccount, AuthUser authUser, Integer userId) throws BadRequestException {
         if (StringUtils.isNotEmpty(nihAccount.getNihUsername()) && !nihAccount.getNihUsername().isEmpty()) {
             nihAccount.setEraExpiration(generateEraExpirationDates());
             nihAccount.setStatus(true);
-            return researcherService.updateProperties(nihAccount.getNihMap(), user, false);
+            Map<String, String> userProp = researcherService.describeResearcherPropertiesMap(userId);
+            String eraCommonsId = (userProp == null) ? "" : userProp.getOrDefault(UserFields.ERA_COMMONS_ID.getValue(), "");
+            if (!eraCommonsId.equals("")) {
+              libraryCardDAO.updateEraCommonsForUser(userId, eraCommonsId);
```
JVThomas, line 38: Quick question: `eraCommonsId` is the value saved in the UserProperties db right? But since the ResearcherProperties aren't being updated until the end of the function(line 38), is the libraryCard update saving what is essentially the "old" id value instead of the updated value on `nihAccount`?

raejohanek: yeah you are right I just moved the update so it happens first and then the updated era commons id is used to update the library cards

src/main/java/org/broadinstitute/consent/http/service/NihService.java
```diff
@@ -12,23 +13,33 @@
...
     }
 
-    public List<UserProperty> authenticateNih(NIHUserAccount nihAccount, AuthUser user) throws BadRequestException {
+    public List<UserProperty> authenticateNih(NIHUserAccount nihAccount, AuthUser authUser, Integer userId) throws BadRequestException {
         if (StringUtils.isNotEmpty(nihAccount.getNihUsername()) && !nihAccount.getNihUsername().isEmpty()) {
             nihAccount.setEraExpiration(generateEraExpirationDates());
             nihAccount.setStatus(true);
-            return researcherService.updateProperties(nihAccount.getNihMap(), user, false);
+            List<UserProperty> updatedProps = researcherService.updateProperties(nihAccount.getNihMap(), authUser, false);
+            Optional<UserProperty> eraCommonsProp = updatedProps.stream().filter((prop) -> prop.getPropertyKey() == UserFields.ERA_COMMONS_ID.getValue()).findFirst();
```
JVThomas: The commonsID is saved as `nihUsername`, so you'll need to use the `UserFields.ERA_USERNAME` constant
Screen Shot 2021-06-17 at 7 44 31 AM

You also need to use `.equals()` or  `.equalsIgnoreCase()`, (`==` checks if they're the same reference).

JVThomas (dismissed):

src/main/java/org/broadinstitute/consent/http/service/NihService.java
rushtong: Have not reviewed this yet, but having two field names for the same value is going to be a nightmare to debug. We need to consolidate on one value somehow.

rushtong: We have a story that will remove this ambiguity (https://broadworkbench.atlassian.net/browse/DUOS-1095), but it's slated for SO functionality, not LC functionality. I think we can defer unifying the `eraCommonsId` and `nihUsername` fields until that story.

src/main/java/org/broadinstitute/consent/http/service/NihService.java
raejohanek: @JVThomas  I'm not seeing this. When I look in the user properties table I see usernames saved for that property (nihusername) and I see id numbers saved under the eraCommonsId property. Even in the example you show here it looks like that is meant to be a fake/mock username not ID, where are you seeing IDs saved under nihUsername?

JVThomas: This isn't table results, it's the current value of updatedProps at a breakpoint place on line 35. You can mimic the current staging setup by running local instances of the UI and back-end with staging configurations and update the eraCommonsID on the researcherProfile page. (The breakpoint will let you look at the current variables)

src/main/java/org/broadinstitute/consent/http/resources/NihAccountResource.java
```diff
@@ -32,8 +35,9 @@ public NihAccountResource(NihService nihService, UserService userService) {
     @RolesAllowed(RESEARCHER)
     public Response registerResearcher(NIHUserAccount nihAccount, @Auth AuthUser user) {
         try {
-            userService.findUserByEmail(user.getName());
-            return Response.ok(nihService.authenticateNih(nihAccount, user)).build();
+            User u = userService.findUserByEmail(user.getName());
+            List<UserProperty> authUserProps = nihService.authenticateNih(nihAccount, user, u.getDacUserId());
+            return Response.ok(authUserProps).build();
```
rushtong: nit, not your original code, but we could name things a bit better. `@Auth AuthUser user` should be `@Auth AuthUser authUser` and then `User u = userService.findUserByEmail(user.getName());` could be `User user = userService.findUserByEmail(user.getName());`

rushtong: Also not your code, but if we're here, we should fix it. Lines 24-25:
```
    private NihService nihService;
    private UserService userService;
```
should be 
```
    private final NihService nihService;
    private final UserService userService;
```

src/main/java/org/broadinstitute/consent/http/service/NihService.java
rushtong: It's also visible if you follow the era auth link in the ui, then go to consent swagger and hit the /api/user/me endpoint. You'll see that your user property is saved like this:
```
...
  "researcherProperties": [
    {
      "propertyId": 2578,
      "userId": 3351,
      "propertyKey": "nihUsername",
      "propertyValue": "grushton"
    },
...
```

rushtong (changes requested):
See comments inline ... couple minor updates needed.

rushtong (approved):
Code looks good. Please note that until DUOS-1065 is complete, there will be no practical effect here since there are no existing LC entries to update.

PR #1123, opened 2021-06-17
DUOS-1199[risk=no]Add SO Role to the system

raejohanek:
SCOPE:
- Add entry for SO to roles table

- Add SO to RolesAllowed for create user and create library card. I think that is all they need for now, based on the tech doc. SO also needs allowance for APIs that yet to be created for retrieving users and dars. **If there are other endpoints I should be adjusting the roles allowed for, please let me know**

- Add SO to USER_ROLES enum (String, id pair).

- Add SO to static string names in Resource 

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1199

----

- Verify all tests go green


rushtong (approved):
Looks good!

PR #1127, opened 2021-06-21
DUOS-1259[risk=no]Update manage/v2 endpoint with optional role name for SOs

raejohanek:
## SCOPE:
- RESOURCE: Add SO as allowed role, add optional role param to manage/v2 endpoint, and update resource test signatures to match
- - made new test for user not found case
- SERVICE: If the roleName is present and is Signing Official, then use the new DAO call to get just the dars belonging to the SO's institution, otherwise the pre-existing functionality is maintained, using the user object instead of authUser
- - made a helper for filtering out canceled dars since it is now used in two spots
- - made three new tests for roleName = SO functionality
- - service method calls helper filterDataAccessRequestsByDAC so that had to be slightly refactored to take user instead of authuser
- - made a helper DarUtils hasUserRole, used in DAR and DAC service classes
- DAO: made new query that gets DAR info if the users belong to the same institution as the SO
- API DOCS: add optional param documentation

## ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1259

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/User.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/User.java
@@ -319,11 +319,11 @@ public class User {
         return new Gson().toJson(this);
     }
 
-    public static boolean hasUserRole(User user, UserRoles role) {
-        if (Objects.isNull(user) || Objects.isNull(user.getRoles())) {
+    public boolean hasUserRole(UserRoles role) {
+        if (Objects.isNull(this.getRoles())) {
             return false;
         } else {
-            return user.getRoles().stream().anyMatch((r) -> r.getRoleId().equals(role.getRoleId()));
+            return this.getRoles().stream().anyMatch((r) -> r.getRoleId().equals(role.getRoleId()));
         }
     }
 
--- a/src/main/java/org/broadinstitute/consent/http/models/User.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/User.java
@@ -13,6 +13,7 @@ import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.broadinstitute.consent.http.authentication.GoogleUser;
+import org.broadinstitute.consent.http.enumeration.UserRoles;
 import org.slf4j.LoggerFactory;
 import net.gcardone.junidecode.Junidecode;
 
@@ -316,6 +317,14 @@ public class User {
     @Override
     public String toString() {
         return new Gson().toJson(this);
+    }
+
+    public static boolean hasUserRole(User user, UserRoles role) {
+        if (Objects.isNull(user) || Objects.isNull(user.getRoles())) {
+            return false;
+        } else {
+            return user.getRoles().stream().anyMatch((r) -> r.getRoleId().equals(role.getRoleId()));
+        }
     }
 
 }
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -160,12 +160,18 @@ public class DataAccessRequestResource extends Resource {
             User user = userService.findUserByEmail(authUser.getName());
             String roleNameValue = roleName.orElse(null);
             if (Objects.nonNull(roleNameValue)) {
+                if (roleNameValue.equalsIgnoreCase(UserRoles.MEMBER.getRoleName())
+                || roleNameValue.equalsIgnoreCase(UserRoles.CHAIRPERSON.getRoleName())
+                || roleNameValue.equalsIgnoreCase(UserRoles.DATAOWNER.getRoleName())
+                || roleNameValue.equalsIgnoreCase(UserRoles.RESEARCHER.getRoleName())
+                || roleNameValue.equalsIgnoreCase(UserRoles.ALUMNI.getRoleName())) {
+                    throw new BadRequestException("Signing Official Role is the only role supported at this time");
+                }
                 //if the roleName is SO and the user does not have that role throw an exception
-                if (roleNameValue.equals(UserRoles.SIGNINGOFFICIAL.getRoleName())) {
+                if (roleNameValue.equalsIgnoreCase(UserRoles.SIGNINGOFFICIAL.getRoleName())) {
                     if (!user.hasUserRole(UserRoles.SIGNINGOFFICIAL)) {
                         throw new NotFoundException("User: " + user.getDisplayName() + ", " + " does not have Signing Official role.");
                     }
-                //if there is a roleName but it is not SO then throw an exception
                 } else {
                     throw new BadRequestException("Invalid role name: " + roleNameValue);
                 }
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -156,8 +156,13 @@ public class DataAccessRequestResource extends Resource {
     @Path("/manage/v2")
     @RolesAllowed({ADMIN, CHAIRPERSON, MEMBER, SIGNINGOFFICIAL})
     public Response describeManageDataAccessRequestsV2(@Auth AuthUser authUser, @QueryParam("roleName") Optional<String> roleName) {
-        List<DataAccessRequestManage> dars = dataAccessRequestService.describeDataAccessRequestManageV2(authUser, roleName);
-        return Response.ok().entity(dars).build();
+        try {
+            User user = userService.findUserByEmail(authUser.getName());
+            List<DataAccessRequestManage> dars = dataAccessRequestService.describeDataAccessRequestManageV2(user, roleName);
+            return Response.ok().entity(dars).build();
+        } catch(Exception e) {
+            return createExceptionResponse(e);
+        }
     }
 
     @GET
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -154,9 +154,9 @@ public class DataAccessRequestResource extends Resource {
     @GET
     @Produces("application/json")
     @Path("/manage/v2")
-    @RolesAllowed({ADMIN, CHAIRPERSON, MEMBER})
-    public Response describeManageDataAccessRequestsV2(@Auth AuthUser authUser) {
-        List<DataAccessRequestManage> dars = dataAccessRequestService.describeDataAccessRequestManageV2(authUser);
+    @RolesAllowed({ADMIN, CHAIRPERSON, MEMBER, SIGNINGOFFICIAL})
+    public Response describeManageDataAccessRequestsV2(@Auth AuthUser authUser, @QueryParam("roleName") Optional<String> roleName) {
+        List<DataAccessRequestManage> dars = dataAccessRequestService.describeDataAccessRequestManageV2(authUser, roleName);
         return Response.ok().entity(dars).build();
     }
 
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -5,6 +5,7 @@ import com.google.inject.Inject;
 import io.dropwizard.auth.Auth;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.EnumSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
@@ -160,17 +161,19 @@ public class DataAccessRequestResource extends Resource {
             User user = userService.findUserByEmail(authUser.getName());
             String roleNameValue = roleName.orElse(null);
             if (Objects.nonNull(roleNameValue)) {
-                if (roleNameValue.equalsIgnoreCase(UserRoles.MEMBER.getRoleName())
-                || roleNameValue.equalsIgnoreCase(UserRoles.CHAIRPERSON.getRoleName())
-                || roleNameValue.equalsIgnoreCase(UserRoles.DATAOWNER.getRoleName())
-                || roleNameValue.equalsIgnoreCase(UserRoles.RESEARCHER.getRoleName())
-                || roleNameValue.equalsIgnoreCase(UserRoles.ALUMNI.getRoleName())) {
-                    throw new BadRequestException("Signing Official Role is the only role supported at this time");
-                }
-                //if the roleName is SO and the user does not have that role throw an exception
-                if (roleNameValue.equalsIgnoreCase(UserRoles.SIGNINGOFFICIAL.getRoleName())) {
-                    if (!user.hasUserRole(UserRoles.SIGNINGOFFICIAL)) {
-                        throw new NotFoundException("User: " + user.getDisplayName() + ", " + " does not have Signing Official role.");
+                boolean valid = EnumSet.allOf(UserRoles.class)
+                  .stream()
+                  .map(UserRoles::getRoleName)
+                  .map(String::toLowerCase)
+                  .anyMatch(roleNameValue::equalsIgnoreCase);
+                if (valid) {
+                    //if the roleName is SO and the user does not have that role throw an exception
+                    if (roleNameValue.equalsIgnoreCase(UserRoles.SIGNINGOFFICIAL.getRoleName())) {
+                        if (!user.hasUserRole(UserRoles.SIGNINGOFFICIAL)) {
+                           throw new NotFoundException("User: " + user.getDisplayName() + ", " + " does not have Signing Official role.");
+                       }
+                   } else {
+                        throw new BadRequestException("Signing Official Role is the only role supported at this time");
                     }
                 } else {
                     throw new BadRequestException("Invalid role name: " + roleNameValue);
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -157,7 +157,8 @@ public class DataAccessRequestResource extends Resource {
     public Response describeManageDataAccessRequestsV2(@Auth AuthUser authUser, @QueryParam("roleName") Optional<String> roleName) {
         try {
             User user = userService.findUserByEmail(authUser.getName());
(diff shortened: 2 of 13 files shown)


src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -142,12 +143,22 @@ public Integer getTotalUnReviewedDars(AuthUser authUser) {
...
+            if (roleName.get().equalsIgnoreCase(UserRoles.SIGNINGOFFICIAL.getRoleName())) {
+                User user = userDAO.findUserByEmailAndRoleId(authUser.getName(), 7);
+                if (Objects.nonNull(user) && Objects.nonNull(user.getInstitutionId())) {
+                    List<DataAccessRequest> dars = dataAccessRequestDAO.findAllDataAccessRequestsForInstitution(user.getInstitutionId());
+                    List<DataAccessRequest> openDars = filterOutCanceledDars(dars);
+                    return createAccessRequestManageV2(openDars);
+                } else {
+                    throw new NotFoundException("Signing Official user associated to an Institution cannot be found.");
+                }
+            }
```
rushtong: I think the flow should be: 

- Check if role is present
- Find auth user
  - throw a 404 if user not found
  - ensure that the user has the role
  - throw a 403 error if user is trying to access a role they don't have
- query on DARs for the user/institution

However, the user validation work should really happen in the resource class, so consider moving the find-user/throw-error logic there so the service class can focus on doing only the work of finding DARs for a role. This would be easier of the service method took a full `User` object instead of an `AuthUser` object.

src/main/resources/assets/api-docs.yaml
```diff
@@ -1670,6 +1670,13 @@ paths:
       summary: Manage Data Access Requests V2
       description: |
         Returns all the elections created for Data Access Requests relevant to the current user.
+      parameters:
+        - name: roleName
+          in: query
+          description: The optional roleName of the user making the request
```
rushtong: nit: language
```suggestion
          description: An optional roleName that the user is making the request as. A user can have multiple roles, so adding this parameter will filter the list accordingly.
```

rushtong (changes requested):
Minor feedback inline ... this looks good!

src/main/java/org/broadinstitute/consent/http/service/DacService.java
```diff
@@ -306,13 +306,13 @@ private boolean isAuthUserChairOrMember(AuthUser authUser) {
     /**
      * Filter data access requests by the DAC they are associated with.
      */
-    List<DataAccessRequest> filterDataAccessRequestsByDac(List<DataAccessRequest> documents, AuthUser authUser) {
-        if (isAuthUserAdmin(authUser)) {
+    List<DataAccessRequest> filterDataAccessRequestsByDac(List<DataAccessRequest> documents, User user) {
+        if (DarUtil.hasUserRole(user, 4)) {
```
rushtong: Same feedback as below, either mimic `isAuthUserAdmin` or reference the enum values instead of using integer values.

src/main/java/org/broadinstitute/consent/http/service/DacService.java
```diff
@@ -306,13 +306,13 @@ private boolean isAuthUserChairOrMember(AuthUser authUser) {
...
             return documents;
         }
         // Chair and Member users can see data access requests that they have DAC access to
-        if (isAuthUserChairOrMember(authUser)) {
-            List<Integer> accessibleDatasetIds = dataSetDAO.findDataSetsByAuthUserEmail(authUser.getName()).
+        if (DarUtil.hasUserRole(user, 1) || DarUtil.hasUserRole(user, 2)) {
+            List<Integer> accessibleDatasetIds = dataSetDAO.findDataSetsByAuthUserEmail(user.getEmail()).
```
rushtong: We should either mimic what we're doing in `isAuthUserChairOrMember` with a new method like `isUserChairOrMember`, or reference the user roles enums instead of using the integer values of the role ids.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -139,15 +141,27 @@ public Integer getTotalUnReviewedDars(AuthUser authUser) {
...
+                if (Objects.nonNull(user.getInstitutionId())) {
+                    List<DataAccessRequest> dars = dataAccessRequestDAO.findAllDataAccessRequestsForInstitution(user.getInstitutionId());
+                    List<DataAccessRequest> openDars = filterOutCanceledDars(dars);
+                    return createAccessRequestManageV2(openDars);
+                } else {
+                    throw new NotFoundException("Signing Official (user: " + user.getDisplayName() + ", " + user.getDacUserId() + ") "
+                      + "is not associated with an Institution.");
+                }
+            } else {
+                throw new ForbiddenException("User: " + user.getDisplayName() + ", " + user.getDacUserId() + " does not have Signing Official role.");
```
rushtong: nit: remove the user id from the message, we don't need to leak that unnecessarily.

src/main/java/org/broadinstitute/consent/http/util/DarUtil.java
```diff
@@ -22,6 +23,14 @@
         return Collections.emptyList();
     }
 
+    public static boolean hasUserRole(User user, Integer roleId) {
+        if (Objects.isNull(user) || Objects.isNull(user.getRoles())) {
+            return false;
+        } else {
+            return user.getRoles().stream().anyMatch((role) -> role.getRoleId().equals(roleId));
+        }
+    }
```
rushtong: nit: `DarUtils` doesn't seem like the right home for a method that checks on user roles. This could be a method on the `User` model itself if you really need it.

src/test/java/org/broadinstitute/consent/http/service/DataAccessRequestServiceTest.java
```diff
@@ -1,5 +1,7 @@
 package org.broadinstitute.consent.http.service;
 
+import org.broadinstitute.consent.http.enumeration.UserRoles;
+import org.broadinstitute.consent.http.models.*;
```
rushtong: Avoid star imports

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -139,15 +141,27 @@ public Integer getTotalUnReviewedDars(AuthUser authUser) {
     /**
      * Filter DataAccessRequestManage objects on user.
      *
-     * @param authUser Filter on what DACs the auth user has access to.
+     * @param user Filter on what DARs the user has access to.
      * @return List of DataAccessRequestManage objects
      */
-    public List<DataAccessRequestManage> describeDataAccessRequestManageV2(AuthUser authUser) {
+    public List<DataAccessRequestManage> describeDataAccessRequestManageV2(User user, Optional<String> roleName) {
```
rushtong: Optional is fine as an argument for the resource class, but the service class should take a plain `String` ([see link](http://dolszewski.com/java/java-8-optional-use-cases/)) and then handle the null case specifically.

rushtong (changes requested):
Caveat, I did not test this version locally. Please see comments inline. Thank you!

src/test/java/org/broadinstitute/consent/http/service/DataAccessRequestServiceTest.java
rushtong: See https://intellij-support.jetbrains.com/hc/en-us/community/posts/206203659-Turn-off-Wildcard-imports- for how to do this in your IDE

src/main/java/org/broadinstitute/consent/http/models/User.java
```diff
@@ -318,4 +319,12 @@ public String toString() {
         return new Gson().toJson(this);
     }
 
+    public static boolean hasUserRole(User user, UserRoles role) {
```
rushtong: Can't this just take itself?, i.e. `hasUserRole(UserRoles role) { check this.getRoles(), etc. };` It would assume that `user` isn't null so you'd want to ensure that they are non-null in the callers.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -153,10 +153,16 @@ public Response describeManageDataAccessRequests(@Auth AuthUser authUser) {
...
     @Path("/manage/v2")
-    @RolesAllowed({ADMIN, CHAIRPERSON, MEMBER})
-    public Response describeManageDataAccessRequestsV2(@Auth AuthUser authUser) {
-        List<DataAccessRequestManage> dars = dataAccessRequestService.describeDataAccessRequestManageV2(authUser);
-        return Response.ok().entity(dars).build();
+    @RolesAllowed({ADMIN, CHAIRPERSON, MEMBER, SIGNINGOFFICIAL})
+    public Response describeManageDataAccessRequestsV2(@Auth AuthUser authUser, @QueryParam("roleName") Optional<String> roleName) {
+        try {
+            User user = userService.findUserByEmail(authUser.getName());
+            String roleNameValue = roleName.isPresent() ? roleName.get() : null;
```
rushtong: Intellij should be recommending this for you:
```suggestion
            String roleNameValue = roleName.orElse(null);
```

src/main/java/org/broadinstitute/consent/http/service/DacService.java
```diff
@@ -306,13 +306,13 @@ private boolean isAuthUserChairOrMember(AuthUser authUser) {
     /**
      * Filter data access requests by the DAC they are associated with.
      */
-    List<DataAccessRequest> filterDataAccessRequestsByDac(List<DataAccessRequest> documents, AuthUser authUser) {
-        if (isAuthUserAdmin(authUser)) {
+    List<DataAccessRequest> filterDataAccessRequestsByDac(List<DataAccessRequest> documents, User user) {
+        if (User.hasUserRole(user, UserRoles.ADMIN)) {
```
rushtong: Instead of a static, it should be a plain utility method on user, i.e. 
```suggestion
        if (user.hasUserRole(UserRoles.ADMIN)) {
```

src/main/java/org/broadinstitute/consent/http/models/User.java
rushtong: Additionally, it would be cleaner to have it be non-static so an existing `user` could call this method.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
rushtong: If I'm reading this right, I still see the user id in the message.

rushtong (changes requested):
Looking better! Please see comments inline.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -139,15 +141,27 @@ public Integer getTotalUnReviewedDars(AuthUser authUser) {
...
+                if (Objects.nonNull(user.getInstitutionId())) {
+                    List<DataAccessRequest> dars = dataAccessRequestDAO.findAllDataAccessRequestsForInstitution(user.getInstitutionId());
+                    List<DataAccessRequest> openDars = filterOutCanceledDars(dars);
+                    return createAccessRequestManageV2(openDars);
+                } else {
+                    throw new NotFoundException("Signing Official (user: " + user.getDisplayName() + ") "
+                      + "is not associated with an Institution.");
+                }
+            } else {
+                throw new ForbiddenException("User: " + user.getDisplayName() + ", " + user.getDacUserId() + " does not have Signing Official role.");
```
rushtong: I'm getting this exception if I try to use other roles, ie. `Member`, `Chairperson`, etc., which is confusing since I'm not asking for SO dars. I think in an earlier comment, I suggested handling the role validation in the resource method which should help avoid putting too much logic into the service method. If we validated that the user has the role they are asking for there, then here in the service class, we would only need to check the roleName === SO case and then continue on with the rest of the method in all other cases.

rushtong: Also, I'm still seeing the `user.getDacUserId()` in the error message which was another previous PR comment.

rushtong (changes requested):
Please see comments inline, thank you!

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
rushtong: FYI, I also see this error when I try a non-valid role, like `french` or `fries`. We definitely need some input validation in the resource class to prevent this kind of error.

raejohanek: Okay I just moved the SO role validation to the resource class, and added roleName validation in the resource class, as well as corresponding tests

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -153,10 +154,27 @@ public Response describeManageDataAccessRequests(@Auth AuthUser authUser) {
...
+            if (Objects.nonNull(roleNameValue)) {
+                //if the roleName is SO and the user does not have that role throw an exception
+                if (roleNameValue.equals(UserRoles.SIGNINGOFFICIAL.getRoleName())) {
+                    if (!user.hasUserRole(UserRoles.SIGNINGOFFICIAL)) {
+                        throw new NotFoundException("User: " + user.getDisplayName() + ", " + " does not have Signing Official role.");
+                    }
+                //if there is a roleName but it is not SO then throw an exception
+                } else {
+                    throw new BadRequestException("Invalid role name: " + roleNameValue);
+                }
```
rushtong: If I enter "Member", that's a valid role name. I think the message should say, "Signing Official Role is the only role supported at this time" or something like that. If I enter "French Fries", that's an **_invalid_** role name (see comment above) so we really need two levels of validation here.

src/main/resources/assets/api-docs.yaml
```diff
@@ -1742,6 +1742,13 @@ paths:
       summary: Manage Data Access Requests V2
       description: |
         Returns all the elections created for Data Access Requests relevant to the current user.
+      parameters:
+        - name: roleName
+          in: query
+          description: An optional roleName that the user is making the request as. A user can have multiple roles, so adding this parameter will filter the list accordingly.
```
rushtong, line 1748: In swagger, we can specify a list of allowable values here. Since we're only supporting signing official, it might be good to enforce that here. See https://swagger.io/docs/specification/data-models/enums/ for an example.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -153,10 +154,27 @@ public Response describeManageDataAccessRequests(@Auth AuthUser authUser) {
...
-    public Response describeManageDataAccessRequestsV2(@Auth AuthUser authUser) {
-        List<DataAccessRequestManage> dars = dataAccessRequestService.describeDataAccessRequestManageV2(authUser);
-        return Response.ok().entity(dars).build();
+    @RolesAllowed({ADMIN, CHAIRPERSON, MEMBER, SIGNINGOFFICIAL})
+    public Response describeManageDataAccessRequestsV2(@Auth AuthUser authUser, @QueryParam("roleName") Optional<String> roleName) {
+        try {
+            User user = userService.findUserByEmail(authUser.getName());
+            String roleNameValue = roleName.orElse(null);
+            if (Objects.nonNull(roleNameValue)) {
+                //if the roleName is SO and the user does not have that role throw an exception
```
rushtong: Might be helpful to add validation here on the provided role name here so we can throw a different error if the user tries to manage dars for an invalid role name.

rushtong (changes requested):
Getting closer! Please see comments inline.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
raejohanek: Okay I just added a validation check for any of the existing user role names that gives a bad request exception - "Signing Official Role is the only role supported at this time" in addition to the other validation check that gives a bad request exception - "Invalid user role name" I also added the nullable enum to the api-docs

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -153,10 +154,33 @@ public Response describeManageDataAccessRequests(@Auth AuthUser authUser) {
...
+            User user = userService.findUserByEmail(authUser.getName());
+            String roleNameValue = roleName.orElse(null);
+            if (Objects.nonNull(roleNameValue)) {
+                if (roleNameValue.equalsIgnoreCase(UserRoles.MEMBER.getRoleName())
+                || roleNameValue.equalsIgnoreCase(UserRoles.CHAIRPERSON.getRoleName())
+                || roleNameValue.equalsIgnoreCase(UserRoles.DATAOWNER.getRoleName())
+                || roleNameValue.equalsIgnoreCase(UserRoles.RESEARCHER.getRoleName())
+                || roleNameValue.equalsIgnoreCase(UserRoles.ALUMNI.getRoleName())) {
+                    throw new BadRequestException("Signing Official Role is the only role supported at this time");
+                }
```
rushtong: I generally try to stay away from long conditionals like this. What ends up happening is that over time, we insert variations into the conditions that make it more complicated than it needs to be. There are lots of ways to do this that are a little cleaner.

### Streaming
```
boolean valid = EnumSet.allOf(UserRoles.class)
        .stream()
        .map(UserRoles::getRoleName)
        .map(String::toLowerCase)
        .anyMatch(roleNameValue::equalsIgnoreCase);
```

## Collection contains
```
boolean valid = EnumSet.allOf(UserRoles.class)
        .stream()
        .map(UserRoles::getRoleName)
        .map(String::toLowerCase)
        .collect(Collectors.toList())
        .contains(roleNameValue.toLowerCase());
```

## Enum Values Array
```
boolean valid = Arrays
        .stream(UserRoles.values())
        .anyMatch(ur -> ur.getRoleName().equalsIgnoreCase(roleNameValue));
```

src/main/resources/assets/api-docs.yaml
```diff
@@ -1742,6 +1742,17 @@ paths:
...
+          required: false
+          schema:
+            type: String
+            nullable: true
+            enum:
+              - SigningOfficial
+              - null
```
rushtong: We don't need the null value here.

rushtong: If you try and use that, it tries to send `null` as a role name, which obviously fails.

rushtong (approved):
Thank you for the changes!

PR #1131, opened 2021-06-23
DUOS-1323[risk=low] Admins can create Library cards for unregistered users

JVThomas:
Addresses [DUOS-1323](https://broadworkbench.atlassian.net/browse/DUOS-1323)

Per requirement on [DUOS-1068](https://broadworkbench.atlassian.net/browse/DUOS-1068), admins should be able to create Library Cards for unregistered emails by providing an email. In addition, create flow should have validation steps to ensure that duplicate cards are not created if a card already exists. This PR modifies LC Resource, LC Service, and associated tests to account for changes in user and payload validation.

NOTE: May need to introduce further adjustments later on when service class methods are utilized by other roles (Signing Official comes to mind).

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/exceptions/ConsentConflictException.java
+++ b/src/main/java/org/broadinstitute/consent/http/exceptions/ConsentConflictException.java
@@ -1,0 +1,23 @@
+package org.broadinstitute.consent.http.exceptions;
+
+public class ConsentConflictException extends Exception{
+  public ConsentConflictException() {
+    super();
+  }
+
+  public ConsentConflictException(String message) {
+    super(message);
+  }
+
+  public ConsentConflictException(String message, Throwable cause) {
+    super(message, cause);
+  }
+
+  public ConsentConflictException(Throwable cause) {
+    super(cause);
+  }
+
+  public ConsentConflictException(String message, Throwable cause, boolean enableSuppresion, boolean writableStackTrace) {
+    super(message, cause, enableSuppresion, writableStackTrace);
+  }
+}
--- a/src/main/java/org/broadinstitute/consent/http/exceptions/ConsentConflictException.java
+++ b/src/main/java/org/broadinstitute/consent/http/exceptions/ConsentConflictException.java
@@ -1,23 +1,10 @@
 package org.broadinstitute.consent.http.exceptions;
 
-public class ConsentConflictException extends Exception{
-  public ConsentConflictException() {
-    super();
-  }
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ClientErrorException;
 
-  public ConsentConflictException(String message) {
-    super(message);
-  }
-
-  public ConsentConflictException(String message, Throwable cause) {
-    super(message, cause);
-  }
-
-  public ConsentConflictException(Throwable cause) {
-    super(cause);
-  }
-
-  public ConsentConflictException(String message, Throwable cause, boolean enableSuppresion, boolean writableStackTrace) {
-    super(message, cause, enableSuppresion, writableStackTrace);
-  }
+public class ConsentConflictException extends ClientErrorException {
+ public ConsentConflictException() {
+   super(Response.Status.CONFLICT);
+ }
 }
--- a/src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
@@ -3,12 +3,12 @@ package org.broadinstitute.consent.http.service;
 import org.broadinstitute.consent.http.db.InstitutionDAO;
 import org.broadinstitute.consent.http.db.LibraryCardDAO;
 import org.broadinstitute.consent.http.db.UserDAO;
+import org.broadinstitute.consent.http.exceptions.ConsentConflictException;
 import org.broadinstitute.consent.http.models.Institution;
 import org.broadinstitute.consent.http.models.LibraryCard;
 import org.broadinstitute.consent.http.models.User;
 
 import javax.ws.rs.BadRequestException;
-import javax.ws.rs.InternalServerErrorException;
 import javax.ws.rs.NotFoundException;
 
 import java.util.Date;
@@ -28,7 +28,7 @@ public class LibraryCardService {
         this.userDAO = userDAO;
     }
 
-    public LibraryCard createLibraryCard(LibraryCard libraryCard) {
+    public LibraryCard createLibraryCard(LibraryCard libraryCard) throws Exception {
         throwIfNull(libraryCard);
         checkForValidInstitution(libraryCard.getInstitutionId());
         checkIfCardExists(libraryCard);
@@ -130,7 +130,7 @@ public class LibraryCardService {
     }
 
     //helper method for create method, checks to see if card already exists
-    private void checkIfCardExists(LibraryCard payload) {
+    private void checkIfCardExists(LibraryCard payload) throws Exception {
         Integer userId = payload.getUserId();
         String email = payload.getUserEmail();
         Integer institutionId = payload.getInstitutionId();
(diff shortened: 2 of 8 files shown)


src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
```diff
@@ -124,4 +128,62 @@ private void throwIfNull(LibraryCard libraryCard) {
...
+        }
+        foundCard = results.stream().filter(card -> {
+            Boolean sameUserId = Objects.nonNull(userId) && card.getUserId().equals(userId);
+            Boolean sameUserEmail = Objects.nonNull(email) && card.getUserEmail().equalsIgnoreCase(email);
+            Boolean sameInstitution = card.getInstitutionId().equals(institutionId);
+            return (sameUserId || sameUserEmail) && sameInstitution;
+        }).findFirst();
+
+        if(foundCard.isPresent()) {
+            //This exception is a placeholder, what's an exception that can be translated as "CONFLICT"
```
rushtong: We don't have one that directly expresses that, but we sure could use one. Look at `NotFoundException` which extends `ClientErrorException`. We could write our own `ConflictException` and extend `ClientErrorException` as well, and then handle that explicitly in `Resource` where we add dispatchers for known exceptions.

src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
```diff
@@ -124,4 +128,62 @@ private void throwIfNull(LibraryCard libraryCard) {
...
+            } else {
+                try{
+                    //If a user is found, update the card to have the correct userId associated
+                    User user = userDAO.findUserByEmail(card.getUserEmail());
+                    Integer userId = user.getDacUserId();
+                    card.setUserId(userId);
+                } catch(Exception e) {
+                    //Exception here means the email has not been used
+                    //Does not mean card is invalid, it just means it will only have user email provided
+                    //As such card can move forward as is
```
rushtong: `userDAO.findUserByEmail()` doesn't throw anything (aside from database errors), but can return a null that you'll need to handle.

This means there is no user with an email address in our system, which is fine, we don't need to have one in the user table. We can have an LC for an email that we don't know about yet.

src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
```diff
@@ -124,4 +128,62 @@ private void throwIfNull(LibraryCard libraryCard) {
...
+                }
+            }
+        } else {
+            //check if userId exists
+            User user = userDAO.findUserById(card.getUserId());
+            if(Objects.nonNull(user.getEmail()) && !(user.getEmail().equalsIgnoreCase(card.getUserEmail()))) {
+                //throw error here, user is trying to associate incorrect userId with email
+                throw new BadRequestException();
```
rushtong: Not sure how much of an error message we need to provide here ... but we should say something so the client knows what they did was wrong. Could also be a conflict exception, i.e. email+userId conflict with existing email user.

src/test/java/org/broadinstitute/consent/http/resources/LibraryCardResourceTest.java
```diff
@@ -127,37 +126,35 @@ public void testGetLibraryCardByInstitutionIdThrowsNotFoundException() {
...
+  // public void testCreateLibraryCardInvalidEmail() {
+  //   LibraryCard mockCard = mockLibraryCardSetup();
+  //   mockCard.setUserEmail("wrongemail@gmail.com");
+  //   String payload = new Gson().toJson(mockCard);
+  //   when(userService.findUserByEmail(authUser.getName())).thenReturn(user);
+  //   when(libraryCardService.createLibraryCard(any(LibraryCard.class))).thenReturn(mockCard);
+  //   initResource();
+  //   Response response = resource.createLibraryCard(authUser, payload);
+  //   assertEquals(HttpStatusCodes.STATUS_CODE_SERVER_ERROR, response.getStatus());
+  // }
```
rushtong: Don't forget to remove 😄

rushtong (changes requested):
Nice addition to tests! Please see minor comments inline.

src/main/java/org/broadinstitute/consent/http/exceptions/ConsentConflictException.java
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  public ConsentConflictException(Throwable cause) {
+    super(cause);
+  }
+
+  public ConsentConflictException(String message, Throwable cause, boolean enableSuppresion, boolean writableStackTrace) {
+    super(message, cause, enableSuppresion, writableStackTrace);
+  }
+}
```
rushtong, line 10: I think this would be more accurate if it extended ClientErrorException. This would be more than enough of a class:
```
public class ConsentConflictException extends ClientErrorException {
  public ConsentConflictException() {
    super(Response.Status.CONFLICT);
  }
}
```

src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
```diff
@@ -124,4 +128,62 @@ private void throwIfNull(LibraryCard libraryCard) {
...
+        }
+        foundCard = results.stream().filter(card -> {
+            Boolean sameUserId = Objects.nonNull(userId) && card.getUserId().equals(userId);
+            Boolean sameUserEmail = Objects.nonNull(email) && card.getUserEmail().equalsIgnoreCase(email);
+            Boolean sameInstitution = card.getInstitutionId().equals(institutionId);
+            return (sameUserId || sameUserEmail) && sameInstitution;
+        }).findFirst();
+
+        if(foundCard.isPresent()) {
+            //This exception is a placeholder, what's an exception that can be translated as "CONFLICT"
```
rushtong: rm comment

src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
rushtong: There's no exception thrown in the try block as described by the comments. The only exceptions you'll get are postgres ones.

src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
rushtong: Do you not want to address this comment?

src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
```diff
@@ -124,4 +128,62 @@ private void throwIfNull(LibraryCard libraryCard) {
...
+
+    //Helper method to process user data on create LC payload
+    //Needed since CREATE has a unique situation where admins can create LCs without an active user (save with userEmail instead)
+    private LibraryCard processUserOnNewLC(LibraryCard card) {
+        if (Objects.isNull(card.getUserId())) {
+            if (Objects.isNull(card.getUserEmail())) {
+                throw new BadRequestException();
+            } else {
+                try{
+                    //If a user is found, update the card to have the correct userId associated
```
raejohanek: instead of doing this in a try catch, maybe you could just check if the user is null before getting and setting the user id since the only exception being thrown here would be a null pointer

src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
```diff
@@ -124,4 +128,62 @@ private void throwIfNull(LibraryCard libraryCard) {
...
+                } catch(Exception e) {
+                    //Exception here means the email has not been used
+                    //Does not mean card is invalid, it just means it will only have user email provided
+                    //As such card can move forward as is
+                }
+            }
+        } else {
+            //check if userId exists
+            User user = userDAO.findUserById(card.getUserId());
+            if(Objects.nonNull(user.getEmail()) && !(user.getEmail().equalsIgnoreCase(card.getUserEmail()))) {
```
raejohanek, line 179: I think you could use a null check here on user in case the userId does not exist

raejohanek (approved):
changes look good, tested locally confirmed that library cards can be created for users not in the system, tests look good and all pass 👍

rushtong (approved):
Thank you for the changes!

PR #1132, opened 2021-06-24
DUOS-1260[risk=low]Create new get all users endpoint in UserResource with RoleName path param

raejohanek:
SCOPE:
- New API for SO an Admin in User Resource class
- Deprecate old API in DACUser Resource class
- New service class method in User Service
- Deprecate old service class method in User Service
- New DAO call getUsersByInstitution
- tests for all new methods
- update API docs to include documentation of new API

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1260

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
@@ -181,6 +181,6 @@ public interface UserDAO extends Transactional<UserDAO> {
             " LEFT JOIN user_role ur ON ur.user_id = du.dacuserid " +
             " LEFT JOIN roles r ON r.roleid = ur.role_id " +
             " WHERE du.institution_id = :institutionId")
-    List<User> findUsersByInstitution(@BindList("institutionId") Integer institutionId)
+    List<User> findUsersByInstitution(@BindList("institutionId") Integer institutionId);
 
 }
--- a/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
@@ -175,12 +175,14 @@ public interface UserDAO extends Transactional<UserDAO> {
             @BindList("datasetIds") List<Integer> datasetIds,
             @BindList("roleNames") List<String> roleNames);
 
-    @UseRowMapper(UserWithRolesMapper.class)
-    @SqlQuery("SELECT du.*, r.roleid, r.name, ur.dac_id " +
+    @RegisterBeanMapper(value = User.class)
+    @RegisterBeanMapper(value = UserRole.class)
+    @UseRowReducer(UserWithRolesReducer.class)
+    @SqlQuery("SELECT du.*, r.name, ur.role_id, ur.user_role_id, ur.dac_id " +
             " FROM dacuser du " +
             " LEFT JOIN user_role ur ON ur.user_id = du.dacuserid " +
             " LEFT JOIN roles r ON r.roleid = ur.role_id " +
             " WHERE du.institution_id = :institutionId")
-    List<User> findUsersByInstitution(@BindList("institutionId") Integer institutionId);
+    List<User> findUsersByInstitution(@Bind("institutionId") Integer institutionId);
 
 }
--- a/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
@@ -175,4 +175,12 @@ public interface UserDAO extends Transactional<UserDAO> {
             @BindList("datasetIds") List<Integer> datasetIds,
             @BindList("roleNames") List<String> roleNames);
 
+    @UseRowMapper(UserWithRolesMapper.class)
+    @SqlQuery("SELECT du.*, r.roleid, r.name, ur.dac_id " +
+            " FROM dacuser du " +
+            " LEFT JOIN user_role ur ON ur.user_id = du.dacuserid " +
+            " LEFT JOIN roles r ON r.roleid = ur.role_id " +
+            " WHERE du.institution_id = :institutionId")
+    List<User> findUsersByInstitution(@BindList("institutionId") Integer institutionId)
+
 }
--- a/src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
+++ b/src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
@@ -52,7 +52,6 @@ public enum UserRoles {
         return EnumSet.allOf(UserRoles.class)
           .stream()
           .map(UserRoles::getRoleName)
-          .map(String::toLowerCase)
           .anyMatch(roleName::equalsIgnoreCase);
     }
 
--- a/src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
+++ b/src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
@@ -2,6 +2,8 @@ package org.broadinstitute.consent.http.enumeration;
 
 import org.broadinstitute.consent.http.resources.Resource;
 
+import java.util.EnumSet;
+
 public enum UserRoles {
 
     MEMBER(Resource.MEMBER, 1),
@@ -44,6 +46,14 @@ public enum UserRoles {
             }
         }
         return null;
+    }
+
+    public static boolean isValidRole(String roleName) {
+        return EnumSet.allOf(UserRoles.class)
+          .stream()
+          .map(UserRoles::getRoleName)
+          .map(String::toLowerCase)
+          .anyMatch(roleName::equalsIgnoreCase);
     }
 
 }
--- a/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
@@ -24,6 +24,7 @@ import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
@@ -51,6 +52,19 @@ public class UserResource extends Resource {
         this.userService = userService;
         this.libraryCardService = libraryCardService;
         this.gson = new Gson();
+    }
+
+    @GET
+    @Produces("application/json")
+    @RolesAllowed({ADMIN, SIGNINGOFFICIAL})
+    public Response getUsers(@Auth AuthUser authUser, @QueryParam("roleName") String roleName) {
+        try {
+            User user = userService.findUserByEmail(authUser.getName());
+            List<User> users = userService.getUsersByUserRole(user, roleName);
+            return Response.ok().entity(users).build();
+        } catch (Exception e) {
+            return createExceptionResponse(e);
+        }
     }
 
     @GET
--- a/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
@@ -64,26 +64,21 @@ public class UserResource extends Resource {
     public Response getUsers(@Auth AuthUser authUser, @PathParam("roleName") String roleName) {
         try {
             User user = userService.findUserByEmail(authUser.getName());
-            if (Objects.nonNull(roleName)) {
-                boolean valid = UserRoles.isValidRole(roleName);
-                if (valid) {
-                    //if there is a valid roleName but it is not SO or Admin then throw an exception
-                    if (!roleName.equals(UserRoles.ADMIN.getRoleName()) && !roleName.equals(UserRoles.SIGNINGOFFICIAL.getRoleName())) {
-                        throw new BadRequestException("Unsupported role name: " + roleName);
-                    }
-                    if (!user.hasUserRole(UserRoles.getUserRoleFromName(roleName))) {
-                        throw new NotFoundException("User: " + user.getDisplayName() + ", " + " does not have " + roleName + " role.");
-                    }
-                    List<User> users = userService.getUsersByUserRole(user, roleName);
-                    return Response.ok().entity(users).build();
+            boolean valid = UserRoles.isValidRole(roleName);
+            if (valid) {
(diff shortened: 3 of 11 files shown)


src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
```diff
@@ -175,4 +175,12 @@ void updateUser(@Bind("displayName") String displayName,
             @BindList("datasetIds") List<Integer> datasetIds,
             @BindList("roleNames") List<String> roleNames);
 
+    @UseRowMapper(UserWithRolesMapper.class)
+    @SqlQuery("SELECT du.*, r.roleid, r.name, ur.dac_id " +
+            " FROM dacuser du " +
+            " LEFT JOIN user_role ur ON ur.user_id = du.dacuserid " +
+            " LEFT JOIN roles r ON r.roleid = ur.role_id " +
+            " WHERE du.institution_id = :institutionId")
+    List<User> findUsersByInstitution(@BindList("institutionId") Integer institutionId);
```
rushtong: Should just be a `@Bind` since it's a single integer. `BindList` is for collection types.

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -53,6 +57,52 @@ public UserResource(UserService userService, LibraryCardService libraryCardServi
...
+    @RolesAllowed({ADMIN, SIGNINGOFFICIAL})
+    public Response getUsers(@Auth AuthUser authUser, @PathParam("roleName") String roleName) {
+        try {
+            User user = userService.findUserByEmail(authUser.getName());
+            if (Objects.nonNull(roleName)) {
+                boolean valid = EnumSet.allOf(UserRoles.class)
+                  .stream()
+                  .map(UserRoles::getRoleName)
+                  .map(String::toLowerCase)
+                  .anyMatch(roleName::equalsIgnoreCase);
```
rushtong: We just added this for another PR, so now it's time to think about DRY. This could be a static method on `UserRoles`, i.e. `static boolean UserRoles.isValid(String roleName) {... do the work ...};`

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -53,6 +57,52 @@ public UserResource(UserService userService, LibraryCardService libraryCardServi
...
+                if (valid) {
+                    //if the roleName is SO and the user does not have that role throw an exception
+                    if (roleName.equals(UserRoles.SIGNINGOFFICIAL.getRoleName())) {
+                        if (!user.hasUserRole(UserRoles.SIGNINGOFFICIAL)) {
+                            throw new NotFoundException("User: " + user.getDisplayName() + ", " + " does not have Signing Official role.");
```
rushtong: nit: remove all the messaging here and in other places. I'm torn on whether we should throw a not found here too. Logically, it should be a 403 and then we let the dispatch handler in `Resource` translate 403's to 404's, but I don't think we're doing that right now. 

Additionally, we should be validating the user role prior to all of these if blocks ... just after `if(valid) {...` so that we are prepared for any case where a user tries to see a role they shouldn't have.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -83,10 +84,25 @@ public User findUserByEmail(String email) throws NotFoundException {
...
     public Collection<User> describeUsers() {
         return userDAO.findUsers();
     }
 
+    public List<User> getUsersByUserRole(User user, String roleName) {
+        if (roleName.equalsIgnoreCase(UserRoles.SIGNINGOFFICIAL.getRoleName())) {
+            if (Objects.nonNull(user.getInstitutionId())) {
+                return userDAO.findUsersByInstitution(user.getInstitutionId());
+            } else {
+                throw new NotFoundException("Signing Official (user: " + user.getDisplayName() + ") "+ "is not associated with an Institution.");
```
rushtong: nit, unnecessary string concats:
```suggestion
                throw new NotFoundException("Signing Official (user: " + user.getDisplayName() + ") is not associated with an Institution.");
```

src/main/resources/assets/api-docs.yaml
```diff
@@ -3437,6 +3437,37 @@ paths:
...
+      parameters:
+        - name: roleName
+          in: path
+          description: The name of the role of the user
+          required: true
+          schema:
+            type: String
+            enum:
+              - SigningOfficial
+              - Admin
```
rushtong: nit: we should probably sort these alpha

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
rushtong: In the swagger, this case is marked as a 400 error, not a 404.

rushtong:
See currently failing test:
```
Error:  Errors: 
Error:    UserDAOTest.testFindUsersByInstitution:349 » IllegalArgument argument must be ...
[INFO] 
Error:  Tests run: 835, Failures: 0, Errors: 1, Skipped: 0
```

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
raejohanek: "Additionally, we should be validating the user role prior to all of these if blocks" - I'm not sure I understand what you mean by this... because these if block are validating the role name. And do you want me to remove the exception message completely with no replacement or ? I had originally had the error as a 403 and then after our convo yesterday about how that is leaking unnecessary info, I changed it to 400 bad request, and then I changed it to 404 not found, what do you think, leave it as 404?

rushtong: > "Additionally, we should be validating the user role prior to all of these if blocks"

What I mean here is that this code looks at role names one by one looking for only the ones we care about now. Validating that the user has the role they are asking for is a singular action that should be valid for all users asking for any role. After `isValid`, we should ensure that the user has the role they are asking for and immediately throw an error in that case (failing fast) instead of looking at roles one by one (failing slowly).

rushtong: > what do you think, leave it as 404

Yes, let's leave it like that for now. We can have a team conversation about how we translate authorization errors to 403s or 404s and make a decision at another time.

raejohanek: Okay that makes sense thank you!

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -53,6 +57,38 @@ public UserResource(UserService userService, LibraryCardService libraryCardServi
         this.gson = new Gson();
     }
 
+    @GET
+    @Produces("application/json")
+    @Path("/role/{roleName}")
+    @RolesAllowed({ADMIN, SIGNINGOFFICIAL})
+    public Response getUsers(@Auth AuthUser authUser, @PathParam("roleName") String roleName) {
```
JVThomas, line 63: Instead of having UserRole validations/ thrown Exceptions happen on the resource it may be better to have it contained within service methods. It'll keep the business logic contained in reusable functions while making the Resource logic smaller and easier to follow. 

On a side note the other methods on UserResource are good candidates for refactoring for a later tech debt ticket.

raejohanek: I just had a very similar PR to this but for dars instead of users, DUOS-1259, where I had originally done this same logic and validation in the service class bc that was my thought too, but on the suggestion of Greg I moved it to the resource class, so I was continuing to follow that feedback here. It seems the same to me, so thoughts/preferences? @JVThomas @rushtong

rushtong: @JVThomas It was my feedback to put the role validation logic here in the resource. The resource level is the right place to block entry to invalid requests which is why we have the role permissions here and in other resource classes, we validate that the user is associated to an entity like a DAR. The service class shouldn't have to deal with that level of detail and focus on doing the processing of the entities.

JVThomas: The logic seems extractable and reusable for multiple endpoints. You can have the logic contained and testable in a separate function and still use it within the resource so that it (resource) can maintain its responsibility of proper authorization.

It'll also makes testing easier - the encapsulated function can be tested for its business logic (role validation) while the resource can be tested to ensure it returns proper responses.

src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
```diff
@@ -46,4 +48,12 @@ public static UserRoles getUserRoleFromId(Integer roleId) {
         return null;
     }
 
+    public static boolean isValidRole(String roleName) {
+        return EnumSet.allOf(UserRoles.class)
+          .stream()
+          .map(UserRoles::getRoleName)
+          .map(String::toLowerCase)
```
JVThomas: Do you need `toLowerCase` if you're going to use `equalsIgnoreCase`?

JVThomas (dismissed):
This is the only bug that needs to be addressed. The role dropdown is virtually un-selectable even as an authenticated admin user.
Kapture 2021-06-29 at 13 14 28

The other comments (UserRole service method, toLowerCase) are nitpicks

raejohanek:
> This is the only bug that needs to be addressed. The role dropdown is virtually un-selectable even as an authenticated admin user.
Thanks for pointing this out, I just fixed this issue and got the endpoint working on swagger by removing the required flag for the enum parameter. Even though the roleName is required and won't work with the -- option, there seems to be a bug in swagger for required enums similar to the one described here https://github.com/swagger-api/swagger-ui/issues/3706

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -53,6 +57,38 @@ public UserResource(UserService userService, LibraryCardService libraryCardServi
...
+            User user = userService.findUserByEmail(authUser.getName());
+            if (Objects.nonNull(roleName)) {
+                boolean valid = UserRoles.isValidRole(roleName);
+                if (valid) {
+                    //if there is a valid roleName but it is not SO or Admin then throw an exception
+                    if (!roleName.equals(UserRoles.ADMIN.getRoleName()) && !roleName.equals(UserRoles.SIGNINGOFFICIAL.getRoleName())) {
+                        throw new BadRequestException("Unsupported role name: " + roleName);
+                    }
+                    if (!user.hasUserRole(UserRoles.getUserRoleFromName(roleName))) {
+                        throw new NotFoundException("User: " + user.getDisplayName() + ", " + " does not have " + roleName + " role.");
```
rushtong: nit: 
```suggestion
                        throw new NotFoundException("User: " + user.getDisplayName() + ", does not have " + roleName + " role.");
```

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -53,6 +57,38 @@ public UserResource(UserService userService, LibraryCardService libraryCardServi
...
+                    }
+                    List<User> users = userService.getUsersByUserRole(user, roleName);
+                    return Response.ok().entity(users).build();
+                }
+                else {
+                    throw new BadRequestException("Invalid role name: " + roleName);
+                }
+            } else {
+                throw new BadRequestException("No user role specified.");
+            }
```
rushtong: This block is never executed because `/api/user/role/` is not a valid route. Dropwizard will only hit this method with some kind of value as `roleName`. If you want to handle both `/api/user/role` and `/api/user/role/{roleName}`, you'll need two endpoint methods to do so. Otherwise we get a 404 from DW.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -83,10 +84,25 @@ public User findUserByEmail(String email) throws NotFoundException {
...
+            if (Objects.nonNull(user.getInstitutionId())) {
+                return userDAO.findUsersByInstitution(user.getInstitutionId());
+            } else {
+                throw new NotFoundException("Signing Official (user: " + user.getDisplayName() + ") is not associated with an Institution.");
+            }
+        //there are only two cases here since we validate the roleName in the resource class, if it is not SO, then it is Admin
+        } else {
+            List<User> users = new ArrayList<>(userDAO.findUsers());
+            return users;
+        }
```
rushtong: It would be better to handle the Admin case explicitly here and not make this kind of assumption. If the role is admin, then return the dao call. If not, return an empty list.

Even better, you could add a switch on the role and have explicit cases for each role condition.

src/main/resources/assets/api-docs.yaml
```diff
@@ -3437,6 +3437,35 @@ paths:
...
+    get:
+      summary: Get a list of users that you have access to based on your role
+      description: Finds users based on the user role and permissions
+      parameters:
+        - name: roleName
+          in: path
+          description: The name of the role of the user
+          schema:
+            type: String
+            nullable: false
```
rushtong: This will fix the swagger bug you were seeing:
```suggestion
            type: string
            required: true
```

rushtong (changes requested):
Some minor comments and fixes inline ... looking good!

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
raejohanek: Okay makes sense I will remove that block

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -53,6 +54,33 @@ public UserResource(UserService userService, LibraryCardService libraryCardServi
...
+                //if there is a valid roleName but it is not SO or Admin then throw an exception
+                if (!roleName.equals(UserRoles.ADMIN.getRoleName()) && !roleName.equals(UserRoles.SIGNINGOFFICIAL.getRoleName())) {
+                    throw new BadRequestException("Unsupported role name: " + roleName);
+                }
+                if (!user.hasUserRole(UserRoles.getUserRoleFromName(roleName))) {
+                    throw new NotFoundException("User: " + user.getDisplayName() + ", does not have " + roleName + " role.");
+                }
+                List<User> users = userService.getUsersByUserRole(user, roleName);
+                return Response.ok().entity(users).build();
+                }
```
rushtong: nit spacing
```suggestion
            }
```

rushtong (approved):
One minor suggestion inline, then 👍🏽

PR #1134, opened 2021-06-25
DUOS-1065[risk=no] LC Backpopulation

raejohanek:
SCOPE:
- library cards have not been populated yet because currently the resource class tries to find the user by the library card email and that user may not exist so this throws an error, this bug is already being addressed in DUOS-1323, once that PR is merged in, then I can insert all of the library cards 
- this PR just aims to include the script used for population in consent

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1065

----

- Verify all tests go green


rushtong (approved):
Love it!

PR #1136, opened 2021-06-28
DUOS-1324[risk=no] Remove investigator name from dataset metrics API and display

raejohanek:
SCOPE:
- remove investigator from dataset metrics summary object

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1324
----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
@@ -60,7 +60,6 @@ public class MetricsService {
     @JsonProperty final String projectTitle;
     @JsonProperty final String darCode;
     @JsonProperty final String nonTechRus;
-    @JsonProperty final String investigator;
     @JsonProperty final String referenceId;
 
     public DarMetricsSummary(DataAccessRequest dar) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
```diff
@@ -69,14 +68,12 @@ public DarMetricsSummary(DataAccessRequest dar) {
         this.projectTitle = dar.data.getProjectTitle();
         this.darCode =  dar.data.getDarCode();
         this.nonTechRus =  dar.data.getNonTechRus();
-        this.investigator = DarUtil.findPI(userDAO.findUserById(dar.userId));
```
JVThomas, line 72: @greg.rushton Question: Do we want to remove the investigator attribute for dataset only or do we want to remove it altogether? I'm not too familiar with the exact rules of GDPR and this would remove the investigator detail from both the DatasetMetrics and DarMetrics. If we do want to remove it altogether, then this is the right move.

raejohanek: @JVThomas  This does not affect the darMetrics endpoint. DarMetricsSummary is only used to create smaller objects representing dars for the dataset metrics endpoint

JVThomas (approved):
Looks good, code removes investigator key from the model.

PR #1138, opened 2021-06-28
DUOS-1093[risk=no]Update DAR to use Researcher's SO

raejohanek:
SCOPE:
This is the back end component of a UI update which aims to have a drop down for Researchers to pick their SO from a list of SOs belonging to their institution (or if their SO is not on that list they can enter a free text name). We did not have any API that was appropriate for this use (besides getting all users and filtering through them) so I made a new API for researchers only since the UI update is on the dar application page which is for researchers only. 
- New API getSOsByInstitution, two resource tests, add documentation API docs
- New Service class method, two service tests
- New DAO call, one new DAO test

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1093

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
@@ -179,7 +179,7 @@ public interface UserDAO extends Transactional<UserDAO> {
     @SqlQuery("SELECT u.* FROM dacuser u "
       + " LEFT JOIN user_role ur ON ur.user_id = u.dacuserid "
       + " LEFT JOIN roles r ON r.roleid = ur.role_id "
-      + " WHERE r.name = 'SigningOfficial' "
+      + " WHERE LOWER(r.name) = 'signingofficial' "
       + " AND u.institution_id = :institutionId")
     List<User> getSOsByInstitution(@Bind("institutionId") Integer institutionId);
 
--- a/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
@@ -175,4 +175,12 @@ public interface UserDAO extends Transactional<UserDAO> {
             @BindList("datasetIds") List<Integer> datasetIds,
             @BindList("roleNames") List<String> roleNames);
 
+    @RegisterBeanMapper(value = User.class)
+    @SqlQuery("SELECT u.* FROM dacuser u "
+      + " LEFT JOIN user_role ur ON ur.user_id = u.dacuserid "
+      + " LEFT JOIN roles r ON r.roleid = ur.role_id "
+      + " WHERE r.name = 'SigningOfficial' "
+      + " AND u.institution_id = :institutionId")
+    List<User> getSOsByInstitution(@Bind("institutionId") Integer institutionId);
+
 }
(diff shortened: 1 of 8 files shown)


JVThomas (approved):
Looks good, endpoint returns proper array response.

src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
```diff
@@ -175,4 +175,12 @@ void updateUser(@Bind("displayName") String displayName,
             @BindList("datasetIds") List<Integer> datasetIds,
             @BindList("roleNames") List<String> roleNames);
 
+    @RegisterBeanMapper(value = User.class)
+    @SqlQuery("SELECT u.* FROM dacuser u "
+      + " LEFT JOIN user_role ur ON ur.user_id = u.dacuserid "
+      + " LEFT JOIN roles r ON r.roleid = ur.role_id "
+      + " WHERE r.name = 'SigningOfficial' "
```
rushtong: nit: Prefer lower case string comparisons:
```suggestion
      + " WHERE LOWER(r.name) = 'signingofficial' "
```

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -162,6 +163,23 @@ public Response delete(@PathParam("email") String email, @Context UriInfo info)
...
+    @Path("/SigningOfficials")
+    @RolesAllowed(RESEARCHER)
+    public Response getSOsForInstitution(@Auth AuthUser authUser) {
+        try {
+            User user = userService.findUserByEmail(authUser.getName());
+            if (Objects.isNull(user.getInstitutionId())) {
+                throw new NotFoundException("Current user, " + user.getDisplayName() + ", does not have an institution.");
+            }
+            List<User> signingOfficials = userService.findSOsByInstitutionId(user.getInstitutionId());
+            return Response.ok().entity(signingOfficials).build();
```
rushtong: I'm nervous about the security impact here. We're opening up a lot of information about the signing official to all researchers. Let's see if we can trim this down to a list of display names and ids instead of full objects.

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -162,6 +163,23 @@ public Response delete(@PathParam("email") String email, @Context UriInfo info)
         return Response.ok().build();
     }
 
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("/SigningOfficials")
```
rushtong: Prefer all lower case urls. Mixed case urls are weirdly supported on windows systems, but since we aren't on windows, it has no practical effect. It is, however, a best practice to use dashes. 
```suggestion
    @Path("/signing-officials")
```

Another implementation approach is to add limited SO info to the `/me` call and not open up a new endpoint. Let's discuss in standup.

rushtong (changes requested):
See comments inline, I think we need more design work for this task.

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
raejohanek: That's a good idea, display names and user ids? or we could do display names and emails, or just display names, if we don't want to possibly expose the userId. I can just update the underlying query to only select for those fields and the rest of the fields will be null. Does that sound good? I can also create a new object in the service class if we don't want to pass along the null values

rushtong (approved):
Thank you for the changes, much better approach

PR #1140, opened 2021-06-30
DUOS-1329[risk=no]Consent: Update User include updates to status and rationale

raejohanek:
SCOPE:
- check for updates to status and rationale in user service method
- include status and rationale as arguments in updateUser dao call
- deprecate updateStatus and updateRationale methods in user service and user dao
- update tests for changes and add checks for updated status and rationale to userDAO test

This PR allows the UI to use the DACUser.update method exclusively (accomplished in DUOS-1314) and allows for the now unused code paths to be deleted (accomplished in DUOS-1315) thus this PR must be merged prior to those two PRs.

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1329

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/UserService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/UserService.java
@@ -1,8 +1,6 @@
 package org.broadinstitute.consent.http.service;
 
 import com.google.inject.Inject;
-
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -156,8 +156,19 @@ public User updateDACUserById(Map<String, User> dac, Integer id) throws IllegalA
...
+        if (Objects.isNull(updatedUser.getStatus())) {
+            updatedUser.setStatus(existingUser.getStatus());
+        }
+
+        Integer statusId = RoleStatus.getValueByStatus(updatedUser.getStatus());
+
+        if (Objects.isNull(updatedUser.getRationale())) {
+            updatedUser.setRationale(existingUser.getRationale());
+        }
+
```
rushtong, line 168: By merging the separate update status and rationale functions into a single update user code path, we are allowing the user to update their own status and rationale which is something not previously allowed.

rushtong (changes requested):
I think we need to revisit this implementation choice.

PR #1157, opened 2021-07-09
DUOS-1095[risk=no]Move EraCommons Id to User from User Property table

rushtong:
## Scope
- Create new field for eraCommonsId on User
- Add eraCommonsId column to table
- Add this column to all dao calls that get the other columns from dacuser
- Backpopulate the eraCommonsId column on dacuser
- Delete and replace all references to deprecated user fields
- Add userDAO call to update the new column on authentication

## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1095

## Depends On:
https://github.com/DataBiosphere/duos-ui/pull/1168

----

- Verify all tests go green


src/main/resources/changesets/changelog-consent-71.0.xml
```diff
@@ -0,0 +1,26 @@
...
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="71.0" author="grushton">
+        <validCheckSum>ANY</validCheckSum>
+
+        <addColumn tableName="dacuser">
+            <column name="era_commons_id" type="varchar(255)"/>
+        </addColumn>
+
+        <sql>
```
JVThomas, line 22: Didn't realize you could do this with the changelogs. This makes simple backfills easy to do.

JVThomas (approved):
Code looks good, this should make eraCommonsId easier to work with on queries.

src/main/resources/changesets/changelog-consent-71.0.xml
rushtong: Re-reading this, I think the query needs to be tested on each of the databases. It could be the case that the subquery returns multiple rows for a user id. Not likely, but certainly possible given the query.

JVThomas (approved):
The queries look good. I didn't realize `eraCommonsId` was still being used/referenced within UserProperties, it makes sense to run a check on that before doing the same with `nihUsername`.

Ran consent locally and saw the new column populated for users who had values saved on the db.

PR #1161, opened 2021-07-11
[DUOS-1338][risk=no] Use DataUse for translation

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1338

## Changes
* New ontology call to get the plain language translation of a `DataUse`
* Use data use for the consent's translated use restriction when possible
* Handle NPE cases
* Make updates to the translation when possible
* Fixed bug in `ElectionService.submitFinalAccessVoteDataRequestElection` where we were not honoring the vote that was being passed in.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/ElectionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/ElectionService.java
@@ -22,6 +22,7 @@ import org.broadinstitute.consent.http.models.Consent;
 import org.broadinstitute.consent.http.models.Dac;
 import org.broadinstitute.consent.http.models.DataAccessRequest;
 import org.broadinstitute.consent.http.models.DataSet;
+import org.broadinstitute.consent.http.models.DataUse;
 import org.broadinstitute.consent.http.models.DatasetAssociation;
 import org.broadinstitute.consent.http.models.DatasetDetailEntry;
 import org.broadinstitute.consent.http.models.Election;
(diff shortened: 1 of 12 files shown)


src/main/java/org/broadinstitute/consent/http/service/ElectionService.java
```diff
@@ -205,13 +208,6 @@ public Election submitFinalAccessVoteDataRequestElection(Integer electionId, Boo
         }
         Integer userId = dar.getUserId();
         List<LibraryCard> libraryCards = Objects.nonNull(userId) ? libraryCardDAO.findLibraryCardsByUserId(userId) : Collections.emptyList();
-        List<Vote> finalVotes = voteDAO.findFinalVotesByElectionId(electionId);
-        // The first final vote to be submitted is what determines the approval/denial of the election
-        // isApproved represents whether or not a final vote has already been submitted
-        boolean isApproved = finalVotes.stream().
-                filter(Objects::nonNull).
-                filter(v -> Objects.nonNull(v.getVote())).
-                anyMatch(Vote::getVote);
```
rushtong, line 214: This was a bug in the original logic. This method takes an election and a true/false approved value so we know up front whether this election is approved or not. There is the possibility of a race condition as sometimes the final votes had been saved, sometimes they were not, and in those cases, the notifications would not go out.

JVThomas (approved):
Code looks good, translatedUseRestriction updates accordingly so as long as the data use object is present.. Race condition fix makes sense, especially with the more reliable function argument.

PR #1175, opened 2021-07-19
DUOS-1327[risk=no]Add Researcher functionality to Manage V2 endpoint

raejohanek:
SCOPE:
- Modify API in resource class to allow researchers and update exception checks to account for newly supported role
- Modify service method by adding if statement for researchers and returning only their dars
- Create new DAO call that gets dars by user Id
- Create tests for all new and modified methods
- Update api-docs with new supported value in enum

ADDRESSES: 
https://broadworkbench.atlassian.net/browse/DUOS-1327
----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -125,7 +125,7 @@ public class DataAccessRequestResource extends Resource {
     @GET
     @Produces("application/json")
     @Path("/manage/v2")
-    @RolesAllowed({ADMIN, CHAIRPERSON, MEMBER, SIGNINGOFFICIAL})
+    @RolesAllowed({ADMIN, CHAIRPERSON, MEMBER, SIGNINGOFFICIAL, RESEARCHER})
     public Response describeManageDataAccessRequestsV2(@Auth AuthUser authUser, @QueryParam("roleName") Optional<String> roleName) {
         try {
             User user = userService.findUserByEmail(authUser.getName());
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -133,13 +133,13 @@ public class DataAccessRequestResource extends Resource {
             if (Objects.nonNull(roleNameValue)) {
                 boolean valid = UserRoles.isValidRole(roleNameValue);
                 if (valid) {
-                    //if the roleName is SO and the user does not have that role throw an exception
-                    if (roleNameValue.equalsIgnoreCase(UserRoles.SIGNINGOFFICIAL.getRoleName())) {
-                        if (!user.hasUserRole(UserRoles.SIGNINGOFFICIAL)) {
-                           throw new NotFoundException("User: " + user.getDisplayName() + ", " + " does not have Signing Official role.");
-                       }
-                   } else {
-                        throw new BadRequestException("Signing Official Role is the only role supported at this time");
+                    //if the user does not have the given roleName throw NotFoundException
+                    if (!user.hasUserRole(UserRoles.getUserRoleFromName(roleNameValue))) {
+                        throw new NotFoundException("User: " + user.getDisplayName() + ", does not have " + roleName + " role.");
+                    }
+                    //if there is a valid roleName but it is not SO or Researcher then throw an exception
+                    if (!roleNameValue.equals(UserRoles.RESEARCHER.getRoleName()) && !roleNameValue.equals(UserRoles.SIGNINGOFFICIAL.getRoleName())) {
+                        throw new BadRequestException("Unsupported role name: " + roleName);
                     }
                 } else {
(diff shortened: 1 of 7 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -125,21 +125,21 @@ public Response describeManageDataAccessRequests(@Auth AuthUser authUser) {
...
-                   } else {
-                        throw new BadRequestException("Signing Official Role is the only role supported at this time");
+                    //if the user does not have the given roleName throw NotFoundException
+                    if (!user.hasUserRole(UserRoles.getUserRoleFromName(roleNameValue))) {
+                        throw new NotFoundException("User: " + user.getDisplayName() + ", does not have " + roleName + " role.");
+                    }
+                    //if there is a valid roleName but it is not SO or Researcher then throw an exception
+                    if (!roleNameValue.equals(UserRoles.RESEARCHER.getRoleName()) && !roleNameValue.equals(UserRoles.SIGNINGOFFICIAL.getRoleName())) {
+                        throw new BadRequestException("Unsupported role name: " + roleName);
                     }
```
rushtong: This is where if-else cases really break down. As more cases are added/removed/handled/etc., this becomes unwieldy. Better to fail fast in a switch block rather than throw iterate over different cases, which is what a switch block is designed to do. Something like this should clean that logic up a lot:

```
                    UserRoles queriedRole = UserRoles.getUserRoleFromName(roleNameValue);
                    switch (queriedRole) {
                        case ADMIN:
                        case CHAIRPERSON:
                        case MEMBER:
                            throw new BadRequestException("Unsupported role name: " + roleNameValue);
                        default:
                            break;
                    }

```
Caveat, there are some more optimizations to be made with the above code, but this is a lot easier to read than the multi-case if statement.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -155,6 +155,10 @@ public Integer getTotalUnReviewedDars(AuthUser authUser) {
                       + "is not associated with an Institution.");
                 }
             }
+            if (roleName.equalsIgnoreCase(UserRoles.RESEARCHER.getRoleName())) {
+                    List<DataAccessRequest> dars = dataAccessRequestDAO.findAllDarsByUserId(user.getDacUserId());
+                    return createAccessRequestManageV2(dars);
+            }
```
rushtong: Similar comment here. Instead of passing in a string, if we passed in the UserRoles enum, we would be able to better use switch blocks on the enum and clean up the logic.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -155,6 +155,10 @@ public Integer getTotalUnReviewedDars(AuthUser authUser) {
...
         }
         //if there is no roleName then user is a member, chair, or admin
         List<DataAccessRequest> allDars = findAllDataAccessRequests();
```
rushtong: So, this method actually DOES support member/chair/admin but the null case has semantic meaning. A better approach to this method might be to handle the member/chair/admin cases all in the same switch block since they are processed using the same code.

rushtong:
@raejohanek This looks like a real failing test:
```
Error:  Failures: 
Error:    DataAccessRequestDAOTest.testFindAllDarsByUserId:61
```

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -125,27 +125,37 @@ public Response describeManageDataAccessRequests(@Auth AuthUser authUser) {
...
+                if (Objects.isNull(queriedUserRole)) {
                     throw new BadRequestException("Invalid role name: " + roleNameValue);
+                } else {
+                    //if there is a valid roleName but it is not SO or Researcher then throw an exception
+                    if (queriedUserRole != UserRoles.RESEARCHER && queriedUserRole != UserRoles.SIGNINGOFFICIAL) {
+                        throw new BadRequestException("Unsupported role name: " + roleName);
+                    }
+                    //if the user does not have the given roleName throw NotFoundException
+                    if (!user.hasUserRole(UserRoles.getUserRoleFromName(roleNameValue))) {
+                        throw new NotFoundException("User: " + user.getDisplayName() + ", does not have " + roleName + " role.");
```
rushtong: This comes through as:
>   "message": "User: Gregory Rushton, does not have Optional[SigningOfficial] role.",

Due to `roleName` being an `Optional`. Should dereference it with `roleName.get()`
Same thing on line 141.

rushtong: Even better, refer to the actual enum:
```
throw new BadRequestException("Unsupported role name: " + queriedUserRole.getRoleName());
```

rushtong (approved):
Minor error case suggestions inline that shouldn't hold up the PR. Looks good 👍🏽

rushtong (approved):
👍🏽

PR #1178, opened 2021-07-20
DUOS-1335[risk=no]Add Library Cards to findUserById response

JVThomas:
Addresses [DUOS-1335](https://broadworkbench.atlassian.net/browse/DUOS-1335) (back-end component of the ticket)

PR adds library cards to the User object within the `findUserById` method for the front-end response. PR also includes updates to the User class and associated tests.

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/UserService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/UserService.java
@@ -84,6 +84,8 @@ public class UserService {
         if (user == null) {
             throw new NotFoundException("Unable to find user with id: " + id);
         }
+        List<LibraryCard> cards = libraryCardDAO.findLibraryCardsByUserId(user.getDacUserId());
+        user.setLibraryCards(cards);
         return user;
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -84,6 +84,8 @@ public User findUserById(Integer id) throws NotFoundException {
         if (user == null) {
             throw new NotFoundException("Unable to find user with id: " + id);
         }
+        List<LibraryCard> cards = libraryCardDAO.findLibraryCardsByUserId(user.getDacUserId());
+        user.setLibraryCards(cards);
```
rushtong, line 88: This works, but it would be more efficient to pull the LC info from the `userDAO.findUserById(id);` call so we don't have to make the extra db query.

rushtong: Conversation resolved in slack. The nested institution subqueries make this onerous.

rushtong (approved):
👍🏽

PR #1190, opened 2021-07-23
[DUOS-1361][risk=no] Sort institution on name by default

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1361

## Changes
* Sort on institution name, asc, by default
* Fix NPE when a database error occurs and we have not captured the vendor code in a map

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/Resource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/Resource.java
@@ -1,22 +1,5 @@
 package org.broadinstitute.consent.http.resources;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.sql.SQLException;
-import java.sql.SQLSyntaxErrorException;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.AbstractMap;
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/consent/http/resources/Resource.java
```diff
@@ -153,7 +153,9 @@ private static Response unableToExecuteExceptionHandler(Exception e) {
         try {
             if (e.getCause() instanceof PSQLException) {
                 String vendorCode = ((PSQLException) e.getCause()).getSQLState();
-                status = vendorCodeStatusMap.get(vendorCode);
+                if (vendorCodeStatusMap.containsKey(vendorCode)) {
+                    status = vendorCodeStatusMap.get(vendorCode);
+                }
```
rushtong, line 158: If we didn't have a vendor code in the map, this would reset the status to null which triggered an NPE in the returned response object.

raejohanek (approved):
tested locally 👍 institutions return in alphabetical order

JVThomas (approved):

PR #1191, opened 2021-07-26
DUOS-763[risk=no]Dar Collection Modelling

raejohanek:
SCOPE:
- Create darCollection model
- - I used timestamp instead of date for createDate and updateDate
- - I included the entire User object for createUser and updateUser on the model, but only the userIds are in the table, so population of these fields will require userDAO calls in the mapper/reducer or in the dao call itself
- Create darCollection table in database
- Add collection_id column to data access request table in database
- Add collection_id to data access request dao calls

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-763

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/mapper/DataAccessRequestMapper.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/mapper/DataAccessRequestMapper.java
@@ -21,7 +21,9 @@ public class DataAccessRequestMapper implements RowMapper<DataAccessRequest>, Ro
         DataAccessRequest dar = new DataAccessRequest();
         dar.setId(resultSet.getInt("id"));
         dar.setReferenceId(resultSet.getString("reference_id"));
-        dar.setCollectionId(resultSet.getInt("collection_id"));
+        if (hasColumn(resultSet, "collection_id")) {
+            dar.setCollectionId(resultSet.getInt("collection_id"));
+        }
         dar.setDraft(resultSet.getBoolean("draft"));
         dar.setUserId(resultSet.getInt("user_id"));
         dar.setCreateDate(resultSet.getTimestamp("create_date"));
--- a/src/main/java/org/broadinstitute/consent/http/db/mapper/DataAccessRequestMapper.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/mapper/DataAccessRequestMapper.java
@@ -21,6 +21,7 @@ public class DataAccessRequestMapper implements RowMapper<DataAccessRequest>, Ro
         DataAccessRequest dar = new DataAccessRequest();
         dar.setId(resultSet.getInt("id"));
         dar.setReferenceId(resultSet.getString("reference_id"));
+        dar.setCollectionId(resultSet.getInt("collection_id"));
         dar.setDraft(resultSet.getBoolean("draft"));
         dar.setUserId(resultSet.getInt("user_id"));
         dar.setCreateDate(resultSet.getTimestamp("create_date"));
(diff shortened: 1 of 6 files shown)


src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
```diff
@@ -0,0 +1,77 @@
...
+public class DarCollection {
+
+  @JsonProperty
+  Integer darCollectionId;
+
+  @JsonProperty
+  String darCode;
+
+  @JsonProperty
+  Timestamp createDate;
```
rushtong, line 16: These should be dates for consistency

src/main/resources/changesets/changelog-consent-72.0.xml
```diff
@@ -0,0 +1,40 @@
...
+            <column name="update_user" type="timestamp"/>
+        </createTable>
+
+        <rollback>
+            <dropColumn tableName="data_access_request" columnName="collection_id"/>
+        </rollback>
+
+        <rollback>
+            <dropTable tableName="dar_collection"/>
+        </rollback>
```
rushtong: These can be in the same `rollback` block.

src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
raejohanek: Okay I can make that change! but I'm wondering what you mean, consistency with what? Because the data access requests use timestamps

rushtong: You are correct, we are actually all over the place. `Dac`, `Institution`, `Vote`, `Dataset`, `User`, `Match`, `Election`, and others all use `Date`, but `Consent` and `DataAccessRequest` use `Timestamp`. 

I'm not sure if we ever had an intentional approach to the actual class, but reading up on it, both are flawed and [`Instant`](https://docs.oracle.com/javase/10/docs/api/java/time/package-summary.html) is [now the preferred time storage class](https://www.baeldung.com/migrating-to-java-8-date-time-api). Let's stick with `Timestamp` for now and we can file a ticket to explore looking at migrating all of these fields to `Instant` if that makes more sense.

src/main/java/org/broadinstitute/consent/http/db/mapper/DataAccessRequestMapper.java
```diff
@@ -21,6 +21,7 @@ public DataAccessRequest map(ResultSet resultSet, StatementContext statementCont
         DataAccessRequest dar = new DataAccessRequest();
         dar.setId(resultSet.getInt("id"));
         dar.setReferenceId(resultSet.getString("reference_id"));
+        dar.setCollectionId(resultSet.getInt("collection_id"));
```
rushtong: This could be written more safely if we missed a query. Look at the `hasColumn` method where we could check if the query has the field, then we would set the value.

src/main/resources/changesets/changelog-consent-72.0.xml
```diff
@@ -0,0 +1,41 @@
...
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="72.0" author="raejohanek">
+        <validCheckSum>ANY</validCheckSum>
+
+        <addColumn tableName="data_access_request">
+            <column name="collection_id" type="bigint"/>
+        </addColumn>
```
rushtong, line 12: We might consider a FK constraint on this to point to the `dar_collection` table.

rushtong (approved):
Nice work 👍🏽

PR #1192, opened 2021-07-26
[DUOS-1364][risk=no] Fix thread safety

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1364

Note to reviewers, it is recommended to hide whitespace changes.

## Changes
Update the query to update + select the max value so we don't have to run two queries to get the max count. Integrating them into a single query makes it a single transaction and protects against simultaneous calls making out-of-order updates+selects.

See also: https://www.postgresql.org/docs/9.3/queries-with.html#QUERIES-WITH-MODIFYING

----

- Verify all tests go green


raejohanek (commented):
this looks good! I think this approach makes more sense than the one we talked about earlier. Now, getMaxCountByName is unused, except for in tests, so we could probably delete that and update the incrementCountByName tests as well

raejohanek (approved):
tests look good, updated dao call works as expected

PR #1197, opened 2021-07-29
DUOS-1360[risk=no] Manage Institutes: Add SOs to page

raejohanek:
## SCOPE:
- This PR updates the findAllInstitutions DAO call to include the SOs for each institution if they have them. 

- It also adds a test for institutionDAO.findAllInstitutions() for institutions that have SOs, and updates the api-docs to describe the new functionality.

- This PR uses simplifiedUser to represent the SOs using a limited amount of info, so it requires some changes to that class in order to get the class to work in the bean mapper.
- - classes must have setters and thus cannot be final

- updates API docs to include information about the included users (create, update, and so users)

- This PR must be merged prior to a corresponding PR DUOS-1360 on the UI that updates the handling of the updated API response on the manageInstitutes page.

## ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1360

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/models/Institution.java
```diff
@@ -10,11 +13,11 @@
   private String name;
   private String itDirectorName;
   private String itDirectorEmail;
+  private List<User> signingOfficials;
```
rushtong: Could this be `SimplifiedUser`?

src/main/java/org/broadinstitute/consent/http/db/mapper/InstitutionWithUsersReducer.java
```diff
@@ -58,6 +58,15 @@ public void accumulate(Map<Integer, Institution> map, RowView rowView) {
 
     institution.setCreateUser(create_user);
     institution.setUpdateUser(update_user);
+
+    if (Objects.nonNull(rowView.getColumn("so_dacuserid", Integer.class))) {
+      User so_user = new User();
+      so_user.setDacUserId(rowView.getColumn("so_dacuserid", Integer.class));
+      so_user.setEmail(rowView.getColumn("so_email", String.class));
+      so_user.setDisplayName(rowView.getColumn("so_displayname", String.class));
```
rushtong: Would be safer to have column checks on `so_email` and `so_displayname` as well. 

Another thought ... if we used `SimplifiedUser` here instead of `User`, then added a BeanMapper for that class on the DAO method, we could simplify this and let jdbi handle the class construction from whatever row values existed.

rushtong (approved):
Some minor comments inline, but this looks good. Your call if you want to do any refactoring.

JVThomas (approved):
Back-end institution call attaches simplified SO data to institutions (where relevant) within response.

PR #1212, opened 2021-08-05
DUOS-1113[risk=medium]We need a delete user role API

raejohanek:
SCOPE:
- Create service method that simply calls pre-existing DAO method
- Create new API, following structure of addRole API
- - Create helpers in UserRoles so that logic and code are not repeated in these two APIs
- - Create Tests
- - Add to api-docs

ADDRESSES:
- https://broadworkbench.atlassian.net/browse/DUOS-1113

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
+++ b/src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
@@ -1,12 +1,8 @@
 package org.broadinstitute.consent.http.enumeration;
 
-import org.broadinstitute.consent.http.models.User;
-import org.broadinstitute.consent.http.models.UserRole;
 import org.broadinstitute.consent.http.resources.Resource;
 
-import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.stream.Collectors;
 
 import java.util.EnumSet;
 import java.util.List;
@@ -70,13 +66,6 @@ public enum UserRoles {
         List<Integer> listOfNonDacRoleIds = Arrays.asList(
           ALUMNI.getRoleId(), ADMIN.getRoleId(), RESEARCHER.getRoleId(), DATAOWNER.getRoleId(), SIGNINGOFFICIAL.getRoleId());
         return !Objects.isNull(roleId) && listOfNonDacRoleIds.contains(roleId);
-    }
-
-    public static List<Integer> getUserRoleIdsFromUser(User user) {
-        return user.getRoles()
-          .stream()
-          .map(UserRole::getRoleId)
-          .collect(Collectors.toList());
     }
 
 }
--- a/src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
+++ b/src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
@@ -3,6 +3,9 @@ package org.broadinstitute.consent.http.enumeration;
 import org.broadinstitute.consent.http.models.User;
 import org.broadinstitute.consent.http.models.UserRole;
 import org.broadinstitute.consent.http.resources.Resource;
+
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.stream.Collectors;
 
 import java.util.EnumSet;
@@ -64,10 +67,9 @@ public enum UserRoles {
     }
 
     public static boolean isValidNonDACRoleId(Integer roleId) {
-        if (Objects.isNull(roleId) || roleId > 7 || roleId < 3) {
-            return false;
-        }
-        return true;
+        List<Integer> listOfNonDacRoleIds = Arrays.asList(
+          ALUMNI.getRoleId(), ADMIN.getRoleId(), RESEARCHER.getRoleId(), DATAOWNER.getRoleId(), SIGNINGOFFICIAL.getRoleId());
+        return !Objects.isNull(roleId) && listOfNonDacRoleIds.contains(roleId);
     }
 
     public static List<Integer> getUserRoleIdsFromUser(User user) {
--- a/src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
+++ b/src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
@@ -1,8 +1,12 @@
 package org.broadinstitute.consent.http.enumeration;
 
+import org.broadinstitute.consent.http.models.User;
+import org.broadinstitute.consent.http.models.UserRole;
 import org.broadinstitute.consent.http.resources.Resource;
+import java.util.stream.Collectors;
 
 import java.util.EnumSet;
+import java.util.List;
 import java.util.Objects;
 
 public enum UserRoles {
@@ -57,6 +61,20 @@ public enum UserRoles {
           .stream()
           .map(UserRoles::getRoleName)
           .anyMatch(roleName::equalsIgnoreCase);
+    }
+
+    public static boolean isValidRoleId(Integer roleId) {
+        if (Objects.isNull(roleId) || roleId > 7 || roleId < 1) {
+            return false;
+        }
+        return true;
+    }
+
+    public static List<Integer> getUserRoleIdsFromUser(User user) {
+        return user.getRoles()
+          .stream()
+          .map(UserRole::getRoleId)
+          .collect(Collectors.toList());
     }
 
 }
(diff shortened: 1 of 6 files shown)


src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
```diff
@@ -59,4 +63,18 @@ public static boolean isValidRole(String roleName) {
           .anyMatch(roleName::equalsIgnoreCase);
     }
 
+    public static boolean isValidNonDACRoleId(Integer roleId) {
+        if (Objects.isNull(roleId) || roleId > 7 || roleId < 3) {
```
JVThomas: It's better to check for specific roles (could be a constant or an enum) rather than make general assumptions on id values. We don't know if future roles (if any) will follow the above trend.

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -148,6 +137,29 @@ public Response addRoleToUser(@Auth AuthUser authUser, @PathParam("userId") Inte
...
+    @RolesAllowed({ADMIN})
+    public Response deleteRoleFromUser(@Auth AuthUser authUser, @PathParam("userId") Integer userId, @PathParam("roleId") Integer roleId) {
+        try {
+            User user = userService.findUserById(userId);
+            if (!UserRoles.isValidNonDACRoleId(roleId)) {
+                throw new BadRequestException("Role Id must be an Integer representing an allowed role (3 to 7)");
+            }
+            List<Integer> currentUserRoleIds = UserRoles.getUserRoleIdsFromUser(user);
+            if (!currentUserRoleIds.contains(roleId)) {
+                return Response.notModified().build();
```
JVThomas: 304 relates to endpoints and caching, it's not used to describe an unchanged model or entity. Use a 404 with a message indicating the role was not found, which is safe to do here since it's an Admin only endpoint.

JVThomas (dismissed):

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
raejohanek: I am following the precedent set by the AddRole API which uses the not modified response. Do you think I should change them both then for consistency? It looks like the add role endpoint was written by greg about a year ago

rushtong: Will look into this soon!

rushtong:
This is a Medium risk ticket since we're adding an API.

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
rushtong: @raejohanek Justin is correct, we're using the 304 incorrectly in other places. Those should be updated to be a 200 response. Will file a ticket for that.

src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
```diff
@@ -59,4 +66,17 @@ public static boolean isValidRole(String roleName) {
...
+          ALUMNI.getRoleId(), ADMIN.getRoleId(), RESEARCHER.getRoleId(), DATAOWNER.getRoleId(), SIGNINGOFFICIAL.getRoleId());
+        return !Objects.isNull(roleId) && listOfNonDacRoleIds.contains(roleId);
+    }
+
+    public static List<Integer> getUserRoleIdsFromUser(User user) {
+        return user.getRoles()
+          .stream()
+          .map(UserRole::getRoleId)
+          .collect(Collectors.toList());
+    }
```
rushtong: This seems more like a `User` helper method than a `UserRoles` helper method.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -201,6 +201,10 @@ public void updateEmailPreference(boolean preference, Integer userId) {
         return users.stream().map(SimplifiedUser::new).collect(Collectors.toList());
     }
 
+    public void deleteUserRole(Integer dacUserId, Integer roleId) {
+        userRoleDAO.removeSingleUserRole(dacUserId, roleId);
+    }
```
rushtong: Let's add some logging here, this is something we need to do more of. Let the method take a User as the user doing the action, and log (INFO) something like:
```
User 'Gregory Rushton' deleting Role Id: roleId from User ID: dacUserId
```

src/main/resources/assets/api-docs.yaml
```diff
@@ -3336,6 +3336,46 @@ paths:
...
+            The id of the role. Members and Chairpersons cannot be deleted this way.
+            See DAC endpoints for those roles.
+            ALUMNI: 3
+            ADMIN: 4
+            RESEARCHER: 5
+            DATAOWNER: 6
+            SIGNINGOFFICIAL: 7
+          required: true
+          schema:
+            type: number
```
rushtong, line 3362: I understand that this is essentially the same as the add role API, but we can make this better in both places. This could be an enum with a selectable value from the UI. I think this will work, but haven't tested:
```
          schema:
            type: number
            nullable: false
            enum:
              - 3
              - 4
              - 5
              - 6
              - 7
```

src/main/resources/assets/api-docs.yaml
```diff
@@ -3336,6 +3336,46 @@ paths:
...
+        - User
+      responses:
+        200:
+          description: Updated user along with researcher properties and library cards
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/User'
+        304:
+          description: Not modified, User does not have given role
```
rushtong: rm

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
rushtong: Added https://broadworkbench.atlassian.net/browse/DUOS-1393 to cover the add role endpoint.

rushtong (changes requested):
Looking good!

src/main/java/org/broadinstitute/consent/http/models/User.java
```diff
@@ -362,4 +363,11 @@ public boolean hasUserRole(UserRoles role) {
         }
     }
 
+    public static List<Integer> getUserRoleIdsFromUser(User user) {
+        return user.getRoles()
+          .stream()
+          .map(UserRole::getRoleId)
+          .collect(Collectors.toList());
+    }
```
rushtong: This doesn't need the user ... it already is the user. 

```
    public List<Integer> getUserRoleIds() {
        if (Objects.isNull(this.getRoles())) {
        } else {
            return this.getRoles()
                    .stream()
                    .map(UserRole::getRoleId)
        }
    }
```

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -148,6 +137,31 @@ public Response addRoleToUser(@Auth AuthUser authUser, @PathParam("userId") Inte
...
 
+    @DELETE
+    @Path("/{userId}/{roleId}")
+    @Produces("application/json")
+    @RolesAllowed({ADMIN})
+    public Response deleteRoleFromUser(@Auth AuthUser authUser, @PathParam("userId") Integer userId, @PathParam("roleId") Integer roleId) {
+        try {
+            User user = userService.findUserById(userId);
+            if (!UserRoles.isValidNonDACRoleId(roleId)) {
+                throw new BadRequestException("Role Id must be an Integer representing an allowed role (3 to 7)");
```
rushtong: Let's simplify this message - we clearly spell out the requirements in Swagger.
```suggestion
                throw new BadRequestException("Invalid Role Id");
```

rushtong (changes requested):
Two minor suggestions, looking better!

rushtong:
Still seeing [failed tests](https://github.com/DataBiosphere/consent/pull/1212/checks?check_run_id=3301661534), but they're passing for me locally.

raejohanek:
> Still seeing [failed tests](https://github.com/DataBiosphere/consent/pull/1212/checks?check_run_id=3301661534), but they're passing for me locally.

@rushtong I was just going to say the same thing! I'm not sure why they pass locally and fail here, any ideas?

rushtong (approved):
Thank you for the changes!

PR #1220, opened 2021-08-12
DUOS-1137[risk=med] API adjustments for SO Console

JVThomas:
Back-end adjustments that address [DUOS-1137](https://broadworkbench.atlassian.net/browse/DUOS-1137)

PR adds new endpoints and adjusts permissions and/or processing steps of some Admin LC cards so that it can also be used by Signing Officials (for the purposes of their console on the front-end).

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/db/mapper/UserWithRolesReducer.java
```diff
@@ -44,6 +51,19 @@ public void accumulate(Map<Integer, User> map, RowView rowView) {
...
     }
+    //user role join can cause duplication of data if done in tandem with joins on other tables
+    //ex) The same LC can end up being repeated multiple times
+    //Below only adds LC if not currently saved on the array
+    try {
+      if(Objects.nonNull(rowView.getColumn("lc_id", Integer.class))) {
+        LibraryCard lc = rowView.getRow(LibraryCard.class);
+        if(!user.getLibraryCards().stream().anyMatch(card -> card.getId() == lc.getId())) {
+          user.getLibraryCards().add(lc);
+        }
```
rushtong: We started, at one point, to use the add to list pattern which obviates the need for initialization in lines 26-29. See the User addRole and addProperty methods for examples.

src/main/java/org/broadinstitute/consent/http/db/mapper/UsersAndCardsReducer.java
```diff
@@ -0,0 +1,82 @@
...
+import org.broadinstitute.consent.http.enumeration.RoleStatus;
+import org.broadinstitute.consent.http.models.LibraryCard;
+import org.broadinstitute.consent.http.models.Institution;
+import org.broadinstitute.consent.http.models.User;
+import org.broadinstitute.consent.http.models.UserRole;
+import org.jdbi.v3.core.mapper.MappingException;
+import org.jdbi.v3.core.result.LinkedHashMapRowReducer;
+import org.jdbi.v3.core.result.RowView;
+
+public class UsersAndCardsReducer implements LinkedHashMapRowReducer<String, User> {
```
rushtong: maybe this is better defined as `CardsWithUserReducer`? The primary point being to collect library cards and users, if they exist. 

update: nevermind ... after rethinking this, I think you're on the right track here.

src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
```diff
@@ -180,7 +192,14 @@ private LibraryCard processUserOnNewLC(LibraryCard card) throws Exception {
...
+            card.setUserName(user.getDisplayName());
         }
         return card;
     }
+
+    private Boolean checkIsAdmin(User user) {
+        return user.getRoles()
+            .stream()
+            .anyMatch(role -> role.getName().equalsIgnoreCase(UserRoles.ADMIN.getRoleName()));
+    }
```
rushtong, line 200: This method also exists in `InstitutionUtil` and similar methods in `DacService`. We should start collecting those into a user utility. Maybe not for this PR due to its size - a new ticket seems appropriate.

src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
```diff
@@ -66,9 +73,14 @@ public LibraryCard updateLibraryCard(LibraryCard libraryCard, Integer id, Intege
...
-        LibraryCard libraryCard = this.libraryCardDAO.findLibraryCardById(id);
-        throwIfNull(libraryCard);
+    public void deleteLibraryCardById(Integer id, User user) {
+        Boolean isAdmin = checkIsAdmin(user);
+        LibraryCard card = findLibraryCardById(id);
+        throwIfNull(card);
+        //If user is not an admin and LC institutionID doesn't match the users's throw a ForbiddenException
+        if(!isAdmin && !card.getInstitutionId().equals(user.getInstitutionId())) {
+            throw new ForbiddenException("You are not authorized to delete this library card");
+        }
```
rushtong: We should move the authorization logic in the resource layer which is currently where we enforce these kinds of restrictions. At this layer, we're just concerned with business logic.

src/main/resources/assets/api-docs.yaml
```diff
@@ -3224,6 +3224,21 @@ paths:
...
+      description: Find users that do not belong to an institution
+      tags: 
+        - User
+      responses: 
+        200:
+          description: An array of users that do not belong to an institution
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/User'
```
rushtong: nit, this should be an array of users, not a single user:
```suggestion
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/User'
```

src/test/java/org/broadinstitute/consent/http/service/LibraryCardServiceTest.java
```diff
@@ -252,13 +272,28 @@ public void testUpdateLibraryCard_InvalidInstitution() {
...
+        Institution institution = testInstitution();
+        Integer modifiedId = institution.getId() + 1;
+        User user = testUser(institution.getId());
+        UserRole soRole = new UserRole(UserRoles.SIGNINGOFFICIAL.getRoleId(), UserRoles.SIGNINGOFFICIAL.getRoleName());
+        user.addRole(soRole);
+        LibraryCard libraryCard = testLibraryCard(modifiedId, user.getDacUserId());
+        when(libraryCardDAO.findLibraryCardById(any())).thenReturn(libraryCard);
+        initService();
+        service.deleteLibraryCardById(libraryCard.getId(), user);
     }
```
rushtong: Once this logic is in the resource class, this test should move to the resource class test.

src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
```diff
@@ -171,12 +172,57 @@ void updateUser(@Bind("displayName") String displayName,
     @RegisterBeanMapper(value = User.class)
     @RegisterBeanMapper(value = UserRole.class)
     @UseRowReducer(UserWithRolesReducer.class)
-    @SqlQuery("SELECT du.*, r.name, ur.role_id, ur.user_role_id, ur.dac_id " +
+    @SqlQuery("SELECT du.*, r.name, ur.role_id, ur.user_role_id, ur.dac_id " + " FROM dacuser du "
+            + " LEFT JOIN user_role ur ON ur.user_id = du.dacuserid " + " LEFT JOIN roles r ON r.roleid = ur.role_id "
+            + " WHERE du.institution_id = :institutionId")
+    List<User> findUsersByInstitution(@Bind("institutionId") Integer institutionId);
```
rushtong: nit: break these up one line per query segment so its easier to read. 
```suggestion
    @SqlQuery("SELECT du.*, r.name, ur.role_id, ur.user_role_id, ur.dac_id "
            + " FROM dacuser du "
            + " LEFT JOIN user_role ur ON ur.user_id = du.dacuserid "
            + " LEFT JOIN roles r ON r.roleid = ur.role_id "
            + " WHERE du.institution_id = :institutionId")
    List<User> findUsersByInstitution(@Bind("institutionId") Integer institutionId);
```

src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
```diff
@@ -171,12 +172,57 @@ void updateUser(@Bind("displayName") String displayName,
...
+                " lc.id AS lc_id , lc.user_id AS lc_user_id, lc.institution_id AS lc_institution_id, " +
+                " lc.era_commons_id AS lc_era_commons_id, lc.user_name AS lc_user_name, lc.user_email AS lc_user_email, " +
+                " lc.create_user_id AS lc_create_user_id, lc.create_date AS lc_create_date, " +
+                " lc.update_user_id AS lc_update_user_id " +
+                " FROM dacuser du " +
+                " LEFT JOIN user_role ur ON ur.user_id = du.dacuserid " +
+                " LEFT JOIN roles r ON r.roleid = ur.role_id " +
+                " LEFT JOIN library_card lc ON lc.user_id = du.dacuserid AND lc.institution_id = :institutionId " +
+                " WHERE du.institution_id = :institutionId")
+    List<User> getUsersAndCardsForSO(@Bind("institutionId") Integer institutionId);
```
rushtong: I have mixed opinions on this query. It will be challenging to maintain over time. I like that is efficient, but that efficiency obscures clarity. I can see this as two separate queries merged in the service class.

src/main/java/org/broadinstitute/consent/http/resources/LibraryCardResource.java
```diff
@@ -79,10 +79,10 @@ public Response getLibraryCardsByInstitutionId(@Auth AuthUser authUser, @PathPar
   @RolesAllowed({ADMIN, SIGNINGOFFICIAL})
   public Response createLibraryCard(@Auth AuthUser authUser, String libraryCard) {
     try{
-      User admin = userService.findUserByEmail(authUser.getName());
+      User user = userService.findUserByEmail(authUser.getName());
```
rushtong: nice change

rushtong (changes requested):
Works well! See comments inline - there are a couple of updates that I think are important like moving the authorization logic and potentially splitting up the uber-query.

src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
```diff
@@ -171,12 +172,69 @@ void updateUser(@Bind("displayName") String displayName,
...
+        + " lc.id AS lc_id , lc.user_id AS lc_user_id, lc.institution_id AS lc_institution_id, "
+        + " lc.era_commons_id AS lc_era_commons_id, lc.user_name AS lc_user_name, lc.user_email AS lc_user_email, "
+        + " lc.create_user_id AS lc_create_user_id, lc.create_date AS lc_create_date, "
+        + " lc.update_user_id AS lc_update_user_id " 
+    + " FROM dacuser du"
+    + " LEFT JOIN user_role ur ON ur.user_id = du.dacuserid " 
+    + " LEFT JOIN roles r ON r.roleid = ur.role_id "
+    + " INNER JOIN library_card lc ON lc.user_id = du.dacuserid "
+    + " WHERE (du.institution_id != :institutionId OR du.institution_id IS NULL) AND lc.institution_id = :institutionId")
+    List<User> getUsersOutsideInstitutionWithCards(@Bind("institutionId") Integer institutionId);
```
rushtong, line 203: I like the naming here, well thought out

src/main/java/org/broadinstitute/consent/http/db/mapper/UserWithRolesReducer.java
```diff
@@ -44,6 +46,19 @@ public void accumulate(Map<Integer, User> map, RowView rowView) {
...
+    //Below only adds LC if not currently saved on the array
+    try {
+      if(Objects.nonNull(rowView.getColumn("lc_id", Integer.class))) {
+        LibraryCard lc = rowView.getRow(LibraryCard.class);
+        if(Objects.isNull(user.getLibraryCards()) || !user.getLibraryCards().stream().anyMatch(card -> card.getId() == lc.getId())) {
+          user.addLibraryCard(lc);
+        }
+      }
+    } catch(MappingException e) {
+      //Ignore exceptions here, user may not have a library card issued under this instiution
```
rushtong, line 60: Not for this PR, but I think we should start logging these as errors so we can follow up on them.

src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
```diff
@@ -27,10 +29,15 @@ public LibraryCardService(LibraryCardDAO libraryCardDAO, InstitutionDAO institut
...
 
-    public LibraryCard createLibraryCard(LibraryCard libraryCard) throws Exception {
+    public LibraryCard createLibraryCard(LibraryCard libraryCard, User user) throws Exception {
         throwIfNull(libraryCard);
-        checkForValidInstitution(libraryCard.getInstitutionId());
+        Boolean isAdmin = checkIsAdmin(user);
+        //If user is not an admin, use user's institutionId rather than the value provided in the payload
+        if (!isAdmin) {
+            libraryCard.setInstitutionId(user.getInstitutionId());
+        }
```
rushtong: This is a nice check, but it might be better to throw a bad request exception here. Just thinking out loud here.

rushtong (approved):
Thank you for the changes, this looks good 👍🏽

PR #1224, opened 2021-08-13
DUOS-1252[risk=no]DarCollection DAO classes

raejohanek:
SCOPE: 
- This adds findAll, findByCollectionId, findByReferenceId, Insert, Delete, and Update  in DarCollection DAO 
- - Adds tests
- This adds delete By CollectionId and Insert Dar with collectionId DataAccessRequestDAO
- - Adds tests
- This Modifies DataAccessRequestService to update the dar collection when the dar is updated 
- - Adds test
- Adds helper in RowMapperHelper and replaces repeated code in Dar Mapper and Dar Data Mapper
- Adds a changeset that fixes an error made in changeset 72

UPDATE:
- Adds `findDarCollectionsCreatedByUserId` with associated tests
- Adds `findAllDarCollectionsWithFilters` with associated tests

ADDRESSES:
https://broadworkbench.atlassian.net/browse/DUOS-1252

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DAOContainer.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DAOContainer.java
@@ -13,6 +13,7 @@ public class DAOContainer {
   private CounterDAO counterDAO;
   private DacDAO dacDAO;
   private DataAccessRequestDAO dataAccessRequestDAO;
+  private DarCollectionDAO darCollectionDAO;
   private DatasetAssociationDAO datasetAssociationDAO;
   private DatasetDAO datasetDAO;
   private ElectionDAO electionDAO;
@@ -74,6 +75,15 @@ public class DAOContainer {
   public void setDataAccessRequestDAO(
       DataAccessRequestDAO dataAccessRequestDAO) {
     this.dataAccessRequestDAO = dataAccessRequestDAO;
+  }
+
+  public DarCollectionDAO getDarCollectionDAO() {
+    return darCollectionDAO;
+  }
+
+  public void setDarCollectionDAO(
+    DarCollectionDAO darCollectionDAO) {
+    this.darCollectionDAO = darCollectionDAO;
   }
 
   public DatasetAssociationDAO getDatasetAssociationDAO() {
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
@@ -17,20 +17,20 @@ import java.util.List;
 public interface DarCollectionDAO {
 
   final String getCollectionAndDars =
-      "SELECT c.*, dar.id AS dar_id, dar.reference_id AS dar_reference_id, dar.collection_id AS dar_collection_id, " +
+      "SELECT c.*, i.institution_name, u.displayname AS researcher, dar.id AS dar_id, dar.reference_id AS dar_reference_id, dar.collection_id AS dar_collection_id, " +
           "dar.draft AS dar_draft, dar.user_id AS dar_userId, dar.create_date AS dar_create_date, " +
           "dar.sort_date AS dar_sort_date, dar.submission_date AS dar_submission_date, " +
           "dar.update_date AS dar_update_date, (dar.data #>> '{}')::jsonb AS data, " +
-          "(dar.data #>> '{}')::jsonb ->> 'institution' as institution, " + 
-          "(dar.data #>> '{}')::jsonb ->> 'researcher' as researcher, " +
           "(dar.data #>> '{}')::jsonb ->> 'projectTitle' as projectTitle " +
       "FROM dar_collection c " + 
+      "INNER JOIN dacuser u ON u.dacuserid = c.create_user " +
+      "LEFT JOIN institution i ON i.institution_id = u.institution_id " +
       "INNER JOIN data_access_request dar ON c.collection_id = dar.collection_id ";
   
   final String filterQuery = 
-    "WHERE (dar.data #>> '{}')::jsonb ->> 'institution' ~* :institutionSearchTerm " +
+    "WHERE COALESCE(i.institution_name, '') ~* :institutionSearchTerm " +
     "AND (dar.data #>> '{}')::jsonb ->> 'projectTitle' ~* :projectSearchTerm " +
-    "AND (dar.data #>> '{}')::jsonb ->> 'researcher' ~* :researcherSearchTerm " +
+    "AND u.displayname ~* :researcherSearchTerm " +
     "AND c.dar_code ~* :darCodeSearchTerm " + 
     "AND EXISTS " +
         "(SELECT FROM jsonb_array_elements((dar.data #>> '{}')::jsonb -> 'datasets') dataset " +
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
@@ -3,6 +3,7 @@ package org.broadinstitute.consent.http.db;
 import org.broadinstitute.consent.http.db.mapper.DarCollectionReducer;
 import org.broadinstitute.consent.http.db.mapper.DataAccessRequestMapper;
 import org.broadinstitute.consent.http.models.DarCollection;
+import org.broadinstitute.consent.http.models.DataAccessRequest;
 import org.jdbi.v3.sqlobject.config.RegisterBeanMapper;
 import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
 import org.jdbi.v3.sqlobject.customizer.Bind;
@@ -20,8 +21,8 @@ public interface DarCollectionDAO {
    *
    * @return List<DarCollection>
    */
-  @RegisterBeanMapper(DarCollection.class)
-  @RegisterRowMapper(DataAccessRequestMapper.class)
+  @RegisterBeanMapper(value = DarCollection.class)
+  @RegisterBeanMapper(value = DataAccessRequest.class)
   @UseRowReducer(DarCollectionReducer.class)
   @SqlQuery(
     "SELECT * FROM dar_collection c "
@@ -48,7 +49,7 @@ public interface DarCollectionDAO {
       + "            submission_date, update_date AS dar_update_date, (data #>> '{}')::jsonb AS data "
       + "    FROM data_access_request) dar "
       + " ON c.collection_id = dar.dar_collection_id ")
-  List<DarCollection> findAllDARCollectionsWithFilters();
+  List<DarCollection> findAllDARCollectionsWithFilters(String sortField, String sortDirection, List<String> filterTerms);
   /**
    * Find the DARCollection and all of its Data Access Requests that contains the DAR with the given referenceId
    *
@@ -99,7 +100,15 @@ public interface DarCollectionDAO {
                             @Bind("createDate") Date createDate);
 
   /**
-   * Delete the DAR Collection with the given colelction ID
+   * Update the update user and update date of the DAR Collection with the given collection ID
+   */
+  @SqlUpdate("UPDATE dar_collection SET update_user = :updateUserId, update_date = :updateDate WHERE collection_id = :collectionId")
+  void updateDarCollection(@Bind("collectionId") Integer collectionId,
+                           @Bind("updateUserId") Integer updateUserId,
+                           @Bind("updateDate") Date updateDate);
+
+  /**
+   * Delete the DAR Collection with the given collection ID
    */
   @SqlUpdate("DELETE FROM dar_collection WHERE collection_id = :collectionId")
   void deleteByCollectionId(@Bind("collectionId") Integer collectionId);
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
@@ -5,6 +5,7 @@ import org.broadinstitute.consent.http.models.DarCollection;
 import org.broadinstitute.consent.http.models.DataAccessRequest;
 import org.jdbi.v3.sqlobject.config.RegisterBeanMapper;
 import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.customizer.Define;
 import org.jdbi.v3.sqlobject.statement.GetGeneratedKeys;
 import org.jdbi.v3.sqlobject.statement.SqlQuery;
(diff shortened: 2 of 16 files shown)


src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
```diff
@@ -0,0 +1,157 @@
...
+
+public interface DarCollectionDAO {
+
+  final String getCollectionAndDars =
+      "SELECT c.*, dar.id AS dar_id, dar.reference_id AS dar_reference_id, dar.collection_id AS dar_collection_id, " +
+          "dar.draft AS dar_draft, dar.user_id AS dar_userId, dar.create_date AS dar_create_date, " +
+          "dar.sort_date AS dar_sort_date, dar.submission_date AS dar_submission_date, " +
+          "dar.update_date AS dar_update_date, (dar.data #>> '{}')::jsonb AS data, " +
+          "(dar.data #>> '{}')::jsonb ->> 'institution' as institution, " + 
+          "(dar.data #>> '{}')::jsonb ->> 'researcher' as researcher, " +
```
rushtong: Shouldn't these values come from the user and not the dar?

src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
```diff
@@ -0,0 +1,157 @@
...
+  /**
+   * Find the DARCollection and all of its Data Access Requests that has the given collectionId
+   *
+   * @return DarCollection
+   */
+  @RegisterBeanMapper(value = DarCollection.class)
+  @RegisterBeanMapper(value = DataAccessRequest.class, prefix = "dar")
+  @UseRowReducer(DarCollectionReducer.class)
+  @SqlQuery(
+      getCollectionAndDars +  " WHERE c.collection_id = dar.collection_id AND c.collection_id = :collectionId ")
```
rushtong: I didn't know this could work, nice find!

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -175,6 +183,32 @@ void insertVersion2(
       @Bind("updateDate") Date updateDate,
       @Bind("data") @Json DataAccessRequestData data);
 
+  /**
+   * Create new DataAccessRequest.
+   * This version supercedes `insertV2`
```
rushtong, line 189: Can we add `@Deprecated` to the v2 method?

src/main/java/org/broadinstitute/consent/http/db/mapper/RowMapperHelper.java
```diff
@@ -27,7 +35,26 @@ default boolean hasColumn(ResultSet rs, String columnName) throws SQLException {
     return false;
   }
 
-  default String unescapeJava(String value) {
+  static String unescapeJava(String value) {
     return StringEscapeUtils.unescapeJava(StringEscapeUtils.unescapeJava(value));
   }
+
+  static DataAccessRequestData translate(String darDataString) {
```
rushtong: This doesn't have to be static. If you make it a default, anyone that implements this class gets it for free, i.e. line 28 in `DataAccessRequestMapper` can just be 
```
DataAccessRequestData data = translate(darDataString);
```

src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
```diff
@@ -16,13 +22,16 @@
   Timestamp createDate;
 
   @JsonProperty
-  User createUser;
+  Integer createUser;
```
rushtong: This is going to be confusing over time. We should bite the bullet soon and rename this to `createUserId`

src/main/resources/changesets/changelog-consent-74.0.xml
```diff
@@ -0,0 +1,22 @@
...
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="74.0" author="raejohanek">
+        <validCheckSum>ANY</validCheckSum>
+
+        <!-- fix mistake made in changeSet 72.0
+        column types cannot be cast to new type so must be dropped and re-added -->
+        <dropColumn tableName="dar_collection" columnName="update_user"/>
+        <dropColumn tableName="dar_collection" columnName="update_date"/>
```
rushtong, line 13: I think we had some pre-population of these columns in a previous changeset. This will drop all of that data.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
JVThomas: They do, it's defined in the `filterQuery` string. However you still need to declare those values as columns so that they can be referenced in the `ORDER BY` clause

rushtong: Looking at `filterQuery`, they look like they're querying from the dar data json, not from the user's institution or name:
```
  final String filterQuery = 
    "WHERE (dar.data #>> '{}')::jsonb ->> 'institution' ~* :institutionSearchTerm " +
    ...
    "AND (dar.data #>> '{}')::jsonb ->> 'researcher' ~* :researcherSearchTerm " +
```

JVThomas: Right, because we're filtering darCollections based on terms. If it's a researcher, then the term passed in will be based off of the users' institution/name on the service level. But if it's an admin querying (I assume admins will have a manage collection console), then the terms could be anything.

JVThomas: Oh, I see, the fields are deprecated on DataAccessRequestData. So it's a holdover from older records.

src/main/resources/changesets/changelog-consent-74.0.xml
JVThomas: No I checked. The pre-population is only for `create_date` and `create_user`. The update data wasn't filled in, so Rae's add/drop change is valid since the columns won't have any data.

src/main/java/org/broadinstitute/consent/http/db/mapper/RowMapperHelper.java
JVThomas: Can't do that. The method is being used within `DarCollectionReducer` so it has to remain static.

rushtong: DarCollectionReducer can implement RowMapperHelper too, and get it for free, just like the other classes that do this can.

JVThomas: Java classes can implement more than one interface? It's already implementing `LinkedHashMapRowReducer`

rushtong: Sure can!
```
public class DarCollectionReducer implements LinkedHashMapRowReducer<Integer, DarCollection>, RowMapperHelper {
```

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -410,6 +413,9 @@ public DataAccessRequest updateByReferenceIdVersion2(User user, DataAccessReques
             dar.getSubmissionDate(),
             now,
             dar.getData());
+        if (Objects.nonNull(dar.getCollectionId())) {
+            darCollectionDAO.updateDarCollection(dar.getCollectionId(), user.getDacUserId(), now);
```
rushtong, line 417: This is where the automated test is failing. The flow is to create a DAR, vote on it, etc. The failure here is that there is no saved collection with the id that's saved on the DAR. 

Flow:
* ElectionService updateElectionById ->
  * ElectionService updateSortDate ->
    * DataAccessRequestService updateByReferenceIdVersion2 ->  
      * DarCollectionDAO updateDarCollection

src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
rushtong: In the get query, there's an inner join on `c.collection_id = dar.collection_id` so you don't need to add that again here. 
```suggestion
  @SqlQuery(
      getCollectionAndDars +  " WHERE c.collection_id = :collectionId ")
```

src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
```diff
@@ -0,0 +1,157 @@
...
+
+  @RegisterBeanMapper(value = DarCollection.class)
+  @RegisterBeanMapper(value = DataAccessRequest.class, prefix = "dar")
+  @UseRowReducer(DarCollectionReducer.class)
+  @SqlQuery(" SELECT c.*, dar.id AS dar_id, dar.reference_id AS dar_reference_id, dar.collection_id AS dar_collection_id, "
+      + "dar.draft AS dar_draft, dar.user_id AS dar_userId, dar.create_date AS dar_create_date, "
+      + "dar.sort_date AS dar_sort_date, dar.submission_date AS dar_submission_date, "
+      + "dar.update_date AS dar_update_date, (dar.data #>> '{}')::jsonb AS data "
+      + "FROM dar_collection c, data_access_request dar " 
+      + "WHERE c.collection_id = dar.collection_id "
```
rushtong: I think this is better expressed as an inner join like you've done in other queries:
```
    "FROM dar_collection c " +
    "INNER JOIN data_access_request dar ON c.collection_id = dar.collection_id " +
```

src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
```diff
@@ -0,0 +1,157 @@
...
+  @RegisterBeanMapper(value = DarCollection.class)
+  @RegisterBeanMapper(value = DataAccessRequest.class, prefix = "dar")
+  @UseRowReducer(DarCollectionReducer.class)
+  @SqlQuery(
+    " SELECT c.*, dar.id AS dar_id, dar.reference_id AS dar_reference_id, dar.collection_id AS dar_collection_id, " +
+        "dar.draft AS dar_draft, dar.user_id AS dar_userId, dar.create_date AS dar_create_date, " +
+        "dar.sort_date AS dar_sort_date, dar.submission_date AS dar_submission_date, " +
+        "dar.update_date AS dar_update_date, (dar.data #>> '{}')::jsonb AS data " +
+        "FROM dar_collection c, data_access_request dar " +
+    " WHERE c.collection_id = dar.collection_id"
```
rushtong: Same feedback here, I think an inner join would be more expressive
```
    "FROM dar_collection c " +
    "INNER JOIN data_access_request dar ON c.collection_id = dar.collection_id " +

File: src/main/java/org/broadinstitute/consent/http/db/mapper/DataAccessRequestMapper.java [hunk 10]
```diff
@@ -31,21 +25,8 @@ public DataAccessRequest map(ResultSet resultSet, StatementContext statementCont
...
-                data.setReferenceId(dar.getReferenceId());
-                dar.setData(data);
-            } catch (JsonSyntaxException | NullPointerException e) {
-                String message = "Unable to parse Data Access Request, reference id: " + dar.getReferenceId() + "; error: " + e.getMessage();
-                logger.error(message);
-                throw new SQLException(message);
-            }
-        }
+        DataAccessRequestData data = translate(darDataString);
+        dar.setData(data);
```
rushtong, line 33: Not your code, but on line 19, we have a small bug. We need to check both for the column existence, and if the value is not null. `resultSet.getInt("collection_id")` returns 0 in the case of nulls.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
rushtong: This code is triggered when saving a draft DAR as well. Since we have DARs with a collection id of `0` from a different bug, the update fails with an NPE

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -410,6 +413,9 @@ public DataAccessRequest updateByReferenceIdVersion2(User user, DataAccessReques
             dar.getSubmissionDate(),
             now,
             dar.getData());
+        if (Objects.nonNull(dar.getCollectionId())) {
```
rushtong, line 416: For the time being, until the mapper bug is fixed:

```suggestion
        if (Objects.nonNull(dar.getCollectionId()) && dar.getCollectionId() > 0) {
```

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
JVThomas: So the mapper bug ultimately causes the failure because it assigns a 0 to `collection_id`, which gets past all of the null checks in the code, allowing any query on `collection_id = 0` to fire off (and fail)?

rushtong: It causes one of the failures that I've seen. This point of failure is being hit from two code paths. One is from saving draft DARs (and the mapper bug is the source). The second one is from updating an election which then tries to subsequently update the dar collection. I'm still trying to understand why that's failing. It's calling this line with a collection id when it shouldn't have one to begin with ... because we're not saving DARs with collection ids yet.

src/main/java/org/broadinstitute/consent/http/db/DAOContainer.java
```diff
@@ -76,6 +77,15 @@ public void setDataAccessRequestDAO(
...
   }
 
+  public DarCollectionDAO getDarCollectionDAO() {
+    return darCollectionDAO;
+  }
+
+  public void setDarCollectionDAO(
+    DarCollectionDAO darCollectionDAO) {
+    this.darCollectionDAO = darCollectionDAO;
+  }
```
rushtong, line 87: This is only called in a test. It needs to be used in `ConsentModule.providesDAOContainer()` as well so that it has a value. Some of the NPEs I'm seeing are from this not being set.

rushtong (approved):
Here's something we can do. This does not currently interfere with any existing functionality, so is good to go for now, even if we don't have the query exactly right. Once we move to the service level, we can refine it if we need to in subsequent PRs.

JVThomas:
Updated the query with the correction and offset/limit removal

rushtong (approved):
Thank you!

PR #1230, opened 2021-08-17
[DUOS-1356][risk=no] Query for Sam Resource Types

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1356

See also: 
* https://github.com/broadinstitute/terra-helmfile/pull/1560
* https://github.com/broadinstitute/terra-helm/pull/449

## Changes
* New Sam service and resource class
* Passthroughs to get Sam resource types. These will only become necessary for Sam DAR authorization later on down the line, but serves as a model for what to do for the more immediate ToS service calls we'll need to make.
* Model classes for resource types
* Minor updates to configuration classes to ensure we don't hit future config mismatches between the app, the helm chart, and the helm values files.
* Resource and Service class tests

----

- Verify all tests go green


src/main/resources/assets/api-docs.yaml
```diff
@@ -3572,6 +3572,21 @@ paths:
...
           description: The user doesn't exist in the system
+  /api/sam/resource-types:
+    get:
+      summary: Sam Resource Types
+      description: Sam Resource Types
+      tags:
+        - Sam
+      responses:
+        200:
+          description: Dataset Metrics
```
JVThomas: Leftover from copy and paste?

JVThomas (approved):
Other than the doc comment the code looks good, with the models, services, and resource classes following expected output as outlined in the integration doc and SAM's Swagger API

src/main/resources/assets/api-docs.yaml
rushtong: Hah, yes, copy/pasta

PR #1233, opened 2021-08-19
[DUOS-1419][risk=no] Update Jenkins Test Runner

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1419

## Changes
* Update the jenkins test runner to run the same way we do for github actions. Jenkins jobs are now passing and uploading results to BQ as they previously were, but generated the same way we do in the github action.
* Removed unused scripts and config files previously used
* Minor updates to the action so it is closely comparable to the jenkins job

----

- Verify all tests go green


JVThomas (approved):
Changes look good, bonus for getting rid of some jenkins overhead

PR #1232, opened 2021-08-18
[DUOS-1411][risk=no] Add Sam Status

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1411

## Changes
* Add a new status check for Sam
* Add a new degraded state which is triggered if any of the systems are not `OK` - except postgres, we're dead in the water with the db
* Added the expanded response from Ontology and Sam to the details of the results message.
* Expanded the tests

## Example response:
```
{
  "ok": true, 
  "degraded": false, 
  "deadlocks": {
    "details": null, 
    "duration": 0, 
    "error": null, 
    "healthy": true, 
    "message": null, 
    "time": 1629323820192, 
    "timestamp": "2021-08-18T21:57:00.192Z"
  }, 
  "elastic-search": {
    "details": null, 
    "duration": 501, 
    "error": null, 
    "healthy": true, 
    "message": "ClusterHealth is GREEN", 
    "time": 1629323818633, 
    "timestamp": "2021-08-18T21:56:58.633Z"
  }, 
  "google-cloud-storage": {
    "details": null, 
    "duration": 1294, 
    "error": null, 
    "healthy": true, 
    "message": null, 
    "time": 1629323820192, 
    "timestamp": "2021-08-18T21:57:00.192Z"
  }, 
  "ontology": {
    "details": {
      "ok": true, 
      "systems": {
        "deadlocks": {
          "details": null, 
          "duration": 0, 
          "error": null, 
          "healthy": true, 
          "message": null, 
          "time": 1629323820020, 
          "timestamp": "2021-08-18T21:57:00.020Z"
        }, 
        "elastic-search": {
          "details": null, 
          "duration": 1, 
          "error": null, 
          "healthy": true, 
          "message": "ClusterHealth is GREEN", 
          "time": 1629323820004, 
          "timestamp": "2021-08-18T21:57:00.004Z"
        }, 
        "google-cloud-storage": {
          "details": null, 
          "duration": 15, 
          "error": null, 
          "healthy": true, 
          "message": null, 
          "time": 1629323820020, 
          "timestamp": "2021-08-18T21:57:00.020Z"
        }
      }
    }, 
    "duration": 440, 
    "error": null, 
    "healthy": true, 
    "message": null, 
    "time": 1629323820633, 
    "timestamp": "2021-08-18T21:57:00.633Z"
  }, 
  "postgresql": {
    "details": null, 
    "duration": 263, 
    "error": null, 
    "healthy": true, 
    "message": null, 
[...]

PR #1234, opened 2021-08-19
Revert "[DUOS-1411][risk=no] Add Sam Status"

rushtong:
Reverts DataBiosphere/consent#1232

We're seeing issues in FiaB starts related to Consent status so will revert this and see if it has any effect.


JVThomas (approved):
Revert checks out

PR #1256, opened 2021-09-03
[DUOS-1436][risk=no] Service class for Dar Collections

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1436

## Changes
* Minor updates to the filter queries so we can filter by user and filter term
* New pagination response object that holds all information the UI will need to display a page of data
* Minor updates to pagination token

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/PaginationToken.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/PaginationToken.java
@@ -11,6 +11,8 @@ import java.util.Base64;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 public class PaginationToken {
 
@@ -98,7 +100,7 @@ public class PaginationToken {
   public Integer getFilteredCount() {
     return this.filteredCount;
   }
-  
+
   public void setFilteredCount(Integer filteredCount) {
     this.filteredCount = filteredCount;
   }
@@ -127,7 +129,7 @@ public class PaginationToken {
     }
   }
 
-  //could result in bad request if page is greater than page count but unable to do validation 
+  //could result in bad request if page is greater than page count but unable to do validation
   //in every case here because pagecount is only populated in response tokens
   public void generateNext() {
     if (Objects.nonNull(filteredPageCount) && page == filteredPageCount) {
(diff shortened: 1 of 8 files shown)


src/main/java/org/broadinstitute/consent/http/models/PaginationToken.java
```diff
@@ -91,16 +93,21 @@ public String getSortDirection() {
...
   }
 
   public Integer getFilteredCount() {
     return this.filteredCount;
   }
-  
+
   public void setFilteredCount(Integer filteredCount) {
     this.filteredCount = filteredCount;
+    if (((filteredCount/this.getPageSize()) % this.getPageSize()) == 0) {
```
JVThomas: You need a check where `filteredCount > 0 && filteredCount < pageSize`, otherwise the filteredPageCount is going to be 0 for this scenario

Example: `filteredCount = 4`, `pageSize = 10`
if condition -> `(4/10) % 10` equals 0, `0 == 0` is true,  therefore `filteredPageCount = filteredCount/this.getPageSize()` or `4/10 = 0`

JVThomas (changes requested):
One small change for the filteredPageCount calculation, but otherwise the code looks good.

src/main/java/org/broadinstitute/consent/http/models/PaginationToken.java
rushtong: Nice catch!

rushtong: I found more bugs when looking at this. PTAL when you get a chance, thank you!

src/main/java/org/broadinstitute/consent/http/models/PaginationToken.java
```diff
@@ -91,16 +86,29 @@ public String getSortDirection() {
...
+
   public void setFilteredCount(Integer filteredCount) {
     this.filteredCount = filteredCount;
+    /*
+     * Three cases in determining filtered page count
+     * 1. The count is less than the page size: page size = 1
+     * 2. The count/page size modulo page size is 0, i.e. count of 50, page size of 10, page count is 5
+     * 3. Anything else, i.e. count of 55, page size of 10, page size is 6
+     */
+    if (filteredCount <= this.getPageSize()) {
```
JVThomas, line 105: For pageSize (or any user submitted Integer for that matter), do we want to catch and throw a BadRequestException for values that are 0 or less?

rushtong: We can do something like that. This class should be able to validate itself and if it's not valid, then the resource class should do exactly that. That would prevent bad data from making its way to the service class. I suggest we handle that when we write the API that handles parsing a `PaginationToken`.

JVThomas (approved):
Code looks good, additional validations can be added on (for resource or service) as the endpoints are made.

PR #1269, opened 2021-09-20
DUOS-1253[risk=med] Added DAR Collection Resource Cancel endpoint

JVThomas:
Addresses [DUOS-1253](https://broadworkbench.atlassian.net/browse/DUOS-1253)

PR creates cancel endpoint that allows researchers to cancel a DAR Collection with expected additions to DarCollectionService and DarCollectionDAO classes. PR also includes a new cancel DAO method to DataAccessRequestDAO to help facilitate cancel updates.

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
@@ -81,7 +81,7 @@ public class DarCollectionResource extends Resource {
   }
 
   @PUT
-  @Path("dar/{id}/cancel")
+  @Path("{id}/cancel")
   @Produces("application/json")
   @RolesAllowed(RESEARCHER)
   public Response cancelDarCollectionByCollectionId(@Auth AuthUser authUser, @PathParam("id") Integer collectionId) {
--- a/src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
@@ -4,6 +4,8 @@ import com.google.inject.Inject;
 import io.dropwizard.auth.Auth;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
+
 import javax.annotation.security.PermitAll;
 import javax.annotation.security.RolesAllowed;
 import javax.ws.rs.ForbiddenException;
@@ -82,15 +84,22 @@ public class DarCollectionResource extends Resource {
   @Path("dar/{id}/cancel")
   @Produces("application/json")
   @RolesAllowed(RESEARCHER)
-  public Response cancelDarCollection(@Auth AuthUser authUser, @PathParam("id") Integer collectionId) {
+  public Response cancelDarCollectionByCollectionId(@Auth AuthUser authUser, @PathParam("id") Integer collectionId) {
     try {
       User user = userService.findUserByEmail(authUser.getEmail());
       DarCollection collection = darCollectionService.getByCollectionId(collectionId);
+      isCollectionPresent(collection);
       validateUserIsCreator(user, collection);
       DarCollection cancelledCollection = darCollectionService.cancelDarCollection(collection, user);
       return Response.ok().entity(cancelledCollection).build();
     } catch(Exception e) {
       return createExceptionResponse(e);
+    }
+  }
+
+  private void isCollectionPresent(DarCollection collection) {
+    if(Objects.isNull(collection)) {
+      throw new NotFoundException("Collection not found");
     }
   }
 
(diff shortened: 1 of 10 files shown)


rushtong:
Side note, it looks like coveralls is down at the moment which is causing some test failures:
```
Error:  Failed to execute goal org.eluder.coveralls:coveralls-maven-plugin:4.3.0:report (default-cli) on project consent: I/O operation failed: Report submission to Coveralls API failed with HTTP status 503: Service Unavailable (Coveralls API internal error) -> [Help 1]
```

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -144,4 +155,30 @@ public DarCollection updateDarCollection(Integer collectionId, User user) {
     return collections;
   }
 
+  // fetch elections in bulk via dar reference ids
+  // if any election exists, throw a bad request
+  // else collect the dar reference ids where the status is NOT cancelled
+  // once collected, perform bulk update where status is updated with cancelled
```
rushtong: I keep questioning our own logic in making this choice about ANY ELECTION that ever existed, and then re-convince myself this is the right choice. Can we add some extra commentary here, anything that will help me not re-question myself again. Maybe something like:

> If an election has ever been created for any individual DAR in a collection, a researcher cannot 
> move the collection into canceled state. That task would need to be completed by an admin or
> DAC Chairperson.

src/main/resources/assets/paths/cancelCollectionByCollectionId.yaml
```diff
@@ -0,0 +1,25 @@
+put:
```
rushtong, line 1: The ticket calls for a DELETE, but I think PUT is more accurate here, thanks for the catch. Can we also update the ticket so we can track this decision better?

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -144,4 +155,30 @@ public DarCollection updateDarCollection(Integer collectionId, User user) {
...
+    }
+    List<String> nonCanceledIds = dars.stream()
+      .filter(d -> {
+        String status = d.getData().getStatus();
+        return Objects.nonNull(status) && status.toLowerCase() != "canceled";
+      })
+      .map(d -> d.getReferenceId())
+      .collect(Collectors.toList());
+    
+    dataAccessRequestDAO.cancelByReferenceIds(nonCanceledIds);
```
rushtong: Will need to catch the empty collection problem as this can throw a sql exception when that's true.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -144,4 +155,30 @@ public DarCollection updateDarCollection(Integer collectionId, User user) {
...
+  // if any election exists, throw a bad request
+  // else collect the dar reference ids where the status is NOT cancelled
+  // once collected, perform bulk update where status is updated with cancelled
+  public DarCollection cancelDarCollection(DarCollection collection, User user) {
+    List<DataAccessRequest> dars = collection.getDars();
+    List<String> referenceIds = dars.stream()
+      .map(d -> d.getReferenceId())
+      .collect(Collectors.toList());
+    
+    List<Election> elections = electionDAO.findLastElectionsByReferenceIdsAndType(referenceIds, ElectionType.DATA_ACCESS.getValue());
```
rushtong, line 172: same thing here, `referenceIds` should never be empty, but we should be safer about catching that case.

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
```diff
@@ -77,6 +80,29 @@ public Response getCollectionByReferenceId(
...
+  @PUT
+  @Path("{id}/cancel")
+  @Produces("application/json")
+  @RolesAllowed(RESEARCHER)
+  public Response cancelDarCollectionByCollectionId(@Auth AuthUser authUser, @PathParam("id") Integer collectionId) {
+    try {
+      User user = userService.findUserByEmail(authUser.getEmail());
+      DarCollection collection = darCollectionService.getByCollectionId(collectionId);
+      isCollectionPresent(collection);
+      validateUserIsCreator(user, collection);
```
rushtong, line 92: Nice checks!

rushtong (commented):
Looks good overall, just a couple minor nits that need to be updated, then 👍🏽

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
JVThomas: Should an error be thrown here? Logically I can end the function by just returning collection, but the code is discovering bad data which the user should know about.

rushtong: I would log a warning instead of pulling that up to the client.

rushtong (approved):
Thank you for the changes 👍🏽  !

PR #1273, opened 2021-09-22
DUOS-1437[risk=med] Added DAR Collection initial filter query endpoint

JVThomas:
Addresses [DUOS-1437](https://broadworkbench.atlassian.net/browse/DUOS-1437)

PR adds a initialQuery endpoint that executes the initial query for a new filtered Collections GET request alongside tests for endpoint responses. PR also adds adjustments to the model, service and DAO classes with associated updates to their respective tests. 

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
@@ -66,7 +66,10 @@ public interface DarCollectionDAO {
   @RegisterBeanMapper(value = DarCollection.class)
   @RegisterBeanMapper(value = DataAccessRequest.class, prefix = "dar")
   @UseRowReducer(DarCollectionReducer.class)
-  @SqlQuery(getCollectionAndDars + " WHERE c.collection_id in (<collectionIds>)")
+  @SqlQuery(
+    getCollectionAndDars 
+    + " WHERE c.collection_id in (<collectionIds>)"
+    +  " ORDER BY <sortField> <sortOrder>")
   List<DarCollection> findDARCollectionByCollectionIds(
           @BindList("collectionIds") List<Integer> collectionIds,
           @Define("sortField") String sortField,
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
@@ -33,9 +33,6 @@ public interface DarCollectionDAO {
       " AND (" +
         "COALESCE(i.institution_name, '') ~* :filterTerm " +
         " OR (dar.data #>> '{}')::jsonb ->> 'projectTitle' ~* :filterTerm " +
-        //NOTE: since users can change their name, however the dar researcher attribute will not be updated
-        //Search on both dar.data and user displayname for proper name matching
-        " OR (dar.data #>> '{}')::jsonb ->> 'researcher' ~* :filterTerm " +
         " OR u.displayname ~* :filterTerm " +
         " OR c.dar_code ~* :filterTerm " +
         " OR EXISTS " +
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
@@ -67,10 +67,18 @@ public interface DarCollectionDAO {
   @RegisterBeanMapper(value = DataAccessRequest.class, prefix = "dar")
   @UseRowReducer(DarCollectionReducer.class)
   @SqlQuery(
+    getCollectionAndDars + " WHERE c.collection_id in (<collectionIds>)")
+  List<DarCollection> findDARCollectionByCollectionIds(
+          @BindList("collectionIds") List<Integer> collectionIds);
+
+  @RegisterBeanMapper(value = DarCollection.class)
+  @RegisterBeanMapper(value = DataAccessRequest.class, prefix = "dar")
+  @UseRowReducer(DarCollectionReducer.class)
+  @SqlQuery(
     getCollectionAndDars 
     + " WHERE c.collection_id in (<collectionIds>)"
     +  " ORDER BY <sortField> <sortOrder>")
-  List<DarCollection> findDARCollectionByCollectionIds(
+  List<DarCollection> findDARCollectionByCollectionIdsWithOrder(
           @BindList("collectionIds") List<Integer> collectionIds,
           @Define("sortField") String sortField,
           @Define("sortOrder") String sortOrder);
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
@@ -63,7 +63,9 @@ public interface DarCollectionDAO {
           @Define("sortField") String sortField,
           @Define("sortOrder") String sortOrder);
 
-
+  @RegisterBeanMapper(value = DarCollection.class)
+  @RegisterBeanMapper(value = DataAccessRequest.class, prefix = "dar")
+  @UseRowReducer(DarCollectionReducer.class)
   @SqlQuery(getCollectionAndDars + " WHERE c.collection_id in (<collectionIds>)")
   List<DarCollection> findDARCollectionByCollectionIds(
           @BindList("collectionIds") List<Integer> collectionIds,
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
@@ -33,6 +33,9 @@ public interface DarCollectionDAO {
       " AND (" +
         "COALESCE(i.institution_name, '') ~* :filterTerm " +
         " OR (dar.data #>> '{}')::jsonb ->> 'projectTitle' ~* :filterTerm " +
+        //NOTE: since users can change their name, however the dar researcher attribute will not be updated
+        //Search on both dar.data and user displayname for proper name matching
+        " OR (dar.data #>> '{}')::jsonb ->> 'researcher' ~* :filterTerm " +
         " OR u.displayname ~* :filterTerm " +
         " OR c.dar_code ~* :filterTerm " +
         " OR EXISTS " +
--- a/src/main/java/org/broadinstitute/consent/http/models/PaginationToken.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/PaginationToken.java
@@ -50,9 +50,9 @@ public class PaginationToken {
     checkSortField(sortField);
     checkSortDirection(sortDirection);
     this.page = page;
-    this.pageSize = pageSize;
-    this.sortField =  Objects.nonNull(sortField) ? sortField : "darCode";
-    this.sortDirection = Objects.nonNull(sortDirection) ? sortDirection : "DESC";
+    this.pageSize = Objects.nonNull(pageSize) ? pageSize : 10;
+    this.sortField = sortField;
+    this.sortDirection = sortDirection;
     this.filterTerm = filterTerm;
     checkForValidNumbers();
   }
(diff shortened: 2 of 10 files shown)


src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
```diff
@@ -33,6 +33,9 @@
       " AND (" +
         "COALESCE(i.institution_name, '') ~* :filterTerm " +
         " OR (dar.data #>> '{}')::jsonb ->> 'projectTitle' ~* :filterTerm " +
+        //NOTE: since users can change their name, however the dar researcher attribute will not be updated
+        //Search on both dar.data and user displayname for proper name matching
+        " OR (dar.data #>> '{}')::jsonb ->> 'researcher' ~* :filterTerm " +
```
rushtong: We really don't want to be using the `researcher` field in `dar.data` ... it will go away fairly soon. All user-specific data should be pulled from the user's properties. Especially since a user _can_ change their name, their updated name will not be in `dar.data`, but will be on the user.

src/main/java/org/broadinstitute/consent/http/models/PaginationToken.java
```diff
@@ -7,16 +7,16 @@
...
+import java.util.Map;
 import java.util.Objects;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 public class PaginationToken {
 
   private static final Charset UTF_8 = StandardCharsets.UTF_8;
-  private static final List<String> acceptableSortFields = Collections.emptyList();
+  // private static final Map<String> acceptableSortFields = Collections.emptyLi();
```
rushtong: nit: rm

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
```diff
@@ -87,4 +115,19 @@ private void validateUserIsCreator(User user, DarCollection collection) {
...
   }
+
+  private Map<String, String> defineAcceptableSortFields() {
+    return Map.of(
+      "projectTitle", "projectTitle",
+      "researcher", "researcher",
+      "darCode", "dar_code",
+      "institution", "institution_name"
+    );
+  }
```
rushtong: Conceptually, I think this belongs on the `DarCollection` class.

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
```diff
@@ -77,6 +83,28 @@ public Response getCollectionByReferenceId(
...
+  public Response getCollectionsByInitialQuery(
+    @Auth AuthUser authUser,
+    @QueryParam("filterTerm") String filterTerm,
+    @QueryParam("sortField") String sortField,
+    @QueryParam("sortOrder") String sortOrder,
+    @QueryParam("pageSize") int pageSize
+  ) {
+    try {
+      User user = userService.findUserByEmail(authUser.getEmail());
+      checkIfSortFieldAndSortOrderIsPresent(sortField, sortOrder);
```
rushtong: Can we instead pick a default like `dar code` `ascending` instead of throwing a bad request? That way the user will get something back and can re-query or re-sort depending on preference.

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
```diff
@@ -87,4 +115,19 @@ private void validateUserIsCreator(User user, DarCollection collection) {
...
+      "researcher", "researcher",
+      "darCode", "dar_code",
+      "institution", "institution_name"
+    );
+  }
+
+  private void checkIfSortFieldAndSortOrderIsPresent(String sortField, String sortOrder) {
+    if (Objects.isNull(sortOrder) || Objects.isNull(sortOrder)) {
+      throw new BadRequestException("sortOrder and sortField are required");
+    }
```
rushtong: See above, I think we should choose a sensible default and let the client make different choices if preferred.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -144,4 +150,12 @@ public DarCollection updateDarCollection(Integer collectionId, User user) {
     return collections;
   }
 
+  private PaginationResponse<DarCollection> createEmptyPaginationResponse(Integer unfilteredCount) {
+    return new PaginationResponse<DarCollection>()
+          .setUnfilteredCount(unfilteredCount)
+          .setFilteredCount(0)
+          .setFilteredPageCount(1) //should this be 0 or 1?
```
rushtong, line 167: I'm conflicted on this point. `1` seems safer from a UI perspective, but `0` is more accurate. Let's stick with what you have and we can adjust if it causes problems.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -57,13 +57,19 @@ public DarCollectionService(DarCollectionDAO darCollectionDAO, DatasetDAO datase
    * @return A PaginationResponse object
    */
   public PaginationResponse<DarCollection> getCollectionsWithFilters(PaginationToken token, User user) {
-
     List<DarCollection> unfilteredDars = darCollectionDAO.findDARCollectionsCreatedByUserId(user.getDacUserId());
+    if(unfilteredDars.isEmpty()) {
```
rushtong: nit: spacing. Here and below, see our style guide for whitespace
* https://github.com/DataBiosphere/consent/blob/develop/CONTRIBUTING.md
* https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace
```suggestion
    if (unfilteredDars.isEmpty()) {
```

src/main/resources/assets/api-docs.yaml
```diff
@@ -124,6 +124,8 @@ paths:
     $ref: './paths/collection.yaml'
   /api/collections/dar/{referenceId}:
     $ref: './paths/collectionByReferenceId.yaml'
+  /api/collections/query:
+    $ref: './paths/collectionsByQuery.yaml'
```
rushtong, line 128: I love how this is working out

src/main/resources/assets/paths/collectionsByQuery.yaml
```diff
@@ -0,0 +1,45 @@
...
+    200:
+      description: Returns a single DAR Collection
+      content:
+        application/json:
+          schema:
+            $ref: '../schemas/DarCollection.yaml'
+    400:
+      description: Bad Request
+    401:
+      description: Unauthorized
```
rushtong: In general, we can ignore these for individual paths. This is true of any `/api/*` call

src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
```diff
@@ -187,4 +218,13 @@ private void initWithPaginationToken(PaginationToken token, int unfilteredCount,
...
   }
+
+  private Map<String, String> defineAcceptableSortFields() {
+    return Map.of(
+      "projectTitle", "projectTitle",
+      "researcher", "researcher",
+      "darCode", "dar_code",
+      "institution", "institution_name"
+    );
+  }
```
rushtong: DRY. Related to my comment in the resource class, if you find yourself writing the same code twice, consider moving it to a more useful location. I could envision a custom interface for certain model classes where any class that wanted to have custom sort orders and sort fields would be forced to implement this method. Definitely not necessary at this point, but good to keep in mind.

rushtong (changes requested):
See relatively minor comments inline, I think this looks pretty good.

src/main/java/org/broadinstitute/consent/http/models/PaginationToken.java
```diff
@@ -131,20 +130,30 @@ public String toBase64() {
...
+  }
+
+  public Map<String, String> getAcceptableSortFields() {
+    return acceptableSortFields;
+  }
+
+  private String validateSortField(String sortField) {
+    if(Objects.isNull(sortField) || Objects.isNull(acceptableSortFields.get(sortField))) {
+      return "dar_code";
     }
```
rushtong: I think I missed this on the first pass. Having a default of `dar_code` ties this class directly to `DarCollection`. I think it might be better to throw an exception here if we really can't validate it. Another thought is to add a default field as a second argument, but that seems a little convoluted since this is only used in the constructor.

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
```diff
@@ -77,6 +80,27 @@ public Response getCollectionByReferenceId(
...
+  public Response getCollectionsByInitialQuery(
+    @Auth AuthUser authUser,
+    @QueryParam("filterTerm") String filterTerm,
+    @QueryParam("sortField") String sortField,
+    @QueryParam("sortOrder") String sortOrder,
+    @QueryParam("pageSize") int pageSize
+  ) {
+    try {
+      User user = userService.findUserByEmail(authUser.getEmail());
+      PaginationToken token = new PaginationToken(1, pageSize, sortField, sortOrder, filterTerm, DarCollection.acceptableSortFields);
```
rushtong: Related to my comment above, maybe this is a good place to validate the provided fields (and others as well)? For example, if I called this endpoint with a `sortField` of `electionid`, that might trigger a bad request.

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
```diff
@@ -77,6 +80,27 @@ public Response getCollectionByReferenceId(
...
+      PaginationResponse<DarCollection> paginationResponse = darCollectionService.getCollectionsWithFilters(token, user);
+      return Response.ok().entity(paginationResponse).build();
+    } catch(Exception e) {
```
rushtong: minor nit: spacing
```suggestion
    } catch (Exception e) {
```

rushtong (changes requested):
Code looks really good. One minor suggestion is to pull some validation up to the resource class so we don't have to tie `PaginationToken` to any specific model class.

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
JVThomas: That's what was done initially, these updates were added based on feedback
https://github.com/DataBiosphere/consent/pull/1273#discussion_r715055899

rushtong: The feedback in [#1273](https://github.com/DataBiosphere/consent/pull/1273#discussion_r715055899) was to move the fields themselves, which was a good change that you made. My feedback here is of a different category. The definition of the sort fields is one thing, and then what we do with those fields is another thing. My feedback here is about how we use those fields and validating that the information used to create the token is correct or not.

JVThomas: That's the purpose of acceptableSortFields in the PaginationToken itself which would throw errors initially as seen here (see constructor and helper functions)

https://github.com/DataBiosphere/consent/blob/ce2a7422daf85592c8892f80b7ed3fe8af2c9074/src/main/java/org/broadinstitute/consent/http/models/PaginationToken.java

I updated that based on this comment https://github.com/DataBiosphere/consent/pull/1273#discussion_r715057680, but I'm guessing you meant for that to be applied to `null` arguments? If so, I can bring back the old implementation to cover non-null inputs that are invalid and set up defaults for null inputs.

rushtong: My only real concern is that we are using `dar_code` inside of the `PaginationToken` class which doesn't separate concerns. The token class shouldn't be defaulting to a specific model class just because it's the first one we're implementing. My suggestion here in the resource class was just one way to potentially solve that problem. I'm open to any solution so long as we can fail fast in the resource class and separate `PaginationToken` from the model.

rushtong (approved):
Thank you for the changes, this looks great! 👍🏽

PR #1288, opened 2021-10-01
[DUOS-1445][risk=no] Prevent SOs from changing their institution

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1445

Signing officials are prohibited from updating their institution to prevent users from changing institutions just to issue library cards that might be outside of their normal institution. Signing Officials must go through customer support to accomplish this.

This PR does not complete DUOS-1445 as there will need to be a follow-on UI PR that helps user's navigate this edge case on their profile page.

## Changes
* When updating a user, and the user is a Signing Official, throw an error if the user is trying to change their institution. 

----

- Verify all tests go green


JVThomas (approved):
Code looks good. Tested locally, profile allows SO to update if no institution is assigned or If the institution remains unchanged.

PR #1291, opened 2021-10-06
DUOS-1469[risk=no] Add Elections to DARCollection and findDarCollectionsCreatedByUserID 

JVThomas:
Addresses [DUOS-1469](https://broadworkbench.atlassian.net/browse/DUOS-1469)

PR adds elections to the Collection, populated by the `findDarCollectionsCreatedByUserID` query. This will provide details needed in [DUOS-1256](https://broadworkbench.atlassian.net/browse/DUOS-1256) to enable/disable Collection cancellation buttons on the UI (visually).

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
@@ -3,6 +3,7 @@ package org.broadinstitute.consent.http.db;
 import org.broadinstitute.consent.http.db.mapper.DarCollectionReducer;
 import org.broadinstitute.consent.http.models.DarCollection;
 import org.broadinstitute.consent.http.models.DataAccessRequest;
+import org.broadinstitute.consent.http.models.Election;
 import org.jdbi.v3.sqlobject.config.RegisterBeanMapper;
 import org.jdbi.v3.sqlobject.customizer.Bind;
 import org.jdbi.v3.sqlobject.customizer.BindList;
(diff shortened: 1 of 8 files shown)


src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
```diff
@@ -103,14 +104,19 @@
...
-      + "dar.update_date AS dar_update_date, (dar.data #>> '{}')::jsonb AS data "
+      + "dar.update_date AS dar_update_date, (dar.data #>> '{}')::jsonb AS data, "
+      + "e.electionid AS e_election_id, e.referenceid AS e_reference_id, e.status AS e_status, e.createdate AS e_create_date, "
+      + "e.lastupdate AS e_last_update, e.datasetid AS e_dataset_id, e.electiontype AS e_election_type "
       + "FROM dar_collection c "
       + "INNER JOIN data_access_request dar ON c.collection_id = dar.collection_id "
-      + "AND c.create_user_id = :userId")
+      + "AND c.create_user_id = :userId "
+      + "LEFT JOIN election e ON e.referenceid = dar.reference_id "
+  )
```
rushtong: Not a criticism here, just a note to point out that this will add ALL elections to the results, including previously canceled elections, Access, RP, Data Owner elections. I imagine that the UI will have a good amount of work to do to sort through the volume of info here.

src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
```diff
@@ -138,6 +143,22 @@ public void setDatasets(Set<DataSet> datasets) {
...
+    this.electionMap = electionMap;
+  }
+
+  public Map<String, List<Election>> getElectionMap() {
+    return electionMap;
+  }
+
+  public void addElection(Election election) {
+    String referenceId = election.getReferenceId();
+    if(!electionMap.containsKey(referenceId)){
```
rushtong, line 156: nit spacing:
```suggestion
    if (!electionMap.containsKey(referenceId)) {
```

src/main/resources/assets/api-docs.yaml
```diff
@@ -4101,58 +4101,7 @@ components:
...
-          type: string
-          description: Contains the correspondant Dul name for the given election.
-        version:
-          type: integer
-          format: int32
-          description: Indicates the version number of the given election.
-        archived:
-          type: boolean
-          description: Determines if the election is archived.
+      $ref: "./schemas/Election.yaml"
```
rushtong, line 4104: 👍🏽

src/main/resources/assets/schemas/DarCollection.yaml
```diff
@@ -30,3 +30,11 @@ properties:
     description: List of all datasets referenced across all DARs within the collection
     items:
       $ref: './Dataset.yaml'
+  electionMap:
+    type: object
+    description: Map of referenceIds to their associated elections
+    additionalProperties: 
+      type: array
```
rushtong, line 37: I think what you want here is `object` so you can describe a map type. Swagger docs: https://swagger.io/docs/specification/data-models/dictionaries/

rushtong (approved):
This looks pretty good! One minor suggestion for the swagger spec, but otherwise 👍🏽

src/main/resources/assets/schemas/DarCollection.yaml
JVThomas: `electionMap` itself is defined as the object, the key value pairs contained within it needs to be outlined with `additionalProperties` in order to define values where the keys are strings.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
JVThomas: Good point, I can update the query to only retrieve the latest Access and RP election instead.

src/main/resources/assets/schemas/DarCollection.yaml
rushtong: Ah ... I completely misread this, thank you.

rushtong (approved):
👍🏽

PR #1295, opened 2021-10-08
[DUOS-302][risk=no] Add failure config for owasp checks

rushtong:
Adds a simple count check so we can know if a nightly run is failing.
See also: https://github.com/DataBiosphere/consent-ontology/pull/508

----

- Verify all tests go green


JVThomas (approved):
Looks good, build will trigger a fail state if a vulnerability is found.

PR #1311, opened 2021-10-18
DUOS-1256[risk=no] Correction to DAR Status processing in cancelDarCollection (DarCollectionService)

JVThomas:
Back-end correction needed for [DUOS-1256](https://broadworkbench.atlassian.net/browse/DUOS-1256), adjustment correctly evaluates DAR with `status:null` as a non-cancelled DAR

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
@@ -189,7 +189,7 @@ public class DarCollectionService {
     List<String> nonCanceledIds = dars.stream()
       .filter(d -> {
         String status = d.getData().getStatus();
-        return Objects.nonNull(status) && status.toLowerCase() != "canceled";
+        return Objects.isNull(status) || status.toLowerCase() != "canceled";
       })
       .map(d -> d.getReferenceId())
       .collect(Collectors.toList());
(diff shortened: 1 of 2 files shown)


rushtong (approved):
👍🏽

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -189,7 +189,7 @@ public DarCollection cancelDarCollection(DarCollection collection, User user) {
     List<String> nonCanceledIds = dars.stream()
       .filter(d -> {
         String status = d.getData().getStatus();
-        return Objects.nonNull(status) && status.toLowerCase() != "canceled";
+        return Objects.isNull(status) || status.toLowerCase() != "canceled";
       })
```
rushtong: TOL: just an alternative approach that might be of interest to you. If you moved this logic to a method in the `DataAccessRequest` class, say `isNotCanceled()`, then you could filter on the class method instead, something like:
```
      .filter(DataAccessRequest::isNotCanceled)
```
It would be nice if we could invert that in the filter and check for !isCanceled, but you can't have operations in the :: notation like that. Taking the isCanceled approach would mean re-writing the surrounding logic too which might be more than work than this approach.

PR #1321, opened 2021-10-20
[DUOS-1321][risk=nedium] Resubmit Canceled DAR

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1321

## Changes
* New endpoint to generate a Draft DAR from a canceled `DarCollection`
* Service method to support that endpoint
* Repurposed the deprecated `DataAccessRequestDAO.insertVersion2` method that we still need to create DARs that don't have a collection id. Those are really draft DARs so renamed method and tests accordingly.
* Minor test update/refactoring
  * Dar creation helper methods are getting fairly complex and introducing unintended side effects
* Updated logging in `MatchService` to reduce unnecessary info and improve messaging

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
```diff
@@ -159,6 +166,34 @@ public Response cancelDarCollectionByCollectionId(@Auth AuthUser authUser, @Path
...
+    } catch(Exception e) {
+      return createExceptionResponse(e);
+    }
+  }
+
+  private void validateCollectionIsCanceled(DarCollection collection) {
+    boolean isCanceled =
+        collection.getDars().stream()
+            .anyMatch(
+                d -> d.getData().getStatus().equalsIgnoreCase(ElectionStatus.CANCELED.getValue()));
```
JVThomas: It might be better to have either 

1) A dedicated constant on DarCollection for "Canceled"
2) A shared constant for "Canceled" to be used by multiple models

...to keep the code clear. I can see people misattributing the DAR Canceled status with that of an Election's if we use `ElectionStatus.CANCELED` (I remember I did starting out)

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
```diff
@@ -159,6 +166,34 @@ public Response cancelDarCollectionByCollectionId(@Auth AuthUser authUser, @Path
...
+  @Path("{id}/resubmit")
+  @Produces("application/json")
+  @RolesAllowed(RESEARCHER)
+  public Response resubmitDarCollection(@Auth AuthUser authUser, @PathParam("id") Integer collectionId) {
+    try {
+      User user = userService.findUserByEmail(authUser.getEmail());
+      DarCollection sourceCollection = darCollectionService.getByCollectionId(collectionId);
+      isCollectionPresent(sourceCollection);
+      validateUserIsCreator(user, sourceCollection);
+      validateCollectionIsCanceled(sourceCollection);
```
JVThomas, line 179: Here's a question: What should we do about elections? For now we can just prevent resubmission if an election exists, but down the road should researchers be allowed to resubmit if the latest election was closed/denied?

JVThomas (commented):
I've left two comments on the code (one a suggestion, one a question), but other than that the code looks good (provided that tests are added for the new methods).

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
rushtong: Great idea ... will make this more clear with a constant.

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
rushtong: If a DAR is canceled, then there should be no Open election on that DAR. I can add some validation on that in the service method, but if we found one, then there would indeed be a canceled DAR with an open election, which is an error case I hadn't considered. 

[This is the logic](https://github.com/DataBiosphere/consent/blob/develop/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java#L265) we have now for canceling a DAR so I'll use that same logic for the error case.

To answer the question, I don't believe we will ever support resubmission after denial. That case should be handled with a new DAR submission.

rushtong:
@JVThomas Have added a boatload of tests and some minor fixes, PTAL!

src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
```diff
@@ -30,6 +30,9 @@
     @SqlQuery("select electionId from election where referenceId = :referenceId and lower(status) = 'open'")
     Integer getOpenElectionIdByReferenceId(@Bind("referenceId") String referenceId);
 
+    @SqlQuery("SELECT distinct electionid FROM election WHERE referenceid IN (<referenceIds>) AND LOWER(status) = 'open'")
```
JVThomas: Following up on your comment about elections and DAR resubmissions, I think you could simplify this even further and just search for any election that exists on that referenceId. 

- Researchers can't resubmit DARs with Closed/Denied elections
- Researchers shouldn't be allowed to resubmit a DAR with an Open election
- Is there a reason for a researcher to resubmit a DAR that was approved? (Can't think of one of the top of my head)

If there's no reason for anyone to resubmit on approved DARs then a query on elections by reference id with a `LIMIT 1` should be fast and sufficient enough for the election check.

JVThomas: Other than the above comment/inquiry, the code looks good

rushtong: Spent some time reviewing our DAR-cancel logic. We have two existing code paths to do this.
1. `DataAccessRequestService.cancelDataAccessRequest` which looks for any elections, period as recommended in the feedback.
2. `DarCollectionService.cancelDarCollection` which looks for last elections by `DataAccess` type
3. And then this PR tries to add a third logical condition, open elections.

Seems like we should unify these around the same logic. I think you're right - the correct logic should be any election ever created, period. Will try to make that adjustment across all code paths more clear.

JVThomas (approved):
Code looks good! The Draft DAR return on resubmit is especially useful for updating the Researcher Console Draft table

PR #1323, opened 2021-10-22
DUOS-1478[risk=no] DAR Collection ID Assignment - Bugfix and Backfill

JVThomas:
Addresses [DUOS-1478](https://broadworkbench.atlassian.net/browse/DUOS-1478)

PR corrects collection ID assignment on DAR drafts when they are submitted review. PR also backfills missing collection IDs on DARs that 

- Have the `darCode` key saved in the `data` column
- Have `collection_id = NULL`
- Have `draft = false`

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
@@ -362,7 +362,7 @@ public class DataAccessRequestService {
                 if (idx == 0) {
                     DataAccessRequest alreadyExists = dataAccessRequestDAO.findByReferenceId(dataAccessRequest.getReferenceId());
                     if (Objects.nonNull(alreadyExists)) {
-                        dataAccessRequestDAO.updateDraftByReferenceId(dataAccessRequest.getReferenceId(), false);
+                        dataAccessRequestDAO.updateDraftForCollection(collectionId, dataAccessRequest.getReferenceId());
                         dataAccessRequestDAO.updateDataByReferenceIdVersion2(
                             dataAccessRequest.getReferenceId(),
                             user.getDacUserId(),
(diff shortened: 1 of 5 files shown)


src/main/resources/changesets/changelog-consent-76.0.xml
```diff
@@ -0,0 +1,20 @@
...
+      UPDATE data_access_request AS dar 
+      SET collection_id = c.collection_id
+      FROM dar_collection c
+      WHERE (
+        CONCAT((dar.data #>> '{}')::jsonb ->> 'darCode', '-') ~* c.dar_code
+        OR (dar.data #>> '{}')::jsonb ->> 'darCode' = c.dar_code
+      ) 
+      AND dar.collection_id IS NULL
+      AND dar.draft = false;
+    </sql>
```
rushtong, line 17: This looks like it will work in the case where there are multiple DARs for a collection and the first one was not updated - this will make that update.

I think there is a second case where no collection was ever created for a DAR against a single dataset. In that case, there won't be anything to select against in `FROM dar_collection c`. I think we need a second one that looks a lot like the original back-population sql. See `changelog-consent-73.0.xml` where we did some of this work.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -362,7 +362,7 @@ public DataAccessRequest cancelDataAccessRequest(String referenceId) {
                 if (idx == 0) {
                     DataAccessRequest alreadyExists = dataAccessRequestDAO.findByReferenceId(dataAccessRequest.getReferenceId());
                     if (Objects.nonNull(alreadyExists)) {
-                        dataAccessRequestDAO.updateDraftByReferenceId(dataAccessRequest.getReferenceId(), false);
+                        dataAccessRequestDAO.updateDraftForCollection(collectionId, dataAccessRequest.getReferenceId());
```
rushtong, line 365: Nice, clean fix 👍🏽

rushtong (approved):
I'm good with this as is, but I think we need another round of back-population.

src/main/resources/changesets/changelog-consent-76.0.xml
JVThomas: No it works for single datasets as well. 

- `CONCAT((dar.data #>> '{}')::jsonb ->> 'darCode', '-') ~* c.dar_code` is for multi datasets (format is DAR-[digits]-A..., do a similar/Like check against collection dar_code with hyphen appended)
- `(dar.data #>> '{}')::jsonb ->> 'darCode' = c.dar_code` is for single datasets, given the context that the prior condition is false. (Above failing means no multiset format was found, perform an exact check for single dataset format of DAR-[digits]

The only thing it should fail on is if the DAR doesn't have a `darCode` in the `data` column, which should only happen on bugged out DARs that existed earlier in the year due to an old race condition in the DAR submission process (if I remember correctly).

PR #1334, opened 2021-11-03
[DUOS-1457][risk=no] SendGrid health check

MiriamAlter:
- Updated status endpoint to add health check data from SendGrid
- Added unit test for said health check, accounting for internal and external failures
- Minor typo fixes

Issue: https://broadworkbench.atlassian.net/browse/DUOS-1457

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/ConsentApplication.java
```diff
@@ -28,13 +28,12 @@
 import org.broadinstitute.consent.http.authentication.DefaultAuthenticator;
 import org.broadinstitute.consent.http.authentication.OAuthAuthenticator;
 import org.broadinstitute.consent.http.authentication.OAuthCustomAuthFilter;
-import org.broadinstitute.consent.http.health.GCSHealthCheck;
+import org.broadinstitute.consent.http.health.*;
```
rushtong: We don't use star imports, please revert this change and import your health check directly.

rushtong: See https://blog.marcnuri.com/intellij-idea-how-to-disable-wildcard-imports or https://stackoverflow.com/questions/3348816/intellij-never-use-wildcard-imports for how to do this in intellij

src/test/java/org/broadinstitute/consent/http/health/SendGridHealthCheckTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.entity.StringEntity;
+import org.broadinstitute.consent.http.configurations.MailConfiguration;
+import org.broadinstitute.consent.http.util.HttpClientUtil;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+import static org.junit.Assert.*;
```
rushtong: Once you adjust your IDE, re-optimizing imports should fix this.

src/main/java/org/broadinstitute/consent/http/ConsentApplication.java
MiriamAlter: Fixed

src/test/java/org/broadinstitute/consent/http/health/SendGridHealthCheckTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private SendGridHealthCheck healthCheck;
+
+    @Before
+    public void setUp() {
+        MockitoAnnotations.openMocks(this);
+    }
+
+    private String correctEntity = "{\"page\":{\"id\":\"3tgl2vf85cht\",\"name\":\"SendGrid\",\"url\":\"https://status.sendgrid.com\",\"time_zone\":\"America/Los_Angeles\",\"updated_at\":\"2021-11-03T04:01:21.355-07:00\"},\"status\":{\"indicator\":\"none\",\"description\":\"All Systems Operational\"}}";
+    private String incorrectEntity = "{\"page\":{\"id\":\"3tgl2vf85cht\",\"name\":\"SendGrid\",\"url\":\"https://status.sendgrid.com\",\"time_zone\":\"America/Los_Angeles\",\"updated_at\":\"2021-11-03T04:01:21.355-07:00\"},\"status\":{\"indicator\":\"none\",\"description\":\"FAILURE\"}}";
```
rushtong: If we fleshed out the `SendGridStatus` with more of the fields that we know are in it, these strings could be simplified with something like `new Gson().toJson(correctEntity)`

src/main/java/org/broadinstitute/consent/http/resources/StatusResource.java
```diff
@@ -58,14 +58,16 @@ public Response getStatus() {
             healthy = true;
         }
         formattedResults.put(OK, healthy);
-        HealthCheck.Result gcs = results.getOrDefault(ConsentApplication.GCS_CHECK, HealthCheck.Result.unhealthy("Unable to access Sam"));
-        HealthCheck.Result elasticSearch = results.getOrDefault(ConsentApplication.ES_CHECK, HealthCheck.Result.unhealthy("Unable to access Sam"));
-        HealthCheck.Result ontology = results.getOrDefault(ConsentApplication.ONTOLOGY_CHECK, HealthCheck.Result.unhealthy("Unable to access Sam"));
+        HealthCheck.Result gcs = results.getOrDefault(ConsentApplication.GCS_CHECK, HealthCheck.Result.unhealthy("Unable to access Google Cloud Storage"));
+        HealthCheck.Result elasticSearch = results.getOrDefault(ConsentApplication.ES_CHECK, HealthCheck.Result.unhealthy("Unable to access Elastic Search"));
+        HealthCheck.Result ontology = results.getOrDefault(ConsentApplication.ONTOLOGY_CHECK, HealthCheck.Result.unhealthy("Unable to access Ontology"));
```
rushtong, line 63: Nice catch!

src/main/java/org/broadinstitute/consent/http/configurations/MailConfiguration.java
```diff
@@ -8,13 +8,15 @@
...
 
     @NotNull
-    String googleAccount;
+    private String googleAccount;
 
     @NotNull
-    String sendGridApiKey;
+    private String sendGridApiKey;
+
+    private String sendGridStatusUrl;
```
rushtong: For consistency, we should make this not null as well
```suggestion
    @NotNull
    private String sendGridStatusUrl;
```

automation/configs/consent-config.yml
```diff
@@ -35,6 +35,7 @@ mailConfiguration:
   activateEmailNotifications: false
   googleAccount: test@gmail.com
   sendGridApiKey: test-key
+  sendGridStatusUrl: https://status.sendgrid.com/api/v2/status.json
```
rushtong, line 38: This should be a dummy value, not the real one. Look at the Sam url line, that should give you some guidance.

src/main/java/org/broadinstitute/consent/http/health/SendGridHealthCheck.java
```diff
@@ -0,0 +1,68 @@
...
+    @Override
+    protected Result check() throws Exception {
+        try {
+            String statusUrl = configuration.getSendGridStatusUrl();
+            HttpGet httpGet = new HttpGet(statusUrl);
+            try (CloseableHttpResponse response = clientUtil.getHttpResponse(httpGet)) {
+                if (response.getStatusLine().getStatusCode() == HttpStatusCodes.STATUS_CODE_OK) {
+                    String content = IOUtils.toString(response.getEntity().getContent(), Charset.defaultCharset());
+                    SendGridStatus sgStatus = new Gson().fromJson(content, SendGridStatus.class);
+                    if (sgStatus.status.description.equalsIgnoreCase("All Systems Operational")) {
```
rushtong: Some points to consider
1. Since we now know more about what constitutes a healthy status (see https://status.sendgrid.com/api#status), we can be more accurate about what we're checking. This should be looking at the indicator property for one of these values `none, minor, major, or critical`. 
2. You have defined an object, `SendGridStatus`, that should represent the complete response from Sendgrid. We should flesh that out some more for completeness.
3. The logic of `Am I healthy?` really belongs inside the object itself. We should be parsing a `SendGridStatus` and then getting the boolean healthy value from that object. It's just a calculation once the full object is parsed.

src/main/java/org/broadinstitute/consent/http/health/SendGridHealthCheck.java
```diff
@@ -0,0 +1,68 @@
...
+
+    private static class SendGridStatus {
+        Object page;
+        StatusObject status;
+    }
+
+    private static class StatusObject {
+        String indicator;
+        String description;
+    }
```
rushtong: These two classes should likely be extracted into a public `SendGridStatus` class with inner classes to represent the nested objects. You could put that in the models package, or right here next to this class since it is so tightly coupled. 

`indicator` should likely be an enum of the values we're expecting, ie. `none, minor, major, or critical`

src/main/java/org/broadinstitute/consent/http/health/SendGridHealthCheck.java
```diff
@@ -0,0 +1,68 @@
...
+            String statusUrl = configuration.getSendGridStatusUrl();
+            HttpGet httpGet = new HttpGet(statusUrl);
+            try (CloseableHttpResponse response = clientUtil.getHttpResponse(httpGet)) {
+                if (response.getStatusLine().getStatusCode() == HttpStatusCodes.STATUS_CODE_OK) {
+                    String content = IOUtils.toString(response.getEntity().getContent(), Charset.defaultCharset());
+                    SendGridStatus sgStatus = new Gson().fromJson(content, SendGridStatus.class);
+                    if (sgStatus.status.description.equalsIgnoreCase("All Systems Operational")) {
+                        return Result.builder()
+                                .withDetail("pages", sgStatus.page)
+                                .withDetail("status", sgStatus.status)
```
rushtong: Try to use getters for more idiomatic Java. If the class doesn't have them, please add them. Direct field references are great in other languages, less so in Java.
```suggestion
                                .withDetail("pages", sgStatus.getPage())
                                .withDetail("status", sgStatus.getStatus())
```

src/main/java/org/broadinstitute/consent/http/health/SendGridHealthCheck.java
```diff
@@ -0,0 +1,68 @@
...
+                                .healthy()
+                                .build();
+                    } else {
+                        return Result.unhealthy("SendGrid status is unhealthy: " + sgStatus.status.description);
```
rushtong: Same getter comment. It's also possible that either status or description is null, so you'll need to handle potential NPEs here.

src/main/java/org/broadinstitute/consent/http/health/SendGridHealthCheck.java
```diff
@@ -0,0 +1,68 @@
...
+                    }
+                } else {
+                    return Result.unhealthy("SendGrid status is unhealthy: " + response.getStatusLine());
+                }
+            } catch (Exception e) {
+                return Result.unhealthy(e);
+            }
+        } catch (Exception e) {
+            return Result.unhealthy(e);
+        }
```
rushtong, line 42: I realize that you're copying the pattern from the Sam check here, but we should strive to reduce the nested try/catch blocks. This could be condensed down with some pre-checks on values to ensure that we are throwing the right error in the right place.

rushtong (changes requested):
See comments inline. At present, this isn't functional for me (could be something I missed in configuration), but there is room for improvement beyond that.

automation/configs/consent-config.yml
rushtong: Ignore this comment, this does need to be here since it is for automation. I mistakenly thought this was for the test config, which it isn't.

src/main/java/org/broadinstitute/consent/http/health/SendGridHealthCheck.java
rushtong: I'm changing my mind on this ... I think we really do need the nested try/catch because of the HttpGet's attempt to coerce the status url to a URI.

src/main/java/org/broadinstitute/consent/http/health/SendGridStatus.java
```diff
@@ -0,0 +1,55 @@
+package org.broadinstitute.consent.http.health;
+
+import com.codahale.metrics.health.HealthCheck.Result;
+import com.google.gson.Gson;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class SendGridStatus {
+    private Object page;
```
rushtong, line 6: We don't really need `page` for any logic, so it makes sense that this is just a plain object. Generally, I try to avoid using `Object` - I like clarity in the code and `Object` tells me literally nothing about what this field is really about. If you wanted to go the extra step, we could add an inner class in `SendGridStatus` that maps out the values in `page`. Consider this optional feedback.

src/main/java/org/broadinstitute/consent/http/health/SendGridStatus.java
```diff
@@ -0,0 +1,55 @@
...
+    private Object status;
```
rushtong: Here, we do need more clarity around what a `status` really is so this should be a class with `indicator` and `description` fields. We also discussed `indicator` being an enum so we don't have to do raw string comparisons.

src/main/java/org/broadinstitute/consent/http/health/SendGridStatus.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public Object getStatus() {
+        return status;
+    }
+
+    public void setStatus(String indicator, String description) {
+        Map<String, String> statusMap = new HashMap<>();
+        statusMap.put("indicator", indicator);
+        statusMap.put("description", description);
+        status = new Gson().toJson(statusMap);
```
rushtong: See comment about what `status` really should be. In this example, status is just a `String`, which adds unnecessary complexity in your private methods `getStatusMap()` and `isOk()`

src/main/java/org/broadinstitute/consent/http/health/SendGridStatus.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    private Map<String, String> getStatusMap() {
+        return new Gson().fromJson(getStatus().toString(), Map.class);
+    }
```
rushtong: This is an example of something that should stick out in your mind when writing code. This map is, in essence, a sort of temporary variable. Temp variables are very often replaceable by proper object modeling. In this case, if the `status` field were a proper class with an indicator field, you wouldn't need this method at all.

src/main/java/org/broadinstitute/consent/http/health/SendGridStatus.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public boolean isOk() {
+        return getStatusMap().get("indicator").equalsIgnoreCase("none");
+    }
+
+    public Result getResult() {
```
rushtong, line 25: I like this a lot ... very isolated and concise.

src/main/java/org/broadinstitute/consent/http/health/SendGridHealthCheck.java
```diff
@@ -0,0 +1,50 @@
...
+    @Override
+    protected Result check() throws Exception {
+        try {
+            String statusUrl = configuration.getSendGridStatusUrl();
+            HttpGet httpGet = new HttpGet(statusUrl);
+            try (CloseableHttpResponse response = clientUtil.getHttpResponse(httpGet)) {
+                if (response.getStatusLine().getStatusCode() == HttpStatusCodes.STATUS_CODE_OK) {
+                    String content = IOUtils.toString(response.getEntity().getContent(), Charset.defaultCharset());
+                    SendGridStatus status = new Gson().fromJson(content, SendGridStatus.class);
+                    return status.getResult();
```
rushtong, line 33: Isolating the logic inside `status` makes this much cleaner 👍🏽

src/main/java/org/broadinstitute/consent/http/health/SendGridStatus.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public void setStatus(String indicator, String description) {
+        Map<String, String> statusMap = new HashMap<>();
+        statusMap.put("indicator", indicator);
+        statusMap.put("description", description);
+        status = new Gson().toJson(statusMap);
+    }
+
+    private Map<String, String> getStatusMap() {
+        return new Gson().fromJson(getStatus().toString(), Map.class);
```
rushtong: When I run this locally, this line is triggering an exception.

rushtong (changes requested):
Looking better! See comments inline. This version throws a json parsing exception. Please make sure to test thoroughly before submitting, thank you!

src/main/java/org/broadinstitute/consent/http/health/SendGridStatus.java
```diff
@@ -0,0 +1,68 @@
...
+        return page;
+    }
+
+    public void setPage(Object page) {
+        this.page = page;
+    }
+
+    public String getStatus() {
+        return new Gson().toJson(status);
+    }
```
rushtong: This looks unused now. Alternatively, this should probably be
```
    public StatusObject getStatus() {
        return status;
    }
```

src/main/java/org/broadinstitute/consent/http/health/SendGridStatus.java
```diff
@@ -0,0 +1,68 @@
...
+
+    public void setStatus(String indicator, String description) {
+        status = new StatusObject();
+        status.setIndicator(Indicator.valueOf(indicator));
+        status.setDescription(description);
+    }
```
rushtong: In more idiomatic java, one would have a constructor for `StatusObject` that takes the two arguments. Even better, the setter for this should probably take a `StatusObject` to begin with so you don't have to do the manual construction here. The only usages of this method are in tests.
```
    public void setStatus(StatusObject status) {
        this.status = status;
    }
```
I think this is leftover fallout from how this class was originally trying to do too much work when calculating the `Result` object. That refactoring has left some of the less conventional object construction around which means there's a little more cleanup to do.

rushtong (changes requested):
Looking pretty good! There's just a little bit of cleanup left to do here.

src/main/java/org/broadinstitute/consent/http/health/SendGridStatus.java
```diff
@@ -0,0 +1,71 @@
+package org.broadinstitute.consent.http.health;
+
+import com.codahale.metrics.health.HealthCheck.Result;
+import com.google.gson.Gson;
```
rushtong: Unused import

rushtong (approved):
See one comment inline. This looks good! Feel free to fix the import and then merge after that.

PR #1358, opened 2021-11-16
[DUOS-1511][risk=no] Test coverage for resource classes (Part 1)

MiriamAlter:
- All new tests for DataRequestReportsResource and NihAccountResource
- Recoded existing tests for ConsentVotesResource and DatasetResource

Not included:
- New tests for DatasetResource
- New tests for DataRequestVoteResource

Issue: https://broadworkbench.atlassian.net/browse/DUOS-1511

----

- Verify all tests go green

### DIFF
--- a/src/test/java/org/broadinstitute/consent/http/resources/ConsentVoteResourceTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/resources/ConsentVoteResourceTest.java
@@ -13,11 +13,13 @@ import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
 
 import javax.ws.rs.core.Response;
+import java.io.IOException;
 import java.util.UUID;
 
 import static org.junit.Assert.assertEquals;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.when;
 
 public class ConsentVoteResourceTest {
@@ -40,14 +42,20 @@ public class ConsentVoteResourceTest {
         resource = new ConsentVoteResource(emailNotifierService, electionService, voteService);
     }
 
-    @Test
-    public void testCreateConsentVote() throws Exception {
-        Consent consent = new Consent();
-        consent.setConsentId(UUID.randomUUID().toString());
+    private Vote createTestVote() {
         Vote vote = new Vote();
         vote.setVoteId(RandomUtils.nextInt(100, 1000));
         vote.setVote(false);
         vote.setRationale("Test");
+        return vote;
+    }
+
+    @Test
+    public void testCreateConsentVoteSuccess() throws Exception {
+//        Consent consent = new Consent();
+//        consent.setConsentId(UUID.randomUUID().toString());
+        // Are these lines necessary? The test works without them.
+        Vote vote = createTestVote();
         when(voteService.updateVoteById(any(), any())).thenReturn(vote);
         when(electionService.validateCollectEmailCondition(any())).thenReturn(true);
         doNothing().when(emailNotifierService).sendCollectMessage(any());
@@ -55,6 +63,20 @@ public class ConsentVoteResourceTest {
 
         Response response = resource.firstVoteUpdate(vote, UUID.randomUUID().toString(), vote.getVoteId());
         assertEquals(200, response.getStatus());
+    }
+
+    @Test
+    public void testCreateConsentVoteCollectMessageError() throws Exception {
+        Vote vote = createTestVote();
+        when(voteService.updateVoteById(any(), any())).thenReturn(vote);
+        when(electionService.validateCollectEmailCondition(any())).thenReturn(true);
+        doThrow(new IOException()).when(emailNotifierService).sendCollectMessage(any());
+        initResource();
+
+        Response response = resource.firstVoteUpdate(vote, UUID.randomUUID().toString(), vote.getVoteId());
+        assertEquals(200, response.getStatus());
+
+        // This test is passing, and it looks like it shouldn't. Does something need to be reworked in the resource class?
     }
 
 }
--- a/src/test/java/org/broadinstitute/consent/http/resources/ConsentVoteResourceTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/resources/ConsentVoteResourceTest.java
@@ -10,7 +10,6 @@ import org.broadinstitute.consent.http.service.VoteService;
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.Mock;
-import org.mockito.MockitoAnnotations;
 
 import javax.ws.rs.core.Response;
 import java.io.IOException;
@@ -21,6 +20,7 @@ import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.doNothing;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.when;
+import static org.mockito.MockitoAnnotations.openMocks;
 
 public class ConsentVoteResourceTest {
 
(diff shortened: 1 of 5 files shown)


src/test/java/org/broadinstitute/consent/http/resources/ConsentVoteResourceTest.java
```diff
@@ -30,31 +32,125 @@
...
+    public void testCreateConsentVoteCollectMessageError() throws Exception {
+        when(voteService.updateVoteById(any(), any())).thenReturn(vote);
+        when(electionService.validateCollectEmailCondition(any())).thenReturn(true);
+        doThrow(new IOException()).when(emailNotifierService).sendCollectMessage(any());
+        initResource();
+
+        Response response = resource.firstVoteUpdate(vote, consent.getConsentId(), vote.getVoteId());
         assertEquals(200, response.getStatus());
+
+        // This test is passing, and it looks like it shouldn't. Does something need to be reworked in the resource class?
```
rushtong: This scenario looks like this should pass, which it does. There is a handler for catching IOException that intentionally does NOT trigger a downstream exception to the client. The catch logic goes all the way back to 10/21/2015 and is likely there to prevent unrelated email errors from being shown to the caller.

src/test/java/org/broadinstitute/consent/http/resources/ConsentVoteResourceTest.java
```diff
@@ -30,31 +32,125 @@
...
-        MockitoAnnotations.initMocks(this);
+        openMocks(this);
+
+        consent = new Consent();
+        consent.setConsentId(UUID.randomUUID().toString());
+
+        vote = new Vote();
+        vote.setVoteId(RandomUtils.nextInt(100, 1000));
+        vote.setVote(false);
+        vote.setRationale("Test");
```
rushtong: One thing that I find helpful is to have a generator function instead of using global class variables. What happens over time is that the global var needs to be modified for a new test, which can then unintentionally affect other tests. Instead, if we create a private generator function that works for the current cases, then a new test that needs new info on the same object can call that, modify what it needs, and not break any other tests in the process.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -95,47 +108,78 @@ public void testCreateDataset() throws Exception {
...
+    public void testCreateDatasetNoProperties() {
+        initResource();
+        Response response = resource.createDataset(authUser, uriInfo, "{\"properties\":[]}");
+        assertEquals(400, response.getStatus());
+    }
+
+    @Test(expected = BadRequestException.class)
+    public void testCreateDatasetMissingName() {
+        String json = createPropertiesJson("Dataset Name", null);
+        // TODO: Rewrite createDataset to handle empty name strings and entirely missing name properties
```
rushtong: This comment should be moved to two separate tests, something like `testCreateDatasetEmptyName` and `testCreateDatasetMissingName`. Each should test the class as it is and list a ticket/TODO to address in a subsequent PR. 

Additionally, this test should be renamed to `testCreateDatasetNullName` since it's not really a missing property, just null.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -95,47 +108,78 @@ public void testCreateDataset() throws Exception {
...
-              "{\"properties\":[{\"propertyName\":\"Species\",\"propertyValue\":\"test\"}]}");
-        assertEquals(400, responseMissingDatasetName.getStatus());
+        String json = createPropertiesJson(duplicateProperties);
 
-        Response responseInvalidProperty = resource.createDataset(authUser, uriInfo,
-              "{\"properties\":[{\"propertyName\":\"Invalid Property\",\"propertyValue\":\"test\"}]");
-        assertEquals(400, responseInvalidProperty.getStatus());
+        initResource();
+        Response response = resource.createDataset(authUser, uriInfo, json);
+        assertEquals(400, response.getStatus());
```
rushtong: This test never reaches this line since line 163, `resource.createDataset`, throws the expected exception.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -95,47 +108,78 @@ public void testCreateDataset() throws Exception {
...
+        when(datasetService.getDatasetByName("test")).thenReturn(inUse);
+
+        String json = createPropertiesJson("Dataset Name", "test");
 
-        Response responseNameInUse = resource.createDataset(authUser, uriInfo,
-              "{\"properties\":[{\"propertyName\":\"Dataset Name\",\"propertyValue\":\"test\"}]}");
-        assertEquals(409, responseNameInUse.getStatus());
+        initResource();
+        Response response = resource.createDataset(authUser, uriInfo, json);
+        assertEquals(409, response.getStatus());
```
rushtong: Similar comment - unreachable line.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -1,24 +1,7 @@
...
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import javax.ws.rs.BadRequestException;
-import javax.ws.rs.NotFoundException;
-import javax.ws.rs.core.Response;
-import javax.ws.rs.core.UriBuilder;
-import javax.ws.rs.core.UriInfo;
+import liquibase.pro.packaged.D;
```
rushtong: unused

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -95,47 +108,78 @@ public void testCreateDataset() throws Exception {
...
+        initResource();
+        Response response = resource.createDataset(authUser, uriInfo, "");
+        assertEquals(400, response.getStatus());
+    }
+
+    @Test(expected = BadRequestException.class)
+    public void testCreateDatasetNoProperties() {
+        initResource();
+        Response response = resource.createDataset(authUser, uriInfo, "{\"properties\":[]}");
+        assertEquals(400, response.getStatus());
```
rushtong: unreachable code.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -95,47 +108,78 @@ public void testCreateDataset() throws Exception {
...
+    @Test(expected = BadRequestException.class)
+    public void testCreateDatasetMissingName() {
+        String json = createPropertiesJson("Dataset Name", null);
+        // TODO: Rewrite createDataset to handle empty name strings and entirely missing name properties
 
         initResource();
-        Response responseNoJson = resource.createDataset(authUser, uriInfo, "");
-        assertEquals(400, responseNoJson.getStatus());
+        Response response = resource.createDataset(authUser, uriInfo, json);
+        assertEquals(400, response.getStatus());
```
rushtong: unreachable.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -95,47 +108,78 @@ public void testCreateDataset() throws Exception {
...
+    public void testCreateDatasetInvalidProperty() {
+        List<DataSetPropertyDTO> invalidProperties = new ArrayList<>();
+        invalidProperties.add(new DataSetPropertyDTO("Invalid Property", "test"));
+        when(datasetService.findInvalidProperties(any())).thenReturn(invalidProperties);
+
+        String json = createPropertiesJson(invalidProperties);
+
+        initResource();
+        Response response = resource.createDataset(authUser, uriInfo, json);
+        assertEquals(400, response.getStatus());
```
rushtong: unreachable

rushtong (changes requested):
Mostly minor comments inline - looking good overall!

src/test/java/org/broadinstitute/consent/http/resources/ConsentVoteResourceTest.java
```diff
@@ -31,30 +37,179 @@
...
+    public void testCreateConsentVoteCollectMessageError() throws Exception {
+        Vote vote = createMockVote();
+        Consent consent = createMockConsent();
+
+        Logger logger = Logger.getLogger(ConsentVoteResource.class.getName());
+        LogHandler handler = new LogHandler();
+        handler.setLevel(Level.ALL);
+        logger.setUseParentHandlers(false);
+        logger.addHandler(handler);
+        logger.setLevel(Level.ALL);
```
rushtong, line 102: This is very nice. It would be super helpful to extract this into a default interface so other test classes can use the same approach. Would be happy to pair with you on this suggestion, but you can see a similar pattern in `WithMockServer`. That interface has default methods that can be called by any class that implements it. In this case, we might have a default method `default LogHandler setLoggerForClass(String className) { ... blah ... }` and then you can do the assert below with log handler. Of course, there are other ways of implementing this, feel free to explore as you see fit.

rushtong (approved):
Thank you for the changes. This is good to go as is, but feel free to look at my suggestion.

src/test/java/org/broadinstitute/consent/http/resources/ConsentVoteResourceTest.java
MiriamAlter: Good idea! Would you like to add this as a story to Jira or should I?

rushtong: I created DUOS-1515 for this.

PR #1362, opened 2021-11-22
[DUOS-1521][risk=no] Populate election's final vote

rushtong:
## Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DUOS-1521

Map the final vote value if we have one.
See also: https://github.com/DataBiosphere/duos-ui/pull/1361 where we handle the error case more explicitly.

----

- Verify all tests go green


quazi-h (approved):
Changes make sense, reading the debugging notes on the ticket was helpful. Am I understanding correctly that we are going to make sure that both the DAC Console and Researcher Console are both using the same Election logic here?

rushtong:
@quazi-broad 
> Am I understanding correctly that we are going to make sure that both the DAC Console and Researcher Console are both using the same Election logic here?

Correct. The UI PR does exactly that: https://github.com/DataBiosphere/duos-ui/pull/1361

PR #1366, opened 2021-11-29
[DUOS-511][risk=no] Refactor DataAccessRequest Summary Details

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-511
See also: https://github.com/DataBiosphere/duos-ui/pull/1365

Discovered an underlying bug in the election dao that returns duplicate election rows: https://broadworkbench.atlassian.net/browse/DUOS-1526

## Changes
* Extract `DataAccessRequestSummaryDetail` and `ConsentSummaryDetail` into value objects for easier construction and printing
* Handle the case of `N` DAC members flexibly
* Minor updates to api docs
* Updated tests
* Updated the resource class with a default selection and simplification

Before this update, dev returned only a small fraction of DARs and 0 Consents. After this update, all reviewed DARs and consent summaries are returned. We add extra columns for additional DAC Members when it is necessary. All rows need to know the max number of DAC Members so they can write out the correct number of columns. This was not handled correctly in the original code, but is in this version. This PR also fixes a number of missing `Agreement` votes. We don't need to look at the actual database value of votes of `Agreement` type, we can simply compare the match value to the chairperson final vote value to determine agreement.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
@@ -225,10 +225,6 @@ public interface ElectionDAO extends Transactional<ElectionDAO> {
             "left join vote v on v.electionId = e.electionId and lower(v.type) = 'chairperson' ")
      List<Election> findLastElectionsWithFinalVoteByReferenceIdsTypeAndStatus(@BindList("referenceIds") List<String> referenceIds, @Bind("status") String status);
 
-     @RegisterRowMapper(AccessRPMapper.class)
-     @SqlQuery("select * from access_rp where electionAccessId in (<electionAccessIds>) ")
-     List<AccessRP> findAccessRPbyElectionAccessId(@BindList("electionAccessIds") List<Integer> electionAccessIds); 
-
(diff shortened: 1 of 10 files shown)


JVThomas:
I noticed some odd formatting with my local version of staging

Screen Shot 2021-11-30 at 1 52 31 PM

Upon further inspection it turned out that the rationale comment (Vote) had a blank line separating the two sentences. My guess is that may have been saved and/or interpreted as a newline character which, in turn, would adversely affect the table formatting.

It might be a good idea to strip tabs and blank lines from values that are tied to free text inputs to keep the table format intact since it's all being added to StringBuilder

rushtong:
> It might be a good idea to strip tabs and blank lines from values that are tied to free text inputs to keep the table format intact since it's all being added to StringBuilder

Great idea ... will do that for those kinds of fields.

rushtong:
Latest commit handles line breaks and tabs in user input. However, I see a duplicate row bug that I missed.
Screen Shot 2021-12-01 at 8 19 20 AM

rushtong:
Latest commit fixes the duplicate election bug while preserving the final vote.
Screen Shot 2021-12-01 at 8 59 24 AM

JVThomas (approved):
Code and table outputs look good (tested locally)

PR #1374, opened 2021-12-06
[DUOS-1290][risk=medium] Multi-vote API

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1290

## Changes
* New vote resource specifically to handle votes
* Bulk vote update endpoint for voting on a list of vote ids
* Update election to closed status for votes of "FINAL" type
* New pattern for more complex DAO calls that require transactions across multiple db operations

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/resources/VoteResource.java
```diff
@@ -0,0 +1,102 @@
...
+     * @param vote The boolean value to update votes for
+     * @param rationale The string value for all rationales
+     * @param json The list of vote ids, in json format
+     * @return Response with results of the update.
+     */
+    @PUT
+    @Consumes("application/json")
+    @Produces("application/json")
+    @RolesAllowed({CHAIRPERSON, MEMBER})
+    public Response updateVotes(
```
JVThomas, line 57: Question: What's the plan on updating the election status on a Chairperson vote? Will the front-end issue a separate back-end request to update Election, or should it be handled through this endpoint and/or its associated service methods?

src/main/java/org/broadinstitute/consent/http/resources/VoteResource.java
```diff
@@ -0,0 +1,102 @@
+package org.broadinstitute.consent.http.resources;
+
+
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
+import io.dropwizard.auth.Auth;
+import org.broadinstitute.consent.http.enumeration.ElectionStatus;
```
JVThomas: A couple of these imports aren't being used, namely ElectionStatus, Election, and Collectors

src/main/java/org/broadinstitute/consent/http/resources/VoteResource.java
rushtong: Excellent question and I think this is something I missed. I don't see a ticket for this in the epic, so will do some thinking on what to do here. I'm leaning toward updating the election based on the types of votes passed in.

JVThomas: It's the only gap in the PR right now. Functionally the API makes sense, and the process checks to see if the election is open before updating the votes (so a chairperson can't change their vote if the election is already closed). However if the election is open a chairperson can change their vote however many times they want.

rushtong: Addressed in the service class

rushtong:
#### Edit: This seems to be resolved by re-running the action.

Note that these failing tests in the coverage action are passing locally for me. Will do some debugging, but it is consistently happening in the action.
```
Error:  Errors: 
Error:    UserServiceTest.testCreateUserDuplicateEmail »  Unexpected exception, expected...
Error:    UserServiceTest.testGetUsersAsRoleAdmin:414 » IllegalArgument duplicate elemen...
```

JVThomas (approved):
Code looks good. The election updates now prevent any vote changes after `Final` vote submission.

quazi-h (approved):
This PR looks good to me. I was able to follow the logic in VoteResource and VoteServiceDAO and it looks like it is adding the functionality that the ticket asks for. Tests look good as well. Sorry for the delay, I ended up reading up on DAOs since I was unfamiliar with the concept.

PR #1384, opened 2021-12-10
[DUOS-1534][risk=no] DatasetResource optimizing

MiriamAlter:
- createDataset now returns a BadRequestException when dataset name is empty or missing
- getDataSetSample no longer requires a file

Issue: https://broadworkbench.atlassian.net/browse/DUOS-1534

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -203,15 +209,16 @@ public Response validateDatasetName(@QueryParam("name") String name) {
     public Response getDataSetSample() {
         String msg = "GETting Data Set Sample";
         logger().debug(msg);
-        String fileName = "DataSetSample.tsv";
+        String dataSetSample = "Dataset Name\tData Type\tSpecies\tPhenotype/Indication\t# of participants\tDescription\tdbGAP\tData Depositor\tPrincipal Investigator(PI)\tSample Collection ID\tConsent ID"
+                + "\n(Bucienne Monco) - Muc-1 Kidney Disease\tDNA, whole genome\thuman\tmuc-1, kidney disease\t31\tmuc-1 patients that developed cancer , 5 weeks after treatment\thttp://....\tJohn Doe\tMark Smith\tSC-20658\t1";
```
rushtong: This could be moved to a class level field with getter/setter so it could be customized by a test

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -203,15 +209,16 @@ public Response validateDatasetName(@QueryParam("name") String name) {
...
         InputStream inputStream = null;
         try {
-            inputStream = Resources.getResource(fileName).openStream();
+            inputStream = new ByteArrayInputStream(dataSetSample.getBytes());
         } catch (Exception e) {
             logger().error("Error when GETting dataset sample. Cause: " + e);
             return createExceptionResponse(e);
         }
-        return Response.ok(inputStream).header("Content-Disposition", "attachment; filename=" + fileName).build();
+        return Response.ok(inputStream).header("Content-Disposition", "attachment; filename=DataSetSample.tsv").build();
```
rushtong: `DataSetSample.tsv` could also be moved to a class level field so it could be customized by a test.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -656,7 +653,7 @@ public void testUpdateNeedsReviewDataSetsError() {
 
     @Test
     public void testDownloadDatasetApprovedUsersSuccess() {
-        List header = List.of("attachment; filename =DatasetApprovedUsers.tsv");
+        List<String> header = List.of("attachment; filename =DatasetApprovedUsers.tsv");
```
rushtong: This is actually replicating a bug we have defined somewhere in the backlog. The space in `filename =DatasetApprovedUsers.tsv` breaks the actual download.

rushtong (changes requested):
Looking good, minor comments inline. Look for the file name related bug in the backlog and make sure that is marked as in progress/completed after this is merged.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
MiriamAlter: That's odd. There's no space in the header definition, but the test shows that it's there in the response. It fails when there is no space. Should I do something about this?

rushtong: Yes, it needs to be fixed. 
See: 
* https://stackoverflow.com/questions/93551/how-to-encode-the-filename-parameter-of-content-disposition-header-in-http
* https://broadworkbench.atlassian.net/browse/DUOS-1319

It looks like 1319 is resolved, but the test here is clearly not working as it should. There should be no space between`filename` and `=`.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -59,6 +60,35 @@
...
+        return dataSetSampleContent;
+    }
+
+    void setDataSetSampleContent(String content) {
+        dataSetSampleContent = content;
+    }
+
+    void resetDataSetSampleContent() {
+        dataSetSampleContent = "Dataset Name\tData Type\tSpecies\tPhenotype/Indication\t# of participants\tDescription\tdbGAP\tData Depositor\tPrincipal Investigator(PI)\tSample Collection ID\tConsent ID"
+                + "\n(Bucienne Monco) - Muc-1 Kidney Disease\tDNA, whole genome\thuman\tmuc-1, kidney disease\t31\tmuc-1 patients that developed cancer , 5 weeks after treatment\thttp://....\tJohn Doe\tMark Smith\tSC-20658\t1";
```
rushtong: Instead of copying this twice, make it a private static once. Then, lines 64 and 88 can each point to the same constant.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -91,8 +125,13 @@ public Response createDataset(@Auth AuthUser authUser, @Context UriInfo info, St
         if (duplicateProperties.size() > 0) {
             throw new BadRequestException("Dataset contains multiple values for the same property.");
         }
```
rushtong, line 127: This could catch any geenric exception, not just this one

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -59,12 +60,45 @@
...
+    private String dataSetSampleFileName;
+    private String dataSetSampleContent;
+
+    String getDataSetSampleFileName() {
+        return dataSetSampleFileName;
+    }
+
+    void setDataSetSampleFileName(String fileName) {
+        dataSetSampleFileName = fileName;
+    }
```
rushtong, line 76: "DataSetSample.tsv" could be a private constant in this file so you don't have to copy the string here and on line 63.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -91,8 +125,13 @@ public Response createDataset(@Auth AuthUser authUser, @Context UriInfo info, St
         if (duplicateProperties.size() > 0) {
             throw new BadRequestException("Dataset contains multiple values for the same property.");
         }
-        String name = inputDataset.getPropertyValue("Dataset Name");
-        if (Objects.isNull(name)) {
+        String name = "";
+        try {
+            name = inputDataset.getPropertyValue("Dataset Name");
+        } catch (IndexOutOfBoundsException e) {
```
rushtong, line 131: This could catch any generic exception, not just this one

rushtong (approved):
Looks good!

PR #1388, opened 2021-12-14
[DUOS-1540][risk=no] Clean up automation DAR calls

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1540

Automation test cleanup

----

- Verify all tests go green


quazi-h (approved):
Cleanup looks good. What is the significance of the "Chain" files? I was poking around older commits and it was noted that 'chains structure that allows for reuse of different ChainBuilder objects across different scenarios'. It looks like the methods in those files are used to set up test cases and scenarios but those functions don't exist outside of testing.

rushtong:
> What is the significance of the "Chain" files?

Great question @quazi-broad. You're right that a "chain" represents a discrete block of functionality that can be re-used in different scenarios. All of this code is "test" code so the idea behind all of this is to simulate what happens in the UI when a user follows a particular path. These were constructed with the idea that we would be able to load test a particular user journey simulating N users across a limited time frame.

PR #1401, opened 2022-01-05
[DUOS-1307][risk=low] Allow for Admins, Chairs, to cancel DAR Collection & Associated Elections

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1307

## Changes
* Updated existing researcher only endpoint to be accessible to Admins, Chairs
* Added optional role name param so Admins can cancel as a Chair, or as a Researcher
* Split service logic into three methods since they have very different effects
  * Researchers can only cancel DARs when there are no elections
  * Chairs can only cancel DARs+elections that are against datasets in their DACs
  * Admins can cancel all of the above.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
@@ -51,9 +51,12 @@ public class DataAccessRequest {
     this.elections = new HashMap<>();
   }
 
-  public static boolean isNotCanceled(DataAccessRequest dar) {
-    String status = dar.getData().getStatus();
-    return Objects.isNull(status) || !status.equalsIgnoreCase("canceled");
+  public static boolean isCanceled(DataAccessRequest dar) {
+    return
+      Objects.nonNull(dar) &&
+      Objects.nonNull(dar.getData()) &&
+      Objects.nonNull(dar.getData().getStatus()) &&
+      dar.getData().getStatus().equalsIgnoreCase("canceled");
   }
 
   public Integer getId() {
--- a/src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
@@ -53,7 +53,7 @@ public class DataAccessRequest {
 
   public static boolean isNotCanceled(DataAccessRequest dar) {
     String status = dar.getData().getStatus();
-    return Objects.isNull(status) || status.toLowerCase() != "canceled";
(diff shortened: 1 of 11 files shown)


src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
```diff
@@ -53,7 +53,7 @@ public DataAccessRequest() {
 
   public static boolean isNotCanceled(DataAccessRequest dar) {
```
gabelmattos: While it's not a change in this PR, it's usually best practice to do `isCancelled` as opposed to `isNotCancelled`. When you do an if statement based on that, you want to negate the if statement and not the variable itself.

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
```diff
@@ -148,18 +150,40 @@ public Response getCollectionsByToken(
...
+
+      // Default to the least impactful role if none provided.
+      UserRoles actingRole = UserRoles.RESEARCHER;
+      if (Objects.nonNull(roleName)) {
+        validateUserHasRoleName(user, roleName);
+        actingRole = UserRoles.getUserRoleFromName(roleName);
+      }
+
+      DarCollection cancelledCollection;
+      switch (Objects.requireNonNull(actingRole)) {
```
gabelmattos: This looks like a great example of something we could potentially leverage abstract classes in the factory.

gabelmattos (approved):
The tests look good.
I added a comment with the potential for reviewing it about negating on the if statement instead of on the function name itself, as it may generate confusion. You can choose to modify or not.

Either way, I'm ok with it.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
rushtong: That's a great point - I will look into how much of a refactor it is. If it's small, will handle it here.

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
rushtong: `Objects.requireNonNull` was actually an intellij suggestion mostly due to the fact that on line 170, `actingRole` has the potential to be overwritten with a null value. I think I'll clean that up a little more to prevent the need for this.

JVThomas (approved):
Tested locally, was able to cancel as expected based on user role regardless of the user's role combinations

rushtong:
Thank you @gabedemattos and @JVThomas!

PR #1406, opened 2022-01-11
[DUOS-1306][risk=medium] Create Elections & Votes for DarCollection

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1306

## Changes
Create new elections for a DarCollection
1. Creates only Access and RP elections, one pair per Data Access Request in the collection. We are ignoring potential dataset elections as that is a deprecated feature
2. Create appropriate votes for each election
3. Move all operations to a transaction block so any failures rolls everything back
4. Don't create elections if there are any other kinds of elections (aside from canceled) for the DARs. Canceled elections are the only case where we CAN create new elections for a collection.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
@@ -326,7 +326,7 @@ public class DarCollectionService {
       .filter(DataAccessRequest::isNotCanceled)
       .map(DataAccessRequest::getReferenceId)
       .collect(Collectors.toList());
-    if(!activeDarIds.isEmpty()) {
+    if (!activeDarIds.isEmpty()) {
       dataAccessRequestDAO.cancelByReferenceIds(activeDarIds);
     }
 }
--- a/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
@@ -227,7 +227,7 @@ public class DarCollectionService {
    *
    * If an election exists for a DAR within the collection, that DAR cannot be cancelled by the
    * researcher. Since it's now under DAC review, it's up to the DAC Chair (or admin) to
-   * ultimately decline or cancel the collection.
+   * ultimately decline or cancel the elections for the collection.
    *
    * @param collection The DarCollection
    * @return The canceled DarCollection
--- a/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
@@ -35,6 +35,7 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class DarCollectionService {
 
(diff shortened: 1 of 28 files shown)


src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -320,6 +332,55 @@ public DarCollection cancelDarCollectionElectionsAsChair(DarCollection collectio
...
+            logger.error("Unable to send new case message to DAC members for DAR: " + dar.getReferenceId());
+          }
+        } else {
+          logger.error("Did not find a created access election for DAR: " + dar.getReferenceId());
+        }
+      });
+      // TOL ... is this the right thing to do? When a chair/admin creates an election,
+      // should we really set the update user/date here? Thinking of ditching this, but if we do,
+      // then the user display could be affected where we sort by the sort date.
+      darCollectionDAO.updateDarCollection(collection.getDarCollectionId(), user.getDacUserId(), new Date());
```
rushtong: I'm especially looking for feedback on this choice.

src/main/java/org/broadinstitute/consent/http/service/dao/DarCollectionServiceDAO.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  public void createElectionsForDarCollection(DarCollection collection) throws SQLException {
+    final Date now = new Date();
+    jdbi.useHandle(
+        handle -> {
+          // By default, new connections are set to auto-commit which breaks our rollback strategy.
+          // Turn that off for this connection. This will not affect existing or new connections and
+          // only applies to the current one in this handle.
+          handle.getConnection().setAutoCommit(false);
```
quazi-h, line 38: This would allow the entire transaction block like you mentioned in the changes notes, right?

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
quazi-h: I'm not sure which would be the correct action here.
Once an election is created, is there a next step where the election passes or fails?
If we update the user/date there then I would think that you wouldn't need to update here since you are not changing anything about the DAR itself?

quazi-h (approved):
Changes make sense to me, creation of the Access and RP elections per DAR in the collection looks good.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
JVThomas: So I guess the question is whether or not we consider election generation a Collection update? IMO I don't think that we need to. If we need to track elections generated within a specific period of time, we can look at elections for that info. There's no changes to the Collection itself (DARs are the same), it's just updates to a different model that the collection has ties to.

rushtong: Thank you @JVThomas ... I like that line of thought, will make that change.

src/main/java/org/broadinstitute/consent/http/service/dao/DarCollectionServiceDAO.java
rushtong: In and of itself, no, it just makes it so each sql is not committed after execution. The rollback operation would happen if any of the individual sql statements fail which would happen on the `commit()` line.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rushtong: PR updated, PTAL, thank you!

PR #1445, opened 2022-02-07
DUOS-1477[risk=med] Support Paginated Dar Collection by Role Name Endpoint

JVThomas:
Addresses [DUOS-1477](https://broadworkbench.atlassian.net/browse/DUOS-1477)

PR adds endpoints for search query and pagination requests with role-based permissions in mind. As such PR includes updates to DAO, Service, and Resource classes with extensive updates to the relevant test classes (primarily the DAO tests).

At a high level the PR effectively introduces two new endpoints to the API, one for when a user initially searches by filter parameters or if search parameters are updated, the second being an endpoint that is utilized when the user changes the page.

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
@@ -241,14 +241,6 @@ public interface DatasetDAO extends Transactional<DatasetDAO> {
             " where c.dac_id = :dacId ")
     Set<DatasetDTO> findDatasetsByDac(@Bind("dacId") Integer dacId);
 
-    //TODO: write tests for this function, makes sure it works as expected
-    @SqlQuery(
-        " SELECT d.datasetid FROM dataset d " +
-        " INNER JOIN consentassociations a on a.dataSetId = d.dataSetId " +
-        " INNER JOIN consents c on c.consentId = a.consentId " +
-        " where c.dac_id IN (<dacIds>)")
-    List<Integer> findDatasetIdsByDacIds(@Bind("dacIds") List<Integer> dacIds);
-
     /**
      * DACs -> Consents -> Consent Associations -> DataSets
      * DataSets -> DatasetProperties -> Dictionary
--- a/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
@@ -241,6 +241,14 @@ public interface DatasetDAO extends Transactional<DatasetDAO> {
             " where c.dac_id = :dacId ")
     Set<DatasetDTO> findDatasetsByDac(@Bind("dacId") Integer dacId);
 
+    //TODO: write tests for this function, makes sure it works as expected
+    @SqlQuery(
+        " SELECT d.datasetid FROM dataset d " +
+        " INNER JOIN consentassociations a on a.dataSetId = d.dataSetId " +
+        " INNER JOIN consents c on c.consentId = a.consentId " +
+        " where c.dac_id IN (<dacIds>)")
+    List<Integer> findDatasetIdsByDacIds(@Bind("dacIds") List<Integer> dacIds);
+
     /**
      * DACs -> Consents -> Consent Associations -> DataSets
      * DataSets -> DatasetProperties -> Dictionary
--- a/src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
@@ -24,6 +24,23 @@ public class DarCollection {
 
   public static String defaultTokenSortField = "darCode";
 
+  //This query is specific to DAR Collections, which is why it's defined here
+  public static final String DAR_FILTER_QUERY_COLUMNS = 
+    "dar.id AS dar_id, dar.reference_id AS dar_reference_id, dar.collection_id AS dar_collection_id, " +
+      "dar.draft AS dar_draft, dar.user_id AS dar_userId, dar.create_date AS dar_create_date, " +
+      "dar.sort_date AS dar_sort_date, dar.submission_date AS dar_submission_date, " +
+      "dar.update_date AS dar_update_date, (dar.data #>> '{}')::jsonb AS data, " +
+      "(dar.data #>> '{}')::jsonb ->> 'projectTitle' as projectTitle";
+  
+  public static final String FILTER_TERMS_QUERY = 
+    "COALESCE(i.institution_name, '') ~* :filterTerm " +
+        " OR (dar.data #>> '{}')::jsonb ->> 'projectTitle' ~* :filterTerm " +
+        " OR u.displayname ~* :filterTerm " +
+        " OR c.dar_code ~* :filterTerm " +
+        " OR EXISTS " +
+        " (SELECT FROM jsonb_array_elements((dar.data #>> '{}')::jsonb -> 'datasets') dataset " +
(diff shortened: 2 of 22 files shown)


src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
```diff
@@ -241,6 +241,14 @@
             " where c.dac_id = :dacId ")
     Set<DatasetDTO> findDatasetsByDac(@Bind("dacId") Integer dacId);
 
+    //TODO: write tests for this function, makes sure it works as expected
```
rushtong: I don't see a test for this one. Would be happy to do that in a follow-on ticket/PR

src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
```diff
@@ -23,7 +23,25 @@
...
-
+  public static String defaultTokenSortOrder = "DESC";
+
+  //This query is specific to DAR Collections, which is why it's defined here
+  public static final String DAR_FILTER_QUERY_COLUMNS = 
+    "dar.id AS dar_id, dar.reference_id AS dar_reference_id, dar.collection_id AS dar_collection_id, " +
+      "dar.draft AS dar_draft, dar.user_id AS dar_userId, dar.create_date AS dar_create_date, " +
+      "dar.sort_date AS dar_sort_date, dar.submission_date AS dar_submission_date, " +
+      "dar.update_date AS dar_update_date, (dar.data #>> '{}')::jsonb AS data, " +
+      "(dar.data #>> '{}')::jsonb ->> 'projectTitle' as projectTitle";
```
rushtong: nit: Add a space to the end of this query to protect against other string additions combining it in this way: `projectTitlesomethingElse`

src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
```diff
@@ -23,7 +23,25 @@
...
+  
+  public static final String FILTER_TERMS_QUERY = 
+    "COALESCE(i.institution_name, '') ~* :filterTerm " +
+        " OR (dar.data #>> '{}')::jsonb ->> 'projectTitle' ~* :filterTerm " +
+        " OR u.displayname ~* :filterTerm " +
+        " OR c.dar_code ~* :filterTerm " +
+        " OR EXISTS " +
+        " (SELECT FROM jsonb_array_elements((dar.data #>> '{}')::jsonb -> 'datasets') dataset " +
+        " WHERE dataset ->> 'label' ~* :filterTerm)";
```
rushtong: nit: same idea here ... I can see this becoming `:filterTerm)AND ...`

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -133,48 +133,93 @@ public DarCollectionService(DarCollectionDAO darCollectionDAO, DarCollectionServ
...
+  }
+
+  //Helper function for queryCollectionsByFilterAndUserRoles
+  //Function verifies user role permission and executes query via DAO methods
+  private List<DarCollection> getFilteredCollectionsByUserRole(User user, String userRole, PaginationToken token, List<Integer> dacIds) {
+    String sortOrder = Objects.isNull(token.getSortDirection()) ? DarCollection.defaultTokenSortOrder : token.getSortDirection();
+    String sortField = Objects.isNull(token.getSortField()) ? DarCollection.defaultTokenSortField : token.getSortField();
+    String filterTerm = Objects.isNull(token.getFilterTerm()) ? "" : token.getFilterTerm();
+    List<DarCollection> collections;
+    if(userRole.equalsIgnoreCase(UserRoles.ADMIN.getRoleName())) {
```
rushtong: nit: A switch block is usually a lot easier to read and reason about than a series of if-else statements, please consider that approach

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -133,48 +133,93 @@ public DarCollectionService(DarCollectionDAO darCollectionDAO, DarCollectionServ
...
+            .map(UserRole::getDacId)
+            .distinct()
+            .collect(Collectors.toList());
+  }
+
+  //Helper method for queryCollectionsByFiltersAndUserRoles
+  //Verifies user role and determines unfiltered count total via DAO methods
+  private Integer getCountForUnfilteredQueryByRole(User user, String userRole) {
+    Integer size = 0;
+    if(userRole.equalsIgnoreCase(UserRoles.ADMIN.getRoleName())) {
```
rushtong: nit: similar idea here, please consider a switch block

src/test/java/org/broadinstitute/consent/http/db/DarCollectionDAOTest.java
```diff
@@ -15,44 +8,54 @@
...
 import java.util.stream.Collectors;
+
 import org.apache.commons.lang3.RandomUtils;
 import org.broadinstitute.consent.http.enumeration.UserFields;
 import org.broadinstitute.consent.http.models.*;
 import org.junit.Test;
 import org.postgresql.util.PSQLException;
 import org.postgresql.util.PSQLState;
 
+import static org.junit.Assert.*;
```
rushtong: nit: avoid star imports

src/test/java/org/broadinstitute/consent/http/resources/DarCollectionResourceTest.java
```diff
@@ -197,23 +197,23 @@ public void testGetCollectionByReferenceIdNotFound() {
...
+//  }
+
+//  @Test
+//  public void testGetCollectionsByInitialQuery_BadSortOrder() {
+//    when(userService.findUserByEmail(anyString())).thenReturn(researcher);
+//    initResource();
+//
+//    Response response = resource.getCollectionsByInitialQuery(authUser, "filterTerm", "projectTitle", "badSortOrder", 10);
+//    assertEquals(HttpStatusCodes.STATUS_CODE_BAD_REQUEST, response.getStatus());
+//  }
```
rushtong: nit: rm if we don't need it

rushtong (approved):
This looks pretty good! See comments inline. Most are small and optional, so feel free to implement if you choose, 👍🏽 either way.

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
JVThomas: Ah shoot, can't believed I forgot this.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
JVThomas: Can you use variables/functions for the case statements? The only reason I stuck with the if else is because I wanted to use the defined values in `UserRoles`

rushtong: No, the values have to be static. The string values in `UserRoles` actually come from `Resource` so you could do something like this:
```
    switch (userRole) {
      case Resource.ADMIN:
        break;
      case Resource.SIGNINGOFFICIAL:
        break;
      default:
        break;
    }
```
Another option is to pull in those values as static imports.

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
JVThomas: Actually, I think that's an unused function. I wrote it earlier for the initial approach, but dropped it from use. I can safely remove the method.

PR #1460, opened 2022-02-17
[DUOS-1636][risk=no] Improve DAR Collections By Role Performance

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1636

This PR makes a few small adjustments to help improve performance somewhat. There is likely more work to do here, but this is a start.
* Filter the user properties down to those specifically necessary for the user console UIs
* Add a couple of indexes to help improve lookups.

Empirically, testing as a chairperson, I was able to get the following results:
* Payload reduced from 1.98 MB to 1.59 MB (24% improvement)
* TTFB reduced from 19.13 seconds to 3.35 seconds (5.7x improvement)

Additionally, I tested the approach of simplifying the partition query of the elections into a simple join and then post-processing the older elections out of the response objects in the reducer. That actually increased response times so that is not part of this PR.

### Pre-optimization
Pre-optimization

### Post-optimization
Reference id Index

----

- Verify all tests go green


JVThomas (approved):
This looks great, the time improvement for relatively minimal changes is excellent!

PR #1477, opened 2022-03-08
[DUOS-1643][risk=no] consent rp votes modifiable after election is closed

shaemarks:
[Link to Jira ticket](https://broadworkbench.atlassian.net/browse/DUOS-1643)
Summary of changes:
- Check for Open or RP elections when validating if votes can be updated in VoteServiceDAO.updateVotesWithValue
- Add unit tests to VoteServiceDAOTest for rp elections 
- Add method to DAOTestHelper to change election status 
- Minor refactoring of updateVotesWithValue

----

- Verify all tests go green

### DIFF
--- a/src/test/java/org/broadinstitute/consent/http/service/dao/VoteServiceDAOTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/service/dao/VoteServiceDAOTest.java
@@ -107,12 +107,70 @@ public class VoteServiceDAOTest extends DAOTestHelper {
     assertTrue(votes.isEmpty());
   }
 
-  @Test
+  @Test(expected = IllegalArgumentException.class)
   public void testUpdateVotesWithValue_ClosedElection() throws Exception {
     User user = createUser();
     DataAccessRequest dar = createDataAccessRequestV3();
     DataSet dataset = createDataset();
     Election election = createAccessElection(dar.getReferenceId(), dataset.getDataSetId());
+    closeElection(election);
+    Vote vote = createDacVote(user.getDacUserId(), election.getElectionId());
+    initService();
+
+    serviceDAO.updateVotesWithValue(List.of(vote), true, "rationale");
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testUpdateVotesWithValue_CancelledElection() throws Exception {
+    User user = createUser();
+    DataAccessRequest dar = createDataAccessRequestV3();
+    DataSet dataset = createDataset();
+    Election election = createCancelledAccessElection(dar.getReferenceId(), dataset.getDataSetId());
+    Vote vote = createDacVote(user.getDacUserId(), election.getElectionId());
+    initService();
+
+    serviceDAO.updateVotesWithValue(List.of(vote), true, "rationale");
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testUpdateVotesWithValue_MultipleElectionsDifferentStatuses() throws Exception {
+    User user = createUser();
+    DataAccessRequest dar = createDataAccessRequestV3();
+    DataSet dataset = createDataset();
+    Election election1 = createAccessElection(dar.getReferenceId(), dataset.getDataSetId());
+    Election election2 = createCancelledAccessElection(dar.getReferenceId(), dataset.getDataSetId());
+    Election election3 = createAccessElection(dar.getReferenceId(), dataset.getDataSetId());
+    Vote vote1 = createDacVote(user.getDacUserId(), election1.getElectionId());
+    Vote vote2 = createDacVote(user.getDacUserId(), election2.getElectionId());
+    Vote vote3 = createDacVote(user.getDacUserId(), election3.getElectionId());
+    String rationale = "rationale";
+    initService();
+
+    List<Vote> votes = serviceDAO.updateVotesWithValue(List.of(vote1, vote2, vote3), true, rationale);
+  }
+
+  @Test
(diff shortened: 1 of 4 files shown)


shaemarks:
One possible revision that I wasn't sure about was making the exception message from validateElectionsCanUpdateVotes more general now that there are two acceptable states. Thoughts?

JVThomas:
> One possible revision that I wasn't sure about was making the exception message from validateElectionsCanUpdateVotes more general now that there are two acceptable states. Thoughts?

What do you mean by more general? As in not being specific about the Open and RP constraints in the error message?

shaemarks:
> > One possible revision that I wasn't sure about was making the exception message from validateElectionsCanUpdateVotes more general now that there are two acceptable states. Thoughts?
> 
> What do you mean by more general? As in not being specific about the Open and RP constraints in the error message?

Yes

JVThomas:
> > > One possible revision that I wasn't sure about was making the exception message from validateElectionsCanUpdateVotes more general now that there are two acceptable states. Thoughts?
> > 
> > 
> > What do you mean by more general? As in not being specific about the Open and RP constraints in the error message?
> 
> Yes

I'm fine with a more general statement since I'm not a fan of giving away to many details on the error response. What's important is that the message is specific enough that it can be traced back to the method for development/debugging purposes.

rushtong:
> I'm fine with a more general statement since I'm not a fan of giving away to many details on the error response. What's important is that the message is specific enough that it can be traced back to the method for development/debugging messages.

Agreed. This is a dao level class reporting the error, so I think it's reasonable to break out the specifics. It can percolate up to the client via the service layer - I think that's where we need to be more mindful of what is returned.

src/test/java/org/broadinstitute/consent/http/service/dao/VoteServiceDAOTest.java
```diff
@@ -118,4 +131,106 @@ public void testUpdateVotesWithValue_closedElection() throws Exception {
...
+    changeElectionStatus(election3, ElectionStatus.CLOSED);
+    Vote vote1 = createDacVote(user.getDacUserId(), election1.getElectionId());
+    Vote vote2 = createDacVote(user.getDacUserId(), election2.getElectionId());
+    Vote vote3 = createDacVote(user.getDacUserId(), election3.getElectionId());
+    String rationale = "rationale";
+    initService();
+
+    List<Vote> votes = serviceDAO.updateVotesWithValue(List.of(vote1, vote2, vote3), true, rationale);
+    assertNotNull(votes);
+    assertFalse(votes.isEmpty());
```
JVThomas, line 206: nit: You could also check on the ids to ensure that the votes being updated match those in the request list. I'm guessing there might be other tests where you could run this check as well.

JVThomas (commented):
Code looks good and works locally for me. Only have one nitpick with the tests.

src/test/java/org/broadinstitute/consent/http/service/dao/VoteServiceDAOTest.java
shaemarks: When I tried to re-run the tests locally, I got an error `java.lang.IllegalArgumentException: resource consent-config.yml not found.` (part of the setup in the test helper class) even though I can see a `consent-config.yml ` file in my resource package. This is the case even with the code the exact same as here, so I'm not sure what's wrong.

JVThomas: The tests work fine for me locally, so I wonder if the tests didn't compile correctly on your end. You can try running `mvn test-compile` on the CLI and see if that helps (though I think IntelliJ already does that for you whenever you run a test locally)

src/test/java/org/broadinstitute/consent/http/service/dao/VoteServiceDAOTest.java
```diff
@@ -118,4 +131,106 @@ public void testUpdateVotesWithValue_closedElection() throws Exception {
...
+    Vote vote1 = createDacVote(user.getDacUserId(), rpElection1.getElectionId());
+    Vote vote2 = createDacVote(user.getDacUserId(), rpElection2.getElectionId());
+    Vote vote3 = createDacVote(user.getDacUserId(), accessElection.getElectionId());
+    String rationale = "rationale";
+    initService();
+
+    List<Vote> votes = serviceDAO.updateVotesWithValue(List.of(vote1, vote2, vote3), true, rationale);
+    assertNotNull(votes);
+    assertFalse(votes.isEmpty());
+    votes.forEach(v -> assertTrue(v.getVote()));
```
JVThomas: Forgot to mention this, instead of running two different loops across votes you can use bracket notation and group the assertions together in one loop.

So something like `votes.forEach(v -> { <assertions block> } )`

src/test/java/org/broadinstitute/consent/http/service/dao/VoteServiceDAOTest.java
shaemarks: After running `mvn test-compile` and trying to run the tests, I get `java.io.IOException: Failed to bind to 0.0.0.0/0.0.0.0:8180` due to bind already being in use

rushtong: That means you're running the app at the same time as tests. Quit docker and re-run them.

JVThomas (approved):
Code and tests look good and, as mentioned earlier, functionality worked for me when I tested it locally. Good work!

PR #1489, opened 2022-03-15
[DUOS-1660][risk=no] Require library card for creating a DAR

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1660

Adds a check to the create DAR endpoint to ensure that the create user has at least one library card.

Manually tested with a user that does have a library card and one that does not. There is no change to Library Card users' ability to submit DARs. Non-LC users will see a 403 error and be redirected back to their console. The UI work for handling this case is covered in https://broadworkbench.atlassian.net/browse/DUOS-1657

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
```diff
@@ -92,6 +93,12 @@ public Response createDataAccessRequest(
       @Auth AuthUser authUser, @Context UriInfo info, String dar) {
     try {
       User user = findUserByEmail(authUser.getEmail());
+      if (Objects.isNull(user.getLibraryCards()) || user.getLibraryCards().isEmpty()) {
+        return Response
+          .status(Response.Status.FORBIDDEN)
```
quazi-h, line 98: Is this the piece that would trigger the 403 response in `./paths/darV2.yaml`?

quazi-h (approved):
Changes to check for a library card during DAR submission looks correct to me. This should return a 403 response from the API when attempting to submit a DAR with a user without a library card.

JVThomas (approved):
Tested locally and confirmed that the lc checks are working as expected. Code looks good and tests have been updated to reflect the changes.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
rushtong: Yes, that's correct.

PR #1492, opened 2022-03-16
[DUOS-1677][risk=low] New Update Rationale API

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1677

New endpoint to update the rationale for a list of vote ids for votes in RP and open Data Access elections.

----

- Verify all tests go green


quazi-h (approved):
The new api/votes/rationale endpoint looks good to me. The logic in VoteResource and VoteService.updateRationaleByVoteIds should allow us to update the rationales for the specified vote ids.

rushtong:
@shaemarks @quazi-broad - Please take another look. I made some updates to follow the pattern @shaemarks created.

shaemarks:
Looks good to me! Only thing I noticed is that rationale updates are only including RP and open Data Access elections, while vote updates are currently including RP and any open election, so it would probably be best to keep that logic consistent. This check for updating votes is found within VoteServiceDAO.java

rushtong:
> Looks good to me! Only thing I noticed is that rationale updates are only including RP and open Data Access elections, while vote updates are currently including RP and any open election, so it would probably be best to keep that logic consistent. This check for updating votes is found within VoteServiceDAO.java

I believe the tickets had different logical conditions. Updating rationale was more flexible for RP election votes, but had to be more restrictive for Data Access election votes.

shaemarks:
> > Looks good to me! Only thing I noticed is that rationale updates are only including RP and open Data Access elections, while vote updates are currently including RP and any open election, so it would probably be best to keep that logic consistent. This check for updating votes is found within VoteServiceDAO.java
> 
> I believe the tickets had different logical conditions. Updating rationale was more flexible for RP election votes, but had to be more restrictive for Data Access election votes.

So it's not an issue that elections that are neither RP nor Data Access may have their votes (which could possibly include a rationale) updated through the updateVotes endpoint?

rushtong:
> So it's not an issue that elections that are neither RP nor Data Access may have their votes (which could possibly include a rationale) updated through the updateVotes endpoint?

Votes for non RP and non Data Access elections [trigger an error](https://github.com/DataBiosphere/consent/pull/1492/files#diff-0932ac04bb26e5246f33702b119138240f419192f1044797ba3d2b8b59c1eeeaR319)

shaemarks:
I was wondering if the endpoint I was working on that updates the vote value should follow this rule as well, especially since it can also change the rationale as an optional field. Currently the update vote endpoint checks the type and status of elections in the VoteServiceDAO class, which follows slightly different logic than this.

shaemarks:
Screenshot (4)

rushtong:
> I was wondering if the endpoint I was working on that updates the vote value should follow this rule as well, especially since it can also change the rationale as an optional field. Currently the update vote endpoint checks the type and status of elections in the VoteServiceDAO class, which follows slightly different logic than this.

Ah, good question, I think you might be right. Let's file a ticket for that so we can refine it as a team.

shaemarks (approved):
Looks good to me!

quazi-h (approved):
The new rationale endpoint looks good to me. It is only accessible to chairs and members and is correctly taking in a body of rationale and voteIds. The filtering logic for RP votes and Data Access votes also looks good.

PR #1493, opened 2022-03-16
[DUOS-1678][risk=no] consent update vote api fixes

shaemarks:
[Link to Jira ticket](https://broadworkbench.atlassian.net/browse/DUOS-1678)
Summary of changes:

- Modify updateVotes in VoteResource to take in vote, rationale, and vote ids in json format 
- Create custom model for json object 
- Convert tests to new method signature and add tests for when vote and rationale are null

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/resources/VoteResource.java
```diff
@@ -87,10 +83,56 @@ public Response updateVotes(
...
+            List<Vote> updatedVotes = voteService.updateVotesWithValue(votes, voteUpdateInfo.getVote(), voteUpdateInfo.getRationale());
             return Response.ok().entity(updatedVotes).build();
         } catch (Exception e) {
             return createExceptionResponse(e);
         }
     }
+
+    static class VoteUpdateInfo {
+
+        @JsonProperty
```
rushtong: As a side note, `@JsonProperty` is mostly unnecessary. This is a Jackson annotation that allows for writing/parsing via Jackson libraries. Since we're using Gson here, we don't need them. We should have a tech debt story to deprecate this usage.

src/main/resources/assets/paths/votes.yaml
```diff
@@ -1,31 +1,31 @@
...
     content:
       application/json:
         schema:
           type: string
-          example: [1, 2, 3]
+          example: {
+            "vote": true|false,
+            "rationale": "rationale string",
+            "voteIds": [1, 2, 3]
+          }
```
rushtong, line 28: Very nice!

src/main/java/org/broadinstitute/consent/http/resources/VoteResource.java
```diff
@@ -87,10 +83,56 @@ public Response updateVotes(
...
 
-            List<Vote> updatedVotes = voteService.updateVotesWithValue(votes, vote, rationale);
+            List<Vote> updatedVotes = voteService.updateVotesWithValue(votes, voteUpdateInfo.getVote(), voteUpdateInfo.getRationale());
             return Response.ok().entity(updatedVotes).build();
         } catch (Exception e) {
             return createExceptionResponse(e);
         }
     }
+
+    static class VoteUpdateInfo {
```
rushtong: Thinking out loud here. We have this simple model class, and I have one for a similar method in my PR. It might make sense to pull these out into a separate model package specific to the resources directly. Another possibility would be in `org/broadinstitute/consent/http/models/vote`

rushtong (approved):
See minor comments inline, all are optional, address if you prefer. Looks good!

src/main/java/org/broadinstitute/consent/http/resources/VoteResource.java
```diff
@@ -1,12 +1,13 @@
 package org.broadinstitute.consent.http.resources;
 
 
+import com.fasterxml.jackson.annotation.JsonProperty;
```
JVThomas: You can remove this import since you got rid of the `@JsonProperty` annotation

JVThomas (approved):
One little nitpick, but other than that the code looks good. Tests cover all of the expected edge cases and the functionality works when I tested it locally.

PR #1536, opened 2022-04-11
[DUOS-1442][risk=no] New Fields for Dataset

rushtong:
## Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DUOS-1442

More progress towards deprecating the dataset dto and associated classes/methods. This PR adds new fields to the Dataset object and populates them with the same values we're using to populate the DTO

* New Fields:
  * Deletable: If the dataset id exists in any DAR's list of dataset ids, then it is not deletable
  * Translated Use Restriction: This is the value pulled directly from the consent. Current values for this are populated when the consent is created and we are using the DataUse object to drive the translation. For older consents that came from ORSP, this is not the case - but for anything populated via the current Dataset Registration page, this is true.
  * Dataset Properties: These are now populated and are not the DTO versions. I added a new `propertyName` field to the object so we don't need to dig into the dictionary object.
  * DAC Id: We now pull the DAC's id directly from the consent
  * Consent Id: We now pull the consent's id from the consent
* We now use the DatasetReducer for all Dataset population
* Some minor dao method renaming to move further along the path of `DataSet` -> `Dataset`
* New v2 dataset API that behaves the same way the original does, but returns real dataset objects instead of DTOs
* `isAssociatedToDataOwners` and `updateAssociationToDataOwnerAllowed` are unused in `DatasetDTO` so these are not migrated.

----

- Verify all tests go green


src/main/resources/assets/paths/datasetV2.yaml
```diff
@@ -0,0 +1,39 @@
+get:
+  summary: Get Datasets
+  description: Returns Dataset list filtered by authenticated user.
```
shaemarks: It might be useful to add more detail about what "filtered by authenticated user" means / what is returned for users with different roles. As is, if I were to use this endpoint in the future, I would have to look in the DatasetService or DatasetDAO class to fully understand the different cases, particularly `datasetDAO.getActiveDatasets()` for users without an admin role.

src/main/java/org/broadinstitute/consent/http/models/Dataset.java
```diff
@@ -191,11 +237,63 @@ public void setAlias(Integer alias) {
         this.alias = alias;
     }
```
shaemarks: Nit: move `setAlias` directly underneath `getAlias` on line 224

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -209,6 +193,21 @@ public Response describeDataSets(@Auth AuthUser authUser) {
         }
     }
 
+    @GET
+    @Produces("application/json")
+    @PermitAll
+    @Path("/v2")
+    public Response findAllDatasetsByUser(@Auth AuthUser authUser) {
```
shaemarks: Nit: may be more accurate to name this method `findAllDatasetsAvailableToUser` or something similar since my first assumption when reading this name is that the method would find datasets where the `createUser` matches the provided user. Similar comment for `datasetService.findAllDatasetsByUser(user)`.

shaemarks (changes requested):
The population of the new Dataset fields and added tests look good, as do the `DataSet` -> `Dataset` name changes. I left some comments about clarifying the `/dataset/v2` endpoint, but good otherwise

shaemarks:
Thanks for the changes! Code looks good, but I'm getting a 500 Internal Server Error when I try to test the endpoint on Swagger. The error seems to be originating from the `private Object createProxy(Jdbi db, Class<?> extensionType, Class<?>... extraTypes)`  method in `OnDemandExtensions.class`, which is invoked when `datasetDAO.getAllDatasets()` or `datasetDAO.getActiveDatasets()` is called. I'm not sure if this error has more to do with the code itself or my machine though.

rushtong:
> The error seems to be originating from the private Object createProxy(Jdbi db, Class<?> extensionType, Class<?>... extraTypes) method in OnDemandExtensions.class, which is invoked when datasetDAO.getAllDatasets() or datasetDAO.getActiveDatasets() is called. I'm not sure if this error has more to do with the code itself or my machine though.

Thank you @shaemarks ... something structural has changed since I last tested this, I'll look into this some more.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -429,4 +431,17 @@ private boolean filterDatasetOnProperties(DatasetDTO dataset, String term) {
...
     }
+
+    public List<Dataset> findAllDatasetsByUser(User user) {
+        List<Dataset> datasets = new ArrayList<>();
+        if (user.hasUserRole(UserRoles.ADMIN)) {
+            datasets.addAll(datasetDAO.getAllDatasets());
+        } else {
+            datasets.addAll(datasetDAO.getActiveDatasets());
+            if (user.hasUserRole(UserRoles.CHAIRPERSON)) {
+                datasets.addAll(datasetDAO.findDatasetsByAuthUserEmail(user.getEmail()));
```
shaemarks: If a chairperson has a DAC associated with active datasets, will the datasets appear twice in the returned list (once from `datasetDAO.getActiveDatasets` and a second time from `datasetDAO.findDatasetsByAuthUserEmail`), and if so, would this be an issue?

rushtong: Yes, this is an issue. This is essentially a copy of the original code, but for datasets instead of dataset dtos. The original method collected them in a set which solved this problem. I'll update this, thank you 👍🏽

shaemarks (approved):
Changes look great, and new DatasetService tests do a good job testing the different cases

PR #1545, opened 2022-04-19
DUOS-1730[risk=no] Vote API - Validate LC existence for certain vote scenarios

JVThomas:
Addresses [DUOS-1730](https://broadworkbench.atlassian.net/browse/DUOS-1730)

PR adds an Library Card check when updating votes for the following scenario...

* User vote is of type `Chairperson` or `Final`
* Election type is `DataAccess`
* Vote value is `true`

... in order to ensure that DACs can only vote Yes/True if the data requester has a Library Card associated with their account.

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
@@ -136,6 +136,13 @@ public interface ElectionDAO extends Transactional<ElectionDAO> {
             " ORDER BY e.createdate ASC ")
     List<Election> findLastDataAccessElectionsWithFinalVoteByStatus(@Bind("status") String status);
 
+	@SqlQuery("SELECT e.* FROM election" +
+		"INNER JOIN data_access_request dar ON dar.reference_id = e.referenceid " +
+		"INNER JOIN dacuser du ON du.dacuserid = dar.user_id " +
+        "INNER JOIN library_card lc ON lc.user_id = du.dacuserid " +
+        "WHERE e.electionid IN (<electionIds>) ")
+	List<Election> findElectionsWithCardHoldingUsersByElectionIds(@BindList("electionIds") List <Integer> electionIds);
+
     @SqlQuery("SELECT DISTINCT e.electionId, e.datasetId, v.vote finalVote, e.status, e.createDate, e.referenceId, " +
             "       v.rationale finalRationale, v.createDate finalVoteDate, e.lastUpdate, e.finalAccessVote, " +
             "       e.electionType, e.dataUseLetter, e.dulName, e.archived, e.version " +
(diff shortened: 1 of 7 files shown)


src/main/java/org/broadinstitute/consent/http/service/ElectionService.java
```diff
@@ -426,6 +430,10 @@ public List<Election> createDataSetElections(String referenceId, Map<User, List<
         return CollectionUtils.isEmpty(electionsIds) ? null : electionDAO.findElectionsByIds(electionsIds);
     }
 
+    public List<Election> findElectionsByVoteIdsAndType(List<Integer> voteIds, String electionType) {
+        return electionDAO.findElectionsByVoteIdsAndType(voteIds, electionType);
```
rushtong, line 434: Looking at usages of this, it looks possible to pass in an empty list of vote ids. Might be safer to handle that case before making the DAO call which will throw a sql exception if it is empty.

rushtong: Usage of this look pretty safe, but it might be nice to add an empty list check here to ensure we never make the dao call.

src/main/java/org/broadinstitute/consent/http/service/ElectionService.java
```diff
@@ -384,6 +384,10 @@ public boolean checkDataOwnerToCloseElection(Integer electionId){
         return closeElection;
     }
 
+    public List<Election> findElectionsWithCardHoldingUsersByElectionIds(List<Integer> electionIds) {
+        return electionDAO.findElectionsWithCardHoldingUsersByElectionIds(electionIds);
```
rushtong: Same empty list handling feedback applies here - current usages are safe, this would be a future-proofing kind of thing.

src/main/java/org/broadinstitute/consent/http/resources/VoteResource.java
```diff
@@ -139,4 +150,34 @@ public Response updateVoteRationale(@Auth AuthUser authUser, String json) {
...
         }
     }
+    
+    //Private helper function, checks to see if user has library card for chair votes that are getting an incoming "yes" update
+    private void voteUpdateLCCheck(List<Vote> votes) {
+        //filter for chair or final votes
+        List<Vote> targetVotes = votes.stream()
+            .filter(v -> {
+                String type = v.getType();
+                return type.equalsIgnoreCase("chairperson") || type.equalsIgnoreCase("final");
```
rushtong: Use the `VoteType` enum here to reduce raw string checks

src/main/java/org/broadinstitute/consent/http/resources/VoteResource.java
```diff
@@ -139,4 +150,34 @@ public Response updateVoteRationale(@Auth AuthUser authUser, String json) {
...
+            })
+            .collect(Collectors.toList());
+        //if the filtered list is populated, get the vote ids and get the full vote records for those that have type = 'DataAccess'
+        if(!targetVotes.isEmpty()) {
+            List<Integer> voteIds = targetVotes.stream()
+                .map(Vote::getVoteId)
+                .collect(Collectors.toList());
+            List<Election> targetElections = electionService.findElectionsByVoteIdsAndType(voteIds, "dataaccess");
```
rushtong: Use the `ElectionType` enum here in the service call

rushtong (commented):
See minor feedback inline. Code looks good, logic looks good, 👍🏽  overall

shaemarks (approved):
Code for validating library cards looks good, and tests thoroughly cover different voting scenarios.

src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
```diff
@@ -161,6 +170,17 @@ void updateElectionStatus(@BindList("electionsId") List<Integer> electionsId,
...
 
+    @UseRowMapper(SimpleElectionMapper.class)
+    @SqlQuery("SELECT * FROM election e " + 
+        "INNER JOIN vote v ON v.electionid = e.electionid " +
+        "WHERE LOWER(e.electiontype) = :electionType " + 
+        "AND v.voteid IN (<voteIds>)")
+	List<Election> findElectionsByVoteIdsAndType(
+		@BindList("voteIds") List<Integer> voteIds,
+		@Bind("electionType") String electionType
+	);
```
rushtong: minor nit, indentation looks off.

rushtong (approved):
Thanks for the changes! See one minor indentation comment inline, then 👍🏽

PR #1547, opened 2022-04-20
[DUOS-1722][risk=no] Update email handling logic for DAR Collection

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1722

Updates email handling logic for DAR Collection emails. Now we send one email per collection instead of one email per DAR to each relevant user. In this case, relevant users are admins and chairs (related to the datasets) who have elected to receive email.

### Example email:

Screen Shot 2022-04-22 at 1 09 29 PM

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/ConsentApplication.java
+++ b/src/main/java/org/broadinstitute/consent/http/ConsentApplication.java
@@ -226,7 +226,7 @@ public class ConsentApplication extends Application<ConsentConfiguration> {
 
         // Register standard application resources.
         env.jersey().register(new ApprovalExpirationTimeResource(approvalExpirationTimeService, userService));
-        env.jersey().register(new DataAccessRequestResourceVersion2(dataAccessRequestService, gcsService, userService, matchService));
+        env.jersey().register(new DataAccessRequestResourceVersion2(dataAccessRequestService, emailNotifierService, gcsService, userService, matchService));
         env.jersey().register(new DataAccessRequestResource(dataAccessRequestService, userService, consentService, electionService));
         env.jersey().register(new DatasetResource(consentService, datasetService, userService, dataAccessRequestService));
         env.jersey().register(new DatasetAssociationsResource(datasetAssociationService));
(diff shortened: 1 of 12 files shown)


JVThomas (approved):
Code looks good, I can confirm that I received an email for the collection being submitted.

I technically received two emails, but that's because my local user account was both an Admin and a Chair. It's safe to say that we (dev team) are probably the only ones who'll experience this issue.

If you do want to address it, then it might be a good idea to add user emails to a Set or Map and then use the values there to construct the ArrayList for further processing steps.

rushtong:
> I technically received two emails, but that's because my local user account was both an Admin and a Chair. It's safe to say that we (dev team) are probably the only ones who'll experience this issue.

I think I can fix this by calling distinct on the users ... I wanted to make the email more user specific so we use the user's display name in the email like we do in other emails (hoping for more generic consistency for future email work).

shaemarks (commented):
Here are results from my testing:
- Correctly sends only one email to open an election for admins, chairs with datasets associated to the collection, and users with both admin + chair roles when a multi-dataset DAR is submitted 
- Correctly sends zero emails to users without the necessary roles (ex: DAC member with datasets associated to the collection)
- If a multi-dataset DAR is cancelled and resubmitted through the `/new_researcher_console` before the election is opened, no additional email is sent to admins / chairs to open the election
- If a single dataset DAR is cancelled and resubmitted through the `/researcher_console` before the election is opened, a new email is sent to admins / chairs to open the election 
- When clicking on the "Login to create an election" link in the email, I was redirected to a page with the following error 
 
DUOS-1772 Test

rushtong:
Thank you @shaemarks !
These look like bugs:

> * If a multi-dataset DAR is cancelled and resubmitted through the /new_researcher_console before the election is opened, no additional email is sent to admins / chairs to open the election

This is a use case I didn't test - will dig into that and see if we need to add email handling to that path

> * When clicking on the "Login to create an election" link in the email, I was redirected to a page with the following error

This is the existing behavior - it constructs the link to the destination page. I'm going to remove that since we don't handle deep linking well in the UI, even if that were a correct url.

rushtong:
After looking into this a little more, we do not want to send an email after a user cancels/resubmits. That code path creates a draft dar that then needs to be re-submitted which does go through the same process as any other draft dar.

shaemarks (approved):
Looks good to me!

JVThomas (approved):
Looks good to me

PR #1551, opened 2022-04-22
[DUOS-1519][risk=no] add parent reference id to dar

shaemarks:
[Link to Jira ticket](https://broadworkbench.atlassian.net/browse/DUOS-1519)
Summary of changes:
- Add `parent_id` column (defaults to null) to the `data_access_request` table
- Create a foreign key constraint on `id` so that a non-null `parent_id` must refer to an existing DAR `id`
- Update queries from `data_access_request` table to include `parent_id`
- Add `parentId` field to the `DataAccessRequest` model

----

- Verify all tests go green


rushtong:
I think there's a bit more to do here. Have a look at the following classes for where we would need to query for this value and map it to the object:
* DarCollectionDAO
* DataAccessRequestMapper

I don't think we need to modify any tests yet since we can't populate this value - TBD in future stories.

rushtong (approved):
Thanks for the changes! 👍🏽

quazi-h (approved):
New parent_id looks good to me and FK and updated DAR queries look good.

PR #1564, opened 2022-05-05
DUOS-1785[risk=low] Added MatchResource endpoint for bulk requests for DarCollectionReview

JVThomas:
Partially addresses [DUOS-1785](https://broadworkbench.atlassian.net/browse/DUOS-1785)

Since the codebase has moved away from updating/using `AGREEMENT` votes, the `DarCollectionReview` will need to obtain the match data via endpoint. Since the elections are bucketed by `dataUse`, the front-end can pick out one `DataAccess` election's `referenceId` per bucket and provide that as query param (via comma separated string).

This PR provides the backend adjustments to ensure that occurs by updating and/or creating endpoints, service methods, DAO queries, and tests as needed.

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/MatchDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/MatchDAO.java
@@ -30,8 +30,17 @@ public interface MatchDAO extends Transactional<MatchDAO> {
     @SqlQuery("SELECT * FROM match_entity WHERE matchid = :id ")
     Match  findMatchById(@Bind("id") Integer id);
 
-    @SqlQuery("SELECT * FROM match_entity WHERE purpose IN (<purposeId>)")
-    List<Match> findMatchesForPurposeIds(@BindList("purposeId") List<String> purposeId);
+    @SqlQuery(
+        "SELECT match_entity.* FROM match_entity " +
+        "INNER JOIN (" +
+        "   SELECT election.*, MAX(election.electionid) OVER (PARTITION BY election.referenceid, election.datasetid) AS latest " +
+        "   FROM election " +
+        "   WHERE LOWER(election.electiontype) = 'dataaccess' " +
+        ") AS e " +
+        "ON e.referenceid = match_entity.purpose " +
+        "WHERE match_entity.purpose IN (<purposeIds>) AND e.electionid = latest"
+    )
+    List<Match> findMatchesForPurposeIds(@BindList("purposeIds") List<String> purposeIds);
 
     @SqlUpdate(
             " INSERT INTO match_entity " +
--- a/src/main/java/org/broadinstitute/consent/http/db/MatchDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/MatchDAO.java
@@ -40,7 +40,10 @@ public interface MatchDAO extends Transactional<MatchDAO> {
         "ON e.referenceid = match_entity.purpose " +
         "WHERE match_entity.purpose IN (<purposeIds>) AND e.electionid = latest"
     )
-    List<Match> findMatchesForPurposeIds(@BindList("purposeIds") List<String> purposeIds);
+    List<Match> findMatchesForBatchPurposeIds(@BindList("purposeIds") List<String> purposeIds);
+
+    @SqlQuery("SELECT * FROM match_entity WHERE purpose IN (<purposeId>)")
+    List<Match> findMatchesForPurposeIds(@BindList("purposeId") List<String> purposeId);
 
     @SqlUpdate(
             " INSERT INTO match_entity " +
--- a/src/main/java/org/broadinstitute/consent/http/db/MatchDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/MatchDAO.java
@@ -40,7 +40,7 @@ public interface MatchDAO extends Transactional<MatchDAO> {
         "ON e.referenceid = match_entity.purpose " +
         "WHERE match_entity.purpose IN (<purposeIds>) AND e.electionid = latest"
     )
-    List<Match> findMatchesForBatchPurposeIds(@BindList("purposeIds") List<String> purposeIds);
+    List<Match> findMatchesForLatestDataAccessElectionsByPurposeIds(@BindList("purposeIds") List<String> purposeIds);
 
     @SqlQuery("SELECT * FROM match_entity WHERE purpose IN (<purposeId>)")
     List<Match> findMatchesForPurposeIds(@BindList("purposeId") List<String> purposeId);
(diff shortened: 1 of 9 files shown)


src/main/java/org/broadinstitute/consent/http/db/MatchDAO.java
```diff
@@ -30,8 +30,17 @@ public interface MatchDAO extends Transactional<MatchDAO> {
...
+        "SELECT match_entity.* FROM match_entity " +
+        "INNER JOIN (" +
+        "   SELECT election.*, MAX(election.electionid) OVER (PARTITION BY election.referenceid, election.datasetid) AS latest " +
+        "   FROM election " +
+        "   WHERE LOWER(election.electiontype) = 'dataaccess' " +
+        ") AS e " +
+        "ON e.referenceid = match_entity.purpose " +
+        "WHERE match_entity.purpose IN (<purposeIds>) AND e.electionid = latest"
+    )
+    List<Match> findMatchesForPurposeIds(@BindList("purposeIds") List<String> purposeIds);
```
rushtong: I'm not sure I understand the necessity of this change. `purposeIds` here refers to DAR reference ids (which is also an election reference id). We insert a single dar reference id + consent id (which is a proxy for the dataset) with a match, NOT per election. This change explicitly says each DAR/dataset match can only be for a single Data Access election - which is a big change. We also want to be able to say, Does this DAR match to something else where there may not be an election. For example, we will want to be able to say does this DAR match any other dataset that we create a match result for, even if the user did not request it. By attaching matches to elections, we're reducing the functionality of matching as a whole.

rushtong (changes requested):
API looks good, but one question on the query. Let's have a conversation if it isn't clear.

src/main/java/org/broadinstitute/consent/http/db/MatchDAO.java
JVThomas: Oh, it's generated on DAR updates, not on election generation. So then what happens to the matches regarding DARs that have the same reference id?

rushtong: > So then what happens to the matches regarding DARs that have the same reference id

Every dar has a unique reference id, so any individual match will always apply to only a single DAR/Dataset combination. If there are N matches for a DAR, that represents the DAR having been matched to N different datasets.

JVThomas: That's not true. It's the reason why Shae had to use `id` for a foreign key instead of reference id for [DUOS-1519](https://github.com/DataBiosphere/consent/pull/1551). Reference ids are supposed to be unique, but considering that staging has DARs with the same reference id we can assume the possibility exists for dev and prod as well.

rushtong: > That's not true. It's the reason why Shae had to use id for a foreign key instead of reference id for https://github.com/DataBiosphere/consent/pull/1551. Reference ids are supposed to be unique, but considering that staging has DARs with the same reference id we can assume the possibility exists for dev and prod as well.

I wasn't aware of that. If that is true, it's a very serious bug. They HAVE to be unique so elections can refer to a single DAR and not multiple DARs. We need to file a ticket for this if true.

JVThomas: It's not a current issue, but it was an issue in the past. However, the fact that it existed (with some DARs in staging as evidence) means that I have to navigate the duplicate ids problem somehow.

At the very least I can rename this new query and use it while restoring the old one. That said, the old query may need some digging into to see if it still holds up (different ticket).

src/main/java/org/broadinstitute/consent/http/db/MatchDAO.java
```diff
@@ -30,6 +30,18 @@ public interface MatchDAO extends Transactional<MatchDAO> {
...
+        "SELECT match_entity.* FROM match_entity " +
+        "INNER JOIN (" +
+        "   SELECT election.*, MAX(election.electionid) OVER (PARTITION BY election.referenceid, election.datasetid) AS latest " +
+        "   FROM election " +
+        "   WHERE LOWER(election.electiontype) = 'dataaccess' " +
+        ") AS e " +
+        "ON e.referenceid = match_entity.purpose " +
+        "WHERE match_entity.purpose IN (<purposeIds>) AND e.electionid = latest"
+    )
+    List<Match> findMatchesForBatchPurposeIds(@BindList("purposeIds") List<String> purposeIds);
```
rushtong: Nit, naming. What you're really finding here is matches for DARs that have data access elections. This will filter out all DARs that do not have any election for them. Maybe this is even more specific, `findPurposeMatchesWithMostRecentDataAccessElections`.

src/main/resources/assets/api-docs.yaml
```diff
@@ -2785,6 +2785,33 @@ paths:
                   $ref: '#/components/schemas/MatchResult'
         500:
           description: Server error.
+  /api/match/purpose/batch/{purposeIds}:
```
rushtong: I think this needs to be `/api/match/purpose/batch` since `purposeIds` is not a Path Parameter.

src/main/resources/assets/api-docs.yaml
```diff
@@ -2785,6 +2785,33 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/MatchResult'
+        500:
+          description: Server error.
+        404:
+          description: Bad request
```
rushtong: Can we move this to a yaml file in `src/main/resources/assets/paths`?

src/test/java/org/broadinstitute/consent/http/service/MatchServiceTest.java
```diff
@@ -300,6 +301,15 @@ public void testRemoveMatchesForConsent() throws Exception {
         service.removeMatchesForConsent(sampleConsent1.getConsentId());
     }
 
+    @Test
+    public void testFindMatchesForPurposeIds() {
+        Match mockMatch = new Match();
+        when(matchDAO.findMatchesForPurposeIds(anyList()))
+            .thenReturn(List.of(mockMatch));
+        initService();
+        service.findMatchesForPurposeIds(List.of("test"));
```
rushtong: Are we actually testing anything here? Maybe wrap the service call in a try-catch and fail in the catch block?

rushtong (commented):
Looks pretty good, see minor comments inline.

rushtong:
The overall risk here is "LOW" since we're adding a new endpoint.

rushtong (approved):
Thank you for the changes 👍🏽

shaemarks (approved):
Code and tests look good, and when run locally with the ui, the algorithm match decision was displayed as expected

PR #1593, opened 2022-05-31
[DUOS-1836][risk=no] Deprecate Unused APIs

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1836

Deprecate newly unused APIs from https://github.com/DataBiosphere/duos-ui/pull/1576

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -51,6 +51,7 @@ public class DataAccessRequestResource extends Resource {
         this.userService = userService;
     }
 
+    @Deprecated
     @GET
     @Produces("application/json")
     @Path("/modalSummary/{id}")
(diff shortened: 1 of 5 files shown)


shaemarks (approved):
Looks good! Deprecated APIs match the endpoints removed from the frontend for this ticket

quazi-h (approved):
Changes make sense to me.

Using the @Deprecated tag would allow those elements to still be functional but from looking up the documentation I see that it should present a warning/error when called. I didn't notice any warnings when hitting one of the endpoints below using a local instance through the swagger page but I would only see that in a compiler, right?

rushtong:
> I didn't notice any warnings when hitting one of the endpoints below using a local instance through the swagger page but I would only see that in a compiler, right?

The APIs won't function any differently, this is purely informational to track for deletion in a future story, i.e. https://broadworkbench.atlassian.net/browse/DUOS-1846

PR #1600, opened 2022-06-07
[DUOS-1768][risk=no] Filter Dataset Count on Chair Console

quazi-h:
[DUOS-1768](https://broadworkbench.atlassian.net/browse/DUOS-1768)

Summary of changes:
- Updated the getCollectionsForUserByRole endpoint to pass a boolean var to the getCollectionsByUserDacs function for chair and member roles.
- Updated the getCollectionsByUserDacs function to pass in a list of user datasets from findDatasetsByAuthUserEmail.
- Updated addDatasetsToCollections to filter the provided argument with all datasets the user has access to against the list of datasets for the collections for the user DACs.
- This is ensuring that the DAR Collections returned by the user's associated DACs are included in the list of datasets that are visible to the user via DACs.
- Added a filter to remove null items within addDatasetsToCollections since it was inflating the dataset count by 1.
- Added a test to DarCollectionServiceTest to make sure we are getting the correct count of datasets when providing userDatasets to addDatasetsToCollections.

Notes:
- Do I need any additional test cases? Maybe one to check count difference (testing getCollectionsForUserByRoleName)?
- Updating the swagger documentation?
----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -77,6 +77,15 @@ public List<DarCollection> getCollectionsForUserByRoleName(User user, String rol
...
+          List<Dataset> userDatasets = datasetDAO.findDatasetsByAuthUserEmail(user.getEmail());
+          List<DarCollection> userCollections = getCollectionsByUserDacs(user);
+          userCollections.forEach(c -> {
+            Set<Dataset> relevantDatasets = c.getDatasets();
+            relevantDatasets.retainAll(userDatasets);
+            c.setDatasets(relevantDatasets);
+          });
+          collections.addAll(userCollections);
+          break;
         case MEMBER:
```
shaemarks, line 80: We would want these changes to be present for both chairperson and member roles, so I would put `case MEMBER:` directly below `case CHAIRPERSON:` so both cases in the switch statement use the new block of code  

Edit: Just noticed that the ticket only specifies for the chair console, but I still think this should also apply to the member console, and I don't think there is an existing ticket to address that

shaemarks (changes requested):
Dataset filtering looks good when getting collections using the chairperson role

I think it would be good to test this behavior in `DarCollectionServiceTest`. You could do this by adding two datasets to a collection, mocking out `findDatasetsByAuthUserEmail` to only return one of the datasets, and then checking that the returned collection only contains the one dataset. It might also be worth updating the Swagger documentation for this endpoint.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rushtong: I agree, if we do this for the chairs, we need to do it for the members too.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -77,6 +77,15 @@ public List<DarCollection> getCollectionsForUserByRoleName(User user, String rol
           collections.addAll(getAllCollections());
           break;
         case CHAIRPERSON:
+          List<Dataset> userDatasets = datasetDAO.findDatasetsByAuthUserEmail(user.getEmail());
+          List<DarCollection> userCollections = getCollectionsByUserDacs(user);
+          userCollections.forEach(c -> {
+            Set<Dataset> relevantDatasets = c.getDatasets();
+            relevantDatasets.retainAll(userDatasets);
+            c.setDatasets(relevantDatasets);
```
rushtong: In theory, this could just be 
```
c.getDatasets().retainAll(userDatasets);
```
since `retainAll` mutates the provided set.

rushtong: Update: Tested this, and it does work

rushtong:
I think we're doing the filtering at the wrong level here. If you follow the code in `getCollectionsByUserDacs`, we loop over each collection and add datasets to it. This adds a second level of looping which can only slow down the response. Instead, I think we should look into modifying `getCollectionsByUserDacs` to take an argument that allows filtering on user-DAC-datasets. That can then be passed down to `addDatasetsToCollections` so it can handle the right logic.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -250,21 +254,25 @@ public DarCollection getByCollectionId(Integer collectionId) {
...
       throw new NotFoundException("Collection with the collection id of " + collectionId + " was not found");
     }
-    List<DarCollection> populatedCollections = addDatasetsToCollections(Collections.singletonList(collection));
+    List<DarCollection> populatedCollections = addDatasetsToCollections(Collections.singletonList(collection), null);
     return populatedCollections.stream().findFirst().orElse(null);
   }
 
-  public List<DarCollection> addDatasetsToCollections(List<DarCollection> collections) {
-
+  public List<DarCollection> addDatasetsToCollections(List<DarCollection> collections, List<Dataset> userDatasets) {
```
rushtong: This method could be a lot simpler if we passed a list of IDs to filter instead of full datasets.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -250,21 +254,25 @@ public DarCollection getByCollectionId(Integer collectionId) {
...
       .map(d-> d.getDars().values())
       .flatMap(Collection::stream)
       .map(d -> d.getData().getDatasetIds())
       .flatMap(Collection::stream)
       .collect(Collectors.toList());
-
     if(!datasetIds.isEmpty()) {
-      Set<Dataset> datasets = datasetDAO.findDatasetWithDataUseByIdList(datasetIds);
+      // gets all datasets that should belong to the collection
+      Set<Dataset> datasets = (Objects.nonNull(userDatasets)) ?
```
rushtong: Here, for example, we could simplify this quite a bit. If `userDatasetIds` has any values in it, we could just get the intersection between that and `datasetIds`. Then, look up the datasets by 
```
Set<Dataset> datasets = datasetDAO.findDatasetWithDataUseByIdList(datasetIdIntersection);
```

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rushtong: Additionally, this method could use some documentation now. What I'd like to suggest is that we describe `List<Integer> userDatasetIds` as an optionally populated list of IDs. If there are any ids, then this method will filter the Collection Datasets by those ids. This is also a good place to try and name this appropriately. Maybe `filterDatasetIds` sounds better (just a suggestion)? `userDatasets` just by itself doesn't really tell me much about what the method is intending to do. Something that implies filtering would go a long way to clearing that up.

What that lets you do is stop checking for null and also to not explicitly pass nulls as arguments to this method. Instead, callers should pass `List.of()` instead of `null` which is a little bit more clear in terms of the intention of the (potentially renamed) argument.

rushtong (changes requested):
This is coming along, thank you for the changes so far. Please see comments inline.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -250,20 +256,33 @@ public DarCollection getByCollectionId(Integer collectionId) {
...
-    List<DarCollection> populatedCollections = addDatasetsToCollections(Collections.singletonList(collection));
+    List<DarCollection> populatedCollections = addDatasetsToCollections(Collections.singletonList(collection), List.of());
     return populatedCollections.stream().findFirst().orElse(null);
   }
 
-  public List<DarCollection> addDatasetsToCollections(List<DarCollection> collections) {
-
+  // This method could be a lot simpler if we passed a list of IDs to filter instead of full datasets.
+  // If there are any ids, then this method will filter the Collection Datasets by those ids.
+  // What that lets you do is stop checking for null and also to not explicitly pass nulls as arguments to this method.
```
rushtong: nit: I don't think we need this comment here

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
quazi-h: I used the retainAll function to mutate datasetIds and find the overlap - is that acceptable?

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
quazi-h: Sorry, didn't mean to leave those in, cleaned up the comments.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rushtong: That works fine

rushtong (approved):
Thank you for the changes! 👍🏽

src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
```diff
@@ -286,6 +287,39 @@ public void testAddDatasetsToCollection() {
...
+    collections.add(generateMockDarCollection(datasets));
+    List<Integer> datasetIds = datasets.stream()
+            .map(Dataset::getDataSetId)
+            .sorted()
+            .collect(Collectors.toList());
+
+    Dataset dataset = new Dataset();
+    dataset.setDataSetId(datasetIds.get(0));
+
+    // mocking out findDatasetsByAuthUserEmail to only return one of the datasets
```
shaemarks: nit: Update comment since only `findDatasetWithDataUseByIdList` is being mocked out now

shaemarks (approved):
Small nit about comment in tests, but changes look good otherwise 👍 
When tested locally, collection that showed 5 datasets under the admin console correctly showed 2 datasets under DAC chair and member consoles

src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
quazi-h: Thank you for spotting that! And thanks for your help getting the test working!

PR #1611, opened 2022-06-15
[DUOS-1761][risk=no] Update Join Table

phendriksen100:
----
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1761

Now stores reference_id and dataset_id in the joins table

### Testing.
UI - Create a DAR and validate the correct reference_id and dataset_ids are being stored.
UnitTest - Run "mvn clean test package"

- Verify all tests go green


dependency-reduced-pom.xml
```diff
@@ -0,0 +1,356 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
```
rushtong: We don't need this file for this PR.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -241,9 +243,29 @@ public DataAccessRequest insertDraftDataAccessRequest(User user, DataAccessReque
...
+     * @param referenceId ReferenceId of the corresponding DAR
+     */
+    private void syncDataAccessRequestDatasets(List<Integer> datasetIds, String referenceId) {
+        dataAccessRequestDAO.deleteDARDatasetRelationByReferenceId(referenceId);
+
+        if (!datasetIds.isEmpty()) {
+            datasetIds.forEach(id -> {
+                dataAccessRequestDAO.insertDARDatasetRelation(referenceId, id);
+            });
+        }
```
rushtong: Suggestion: Batch insert these instead of doing inserts in a loop.

1. Look at MatchDAO.insertAll for a model for how to do this
2. Create a value class that we can use as an insert

```
public class DarDataset() {
    public String referenceId;
    public Integer datasetId;
    ....
}
```

```
    @SqlBatch("INSERT INTO dar_dataset (reference_id, dataset_id) VALUES (:referenceId, :datasetId)")
    void insertAllDarDatasets(@BindBean List<DarDataset> darDatasets);
```

src/main/resources/changesets/changelog-consent-82.0.xml
```diff
@@ -1,6 +1,8 @@
 <databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
                    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
          http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <!--Create the dar_dataset table-->
```
rushtong: We can't make modifications to this file after it has already been executed on a database. Any schema changes need to be in a separate changelog file.

rushtong (changes requested):
Still reviewing, but there are a couple of changes that can be made now. See them inline.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -313,6 +335,11 @@ public DataAccessRequest createDraftDarFromCanceledCollection(User user, DarColl
             now,
             newData
         );
+
+        // Update the Joins table *dar_dataset
+        dataAccessRequestDAO.deleteDARDatasetRelationByReferenceIds(canceledReferenceIds);
```
rushtong: I'm pretty sure we don't want to do this. Even though a canceled DAR is canceled, that doesn't mean it should not be associated to the dataset that the user requested it to be associated to. We do want to sync the new DAR, but that's really all we want to do here.

src/main/resources/changesets/changelog-consent-82.0.xml
phendriksen100: Ooopps sorry forgot to remove the comment

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
```diff
@@ -2,11 +2,7 @@
 
 import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.commons.lang3.RandomUtils;
-import org.broadinstitute.consent.http.models.DarCollection;
-import org.broadinstitute.consent.http.models.DataAccessRequest;
-import org.broadinstitute.consent.http.models.DataAccessRequestData;
-import org.broadinstitute.consent.http.models.Dataset;
-import org.broadinstitute.consent.http.models.User;
+import org.broadinstitute.consent.http.models.*;
```
rushtong: Avoid * imports. There is a setting in intellij that lets you configure the value where it starts doing this for you. I set mine to 9999 or something crazy.

Screen Shot 2022-06-16 at 11 10 32 AM

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -241,9 +232,30 @@ public DataAccessRequest insertDraftDataAccessRequest(User user, DataAccessReque
...
+     * First delete any rows with the current reference id. This will allow us to keep (referenceId, dataset_id) unique
+     * Takes in a list of datasetIds and a referenceId and adds them to the dar_dataset collection
+     *
+     * @param datasetIds List of Integers that represent the datasetIds
+     * @param referenceId ReferenceId of the corresponding DAR
+     */
+    private void syncDataAccessRequestDatasets(List<Integer> datasetIds, String referenceId) {
+        List<DarDataset> darDatasets = datasetIds.stream()
+                .map(datasetId -> new DarDataset(referenceId, datasetId))
+                .collect(Collectors.toUnmodifiableList());
```
rushtong: Nice! I have seen wonkiness in the past with `toUnmodifiableList` ... if you find any, switch it out to a plain list.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -17,17 +17,7 @@
...
-import org.broadinstitute.consent.http.models.Dac;
-import org.broadinstitute.consent.http.models.DarCollection;
-import org.broadinstitute.consent.http.models.DataAccessRequest;
-import org.broadinstitute.consent.http.models.DataAccessRequestData;
-import org.broadinstitute.consent.http.models.DataAccessRequestManage;
-import org.broadinstitute.consent.http.models.Dataset;
-import org.broadinstitute.consent.http.models.Election;
-import org.broadinstitute.consent.http.models.User;
-import org.broadinstitute.consent.http.models.Vote;
+import org.broadinstitute.consent.http.models.*;
```
rushtong: nit: revert * import

src/main/java/org/broadinstitute/consent/http/service/SummaryService.java
```diff
@@ -153,7 +153,7 @@ public List<ConsentSummaryDetail> describeConsentSummaryDetail() {
             Stream.of(ElectionStatus.CLOSED.getValue(), ElectionStatus.CANCELED.getValue())
                 .map(String::toLowerCase)
                 .collect(Collectors.toList());
-        List<Election> reviewedElections = electionDAO.findElectionsWithFinalVoteByTypeAndStatus(ElectionType.TRANSLATE_DUL.getValue(), statuses).stream().filter(e -> Objects.nonNull(e.getFinalVote())).distinct().collect(Collectors.toUnmodifiableList());
+        List<Election> reviewedElections = electionDAO.findElectionsWithFinalVoteByTypeAndStatus(ElectionType.TRANSLATE_DUL.getValue(), statuses).stream().filter(e -> Objects.nonNull(e.getFinalVote())).distinct().collect(Collectors.toList());
```
rushtong: nit: Changes here seem unrelated to the intent of the PR, I would revert these. It is true, that in the past, I've had issues with `toUnmodifiableList`, that's not the case here.

src/main/resources/changesets/changelog-consent-82.0.xml
```diff
@@ -10,18 +11,14 @@
...
+                                 deferrable="false" initiallyDeferred="false" onDelete="CASCADE" onUpdate="NO ACTION"
                                  referencedColumnNames="datasetid" referencedTableName="dataset"
-                                 referencesUniqueColumn="true"/>
+                                 referencesUniqueColumn="false"/>
         <addForeignKeyConstraint baseColumnNames="reference_id" baseTableName="dar_dataset" constraintName="fkReferenceId"
-                                 deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION"
+                                 deferrable="false" initiallyDeferred="false" onDelete="CASCADE" onUpdate="NO ACTION"
                                  referencedColumnNames="reference_id" referencedTableName="data_access_request"
-                                 referencesUniqueColumn="true"/>
+                                 referencesUniqueColumn="false"/>
```
rushtong: All schema changes you want to make need to be in a new, separate changelog file. This file has already been committed to prod so updating it here will have no effect.

src/test/java/org/broadinstitute/consent/http/service/SummaryServiceTest.java
```diff
@@ -255,7 +255,7 @@ private List<Vote> createVotes(Integer electionId, Integer userId) {
                 v.setDacUserId(userId);
                 return v;
             }
-        ).collect(Collectors.toUnmodifiableList());
+        ).collect(Collectors.toList());
```
rushtong: Same feedback here, changes in this file are unrelated to the PR.

rushtong (changes requested):
This is coming along nicely! See comments inline.

src/main/resources/changesets/changelog-consent-82.0.xml
```diff
@@ -10,10 +11,6 @@
                 <constraints nullable="false"/>
             </column>
         </createTable>
-        <addPrimaryKey
-                columnNames="dataset_id, reference_id"
-                constraintName="join_table_key"
-                tableName="dar_dataset"/>
```
rushtong: Please revert changes to this file. Since this changeset has been run on all environments, it will not be re-run again so any changes here will not be applied to the database.

src/main/resources/changesets/changelog-consent-83.0.xml
```diff
@@ -0,0 +1,15 @@
...
+
+    <changeSet id="83.0" author="phendriksen">
+        <addForeignKeyConstraint baseColumnNames="dataset_id" baseTableName="dar_dataset" constraintName="fkDatasetId"
+                                 deferrable="false" initiallyDeferred="false" onDelete="CASCADE" onUpdate="NO ACTION"
+                                 referencedColumnNames="datasetid" referencedTableName="dataset"
+                                 referencesUniqueColumn="false"/>
+        <addForeignKeyConstraint baseColumnNames="reference_id" baseTableName="dar_dataset" constraintName="fkReferenceId"
+                                 deferrable="false" initiallyDeferred="false" onDelete="CASCADE" onUpdate="NO ACTION"
+                                 referencedColumnNames="reference_id" referencedTableName="data_access_request"
+                                 referencesUniqueColumn="false"/>
```
rushtong: I would prefer not to have any cascading deletes. We generally prefer not to hard delete entities and prefer soft deletes. I don't think we need this changelog file.

rushtong (changes requested):
Looking better!

rushtong (approved):
Thank you for all of the changes! This works perfectly 👍🏽

JVThomas (approved):
Looks good to me

PR #1615, opened 2022-06-24
[DUOS-344][risk=no] Standardize the Users table.

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-344

* Migrate`dacuser` table to `users`; `user` is a reserved word in postgresql so we can't use that.
* Migrate `dacuserid` to `user_id`
* Migrate `displayname` to `display_name`
* Migrate `createdate` to `create_date`
* Significant DAO refactoring. There are a number of cases where other tables have a `user_id` column so affected queries needed some rework.

----

- Verify all tests go green


JVThomas:
I noticed this issue when pairing the API with the UI. The `currentUser` data in Storage is carrying a user object with `userId` present and not `dacUserId`, causing validation/rendering issues wherever `dacUserId` is expected.

In the example below, the vote button never shows up on the Chair Console because it can't check `currentUser.dacUserId` against `vote.dacUserId` (essentially undefined === vote.dacUserId check).

Screen Shot 2022-06-27 at 11 54 34 AM

rushtong:
> I noticed this issue when pairing the API with the UI. The `currentUser` data in Storage is carrying a user object with `userId` present and not `dacUserId`, causing validation/rendering issues wherever `dacUserId` is expected.

Oof, yeah, I missed some legacy population when doing some cleanup. These changes ensure that we always set the value of `dacUserId` which fixes the underlying problem

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -63,19 +63,27 @@ public UserService(UserDAO userDAO, UserPropertyDAO userPropertyDAO, UserRoleDAO
...
     public static class SimplifiedUser {
+        public Integer userId;
+        @Deprecated
         public Integer dacUserId;
         public String displayName;
         public String email;
 
         public SimplifiedUser(User user) {
+            this.userId = user.getUserId();
+            this.dacUserId = user.getUserId();
```
quazi-h, line 74: Was wondering if retaining the dacUserId here was intentional (lines 68, 74, 87, 88)

quazi-h (commented):
I found a few files where "dacUserId" showed up when searching the consent files.

yaml files:

api-docs.yaml
line #3402 dacuserId -> userId
line #3513 dacUserId -> userId

User.yaml
 line #6 dacUserId -> userId
 
Vote.yaml
 line #10 dacUserId -> userId
 
 Working my way through the test and other project files still.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
rushtong: This is very intentional. Until the UI can rely on only `userId`, we need to continue populating the `dacUserId` value.

rushtong:
> I found a few files where "dacUserId" showed up when searching the consent files.
> ...
> Vote.yaml line #10 dacUserId -> userId
> 
> Working my way through the test and other project files still.

There are two objects, `Vote` and `Match` that still have a `dacUserId` column. This PR doesn't try to address those, just the usages of User.dacUserId

rushtong:
> I found a few files where "dacUserId" showed up when searching the consent files.
> 
> yaml files:
> 
> api-docs.yaml line #3402 dacuserId -> userId line #3513 dacUserId -> userId
> 
> User.yaml line #6 dacUserId -> userId
> 

Nice catch - I'll update the User.yaml and double check api-docs.yaml

rushtong:
@quazi-broad - All existing references to `dacUserId` or `dacuserId` in the api-docs are intentional. In the User case, we need to continue populating that field until the UI is migrated over. In the other cases, there are other objects that have not been converted.

JVThomas (approved):
Other than pending API doc updates (like a brief explanation of why both ids are used for user) the changes seem good to me.

quazi-h (approved):
PR looks good to me, makes sense that we should still populate dacUserId for the UI until we update it to point at userId.

PR #1628, opened 2022-06-30
[DUOS-1892][risk=no] update get institution by to include signing officials 

shaemarks:
Addresses [DUOS-1892](https://broadworkbench.atlassian.net/browse/DUOS-1892)
Summary of Changes:
- Modified GET `/api/institutions/{id}` endpoint to include an array of signing officials belonging to the institution in its response 
- Added tests in `InstitutionServiceTest`
- Moved yaml docs for the endpoint to a separate path and documented change to endpoint  
----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
@@ -65,10 +65,13 @@ public class InstitutionService {
   public Institution findInstitutionById(Integer id) {
     Institution institution = institutionDAO.findInstitutionById(id);
     isInstitutionNull(institution);
+
     List<SimplifiedUser> signingOfficials = userDAO.getSOsByInstitution(id).stream()
             .map(SimplifiedUser::new)
             .collect(Collectors.toList());
-    return institutionDAO.findInstitutionById(id);
+    institution.setSigningOfficials(signingOfficials);
+
+    return institution;
   }
 
   public List<Institution> findAllInstitutions() {
--- a/src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
@@ -3,21 +3,26 @@ package org.broadinstitute.consent.http.service;
 import com.google.inject.Inject;
 
 import org.broadinstitute.consent.http.db.InstitutionDAO;
+import org.broadinstitute.consent.http.db.UserDAO;
 import org.broadinstitute.consent.http.models.Institution;
+import org.broadinstitute.consent.http.service.UserService.SimplifiedUser;
 
 import java.util.Date;
 import java.util.List;
 import java.util.Objects;
+import java.util.stream.Collectors;
 
 import javax.ws.rs.NotFoundException;
 
 public class InstitutionService {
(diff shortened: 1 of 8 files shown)


src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
```diff
@@ -60,7 +65,13 @@ public void deleteInstitutionById(Integer id) {
   public Institution findInstitutionById(Integer id) {
     Institution institution = institutionDAO.findInstitutionById(id);
     isInstitutionNull(institution);
-    return institutionDAO.findInstitutionById(id);
+
+    List<SimplifiedUser> signingOfficials = userDAO.getSOsByInstitution(id).stream()
+            .map(SimplifiedUser::new)
+            .collect(Collectors.toList());
+    institution.setSigningOfficials(signingOfficials);
```
rushtong, line 72: An optimization here would be to combine the two DAO calls into one. That would be a somewhat larger refactor as it would touch the mappers and reducers. This works, however.

src/main/resources/assets/paths/institutionById.yaml
```diff
@@ -0,0 +1,67 @@
...
+      required: true
+      schema:
+        type: string
+  requestBody:
+    description: Required JSON payload representing updated Institution
+    required: true
+    content:
+      application/json:
+        schema:
+          $ref: '#/components/schemas/Institution'
```
rushtong: We could add `signingOfficials` as a field in `Institution`. Additionally, since we're moving the path file here (nice work!) we could refer to the yaml directly.
```suggestion
          $ref: '../schemas/Institution.yaml'
```

rushtong: See also this error when loading the swagger UI:

Screen Shot 2022-07-05 at 8 40 22 AM

rushtong (changes requested):
Once the swagger issue is fixed, this should be good to go.

quazi-h (approved):
Changes look good to me, calling the /api/institutions/{id} endpoint returns a list of signing officials in the response. I did not encounter the Resolver Error Greg mentioned seeing when accessing the swagger UI locally.

I noticed the updated endpoint is returning an extra field for each SO that isn't specified in the ticket - dacUserId but after tracing that I think that is expected while we are working to deprecate dacUserId. (The setSigningOfficials call returns a List<SimplifiedUser> which includes the @Deprecated dacUserId field.)

src/main/resources/assets/paths/institutionById.yaml
shaemarks: Since `PUT api/institutions/{id}` displays the `Institution` schema for the request body but does not use `signingOfficials` as part of the institution update, is it best to 
1) not add `signingOfficials` as a field to `Institution` schema 
2) include `signingOfficials` as a field anyway
3) have separate `Institution` schemas: one with `signingOfficials`, one without

rushtong: Great observation. I would add it as a field, but would add to the description that it is not an updatable field. We have this problem in some other cases as well.

rushtong (approved):
Thank you for the updates! 👍🏽

PR #1633, opened 2022-07-05
[DUOS-1638][risk=no] Translate data use based on type when creating a dataset

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1638

Fixes a bug where we were using the wrong translation endpoint when creating a `UseRestriction` for a dataset.

----

- Verify all tests go green


quazi-h (approved):
New translation url for consents matches what's specified in the ticket and the parseUseRestriction method was updated to take an additional translation type argument that is used to decide whether to call the DarTranslate url vs the getConsentTranslate url. This was a cleaner approach with minimal changes to the DatasetService createConsentForDataset endpoint.

shaemarks (approved):
Looks good to me! Tested by creating a dataset and creating a dar, and they correctly used their respective data use translation urls upon submission

PR #1640, opened 2022-07-07
[DUOS-1814][risk=low] New API for updating select user fields

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1814

----

- Verify all tests go green

### DIFF
--- a/src/main/resources/assets/paths/userById.yaml
+++ b/src/main/resources/assets/paths/userById.yaml
@@ -1,4 +1,5 @@
 put:
+  deprecated: true
   summary: Update a limited set of user fields
   description: |
     Updates the user identified by the ID. Requires the authenticated user to have the same id as that of the user being updated except in the admin case - admins can update information for other users. A user's email cannot be updated. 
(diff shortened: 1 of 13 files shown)


shaemarks (approved):
Code looks and functions great! Can confirm that I was able to update the fields listed in the ticket and I was not able to delete the researcher role, alter chair or member roles, or do any updates without an admin role

src/main/resources/assets/paths/userById.yaml
```diff
@@ -1,4 +1,5 @@
 put:
+  deprecated: true
```
shaemarks, line 2: My Swagger page isn't showing this endpoint as deprecated for some reason

rushtong: Let me look into that - I don't think I checked for the deprecation in the ui.

rushtong: I'm seeing it as deprecated here:

Screen Shot 2022-07-11 at 5 24 49 PM

quazi-h (approved):
Changes here look good to me. I was able to use the new endpoint to update the specified fields. I tested by using an admin user to update the institutionId, suggestedInstitution, and eraCommonsId and confirned the changes applied to the user when querying /user/me.

PR #1643, opened 2022-07-08
[DUOS-1818][risk=no] Remove additionalEmail, academicEmail, and checkNotifications

connorlbark:
## Addresses 

https://broadworkbench.atlassian.net/browse/DUOS-1818

----

- Verify all tests go green


src/test/java/org/broadinstitute/consent/http/db/UserDAOTest.java
```diff
@@ -169,22 +169,6 @@ public void testInsertDACUser() {
...
-        userDAO.updateUser(
-                "Dac User Test",
-                user.getUserId(),
-                newEmail,
-                firstInstitute.getId()
-                );
-        User user2 = userDAO.findUserById(user.getUserId());
-        assertEquals(user2.getAdditionalEmail(), newEmail);
-        assertEquals(user2.getInstitution().getId(), firstInstitute.getId());
-    }
```
rushtong: I think we still need this test case,  but without the email component. This test is actually testing two cases, institution and email updates. We should maintain the institution part of this test. I do see usages in the helper class, but that's not sufficient - we should have a test for `userDAO.updateUser` here too.

src/main/java/org/broadinstitute/consent/http/service/EmailNotifierService.java
```diff
@@ -263,19 +263,11 @@ public void sendDataCustodianApprovalMessage(String toAddress,
     }
 
     private Set<String> getEmails(List<User> users) {
```
rushtong, line 264: Lines 151 and 163 also make reference to `additionalEmail` so we should refactor those a bit. It looks like we can simplify that quite a bit in both cases:
```
Set<String> emails = Set.of(data.get("email"));
```

src/main/java/org/broadinstitute/consent/http/enumeration/UserFields.java
```diff
@@ -4,7 +4,6 @@
 import java.util.List;
 
 public enum UserFields {
-  ACADEMIC_BUSINESS_EMAIL("academicEmail", true),
```
rushtong, line 7: The ticket misses this, but I do see this in the code. We should deprecate this from `DataAccessRequestData` as well, my apologies for not covering that in DUOS-1818. You'll see some convoluted logic around this in `ElectionService` too. 

```
String researcherEmail = Objects.nonNull(researcher) ?
        researcher.getEmail() :
        Objects.nonNull(dar.getData().getAcademicEmail()) ?
                dar.getData().getAcademicEmail() :
                dar.getData().getResearcher();
```
should probably be:
```
String researcherEmail = Objects.nonNull(researcher) ? researcher.getEmail() : null;
```
with an additional null check around 
```
emailNotifierService.sendDataCustodianApprovalMessage(custodian.getEmail(), darCode, mailDTOS, custodian.getDisplayName(), researcherEmail);
```
so we're not trying to send an email to someplace we cannot send it. This would deserve a log.error call as well so we can catch these kinds of things.

rushtong (changes requested):
Have not tested functionally, but there are a couple things we still need to do here. Nice work so far!

src/test/java/org/broadinstitute/consent/http/db/UserDAOTest.java
connorlbark: ahh, didn't see the institution change - brought the test back!

src/main/java/org/broadinstitute/consent/http/enumeration/UserFields.java
connorlbark: thanks for the tips! removed the field; seemed it wasn't used anywhere else.

quazi-h (commented):
Looking good so far, when loading the changes and running in a local instance, everything compiled okay. When hitting the /user/me endpoint academicEmail and checkNotifications was being returned under "researcherProperties" - is that expected?

I did a broad search across the consent repo for all references to academicEmail, additionalEmail, and checkNotifications and found the following occurences:

DarService.scala
line 55: academicEmail

User.java
line 30: additional_email

InstititutionDAO.java
lines 64: additional_email
lines 68: additional_email

DacDAO.java
line 75: additional_email

JsonProtocols.scala
line 146: checkNotifications
line 120: academicEmail
line 213: academicEmail


DataAccessRequestData.java
line 29: add "additionalEmail" and "checkNotifications" to the list of DEPRECATED_PROPS?

ResearcherModels.scala
line 15: academicEmail
line 41: checkNotifications


DataAccessRequestModels.scala
line 65: academicEmail

rushtong:
@quazi-broad - https://broadworkbench.atlassian.net/browse/DUOS-1896 should filter out the deprecated fields from what we have in the db presently.

Nice catch on the dao calls, I missed those completely

rushtong:
We don't need to modify the scala classes for this PR - those are all optional and should remain so.

connorlbark:
thanks for all the great feedback! i believe i addressed everything: DAO calls updated, additionalEmail and checkNotifications added to deprecated fields, also fixed merge conflicts with develop.

quazi-h (approved):
Thanks for the changes! I think this is looking good.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequestData.java
```diff
@@ -26,7 +26,7 @@ public class DataAccessRequestData {
             "profileName", "pubmedId", "scientificUrl", "eraExpiration", "academicEmail",
             "eraAuthorized", "nihUsername", "linkedIn", "orcid", "researcherGate", "datasetDetail",
             "datasets", "datasetId", "validRestriction", "restriction", "translatedUseRestriction",
-            "createDate", "sortDate");
+            "createDate", "sortDate", "additionalEmail", "checkNotifications" );
```
rushtong, line 29: This is a non-issue, but I thought I would describe it for posterity. `DEPRECATED_PROPS` is used to create a simplified version of the DAR where it filters out all of these values. I initially thought we would still need to do that filtering, but it turns out that since you removed the getters/setters, these fields are never populated in any db call. So, nice work.

rushtong (approved):
Nice work, thank you!

PR #1645, opened 2022-07-11
[DUOS-1890][risk=no] add support for zendesk tickets

shaemarks:
Addresses [DUOS-1890](https://broadworkbench.atlassian.net/browse/DUOS-1890)
Summary of Changes:
- Create classes to recreate structure of a ticket for posting to zendesk api 
- Create SupportRequestService class for creating tickets and posting them to the api to submit support requests 
- Add tests to SupportRequestServiceTest

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/models/supportticket/CustomRequestField.java
```diff
@@ -0,0 +1,43 @@
+package org.broadinstitute.consent.http.models.supportticket;
```
rushtong: minor package naming nit, can we rename this to `org.broadinstitute.consent.http.models.support` instead?

rushtong (changes requested):
This looks good, but I'm worried that the format isn't exactly right. We're posting JSON with snake case naming and I think we're not specifying that when we JSON-ify the ticket object. For example, will `customFields` (on `SupportRequest`) translate to `customFields` or `custom_fields`? Zendesk is expecting `custom_fields`.  Same concept applies to `ticket_form_id`. Can we add a test that looks for the body of the post and ensure the format is accurate?

src/main/java/org/broadinstitute/consent/http/service/SupportRequestService.java
```diff
@@ -0,0 +1,91 @@
...
+        customFields.add(new CustomRequestField("360012782111", email));
+        customFields.add(new CustomRequestField("360018545031", email));
+        SupportRequestComment comment = new SupportRequestComment(description, url);
+
+        return new SupportTicket(requester, subject, customFields, comment);
+    }
+
+    public void postTicketToSupport(SupportTicket ticket, AuthUser authUser) throws Exception {
+        GenericUrl genericUrl = new GenericUrl(configuration.postSupportRequestUrl());
+        JsonHttpContent content = new JsonHttpContent(new GsonFactory(), ticket);
```
rushtong: I think this is where we need to do some extra work. Standard Gson conversion of the object will camel-case the field names. I was able to get the right format with this code:
```
System.out.println(new GsonBuilder()
    .setPrettyPrinting()
    .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
    .create()
    .toJson(ticket));
```
but then we need to somehow use the naming policy in `GsonFactory` which I don't think we can do. Looking into workarounds in the meantime.

shaemarks: I tried passing the formatted string created by `GsonBuilder` as the data for `JsonHttpContent` hoping that since it's just a string, `GsonFactory` wouldn't do anything else to it. Now `request.content.data` looks the same as `expectedBody` in `testPostTicketToSupport`, but the test doesn't pass if I make `mockServerClient` look for `expectedBody` in the request. I'm not sure if this is an issue with the request itself not being structured properly or me not setting up `mockServerClient` correctly to check the request body

shaemarks: If I use this in the test: 
``` 
 mockServerClient.when(request())
                .respond(response()
                        .withHeader(Header.header("Content-Type", "application/json"))
                        .withStatusCode(HttpStatusCodes.STATUS_CODE_OK));

 service.postTicketToSupport(ticket, authUser);
 LogEventRequestAndResponse[] requests = mockServerClient.retrieveRecordedRequestsAndResponses(null);
```
to see all requests and responses but not require a specific request body for `mockServerClient `to work, then this is the request sent:
``` {
  "method" : "POST",
  "path" : "/",
  "headers" : {
    "x-app-id" : [ "DUOS" ],
    "content-length" : [ "435" ],
    "User-Agent" : [ "Google-HTTP-Java-Client/1.42.0 (gzip)" ],
    "Host" : [ "localhost:53764" ],
    "Content-Type" : [ "application/json; charset=UTF-8" ],
    "Connection" : [ "keep-alive" ],
    "Authorization" : [ "Bearer null" ],
    "Accept-Encoding" : [ "gzip" ],
    "Accept" : [ "application/json" ]
  },
  "keepAlive" : true,
  "secure" : false,
  "remoteAddress" : "172.17.0.1",
  "body" : "{\n  \"request\": {\n    \"requester\": {\n      \"name\": \"nVuVnlRfWH\",\n      \"email\": \"ppgJuTWFkI\"\n    },\n    \"subject\": \"PDctpZQwpK\",\n    \"custom_fields\": [\n      {\n        \"id\": 6428364382081724521,\n        \"value\": \"JXwYyyoKmM\"\n      }\n    ],\n    \"comment\": {\n      \"body\": \"WahAaCBerC\\n\\n------------------\\nSubmitted from: FVYhtibdrv\"\n    },\n    \"ticket_form_id\": 360000669472\n  }\n}"
}
```

src/main/java/org/broadinstitute/consent/http/models/support/SupportRequestComment.java
```diff
@@ -0,0 +1,20 @@
+package org.broadinstitute.consent.http.models.support;
+
+/**
+ * Represents comment describing issue for a Zendesk support ticket https://broadinstitute.zendesk.com/api/v2/requests.json
+ */
+public class SupportRequestComment {
+    private String body;
+
+    public SupportRequestComment(String description, String url) {
+        this.body = description + "\n\n------------------\nSubmitted from: " + url;
```
connorlbark: optional small nit: maybe description and url should be class fields, then getBody combines them to a string? just in case we need that data separately in the future.

connorlbark (commented):
looks great! i'll check it out again once the tests are fixed and approve then

rushtong (approved):
Looks great, glad we're able to test the body content.

src/main/java/org/broadinstitute/consent/http/service/SupportRequestService.java
shaemarks: Update: Still wasn't able to get `mockServerClient.when(request().with(expectedBody))...` to work for `testPostTicketToSupport`, but I got the test working by retrieving the request afterwards with `mockServerClient.retrieveRecordedRequests` and comparing the expected and actual request bodies.

shaemarks: Update: Still wasn't able to get `mockServerClient.when(request().with(expectedBody))...` to work for `testPostTicketToSupport`, but I got the test working by retrieving the request afterwards with `mockServerClient.retrieveRecordedRequests` and comparing the expected and actual request bodies.

src/main/java/org/broadinstitute/consent/http/models/support/SupportRequestComment.java
shaemarks: I think adding class fields would require putting extra customizations on the `GsonBuilder` to make sure those fields are excluded when the ticket is converted to json. 
Also, I think sometime between you starting and posting your code review I actually switched to putting the `description + "\n\n------------------\nSubmitted from: " + url` formatting in `SupportRequestService.createSupportTicket`, partly to make it easier to test the request body structure, partly because the api would accept any String as the comment body. I don't have a strong opinion either way, so let me know if you think one is best.

shaemarks: I think adding class fields would require putting extra customizations on the `GsonBuilder` to make sure those fields are excluded when the ticket is converted to json. 
Also, I think sometime between you starting and posting your code review I actually switched to putting the `description + "\n\n------------------\nSubmitted from: " + url` formatting in `SupportRequestService.createSupportTicket`, partly to make it easier to test the request body structure, partly because the api would accept any String as the comment body. I don't have a strong opinion either way, so let me know if you think one is best.

connorlbark: ohhh okay, that makes sense. all good!

connorlbark (approved):
looks great to me as well!!

PR #1648, opened 2022-07-12
[DUOS-1764, DUOS-1765][risk=no] Create a single DAR for N Datasets on submission

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1764
https://broadworkbench.atlassian.net/browse/DUOS-1765

## Changes
* Update to the DAR submission process so we only create a single DAR instead of N DARs per dataset requested.
* Updates to resource class and tests to cover this change.

Note that we should immediately follow up on https://broadworkbench.atlassian.net/browse/DUOS-1961 as that story is blocked by this one and we will no longer need to archive a group of DARs (DARCollection) and recreate a new draft from them.

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -318,20 +317,4 @@ void insertDataAccessRequest(
...
-  @SqlQuery("SELECT distinct reference_id, dataset_id FROM dar_dataset WHERE reference_id IN (<referenceIds>)")
-  List<DarDataset> findAllDARDatasets(@BindList("referenceIds") List<String> referenceIds);
-
-  @SqlQuery(
-      " SELECT distinct d.reference_id "
-          + " FROM dar_dataset d "
-          + " INNER JOIN data_access_request dar ON dar.reference_id = d.reference_id AND dar.collection_id = :collectionId "
-          + " WHERE d.dataset_id IN <datasetIds> ")
-  List<String> findReferenceIdsForDatasetIdsWithCollectionId(
-      @BindList("datasetIds") List<Integer> datasetIds, @Bind("collectionId") Integer collectionId);
```
rushtong, line 335: Unused

src/main/java/org/broadinstitute/consent/http/db/ElectionDAO.java
```diff
@@ -268,6 +270,19 @@ List<Election> findLastElectionsByReferenceIds(
     @UseRowMapper(SimpleElectionMapper.class)
     Election findLastElectionByReferenceIdAndType(@Bind("referenceId") String referenceId, @Bind("type") String type);
 
+    @SqlQuery(
+        " SELECT e.* " +
+        " FROM election e " +
+        " INNER JOIN (SELECT referenceid, datasetid, MAX(createdate) max_date FROM election WHERE LOWER(electiontype) = lower(:type) AND datasetid = :datasetId GROUP BY referenceid, datasetid) election_view " +
```
JVThomas, line 276: Why use MAX(createdate) over something like MAX(electionid)?

rushtong: Mostly because we use it a lot in other queries. I believe both approaches will work equally well.

shaemarks (commented):
Looks good overall! When run locally with the ui, I submitted a multi-dataset dar, and the resulting collection contained the single dar with all of the datasets.

However, I think some changes will need to be made to the bucketing logic on the frontend to support this change. The collection I created contained 4 datause buckets, each containing all of the elections since they are processed by dar. The votes are then processed by elections, so voting shouldn't be accurate with these changes.

processDataUseBuckets

rushtong:
> However, I think some changes will need to be made to the bucketing logic on the frontend to support this change. The collection I created contained 4 datause buckets, each containing all of the elections since they are processed by dar. The votes are then processed by elections, so voting shouldn't be accurate with these changes.

I did actually test this case, but the bucketing and votes looked correct to me when I did so. That could have been an edge case where it worked, and there are others where it doesn't ... I'll try to dig more into the front end and figure something out.

src/main/java/org/broadinstitute/consent/http/service/dao/DarCollectionServiceDAO.java
```diff
@@ -79,59 +79,62 @@ public List<String> createElectionsForDarCollection(User user, DarCollection col
           //    4. Member votes for rp election
           //        4a. Chair Vote for chair
           collection.getDars().values().forEach(dar -> {
-                // If there is an existing open election for this DAR, we can ignore it
-                Election lastDataAccessElection = electionDAO.findLastElectionByReferenceIdAndType(dar.getReferenceId(), ElectionType.DATA_ACCESS.getValue());
-                boolean ignore = Objects.nonNull(lastDataAccessElection) && lastDataAccessElection.getStatus().equals(ElectionStatus.OPEN.getValue());
+                dar.getDatasetIds().forEach(datasetId -> {
+                    // If there is an existing open election for this DAR+Dataset, we can ignore it
+                    Election lastDataAccessElection = electionDAO.findLastElectionByReferenceIdDatasetIdAndType(dar.getReferenceId(), datasetId, ElectionType.DATA_ACCESS.getValue());
+                    boolean ignore = Objects.nonNull(lastDataAccessElection) && lastDataAccessElection.getStatus().equals(ElectionStatus.OPEN.getValue());
```
JVThomas: Nit: Just to be safe, can you update the old `equals` usage with `equalsIgnoreCase`?

JVThomas (approved):
Minor nitpick, but the code looks good. That said, Shae is right about the bucketing logic on the front-end. It needs to be updated to account for the "one dar to many elections" relationship. An update there needs to happen before this PR can go to staging/production.

shaemarks (approved):

PR #1649, opened 2022-07-13
[DUOS-1815][risk=low] Add update own user properties endpoint

connorlbark:
## Partially Addresses 

https://broadworkbench.atlassian.net/browse/DUOS-1815

Adds needed `PUT /api/user/` endpoint so non-admin users can update their properties.

Frontend component: https://github.com/DataBiosphere/duos-ui/pull/1672

Also fixes bug wherein signing officials are not fully returned for non-admin users; this is necessary for the linked frontend issue, so I thought it was an appropriate place to fix it.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
@@ -19,6 +19,7 @@ import org.broadinstitute.consent.http.service.ResearcherService;
 import org.broadinstitute.consent.http.service.UserService;
 import org.broadinstitute.consent.http.service.UserService.SimplifiedUser;
 import org.broadinstitute.consent.http.service.sam.SamService;
+import org.glassfish.hk2.utilities.reflection.Logger;
 
 import javax.annotation.security.PermitAll;
 import javax.annotation.security.RolesAllowed;
--- a/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
@@ -177,11 +177,10 @@ public class UserResource extends Resource {
             UserUpdateFields userUpdateFields = gson.fromJson(json, UserUpdateFields.class);
             // Ensure that we have a real user with this ID, fail if we do not.
             userService.findUserById(userId);
-            URI uri = info.getRequestUriBuilder().path("{id}").build(userId);
             User updatedUser = userService.updateUserFieldsById(userUpdateFields, userId);
             Gson gson = new Gson();
             JsonObject jsonUser = userService.findUserWithPropertiesByIdAsJsonObject(authUser, updatedUser.getUserId());
-            return Response.ok(uri).entity(gson.toJson(jsonUser)).build();
+            return Response.ok().entity(gson.toJson(jsonUser)).build();
         } catch (Exception e) {
             return createExceptionResponse(e);
         }
@@ -197,15 +196,14 @@ public class UserResource extends Resource {
             UserUpdateFields userUpdateFields = gson.fromJson(json, UserUpdateFields.class);
 
             if (Objects.nonNull(userUpdateFields.getUserRoleIds()) && !user.hasUserRole(UserRoles.ADMIN)) {
-                throw new BadRequestException("Cannot change user roles");
+                throw new BadRequestException("Cannot change user's roles.");
             }
 
             user = userService.updateUserFieldsById(userUpdateFields, user.getUserId());
             Gson gson = new Gson();
             JsonObject jsonUser = userService.findUserWithPropertiesByIdAsJsonObject(authUser, user.getUserId());
 
-            URI uri = info.getRequestUriBuilder().path("").build();
-            return Response.ok(uri).entity(gson.toJson(jsonUser)).build();
+            return Response.ok().entity(gson.toJson(jsonUser)).build();
         } catch (Exception e) {
             return createExceptionResponse(e);
         }
--- a/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
@@ -175,9 +175,9 @@ public class UserResource extends Resource {
         try {
             UserUpdateFields userUpdateFields = gson.fromJson(json, UserUpdateFields.class);
             // Ensure that we have a real user with this ID, fail if we do not.
-            User user = userService.findUserById(userId);
+            userService.findUserById(userId);
             URI uri = info.getRequestUriBuilder().path("{id}").build(userId);
-            User updatedUser = userService.updateUserFieldsById(user, userUpdateFields, userId);
+            User updatedUser = userService.updateUserFieldsById(userUpdateFields, userId);
             Gson gson = new Gson();
             JsonObject jsonUser = userService.findUserWithPropertiesByIdAsJsonObject(authUser, updatedUser.getUserId());
             return Response.ok(uri).entity(gson.toJson(jsonUser)).build();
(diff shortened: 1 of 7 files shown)


src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -70,7 +72,12 @@ public UserService(UserDAO userDAO, UserPropertyDAO userPropertyDAO, UserRoleDAO
      * @param userId The User's ID
      * @return The updated User
      */
-    public User updateUserFieldsById(UserUpdateFields userUpdateFields, Integer userId) {
+    public User updateUserFieldsById(@NonNull User user, UserUpdateFields userUpdateFields, Integer userId) {
+        // must be updating oneself or be an admin
+        if (!user.hasUserRole(UserRoles.ADMIN) && !user.getUserId().equals(userId)) {
+            throw new NotAuthorizedException("Cannot update user.");
+        }
```
rushtong: We have two general practices that apply here. 
1. If you look at `Resource.validateAuthedRoleUser()`, you'll see what we usually do in this situation where a user is trying to do something they cannot.
2. We like to separate out auth logic at the Resource level since that's where we have permissions annotations. Here in the service class, we like to keep the logic limited to business rules.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -91,8 +98,9 @@ public User updateUserFieldsById(UserUpdateFields userUpdateFields, Integer user
                 userPropertyDAO.deletePropertiesByUserAndKey(userProps);
                 userPropertyDAO.insertAll(userProps);
             }
-            // Handle Roles.
-            if (Objects.nonNull(userUpdateFields.getUserRoleIds())) {
+
+            // Handle Roles; must be admin to update roles.
+            if (user.hasUserRole(UserRoles.ADMIN) && Objects.nonNull(userUpdateFields.getUserRoleIds())) {
```
rushtong: Same comment here, I would prefer not to check for permissions here. If there were business logic, i.e. admins can update roles XYZ, others can update roles EFG, then yes, that makes sense here. I think there are two easy options here. One is to break out the update method into two, one for handling all values in `UserUpdateFields`, and then another separate one for not handling roles. A second option is in the resource class, make sure that there are no roles in the `UserUpdateFields` object. That can be enforced in the resource method and throwing a BadRequestException if a user tries to update their own roles in this manner.

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -175,11 +175,11 @@ public Response update(@Auth AuthUser authUser, @Context UriInfo info, @PathPara
...
-            userService.findUserById(userId);
+            User user = userService.findUserById(userId);
             URI uri = info.getRequestUriBuilder().path("{id}").build(userId);
-            User user = userService.updateUserFieldsById(userUpdateFields, userId);
+            User updatedUser = userService.updateUserFieldsById(user, userUpdateFields, userId);
             Gson gson = new Gson();
-            JsonObject jsonUser = userService.findUserWithPropertiesByIdAsJsonObject(authUser, user.getUserId());
+            JsonObject jsonUser = userService.findUserWithPropertiesByIdAsJsonObject(authUser, updatedUser.getUserId());
             return Response.ok(uri).entity(gson.toJson(jsonUser)).build();
         } catch (Exception e) {
```
rushtong: I would suggest a completely new endpoint here. This one requires `@PathParam("id") Integer userId`. The new one would handle the same path without the ID param. Then you can just get the user from `@Auth AuthUser authUser` instead of looking them up for use in the update method.

rushtong (changes requested):
Nice work - see comments inline. We're kind of designing this API by PR, sorry we don't have more clearly laid out requirements.

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
connorlbark: I believe I added that at line 194; I made a `/me` endpoint which can be called by anyone, whereas the `/{id}` endpoint is admin-only. Let me know if that isn't what you were envisioning

src/main/java/org/broadinstitute/consent/http/service/UserService.java
connorlbark: makes perfect sense! i separated out the logic so that all auth logic is at the resource level, should address both comments. i appreciate the thorough help here!

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -177,9 +178,33 @@ public Response update(@Auth AuthUser authUser, @Context UriInfo info, @PathPara
...
+
+            if (Objects.nonNull(userUpdateFields.getUserRoleIds()) && !user.hasUserRole(UserRoles.ADMIN)) {
+                throw new BadRequestException("Cannot change user roles");
+            }
+
+            user = userService.updateUserFieldsById(userUpdateFields, user.getUserId());
             Gson gson = new Gson();
             JsonObject jsonUser = userService.findUserWithPropertiesByIdAsJsonObject(authUser, user.getUserId());
+
+            URI uri = info.getRequestUriBuilder().path("").build();
```
rushtong: I know this is copied from the other `PUT` method, but both are unnecessary. This is supposed to build a URL to return with the response, but it's not working as expected in both cases. The actual URL should look like host + `/api/user`, but in both cases, it does not. I suggest we remove it from both methods.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -91,7 +91,8 @@ public User updateUserFieldsById(UserUpdateFields userUpdateFields, Integer user
                 userPropertyDAO.deletePropertiesByUserAndKey(userProps);
                 userPropertyDAO.insertAll(userProps);
             }
-            // Handle Roles.
+
+            // Handle Roles; must be admin to update roles.
```
rushtong: nit: no longer relevant in this case, the service class should handle roles regardless at this point.

src/test/java/org/broadinstitute/consent/http/resources/UserResourceTest.java
```diff
@@ -3,7 +3,6 @@
 import com.google.api.client.http.HttpStatusCodes;
 import com.google.gson.Gson;
 import com.google.gson.JsonElement;
-import liquibase.pro.packaged.G;
```
rushtong, line 6: Nice catch, looks like a spurious IDE import.

rushtong (dismissed):
We need an update to the swagger documentation as well. Have a look at `api-docs.yaml`. We've been migrating to describing the body of the api calls in a separate yaml file in `resources/assets/paths`. See other comments inline. This is looking good!

connorlbark:
good catch on the swagger! totally forgot about that; just added it

src/main/resources/assets/paths/user.yaml
```diff
@@ -0,0 +1,64 @@
...
+          schema:
+            $ref: '../schemas/User.yaml'
+    400:
+      description: Unable to validate the user's google identity.
+    409:
+      description: User should be unique
+    500:
+      description: Server error.
+put:
+  summary: Update limited user fields by id
```
shaemarks: nit: remove "by id"

src/test/java/org/broadinstitute/consent/http/resources/UserResourceTest.java
```diff
@@ -386,6 +385,48 @@ public void testUpdate() {
...
+    when(userService.findUserByEmail(any())).thenReturn(user);
+    when(userService.updateUserFieldsById(any(), any())).thenReturn(user);
+    when(userService.findUserWithPropertiesByIdAsJsonObject(any(), any())).thenReturn(gson.toJsonTree(user).getAsJsonObject());
+    initResource();
+    Response response = userResource.updateSelf(authUser, uriInfo, gson.toJson(userUpdateFields));
+    assertEquals(HttpStatusCodes.STATUS_CODE_BAD_REQUEST, response.getStatus());
+  }
+
+  @Test
+  public void testUpdateSelfRolesNotAdmin() {
```
shaemarks: nit: I think the name of this test may be intended for the test above (line 403)? This tests the regular `userResource.update` method, not `updateSelf`

src/main/java/org/broadinstitute/consent/http/util/InstitutionUtil.java
```diff
@@ -38,7 +38,12 @@ private ExclusionStrategy getSerializationExclusionStrategy(Boolean isAdmin) {
       @Override
       public boolean shouldSkipField(FieldAttributes field) {
         String fieldName = field.getName();
-        return !isAdmin && !(fieldName.equals("id") || fieldName.equals("name") || fieldName.equals("signingOfficials"));
+        
+        return !isAdmin && !(fieldName.equals("id")
+                || fieldName.equals("name")
+                || fieldName.equals("signingOfficials")
+                || fieldName.equals("displayName")
+                || fieldName.equals("userId"));
```
shaemarks: It looks like we need to allow for `email` as well based on the mocks and [DUOS-1892](https://broadworkbench.atlassian.net/browse/DUOS-1892) (sorry about that!)

shaemarks (changes requested):
I left a couple of small comments, but looks good overall! I was able to update my own properties, and I got an error response when attempting to update roles without being an admin

src/main/java/org/broadinstitute/consent/http/util/InstitutionUtil.java
connorlbark: good catch!

src/main/java/org/broadinstitute/consent/http/util/InstitutionUtil.java
```diff
@@ -38,7 +38,13 @@ private ExclusionStrategy getSerializationExclusionStrategy(Boolean isAdmin) {
       @Override
       public boolean shouldSkipField(FieldAttributes field) {
         String fieldName = field.getName();
-        return !isAdmin && !(fieldName.equals("id") || fieldName.equals("name") || fieldName.equals("signingOfficials"));
+        
+        return !isAdmin && !(fieldName.equals("id")
+                || fieldName.equals("name")
```
lu-c, line 43: Oh, i LOVE this new formatting, it's really clean

lu-c (approved):
LGTM?

shaemarks (approved):
Looks good to me!

PR #1650, opened 2022-07-14
[DUOS-1896][risk=no] Deprecate user properties

connorlbark:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1896

Deprecates user properties; filters out DAO calls by making user specify all the fields they want, instead of returning all possible fields.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/UserPropertyDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/UserPropertyDAO.java
@@ -33,7 +33,4 @@ public interface UserPropertyDAO extends Transactional<UserPropertyDAO> {
 
     @SqlBatch("DELETE FROM user_property WHERE userid = :userId AND propertykey = :propertyKey")
     void deletePropertiesByUserAndKey(@BindBean Collection<UserProperty> researcherProperties);
-
-    @SqlQuery("SELECT * FROM user_property WHERE userid IN (<userIds>)")
-    List<UserProperty> findResearcherPropertiesByUserIds(@BindList("userIds") List<Integer> userIds);
 }
--- a/src/main/java/org/broadinstitute/consent/http/db/UserPropertyDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/UserPropertyDAO.java
@@ -17,8 +17,9 @@ import java.util.List;
 @RegisterRowMapper(UserPropertyMapper.class)
 public interface UserPropertyDAO extends Transactional<UserPropertyDAO> {
 
-    @SqlQuery("SELECT * FROM user_property WHERE userid = :userId")
-    List<UserProperty> findResearcherPropertiesByUser(@Bind("userId") Integer userId);
+    @SqlQuery("SELECT * FROM user_property WHERE userid = :userId AND propertykey IN (<properties>)")
+    List<UserProperty> findResearcherPropertiesByUser(@Bind("userId") Integer userId,
+                                                      @BindList("properties") List<String> properties);
 
     @SqlQuery("SELECT propertyvalue FROM USER_PROPERTY WHERE userid = :userId AND propertykey = 'completed'")
     String isProfileCompleted(@Bind("userId") Integer userId);
--- a/src/main/java/org/broadinstitute/consent/http/db/UserPropertyDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/UserPropertyDAO.java
@@ -21,6 +21,7 @@ public interface UserPropertyDAO extends Transactional<UserPropertyDAO> {
     List<UserProperty> findResearcherPropertiesByUser(@Bind("userId") Integer userId,
                                                       @BindList("properties") List<String> properties);
 
+    @Deprecated
     @SqlQuery("SELECT propertyvalue FROM USER_PROPERTY WHERE userid = :userId AND propertykey = 'completed'")
     String isProfileCompleted(@Bind("userId") Integer userId);
 
--- a/src/main/java/org/broadinstitute/consent/http/service/ResearcherService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/ResearcherService.java
@@ -15,10 +15,12 @@ import javax.mail.MessagingException;
 import javax.ws.rs.NotFoundException;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 
 public class ResearcherService {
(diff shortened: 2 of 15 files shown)


src/test/java/org/broadinstitute/consent/http/db/UserPropertyDAOTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.broadinstitute.consent.http.enumeration.UserFields;
+import org.broadinstitute.consent.http.enumeration.UserRoles;
+import org.broadinstitute.consent.http.models.User;
+import org.broadinstitute.consent.http.models.UserProperty;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.List;
+
+public class UserPropertyDAOTest extends DAOTestHelper {
```
rushtong, line 13: I'm very surprised we have not had tests for this DAO class for so long - I missed this completely on the initial migration from mysql -> postgresql

src/main/java/org/broadinstitute/consent/http/db/UserPropertyDAO.java
```diff
@@ -17,9 +17,11 @@
 @RegisterRowMapper(UserPropertyMapper.class)
 public interface UserPropertyDAO extends Transactional<UserPropertyDAO> {
 
-    @SqlQuery("SELECT * FROM user_property WHERE userid = :userId")
-    List<UserProperty> findResearcherPropertiesByUser(@Bind("userId") Integer userId);
+    @SqlQuery("SELECT * FROM user_property WHERE userid = :userId AND propertykey IN (<properties>)")
+    List<UserProperty> findResearcherPropertiesByUser(@Bind("userId") Integer userId,
+                                                      @BindList("properties") List<String> properties);
 
+    @Deprecated
```
rushtong, line 24: In addition, it looks like `findResearcherPropertiesByUserIds` is unused. Can we remove that in this PR?

src/main/java/org/broadinstitute/consent/http/service/ResearcherService.java
```diff
@@ -95,7 +91,8 @@ private User validateAuthUser(AuthUser authUser) {
 
     private List<UserProperty> describeResearcherProperties(Integer userId) {
         validateUser(userId);
-        return userPropertyDAO.findResearcherPropertiesByUser(userId);
+        return userPropertyDAO.findResearcherPropertiesByUser(userId,
+                Arrays.stream(UserFields.values()).map(UserFields::getValue).collect(Collectors.toList()));
```
rushtong: Something I like to do is add utility methods to enums to help clean up inline arguments like this. For example, if `UserFields` had this method:
```
  public static List<String> getValues() {
    return Stream.of(UserFields.values()).map(UserFields::getValue).collect(Collectors.toList());
  }
```
then this dao call could be streamlined:
```
return userPropertyDAO.findResearcherPropertiesByUser(userId, UserFields.getValues());
```

src/test/java/org/broadinstitute/consent/http/db/UserPropertyDAOTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Test
+    public void testFindResearcherProperties() {
+        User user = createUserWithRole(UserRoles.RESEARCHER.getRoleId());
+
+        UserProperty suggestedInstitution = new UserProperty();
+        suggestedInstitution.setPropertyKey(UserFields.SUGGESTED_INSTITUTION.getValue());
+        suggestedInstitution.setPropertyValue("asdfasdf");
```
rushtong: Not a requirement, but consider using `RandomStringUtils` for this kind of task.

src/main/java/org/broadinstitute/consent/http/db/UserPropertyDAO.java
```diff
@@ -17,9 +17,11 @@
 @RegisterRowMapper(UserPropertyMapper.class)
 public interface UserPropertyDAO extends Transactional<UserPropertyDAO> {
 
-    @SqlQuery("SELECT * FROM user_property WHERE userid = :userId")
-    List<UserProperty> findResearcherPropertiesByUser(@Bind("userId") Integer userId);
+    @SqlQuery("SELECT * FROM user_property WHERE userid = :userId AND propertykey IN (<properties>)")
+    List<UserProperty> findResearcherPropertiesByUser(@Bind("userId") Integer userId,
+                                                      @BindList("properties") List<String> properties);
```
rushtong, line 22: I really like this approach, it retains the existing outdated data so we don't lose it.

src/main/java/org/broadinstitute/consent/http/service/ResearcherService.java
connorlbark: good idea! just did it

src/main/java/org/broadinstitute/consent/http/db/UserPropertyDAO.java
connorlbark: thanks !!

rushtong:
In testing this out in the UI, there is one UI bug I think we need to solve first. The DAR requires certain profile fields to be filled out with this method in `utils.js`:
```
export const completedResearcherInfoCheck = (properties) => {
  const {
    piName,
    isThePI,
    havePI,
    piEmail,
    institutionId,
  } = properties;

  const piCheck = ({isThePI, havePI, piEmail, piName}) => {
    //conditions listed are invalid checks
    //if all are true, value returned MUST be false, since pi portion of the form is incomplete
    const isThePIFalse = !evaluateTrueString(isThePI);
    const havePITrue = evaluateTrueString(havePI);
    const piAttrEmpty = isEmpty(piName) || isEmpty(piEmail);
    return !(isThePIFalse && havePITrue && piAttrEmpty);
  };

  const stringAttrs = pick(['displayName', 'address1', 'city', 'state', 'zipCode', 'country'])(properties);
  const stringAttrsCompleted = every((string) => !isEmpty(string))(stringAttrs);
  const institutionPresent = !isNil(institutionId);
  const piValid = piCheck({isThePI, havePI, piEmail, piName});
  return piValid && stringAttrsCompleted && institutionPresent;
};
```
In the new world of a reduced profile footprint, the only thing that still needs to be validated is the institution id. Can we tackle that in a companion UI PR? 

Additionally, this PR can't be merged until the updated profile page is complete. This breaks the existing profile page since it is expecting all of these fields to exist on the user.

rushtong (approved):
Great work, thank you! Please hold off on merging this until https://github.com/DataBiosphere/duos-ui/pull/1672 lands.

connorlbark:
> In the new world of a reduced profile footprint, the only thing that still needs to be validated is the institution id. Can we tackle that in a companion UI PR?

got it! will add on a UI PR

quazi-h (approved):
Changes here look good to me. When calling the user endpoint, only the expected 5 user properties are returned. I also confirmed that the PUT/POST /api/user/profile endpoints are deprecated.

Should we also add "deprecated: true" under profileCompleted in the User yaml file? (src/main/resources/assets/schemas/User.yaml)

connorlbark:
Good point on the swagger; deprecated that field.

Created UI PR (https://github.com/DataBiosphere/duos-ui/pull/1683) so that the frontend does not break. Will merge this once that PR is merged in.

PR #1651, opened 2022-07-14
[DUOS-1891][risk=low] Modify PUT /api/user to send support requests

shaemarks:
Addresses  [DUOS-1891](https://broadworkbench.atlassian.net/browse/DUOS-1891)
Summary of Changes:
- Create method in `SupportRequestService` to handle creating and sending a Zendesk support ticket when `PUT /api/user` endpoint has been provided a `suggestedInstitution` and/or `suggestedSigningOfficial`
       - Do nothing if both are null 
       - Different subject and description based on which user update fields are provided 
- Add tests to make sure the request is being sent in the appropriate scenarios with the correct information 

See #1652 for minor PR swapping out a string for an enum in support ticket creation 

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/SupportRequestService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/SupportRequestService.java
@@ -101,53 +101,22 @@ public class SupportRequestService {
         }
     }
 
+
     /**
-     * Sends a support ticket if the user has requested an unfamiliar institution or signing official
+     * Creates and sends a support ticket for a user requesting an unfamiliar institution and/or signing official, if provided
      * @param userUpdateFields A UserUpdateFields object containing update information for the user
      * @param user The user requesting the institution and/or signing official
      * @param authUser AuthUser object used to build POST request
      */
-    public void sendSuggestedPropertiesToSupport(UserUpdateFields userUpdateFields, User user, AuthUser authUser) {
+    public void handleSuggestedUserFieldsSupportRequest(UserUpdateFields userUpdateFields, User user, AuthUser authUser) {
         if (Objects.nonNull(userUpdateFields) || Objects.nonNull(user)) {
             String suggestedInstitution = userUpdateFields.getSuggestedInstitution();
             String suggestedSigningOfficial = userUpdateFields.getSuggestedSigningOfficial();
 
             // Only create and send ticket if either a suggestedInstitution or suggestedSigningOfficial has been provided
             if (Objects.nonNull(suggestedInstitution) || Objects.nonNull(suggestedSigningOfficial)) {
-                String subject;
-                String description;
-
-                // Determine ticket subject and description
-                if (Objects.nonNull(suggestedInstitution) && Objects.nonNull(suggestedSigningOfficial)) {
-                    subject = "New Institution and Signing Official Request";
-                    description = String.format("User %s [%s] has requested a new signing official: {%s} and has requested a new institution: {%s}",
-                            user.getDisplayName(),
-                            user.getEmail(),
-                            suggestedSigningOfficial,
-                            suggestedInstitution);
-                } else if (Objects.nonNull(suggestedInstitution)) {
-                    subject = "New Institution Request";
-                    description = String.format("User %s [%s] has requested a new institution: {%s}",
-                            user.getDisplayName(),
-                            user.getEmail(),
-                            suggestedInstitution);
-                } else {
-                    subject = "New Signing Official Request";
-                    description = String.format("User %s [%s] has requested a new signing official: {%s}",
-                            user.getDisplayName(),
-                            user.getEmail(),
-                            suggestedSigningOfficial);
-                }
-
-                // Create and post ticket to support api
                 try {
-                    SupportTicket ticket = createSupportTicket(
-                            user.getDisplayName(),
-                            TASK_REQUEST_TYPE,
-                            user.getEmail(),
-                            subject,
-                            description,
-                            configuration.postSupportRequestUrl());
+                    SupportTicket ticket = createSuggestedUserFieldsTicket(userUpdateFields, user);
                     postTicketToSupport(ticket, authUser);
                 } catch (Exception e) {
                     throw new ServerErrorException("Unable to send support ticket for user with email: " + user.getEmail(),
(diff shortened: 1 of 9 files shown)


src/main/java/org/broadinstitute/consent/http/service/SupportRequestService.java
```diff
@@ -32,6 +34,8 @@ public class SupportRequestService {
     private final ServicesConfiguration configuration;
     private final Logger logger = LoggerFactory.getLogger(this.getClass());
 
+    private static final String TASK_REQUEST_TYPE = "task";
```
shaemarks: Would be better to create a `RequestType` enum for the `createSupportTicket` to take as an argument instead of `String type`? This is the Zendesk documentation on type: `The type of the request, "question", "incident", "problem", "task"`. If so, I can create a separate PR for this ticket

src/test/java/org/broadinstitute/consent/http/service/SupportRequestServiceTest.java
```diff
@@ -155,6 +160,158 @@ public void testPostTicketToSupportServerError() throws Exception {
...
+        user.setEmail(email);
+        UserUpdateFields updateFields = new UserUpdateFields();
+        updateFields.setSuggestedInstitution(RandomStringUtils.randomAlphabetic(10));
+
+        mockServerClient.when(request())
+                .respond(response()
+                        .withHeader(Header.header("Content-Type", "application/json"))
+                        .withStatusCode(HttpStatusCodes.STATUS_CODE_OK));
+        service.handleSuggestedUserFieldsSupportRequest(updateFields, user, authUser);
+        mockServerClient.verify(request().withMethod("POST"), VerificationTimes.exactly(1));
```
shaemarks, line 189: This tests that a post request is sent if a `suggestedInstitution` update field is provided. The method that creates the actual ticket (`createSuggestedUserFieldsTicket`) and the method that converts the ticket to json and sends the request (`postTicketToSupport`) are tested separately. Is the current test sufficient, or should I check the exact request body here as well?

src/main/java/org/broadinstitute/consent/http/service/SupportRequestService.java
rushtong: @shaemarks - Yes, that is a better approach.

src/test/java/org/broadinstitute/consent/http/service/SupportRequestServiceTest.java
rushtong: @shaemarks - If we have a separate method that test content, then we don't need to repeat that here.

shaemarks: Ok, sounds good!

src/test/java/org/broadinstitute/consent/http/resources/UserResourceTest.java
```diff
@@ -373,45 +380,66 @@ public void testGetUsersByInstitutionSuccess() {
   }
 
   @Test
-  public void testUpdate() {
+  public void testUpdateSelf() {
```
shaemarks, line 383: Some of the changes shown here are just appearing because I moved `testUpdate` down with the other tests for `userResource.update` instead of by the `userResource.updateSelf` tests. Actual change is modifying tests to make sure `supportRequestService.handleSuggestedUserFieldsSupportRequest` is called the appropriate number of times, and that we get an error status if it fails

quazi-h (approved):
Changes here look good to me, the logic for different tickets based on the provided inputs looks correct.

I was trying to hit the /api/user/id on a local instance to see if I could see anything in the console but didn't see anything relevant to these changes - do I have to modify the activateEmailNotifications flag in consent.yaml to see that?

The only thing I can think of adding is maybe modifying the api documentation to mention that it will generate a zendesk ticket for unfamiliar institutions or signing officials.

src/main/java/org/broadinstitute/consent/http/service/SupportRequestService.java
```diff
@@ -97,4 +102,72 @@ public void postTicketToSupport(SupportTicket ticket, AuthUser authUser) throws
...
+
+            // Only create and send ticket if either a suggestedInstitution or suggestedSigningOfficial has been provided
+            if (Objects.nonNull(suggestedInstitution) || Objects.nonNull(suggestedSigningOfficial)) {
+                try {
+                    SupportTicket ticket = createSuggestedUserFieldsTicket(userUpdateFields, user);
+                    postTicketToSupport(ticket, authUser);
+                } catch (Exception e) {
+                    throw new ServerErrorException("Unable to send support ticket for user with email: " + user.getEmail(),
+                            HttpStatusCodes.STATUS_CODE_SERVER_ERROR);
+                }
```
rushtong: Can we add a log.error statement here as well please? We have some examples in other service classes for this and I think it's important so we can follow up on errors.

rushtong (approved):
See one request inline, otherwise looks good!

shaemarks:
> I was trying to hit the /api/user/id on a local instance to see if I could see anything in the console but didn't see anything relevant to these changes - do I have to modify the activateEmailNotifications flag in consent.yaml to see that?
> 

This actually modifies the `PUT /api/user` endpoint, which allows users to modify some of their own properties (sorry, I forgot to change the description of the jira ticket). We currently don't have a configuration like `activateEmailNotifications` for the support requests; think in order to see that the request sent, you would need some sort of account with Zendesk.

shaemarks:
@rushtong Should we have a configuration for this similar to activateEmailNotifications to avoid sending too many Zendesk tickets while testing?

rushtong:
> @rushtong Should we have a configuration for this similar to activateEmailNotifications to avoid sending too many Zendesk tickets while testing?

@shaemarks - I think that is a fantastic idea, please do so. Set it to default  `false` and only override it as `true` if a config value in `consent.yaml` exists. We can add a follow-on PR to broadinstitute/terra-helmfile for a new config value. However, if we don't have a default and require a value in `consent.yaml`, then the deploy will break due to that config not being in the current chart.

rushtong:
> > @rushtong Should we have a configuration for this similar to activateEmailNotifications to avoid sending too many Zendesk tickets while testing?
> 
> @shaemarks - I think that is a fantastic idea, please do so. Set it to default `false` and only override it as `true` if a config value in `consent.yaml` exists. We can add a follow-on PR to broadinstitute/terra-helmfile for a new config value. However, if we don't have a default and require a value in `consent.yaml`, then the deploy will break due to that config not being in the current chart.

To add onto my reply, that suggestion would mean that this won't be functional on prod until we do have that modified config value in terra-helmfile.

rushtong (approved):
These changes look great, thank you! 👍🏽

PR #1656, opened 2022-07-15
[DUOS-1914][risk=low] Resource/service stub for DarCollectionSummary

rushtong:
## Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DUOS-1914

This PR creates a resource and service stub for future queries on `DarCollectionSummary` objects

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
@@ -24,8 +24,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.ws.rs.BadRequestException;
-import javax.ws.rs.NotAcceptableException;
-import javax.ws.rs.NotAuthorizedException;
 import javax.ws.rs.NotFoundException;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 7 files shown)


src/main/resources/assets/paths/collectionsByRoleNameSummary.yaml
```diff
@@ -0,0 +1,34 @@
+get:
+  summary: Get DAR Collection Summaries By Role Name
+  description: |
+    Returns DAR collection summaries by Role Name
+      * Admin - all DAR collection summaries
+      * DAC Chair/Member - collections requesting datasets associated with their DAC 
+      * Signing Official - collections created by researchers belonging to their institution
+      * Researcher - collections created by the researcher 
+    Canceled DAR collections are not returned
```
JVThomas: If it's a researcher canceled collection then we have to show it since they can revise it as a draft on button click. If the collection was canceled by the DAC, then there's merit in relaying that information to the researcher anyway (assuming it's the collection with the latest elections).

rushtong: 👍🏽 Gotcha. This was almost a complete copy/paste from the current get collections by role name docs

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -64,6 +64,42 @@ public DarCollectionService(DarCollectionDAO darCollectionDAO, DarCollectionServ
...
+      case Resource.SIGNINGOFFICIAL:
+        // TODO: Find by creator institution
+        break;
+      case Resource.CHAIRPERSON:
+      case Resource.MEMBER:
+        List<Integer> dacIds = user.getRoles().stream()
+          .map(UserRole::getDacId)
+          .filter(Objects::nonNull)
+          .collect(Collectors.toList());
+        List<Integer> collectionIds = darCollectionDAO.findDARCollectionIdsByDacIds(dacIds);
```
JVThomas, line 92: I'm guessing this is a stub as well so I'm leaving a note here. The DAO method for DAC users takes in datasetIds. `DatasetDAO.findDatasetsByUserId` handles the DAC associations for you and will return the datasets that belong to the user's dac.

JVThomas (approved):
I've added comments, but they're pretty minor and can be adjusted once the DAO PR is merged in.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rushtong: Yes it is just a stub - we can update this in a separate PR.

rushtong:
Merging this so we can make progress on completing the whole ticket.

PR #1658, opened 2022-07-17
DUOS-1914[risk=no] Added service method for DarCollectionSummary

JVThomas:
Partially addresses [DUOS-1912](https://broadworkbench.atlassian.net/browse/DUOS-1914). 
Linked ticket is [DUOS-1930](https://broadworkbench.atlassian.net/browse/DUOS-1930)

PR adds service methods (public and private) for fetching DAR Collection Summaries by role name. PR also includes extensive test suite and a small update to `ElectionStatus.java`, `DarCollectionSummaryDAO`, and `DarCollectionSummaryReducer`

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
@@ -81,20 +81,20 @@ public class DarCollectionResource extends Resource {
     }
   }
 
-  // @GET
-  // @Path("role/{roleName}/summary")
-  // @Produces("application/json")
-  // @RolesAllowed({ADMIN, CHAIRPERSON, MEMBER, SIGNINGOFFICIAL, RESEARCHER})
-  // public Response getCollectionSummariesForUserByRole(@Auth AuthUser authUser, @PathParam("roleName") String roleName) {
-  //   try {
-  //     User user = userService.findUserByEmail(authUser.getEmail());
-  //     validateUserHasRoleName(user, roleName);
-  //     List<DarCollectionSummary> summaries = darCollectionService.getSummariesForRoleName(user, roleName);
-  //     return Response.ok().entity(summaries).build();
-  //   } catch (Exception e) {
-  //     return createExceptionResponse(e);
-  //   }
-  // }
+  @GET
+  @Path("role/{roleName}/summary")
+  @Produces("application/json")
+  @RolesAllowed({ADMIN, CHAIRPERSON, MEMBER, SIGNINGOFFICIAL, RESEARCHER})
+  public Response getCollectionSummariesForUserByRole(@Auth AuthUser authUser, @PathParam("roleName") String roleName) {
+    try {
+      User user = userService.findUserByEmail(authUser.getEmail());
+      validateUserHasRoleName(user, roleName);
+      List<DarCollectionSummary> summaries = darCollectionService.getSummariesForRoleName(user, roleName);
+      return Response.ok().entity(summaries).build();
+    } catch (Exception e) {
+      return createExceptionResponse(e);
+    }
+  }
 
   @GET
   @Path("{collectionId}")
--- a/src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
@@ -89,7 +89,6 @@ public class DarCollectionResource extends Resource {
     try {
       User user = userService.findUserByEmail(authUser.getEmail());
       validateUserHasRoleName(user, roleName);
-      // TODO: Replace with DarCollectionSummary when available
       List<DarCollectionSummary> summaries = darCollectionService.getSummariesForRoleName(user, roleName);
       return Response.ok().entity(summaries).build();
     } catch (Exception e) {
--- a/src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
@@ -7,6 +7,7 @@ import org.broadinstitute.consent.http.enumeration.DarStatus;
 import org.broadinstitute.consent.http.enumeration.UserRoles;
 import org.broadinstitute.consent.http.models.AuthUser;
 import org.broadinstitute.consent.http.models.DarCollection;
+import org.broadinstitute.consent.http.models.DarCollectionSummary;
 import org.broadinstitute.consent.http.models.DataAccessRequest;
 import org.broadinstitute.consent.http.models.Dataset;
 import org.broadinstitute.consent.http.models.PaginationResponse;
@@ -89,7 +90,7 @@ public class DarCollectionResource extends Resource {
       User user = userService.findUserByEmail(authUser.getEmail());
       validateUserHasRoleName(user, roleName);
       // TODO: Replace with DarCollectionSummary when available
-      List<Object> summaries = darCollectionService.getSummariesForRoleName(user, roleName);
+      List<DarCollectionSummary> summaries = darCollectionService.getSummariesForRoleName(user, roleName);
       return Response.ok().entity(summaries).build();
     } catch (Exception e) {
       return createExceptionResponse(e);
--- a/src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
@@ -81,20 +81,20 @@ public class DarCollectionResource extends Resource {
     }
   }
 
-  @GET
-  @Path("role/{roleName}/summary")
-  @Produces("application/json")
-  @RolesAllowed({ADMIN, CHAIRPERSON, MEMBER, SIGNINGOFFICIAL, RESEARCHER})
-  public Response getCollectionSummariesForUserByRole(@Auth AuthUser authUser, @PathParam("roleName") String roleName) {
-    try {
-      User user = userService.findUserByEmail(authUser.getEmail());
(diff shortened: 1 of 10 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
```diff
@@ -89,7 +90,7 @@ public Response getCollectionSummariesForUserByRole(@Auth AuthUser authUser, @Pa
       User user = userService.findUserByEmail(authUser.getEmail());
       validateUserHasRoleName(user, roleName);
       // TODO: Replace with DarCollectionSummary when available
```
rushtong: nit: remove outdated comment

quazi-h (approved):
Changes look good to me. Tested the new endpoint and responses with different summaries for each role look good. Comments were very helpful towards following the logic. Tests are very thorough.

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
```diff
@@ -80,21 +81,20 @@ public Response getCollectionsForUserByRole(@Auth AuthUser authUser, @PathParam(
...
+  // public Response getCollectionSummariesForUserByRole(@Auth AuthUser authUser, @PathParam("roleName") String roleName) {
+  //   try {
+  //     User user = userService.findUserByEmail(authUser.getEmail());
+  //     validateUserHasRoleName(user, roleName);
+  //     List<DarCollectionSummary> summaries = darCollectionService.getSummariesForRoleName(user, roleName);
+  //     return Response.ok().entity(summaries).build();
+  //   } catch (Exception e) {
+  //     return createExceptionResponse(e);
+  //   }
+  // }
```
rushtong: Is this intentional? This breaks the swagger page since the endpoint is removed.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -75,29 +84,227 @@ public DarCollectionService(DarCollectionDAO darCollectionDAO, DarCollectionServ
    * @param userRole The role the user is making the request as
    * @return List of DarCollectionSummary objects
    */
-  public List<Object> getSummariesForRoleName(User user, String userRole) {
-    switch (userRole) {
-      case Resource.ADMIN:
-        // TODO: Find all
+  private void updateStatusCount(Map<String, Integer> statusCount, String status) {
```
rushtong: nit: the javadoc that precedes this is not associated to the method.

rushtong: Minor style nit. I think most of these private void methods could be in a `DarCollectionSummary` utility class instead of sprinkled here in the service class.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -75,29 +84,227 @@ public DarCollectionService(DarCollectionDAO darCollectionDAO, DarCollectionServ
...
+    statusCount.put(status, count + 1);
+  }
+
+  private void determineCollectionStatus(DarCollectionSummary summary, Map<String, Integer> statusCount, Integer datasetCount, Integer electionCount) {
+    //If there are no elections, status is unreviewed
+    //if there are some elections open, status is in process
+    //if all elections are closed or canceled and electionCount == datasetCount, status is complete
+    if(electionCount == 0) {
+      summary.setStatus(DarCollectionStatus.UNREVIEWED.getValue());
+    } else if(electionCount == datasetCount) {
```
rushtong: minor nit: use `.equals()`

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -75,29 +84,227 @@ public DarCollectionService(DarCollectionDAO darCollectionDAO, DarCollectionServ
...
+      Map<String, Integer> statusCount = new HashMap<>();
+      Map<Integer, Election> elections = s.getElections();
+      if(elections.size() == 0) {
+        s.addAction(DarCollectionActions.OPEN.getValue());
+        s.setStatus(DarCollectionStatus.UNREVIEWED.getValue());
+      } else {
+        elections.values().forEach(e -> {
+          String status = e.getStatus();
+          updateStatusCount(statusCount, status);
+          if(status == ElectionStatus.OPEN.getValue()) {
```
rushtong: This could throw an NPE, try using `Objects.equals()` which handles nulls safely.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -75,29 +84,227 @@ public DarCollectionService(DarCollectionDAO darCollectionDAO, DarCollectionServ
...
+    });
+  }
+
+  private void processDarCollectionDraftsAsSummaries(List<DataAccessRequest> drafts, List<DarCollectionSummary> summaries) {
+    drafts.forEach(d -> {
+      try{
+        DarCollectionSummary summary = new DarCollectionSummary();
+        Date createDate = new Date(d.getData().getCreateDate());
+        String darCode = "DRAFT_DAR_" + new SimpleDateFormat("yyyy-MM-dd")
+          .format(createDate).toString();
```
rushtong: nit: `format()` already returns a string
```suggestion
          .format(createDate);
```

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -75,29 +84,227 @@ public DarCollectionService(DarCollectionDAO darCollectionDAO, DarCollectionServ
...
+    summaries.forEach(s -> {
+      Map<Integer, Election> elections = s.getElections();
+      int electionCount = elections.values().size();
+      elections.values().forEach(election -> updateStatusCount(statusCount, election.getStatus()));
+      s.addAction(DarCollectionActions.REVIEW.getValue());
+      //check dar statuses, if they're all canceled show revise (but only if there are no elections)
+      Boolean isCanceled = false;
+      if(electionCount == 0) {
+        Collection<String> darStatuses = s.getDarStatuses().values();
+        isCanceled = darStatuses.size() > 0 && darStatuses.stream()
```
rushtong: `isCanceled` is only used inside the if block so you can instantiate it directly there instead of initializing it here. That way you can avoid an unnecessary variable if `electionCount > 0`
```suggestion
      if(electionCount == 0) {
        Collection<String> darStatuses = s.getDarStatuses().values();
        boolean isCanceled = darStatuses.size() > 0 && darStatuses.stream()
```

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -75,29 +84,227 @@ public DarCollectionService(DarCollectionDAO darCollectionDAO, DarCollectionServ
...
+            .stream()
+            .anyMatch(election -> election.getStatus().equalsIgnoreCase(ElectionStatus.OPEN.getValue()));
+          
+          if(isVotable) {
+            s.setStatus(DarCollectionStatus.IN_PROCESS.getValue());
+            s.addAction(DarCollectionActions.VOTE.getValue());
+          } else {
+            //non-votable states
+              //all canceled (complete)
+              //some datasets do not have elections (in process) (legacy)
```
rushtong: nit: This case can happen for both legacy and new DARs so I don't think we need to limit this (comment-wise) to legacy DARs. For new DARs, this can happen when it is across two DACs, one dac having created elections and the other DAC not having created elections.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -75,29 +84,227 @@ public DarCollectionService(DarCollectionDAO darCollectionDAO, DarCollectionServ
    * @param userRole The role the user is making the request as
    * @return List of DarCollectionSummary objects
    */
-  public List<Object> getSummariesForRoleName(User user, String userRole) {
-    switch (userRole) {
-      case Resource.ADMIN:
-        // TODO: Find all
+  private void updateStatusCount(Map<String, Integer> statusCount, String status) {
+    Integer count = statusCount.get(status);
```
rushtong, line 81: This can throw an NPE if the key is null. Try `containsKey()` instead

rushtong (changes requested):
The overall logic looks correct to me, thank you! See comments inline.

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
JVThomas: Yeah, I didn't want to enable an endpoint with no tests in a PR. I re-enable it in https://github.com/DataBiosphere/consent/pull/1660 since it adds the tests in. Forgot about the Swagger page though.

The PR I linked is branched off of this one, so if that gets approval I can merge that into this branch.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
JVThomas: HashMap supports nulls as keys, so that isn't an issue

rushtong: True, but `Map` doesn't support null keys, so this method is relying on callers to pass in the correct implementation.

JVThomas: Ah, which means that I'll have to check on null to keep Map methods from erroring out.

rushtong (approved):
Thank you for the changes, they look good! 👍🏽

PR #1669, opened 2022-07-26
[DUOS-1959][risk=low] endpoint to fetch collection summaries by role name and collection id 

shaemarks:
Addresses [DUOS-1959](https://broadworkbench.atlassian.net/browse/DUOS-1959)
- Adds endpoint for getting a particular DAR Collection summary by collection id and role name (used to determine actions and if access is allowed)
    - admins can see any collection
    - DACs can only see collections with relevant datasets
    - signing officials can only see collections created by researchers from their institution 
    - researchers can only see collections they created 

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -139,4 +139,75 @@ List<DarCollectionSummary> getDarCollectionSummariesForSO(
...
+
+
+  @RegisterBeanMapper(value = DarCollectionSummary.class)
+  @RegisterBeanMapper(value = DarCollection.class)
+  @RegisterBeanMapper(value = Vote.class, prefix = "v")
+  @RegisterBeanMapper(value = Election.class)
+  @UseRowReducer(DarCollectionSummaryReducer.class)
+  @SqlQuery
+          (
+                  "SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, u.display_name as researcher_name, " +
```
shaemarks: I'm having problems fixing the formatting here. I tried changing the settings for SQL in IntelliJ but haven't been able to fix it. Do we have any standard formatting settings I can follow?

rushtong: I have found intellij's sql formatting very unhelpful. I would suggest manually formatting it similarly to the other queries in this class.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -139,4 +139,77 @@ List<DarCollectionSummary> getDarCollectionSummariesForSO(
...
+    "WHERE c.collection_id= :collectionId " +
+      "AND dd.dataset_id IN (<datasetIds>) " +
+      "AND (e.latest = e.electionid OR e.electionid IS NULL) " +
+      "AND (LOWER(v.type) = 'final' OR (v.dacuserid = :currentUserId OR v.voteid IS NULL)) " +
+      "AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL )"
+  )
+  List<DarCollectionSummary> getDarCollectionSummaryForDACByCollectionId(
+          @Bind("currentUserId") Integer currentUserId,
+          @BindList("datasetIds") List<Integer> datasetIds,
+          @Bind("collectionId") Integer collectionId);
```
rushtong: Question, why are we returning a list instead of a single summary? We know that we should have a single result from the collection id.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -139,4 +139,77 @@ List<DarCollectionSummary> getDarCollectionSummariesForSO(
...
+    ") AS e " +
+      "ON e.referenceid = dar.reference_id " +
+    "INNER JOIN dar_dataset dd " +
+      "ON dar.reference_id = dd.reference_id " +
+    "WHERE c.collection_id = :collectionId " +
+      "AND (e.latest = e.electionid OR e.electionid IS NULL) " +
+      "AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL )"
+  )
+  List<DarCollectionSummary> getDarCollectionSummaryByCollectionId(
+          @Bind("collectionId") Integer collectionId);
```
rushtong: Same question here, I think we should be returning a single summary.

rushtong (changes requested):
Code looks good, works as intended, see comments inline.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
shaemarks: Since we're re-using the collection summary processing methods (processDarCollectionSummariesForAdmin, etc) which take in a list of DarCollectionSummaries, we'd either need to wrap the single summary with List.of() before these calls or do it in the query. If the query doesn't find a summary with the collectionId, we'd need to check that the summary isn't null before wrapping it in List.of() (whereas right now an empty list would be returned by the query and the process methods wouldn't do anything). Would it be better to have the queries return a single summary instead of a list and then use Objects.requireNonNull() or a try-catch?

shaemarks:
Side note: I think line 154 in `DarCollectionService.processDarCollectionSummariesForResearcher`  ( `Map<String, Integer> statusCount = new HashMap<>();` ) should be located within the forEach block  so that it is reset for each summary

JVThomas:
> Side note: I think line 154 in `DarCollectionService.processDarCollectionSummariesForResearcher` ( `Map<String, Integer> statusCount = new HashMap<>();` ) should be located within the forEach block so that it is reset for each summary

Yeah that's right. Good catch!

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
rushtong: Outside of this method, I don't have a preference for what we should do at the moment - it's important to look at them in isolation. When I'm looking at a method, it has to stand on its own and make sense within its own context. `getDarCollectionSummaryForDACByCollectionId` says to me, in plain English, "Find me a single summary from a collection id". If the method does something different than that, my alarm bells go off.

JVThomas: Another option is to redefine the `processDarCollectionSummariesFor...` functions so that they take only a single summary rather than a list of summaries. You would have to iterate through the summary list outside of those process function calls (for the list return, role-based queries), but it'll make them easier to use for these id based queries.

shaemarks: Gotcha, makes sense!

JVThomas (approved):
Functionality is working as intended, I can see the summary format change for a collection as I switch roles. Tests cover positive and negative test cases for queries.

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
```diff
@@ -96,6 +96,47 @@ public Response getCollectionSummariesForUserByRole(@Auth AuthUser authUser, @Pa
...
+          break;
+        default:
+          throw new BadRequestException("Invalid role selection: " + roleName);
+      }
+      if (!allowedAccess) {
+        // user has role but is not allowed to view collection; throw NotFoundException to avoid leaking existence
+        throw new NotFoundException("Collection with the collection id of " + collectionId + " was not found");
+      }
+
+      DarCollectionSummary summary = darCollectionService.getSummaryForRoleNameByCollectionId(user, roleName, collectionId);
```
rushtong: Optional idea here. We're querying for the collection twice, here and on line 108. If there were methods for checking `allowedAccess`  on a summary instead of a collection, you could reduce the work required to complete this task.

rushtong (approved):
Looks good, thank you 👍🏽

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
shaemarks: We can do that. It would involve adding the `institutionId` and the `userId` of the researcher to the summary queries and then copying the logic in existing helper methods (`checkDacPermissionsForCollection`) to determine `allowedAccess`. Would it be better to include the institutionId and userId in the previous summary queries so that these properties aren't null, or to leave them alone since it's not necessary for the consoles?

src/main/java/org/broadinstitute/consent/http/models/DarCollectionSummary.java
```diff
@@ -43,6 +43,8 @@ public class DarCollectionSummary {
   @JsonProperty
   private int datasetCount;
 
+  private Integer researcherId;
```
shaemarks, line 48: I used `researcherId` to match `researcherName`, but I can alternatively use `createUserId` since that's normally what it is referred to as

rushtong (approved):
Still 👍🏽, sorry for the delay on this.

JVThomas (approved):

PR #1670, opened 2022-07-26
[DUOS-1905][risk=no] Send reference ids in DAR Collection Summary

connorlbark:
## Addresses 

https://broadworkbench.atlassian.net/browse/DUOS-1905

Necessary addition for the researcher console page.

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/models/DarCollectionSummary.java
```diff
@@ -19,6 +19,9 @@ public class DarCollectionSummary {
   @JsonProperty
   private Integer darCollectionId;
 
+  @JsonProperty
+  private Set<String> referenceIds;
```
rushtong, line 23: Optional: I generally prefer `List` over `Set` simply because they're easier to work with. Certainly not a blocker if you feel that `Set` is a better fit here.

connorlbark:
Thought: Do we want to always return the reference ids, or only when the dar collection is a draft?

rushtong:
> Thought: Do we want to always return the reference ids, or only when the dar collection is a draft?

In general, I prefer not to conditionally populate properties. If they exist for both types of DARs (and they do), then they should always be there regardless of type. This is something the system does now in some of the older cases (i.e. see DatasetDTO and all of the manage endpoints) and has been problematic because different endpoints ended up returning different properties and to get all of them, the UI needed to call multiple endpoints to get the full picture of the object.

connorlbark:
> > Thought: Do we want to always return the reference ids, or only when the dar collection is a draft?
> 
> In general, I prefer not to conditionally populate properties. If they exist for both types of DARs (and they do), then they should always be there regardless of type. This is something the system does now in some of the older cases (i.e. see DatasetDTO and all of the manage endpoints) and has been problematic because different endpoints ended up returning different properties and to get all of them, the UI needed to call multiple endpoints to get the full picture of the object.

Makes perfect sense to me! That is the current behavior of this PR, so it should be good to go.

src/main/java/org/broadinstitute/consent/http/models/DarCollectionSummary.java
JVThomas: I think, given the context of the query, the Set makes sense. The joins on elections and votes will have referenceId values repeated on multiple rows. With a Set the `DARCollectionSummaryReducer` can add the value and not worry about having duplicates.

JVThomas (approved):
Changes look good, I can see the array of referenceIds on the endpoint response for collections and drafts.

rushtong (approved):
👍🏽 Works as expected!

PR #1678, opened 2022-08-01
[DUOS-1971][risk=no] Use the V2 Algorithm for DAR match generation

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1971

This PR:

- Adds a version to the match table to denote which algorithm we used to store the results
- Updates DAR/Dataset matching logic to query the v2 endpoint and store the results with the v2 version
- Significant match service class test refactoring. The mocking was excessive and unnecessary.
- DOES NOT make any changes to consent/dataset matching. Those code paths are completely deprecated at present and do not need to be touched until we start removing the relevant code altogether.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/MatchService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/MatchService.java
@@ -151,7 +151,7 @@ public class MatchService {
                 } catch (Exception e) {
                     String message = "Error finding single match for purpose: " + dar.getReferenceId();
                     logger.error(message);
-                    matches.add(createMatch(dataset.getDatasetIdentifier(), dar.getReferenceId(), true, false, MatchAlgorithm.V2, List.of(message)));
+                    matches.add(new Match(dataset.getDatasetIdentifier(), dar.getReferenceId(), true, false, MatchAlgorithm.V2, List.of(message)));
                 }
             }
         });
@@ -171,7 +171,7 @@ public class MatchService {
                     matches.add(match);
                 } catch (Exception e) {
                     logger.error("Error finding  matches for consent: " + consentId);
-                    matches.add(createMatch(consentId, dar.getReferenceId(), true, false, MatchAlgorithm.V1, List.of()));
+                    matches.add(new Match(consentId, dar.getReferenceId(), true, false, MatchAlgorithm.V1, List.of()));
                 }
             }
         }
@@ -193,9 +193,9 @@ public class MatchService {
         Response res = matchServiceTargetV1.request(MediaType.APPLICATION_JSON).post(Entity.json(json));
         if (res.getStatus() == Response.Status.OK.getStatusCode()) {
             ResponseMatchingObject entity = res.readEntity(rmo);
-            match = createMatch(consent.getConsentId(), dar.getReferenceId(), false, entity.isResult(), MatchAlgorithm.V1, List.of());
+            match = new Match(consent.getConsentId(), dar.getReferenceId(), false, entity.isResult(), MatchAlgorithm.V1, List.of());
         } else {
-            match = createMatch(consent.getConsentId(), dar.getReferenceId(), true, false, MatchAlgorithm.V1, List.of());
+            match = new Match(consent.getConsentId(), dar.getReferenceId(), true, false, MatchAlgorithm.V1, List.of());
         }
         return match;
     }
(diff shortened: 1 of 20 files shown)


src/main/java/org/broadinstitute/consent/http/service/MatchService.java
```diff
@@ -219,23 +190,48 @@ private Match singleEntitiesMatch(Consent consent, DataAccessRequest dar) {
...
+            GenericType<DataUseResponseMatchingObject> durmo = new GenericType<>(){};
+            DataUseResponseMatchingObject entity = res.readEntity(durmo);
+            match = createMatch(dataset.getDatasetIdentifier(), dar.getReferenceId(), false, entity.isResult(), MatchAlgorithm.V2, entity.getFailureReasons());
+        } else {
+            match = createMatch(dataset.getDatasetIdentifier(), dar.getReferenceId(), true, false, MatchAlgorithm.V2, List.of());
+        }
+        return match;
+    }
+
+    private Match createMatch(String consentId, String purposeId, boolean failed, boolean isMatch, MatchAlgorithm algorithm, List<String> failureReasons) {
```
connorlbark: should this be a constructor

src/main/java/org/broadinstitute/consent/http/service/MatchService.java
```diff
@@ -178,36 +161,24 @@ protected List<Match> createMatchesForDataAccessRequest(DataAccessRequest dar) {
...
         Consent consent = findConsent(consentId);
-        List<Dataset> dataSets = dataSetDAO.getDatasetsForConsent(consentId);
+        List<Dataset> dataSets = datasetDAO.getDatasetsForConsent(consentId);
         List<DataAccessRequest> dars = findRelatedDars(dataSets.stream().map(Dataset::getDataSetId).collect(Collectors.toList()));
         if (consent != null && !dars.isEmpty()) {
             Match match;
             for (DataAccessRequest dar : dars) {
                 try {
-                    match = singleEntitiesMatch(consent, dar);
+                    match = singleEntitiesMatchV1(consent, dar);
```
connorlbark, line 170: out of curiosity - why are we doing V1 matching here and V2 elsewhere? is this a deprecated code path?

src/main/resources/changesets/changelog-consent-14.0.xml
```diff
@@ -28,7 +28,7 @@
     <changeSet author="vvicario" id="58">
         <validCheckSum>ANY</validCheckSum>
         <addForeignKeyConstraint baseColumnNames="consent" baseTableName="matchentity"
-                                 constraintName="fkMatchConsent" deferrable="false" initiallyDeferred="false"
+                                 constraintName="fkmatchconsent" deferrable="false" initiallyDeferred="false"
```
connorlbark, line 31: is changing the old changeset intentional? i'm not sure how these work - would this be updated in prod, too?

connorlbark (commented):
having weird issues with my consent setup (in general) that are making it hard to test - if i figure this out i'll come back to the ticket and test functionality, but without testing this looks good

src/main/java/org/broadinstitute/consent/http/service/MatchService.java
rushtong: It certainly could. Will revisit that.

src/main/resources/changesets/changelog-consent-14.0.xml
rushtong: Ah, good eye. Yes, this is intentional. Originally, in mysql, this was camel cased. When we migrated to postgres, this constraint was all lower cased. The changelog didn't need to be updated since the file had already been run and didn't need to be re-run. However, when we run tests, all changelog files ARE re-run and we hit a naming error because `fkMatchConsent` doesn't match `fkmatchconsent`. The current constraint name is actually lower cased, so this change makes it consistent with what's real.

src/main/java/org/broadinstitute/consent/http/service/MatchService.java
rushtong: In this PR, I am more focused in processing matches for Data Access Requests than for Consents (or datasets). This method is only called when we are creating a new consent from the `POST api/consent` endpoint. To minimize the impact of the overall scope of the PR, I left this in place. That endpoint is no longer used from the UI. It used to be used from ORSP as well, but I'm pretty sure we disabled that so this is essentially a dead code path and on the chopping block.

JVThomas:
Can I ask about the switch from the uuid format to the DUOS formatted dataset id for the `consent` column? (These are the updates that occurred when testing locally, viewed on Postico, for `match-entity`)

Screen Shot 2022-08-11 at 11 18 48 AM

rushtong:
> Can I ask about the switch from the uuid format to the DUOS formatted dataset id for the `consent` column? (These are the updates that occurred when testing locally, viewed on Postico, for `match-entity`)

Datasets don't have a UUID, so we have to store something relatively unique and dataset-centric instead. This was the best choice I could think of at the moment without adding a new UUID column to datasets.

rushtong:
> Can I ask about the switch from the uuid format to the DUOS formatted dataset id for the `consent` column? (These are the updates that occurred when testing locally, viewed on Postico, for `match-entity`)
> 
> Screen Shot 2022-08-11 at 11 18 48 AM

I'm more concerned about the values for V2 `failed` entries. That means the ontology endpoint wasn't able to process the provided JSON.

JVThomas:
> > Can I ask about the switch from the uuid format to the DUOS formatted dataset id for the `consent` column? (These are the updates that occurred when testing locally, viewed on Postico, for `match-entity`)
> > Screen Shot 2022-08-11 at 11 18 48 AM
> 
> I'm more concerned about the values for V2 `failed` entries. That means the ontology endpoint wasn't able to process the provided JSON.

I think the failure here makes sense. I chose 'Other' on the DAR Form to see if the `match_failure_reason` inserts worked as expected (which they did)
Screen Shot 2022-08-11 at 1 00 34 PM
.

rushtong:
> I think the failure here makes sense. I chose 'Other' on the DAR Form to see if the `match_failure_reason` inserts worked as expected (which they did)

There was a legit bug here in the deserialization of ontology's response. I think I introduced it after testing functionality and didn't re-test well enough. The `match` column stores whether there was a match or not. The `failed` column stores whether consent could make a successful call to ontology which in effect tells us that there was some unknown error when making the match.

JVThomas:
> > I think the failure here makes sense. I chose 'Other' on the DAR Form to see if the `match_failure_reason` inserts worked as expected (which they did)
> 
> There was a legit bug here in the deserialization of ontology's response. I think I introduced it after testing functionality and didn't re-test well enough. The `match` column stores whether there was a match or not. The `failed` column stores whether consent could make a successful call to ontology which in effect tells us that there was some unknown error when making the match.

Ah that explains the choice of words in for the error message contents. The corrected version is a bit more specific.

Screen Shot 2022-08-11 at 1 51 33 PM

That said, for this data access request I clicked on `Other`, so the messages are a little confusing, but that's outside the purview of this ticket. With the recent update the PR is saving the match_entity `failed` status correctly and inserting failures correctly into `match_failure_reason`

JVThomas (approved):

connorlbark (approved):
Still having weird issues with my local setup but I trust Justin's review that this works - everything looks solid.

PR #1681, opened 2022-08-02
[DUOS-1921][risk=no] Refactor to remove usages of DataAccessRequestData.getDarCode

shaemarks:
Addresses [DUOS-1921](https://broadworkbench.atlassian.net/browse/DUOS-1921)
----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/service/EmailNotifierService.java
```diff
@@ -28,6 +28,7 @@
 import org.broadinstitute.consent.http.models.UserProperty;
 import org.broadinstitute.consent.http.models.Vote;
 import org.broadinstitute.consent.http.models.dto.DatasetMailDTO;
+import org.jdbi.v3.sqlobject.customizer.Bind;
```
connorlbark: accidental import?

connorlbark (approved):
this looked good to me - the individual changes all looked good - but do we also want to deprecate/remove `getDarCode` on the DAR data, so that we don't run into this issue again?

src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
```diff
@@ -241,7 +246,7 @@ private void validateUserAndVoteId(AuthUser authUser, Integer voteId) throws Not
                 existingVote.getDacUserId());
     }
 
-    private void createDataOwnerElection(Vote vote, DataAccessRequest dar) throws MessagingException, IOException, TemplateException {
+    private void createDataOwnerElection(Vote vote, String referenceId) throws MessagingException, IOException, TemplateException {
```
rushtong, line 248: Thank you for updating this. We don't actually create data owner elections, but the code is now compatible. It really should be in a service class, but we'll be removing this in the near future in favor data submitter functionality.

src/main/java/org/broadinstitute/consent/http/resources/DataRequestVoteResource.java
```diff
@@ -253,18 +258,20 @@ private void createDataOwnerElection(Vote vote, DataAccessRequest dar) throws Me
             finalVote = CollectionUtils.isNotEmpty(finalVotes) ? finalVotes.get(0) : null;
             agreementVote = vote;
         }
-        if((finalVote != null && finalVote.getVote() != null && finalVote.getVote()) && (agreementVote == null || (agreementVote != null && agreementVote.getVote() != null))){
-            List<Dataset> needsApprovedDataSets = datasetService.findNeedsApprovalDataSetByObjectId(dar.getDatasetIds());
+        if((finalVote != null && finalVote.getVote() != null && finalVote.getVote()) && (agreementVote == null || agreementVote.getVote() != null)){
+            DarCollection collection = darCollectionService.getByReferenceId(referenceId);
+            List<Integer> datasetIds = Objects.nonNull(collection) ? collection.getDars().get(referenceId).getDatasetIds() : List.of();
```
rushtong, line 262: This should work, we're pulling the dataset ids in the `findDARCollectionByReferenceId` dao call

src/main/java/org/broadinstitute/consent/http/service/ElectionService.java
```diff
@@ -94,17 +98,13 @@ public List<Election> describeClosedElectionsByType(String type, AuthUser authUs
...
-            List<String> referenceIds = elections.stream().map(Election::getReferenceId).collect(Collectors.toList());
-            List<DataAccessRequest> dataAccessRequests = dataAccessRequestService.
-                    getDataAccessRequestsByReferenceIds(referenceIds);
             elections.forEach(election -> {
-                Optional<DataAccessRequest> darOption = dataAccessRequests.stream().
-                        filter(d -> d.getReferenceId().equals(election.getReferenceId())).
-                        findFirst();
-                darOption.ifPresent(dar -> {
-                    election.setDisplayId(dar.getData().getDarCode());
+                DarCollection collection = darCollectionDAO.findDARCollectionByReferenceId(election.getReferenceId());
```
rushtong: This is a DAO call in a loop and will be pretty expensive. Instead, we should query for all of the collections before iterating over the elections and then pull the right collection from the list while inside the loop.

src/main/java/org/broadinstitute/consent/http/service/EmailNotifierService.java
```diff
@@ -207,9 +208,9 @@ public void sendClosedDataSetElectionsMessage(List<Election> elections) throws M
             Map<String, List<Election>> reviewedDatasets = new HashMap<>();
             for(Election election: elections) {
                 List<Election> dsElections = electionDAO.findLastElectionsByReferenceIdAndType(election.getReferenceId(), ElectionType.DATA_SET.getValue());
-                DataAccessRequest dar = dataAccessRequestDAO.findByReferenceId(election.getReferenceId());
-                String dar_code = Objects.nonNull(dar) && Objects.nonNull(dar.getData()) ? dar.getData().getDarCode() : "";
-                reviewedDatasets.put(dar_code, dsElections);
+                DarCollection collection = collectionDAO.findDARCollectionByReferenceId(election.getReferenceId());
```
rushtong: This is another example of a dao call in a loop, but that's what we were doing before. Consider this optional. It would be nice to pull that call outside of the loop (as a list of collections) and then find the right one for the election being iterated over.

src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
```diff
@@ -223,9 +232,10 @@ public DatasetMetrics generateDatasetMetrics(Integer datasetId) {
     //find dars with the given datasetId in their list of datasetIds, datasetId is a String so it can be converted to jsonb in query
     //convert all dars into smaller objects that only contain the information needed
     List<DataAccessRequest> dars = darDAO.findAllDataAccessRequestsByDatasetId(Integer.toString(datasetId));
-    List<DarMetricsSummary> darInfo = dars.stream().map(dar ->
-      new DarMetricsSummary(dar))
-      .collect(Collectors.toList());
+    List<DarMetricsSummary> darInfo = dars.stream().map(dar -> {
+              DarCollection collection = darCollectionDAO.findDARCollectionByCollectionId(dar.getCollectionId());
```
rushtong: Same dao-loop feedback here. Consider this optional as well - it's a very under-used code path

rushtong (commented):
See feedback inline, thank you, this is a lot of work!

shaemarks:
> do we also want to deprecate/remove `getDarCode` on the DAR data, so that we don't run into this issue again?

Good point!

src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
rushtong: Here's another way to do what you're trying to do here with a single dao call as opposed to N dao calls:
```
    List<Integer> darCollectionIds = dars.stream().map(DataAccessRequest::getCollectionId).collect(Collectors.toList());
    List<DarCollection> darCollections = darCollectionDAO.findDARCollectionByCollectionIds(darCollectionIds);
    Map<Integer, DarCollection> collectionMap = darCollections.stream().collect(Collectors.toMap(DarCollection::getDarCollectionId, Function.identity()));
    List<DarMetricsSummary> darInfo = dars.stream().map(dar -> {
              DarCollection collection = collectionMap.get(dar.getCollectionId());
              return new DarMetricsSummary(dar, collection.getDarCode());
            }).collect(Collectors.toList());
```

shaemarks: Thanks for this suggestion! It has been helpful!

src/main/java/org/broadinstitute/consent/http/service/EmailNotifierService.java
```diff
@@ -205,11 +199,15 @@ public void sendDisabledDatasetsMessage(User user, List<String> disabledDatasets
     public void sendClosedDataSetElectionsMessage(List<Election> elections) throws MessagingException, IOException, TemplateException {
         if(isServiceActive){
             Map<String, List<Election>> reviewedDatasets = new HashMap<>();
+            List<String> referenceIds = elections.stream().map(Election::getReferenceId).collect(Collectors.toList());
+            List<DarCollection> darCollections = collectionDAO.findDARCollectionsByReferenceIds(referenceIds);
```
rushtong: I know this was my suggestion, but it's safer to check if you have an empty list before passing it into a findBy method. Something like this:
```suggestion
            List<DarCollection> darCollections = referenceIds.isEmpty() ? List.of() : collectionDAO.findDARCollectionsByReferenceIds(referenceIds);
```

src/main/java/org/broadinstitute/consent/http/service/EmailNotifierService.java
```diff
@@ -205,11 +199,15 @@ public void sendDisabledDatasetsMessage(User user, List<String> disabledDatasets
...
             for(Election election: elections) {
                 List<Election> dsElections = electionDAO.findLastElectionsByReferenceIdAndType(election.getReferenceId(), ElectionType.DATA_SET.getValue());
-                DataAccessRequest dar = dataAccessRequestDAO.findByReferenceId(election.getReferenceId());
```
rushtong, line 199: It looks like `dataAccessRequestDAO` and `userPropertyDAO` are unused in this class, would be nice to remove them.

rushtong (approved):
👍🏽

connorlbark (approved):
👍

PR #1682, opened 2022-08-02
[DUOS-1961][risk=no] Refactor /resubmit endpoint

connorlbark:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1961

Resubmit now returns a `DarCollectionSummary`. Once put back into a draft status, the draft DAR and the dar collection both exist, and would show up on the table; thus, I added a filter to the DAO call to filter out all DarCollections which have only draft DARs for the researcher console.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
@@ -141,7 +141,7 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
     "WHERE c.create_user_id = :userId " +
     	"AND (e.latest = e.electionid OR e.electionid IS NULL) " +
         "AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL ) " +
-        "AND (EXISTS (SELECT * FROM data_access_request WHERE (collection_id = c.collection_id and draft = false)))"
+        "AND ("+DarCollectionSummary.FILTER_DRAFT_DAR_COLLECTIONS+")"
   )
   List<DarCollectionSummary> getDarCollectionSummariesForResearcher(
       @Bind("userId") Integer userId);
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
@@ -22,8 +22,8 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
   @UseRowReducer(DarCollectionSummaryReducer.class)
   @SqlQuery
   (
-    "SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, u.display_name as researcher_name, " +
-      "i.institution_name, e.electionid, e.status, e.datasetid, e.referenceid, v.voteid as v_vote_id, dd.dataset_id as dd_datasetid, " +
+    "SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, u.display_name as researcher_name, " +
+      "i.institution_name, e.electionid, e.status, e.datasetid, e.referenceid, v.voteid as v_vote_id, dd.dataset_id as dd_datasetid, " +      
       "v.dacuserid as v_dac_user_id, v.vote as v_vote, v.electionid as v_election_id, v.createdate as v_create_date, v.updatedate as v_update_date, v.type as v_type, " +
       "(dar.data #>> '{}')::jsonb ->> 'projectTitle' AS name " +
     "FROM dar_collection c " +
@@ -58,8 +58,8 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
   @UseRowReducer(DarCollectionSummaryReducer.class)
   @SqlQuery
   (
-    "SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, u.display_name as researcher_name, " +
-      "i.institution_name, e.electionid, e.status, e.datasetid, e.referenceid, dd.dataset_id as dd_datasetid, " +
+    "SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, u.display_name as researcher_name, " +
+      "i.institution_name, e.electionid, e.status, e.datasetid, e.referenceid, dd.dataset_id as dd_datasetid, " +      
       "(dar.data #>> '{}')::jsonb ->> 'projectTitle' AS name " +
     "FROM dar_collection c " +
     "INNER JOIN users u " +
@@ -89,8 +89,9 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
   @UseRowReducer(DarCollectionSummaryReducer.class)
   @SqlQuery
   (
-    "SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, u.display_name as researcher_name, " +
-      "i.institution_name, e.electionid, e.status, e.datasetid, e.referenceid, dd.dataset_id as dd_datasetid, " +
+    "SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, u.display_name as researcher_name, " +
+      "i.institution_name, e.electionid, e.status, e.datasetid, e.referenceid, dd.dataset_id as dd_datasetid, " +      
+
       "(dar.data #>> '{}')::jsonb ->> 'projectTitle' AS name " +
     "FROM dar_collection c " +
     "INNER JOIN users u " +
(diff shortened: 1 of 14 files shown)


src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -140,7 +140,8 @@ List<DarCollectionSummary> getDarCollectionSummariesForSO(
       "ON dar.reference_id = dd.reference_id " +
     "WHERE c.create_user_id = :userId " +
     	"AND (e.latest = e.electionid OR e.electionid IS NULL) " +
-        "AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL ) "
+        "AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL ) " +
+        "AND (EXISTS (SELECT * FROM data_access_request WHERE (collection_id = c.collection_id and draft = false)))"
```
rushtong: Super minor optimization suggestion ... the select here could be a lot smaller, you don't need all fields. I _think_ this would work:
```suggestion
        "AND (EXISTS (SELECT 1 FROM data_access_request WHERE (collection_id = c.collection_id and draft = false)))"
```

rushtong: Also, any reason to not use the `DarCollectionSummary.FILTER_DRAFT_DAR_COLLECTIONS` string here?

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -131,23 +133,35 @@ private void processDarCollectionSummariesForAdmin(List<DarCollectionSummary> su
...
-        summary.setStatus(DarCollectionStatus.DRAFT.getValue());
-        summary.setName(d.getData().getProjectTitle());
-        summary.addAction(DarCollectionActions.RESUME.getValue());
-        summary.addAction(DarCollectionActions.DELETE.getValue());
-        summary.addReferenceId(d.referenceId);
-        summaries.add(summary);
+        summaries.add(processDraftAsSummary(d));
       } catch(Exception e) {
         logger.warn("Error processing draft with id: " + d.getId());
+        logger.warn(e.getMessage());
```
rushtong: We should combine the messages into a single log statement - makes it easier to filter in google logs explorer

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -131,23 +133,35 @@ private void processDarCollectionSummariesForAdmin(List<DarCollectionSummary> su
...
+    Date createDate = new Date(d.getData().getCreateDate());
+    String darCode;
+    if (Objects.nonNull(d.getData().getDarCode())) {
+      darCode = d.getData().getDarCode();
+    } else {
+      darCode = "DRAFT_DAR_" + new SimpleDateFormat("yyyy-MM-dd")
+              .format(createDate);
+    }
+
+    summary.setDarCode(darCode);
```
rushtong: See https://github.com/DataBiosphere/consent/pull/1681 where Shae is refactoring the use of `getData().getDarCode()`. I think it would be nice not to re-introduce what she's removing.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -311,6 +325,33 @@ public List<DarCollectionSummary> getSummariesForRoleName(User user, String user
     return summaries;
   }
 
+  public DarCollectionSummary draftDarCollection(DarCollection sourceCollection) {
```
rushtong: nit: naming. I think this should more completely convey what it is doing. It's moving a collection from non-draft to draft, so something like `updateCollectionToDraft` or something similar.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -311,6 +325,33 @@ public List<DarCollectionSummary> getSummariesForRoleName(User user, String user
...
+    this.dataAccessRequestDAO.updateDraftByCollectionId(sourceCollection.getDarCollectionId(), true);
+    sourceCollection.getDars().values().forEach((d) -> {
+      Date now = new Date();
+      DataAccessRequestData newData = new Gson().fromJson(d.getData().toString(), DataAccessRequestData.class);
+      newData.setDarCode(null);
```
rushtong, line 327: I'm not sure we need to do this as this value is being deprecated.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -311,6 +325,33 @@ public List<DarCollectionSummary> getSummariesForRoleName(User user, String user
...
+      newData.setStatus(null);
+      newData.setReferenceId(d.getReferenceId());
+      newData.setCreateDate(now.getTime());
```
rushtong: I don't think we should reset the create date - that should remain what it was from its original submission.

rushtong (changes requested):
Have not functionally tested this yet, but there are some areas that I think we can improve upon. See feedback inline.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
connorlbark: ahh right - I had done this as in the UI it only allows you to resubmit a draft if `isNil(darCode)` so without setting it to null it broke the UI. Should I remove this for now, even though it breaks the UI?

rushtong: Let's leave it in if it breaks the UI. I think the logic of `isNil(darCode)` is no longer valid. In this world, a DAR with a DarCollection (and therefore a dar code) is valid for resubmission. The condition for resubmission is that there are no elections created for it.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
connorlbark: oops! just forgot to good catch

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -140,7 +140,8 @@ List<DarCollectionSummary> getDarCollectionSummariesForSO(
       "ON dar.reference_id = dd.reference_id " +
     "WHERE c.create_user_id = :userId " +
     	"AND (e.latest = e.electionid OR e.electionid IS NULL) " +
-        "AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL ) "
+        "AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL ) " +
+        "AND ("+DarCollectionSummary.FILTER_DRAFT_DAR_COLLECTIONS+")"
```
lu-c: I'm kind of feeling like I would rather see the actual string here, just because the rest of the query is here too. Is the `FILTER_DRAFT_DAR_COLLECTION` variable being referenced anywhere else?

src/main/java/org/broadinstitute/consent/http/models/DarCollectionSummary.java
```diff
@@ -16,6 +16,9 @@
 
 public class DarCollectionSummary {
 
+  public static final String FILTER_DRAFT_DAR_COLLECTIONS =
+          "EXISTS (SELECT 1 FROM data_access_request WHERE (collection_id = c.collection_id and draft = false))";
```
lu-c: Nitpick: I kind of find this language confusing - `filtering` i expect to return all results that match the filter, but `find` is where i expect a single result. The variable name (with the "s" on collections) also implies multiple results possible.

(Sorry, I've mostly had experience working on nosql databases, so if I'm wrong about this, please let me know!) Is this supposed to return a single result?

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
```diff
@@ -385,6 +385,11 @@ private DataAccessRequest populateDarFromJsonString(User user, String json) {
           && existingDar.getUserId().equals(user.getUserId())
           && existingDar.getDraft()) {
         newDar.setReferenceId(data.getReferenceId());
+
+        // if draft was part of a collection, we should use the same collection.
+        if (Objects.nonNull(existingDar.getCollectionId())) {
```
lu-c, line 390: nitpick: what would you think of adding a method on the existing dar for "isDraft()" that returns `Objects.nonNull(existingDar.getCollectionId()`?

I would rec that because I dont know if the null ID will *always* indicate draft status, and it might be a nice safe way for us to update this type of check across the app quickly in the future.

lu-c (approved):
Greg's approval is worth 2x mine :P What he says goes!

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
rushtong: I lean towards this feedback as well. We do have a pattern of extracting large query blocks reused across many queries, but this is only used once and it feels more readable as a string value here.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
connorlbark: ohh to clarify this isn't checking if it is a draft, but that if the DAR which already exists in the db has a collection we should pull that id in. I will clarify the comment above; this just makes sure when the draft is resubmitted it doesn't make a new collection, but keeps the old one (since it's already there).

src/main/java/org/broadinstitute/consent/http/models/DarCollectionSummary.java
connorlbark: Yeah totally! I agree with this feedback and above that it should just be inline with the DAO query string, so that makes the confusing name issue moot.

rushtong (approved):
👍🏽

PR #1684, opened 2022-08-03
[DUOS-1998][risk=low] Get Approved Users TDR endpoint

connorlbark:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1998

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
@@ -58,6 +58,20 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
           + "  AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL) ")
   List<DataAccessRequest> findAllDataAccessRequestsByDatasetId(@Bind("datasetId") String datasetId);
 
+
+  @UseRowReducer(DataAccessRequestReducer.class)
+  @SqlQuery(
+          " SELECT dd.dataset_id, dar.id, dar.reference_id, dar.collection_id, dar.parent_id, dar.draft, dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date, "
+                  + "  (dar.data #>> '{}')::jsonb AS data FROM data_access_request dar "
+                  + "  LEFT JOIN dar_dataset dd ON dd.reference_id = dar.reference_id AND dd.dataset_id = :datasetId::int  "
+                  + "  WHERE dar.draft = false"
+                  + "  AND (EXISTS (SELECT 1 FROM election e"
+                    + "  INNER JOIN vote v on v.electionId = e.electionId and lower(v.type) = 'final'"
+                    + "  WHERE v.vote = true AND lower(e.electionType) = 'dataaccess'"
+                    + "  AND e.referenceId = dar.reference_id))"
+                  + "  AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL) ")
+  List<DataAccessRequest> findAllApprovedDataAccessRequestsByDatasetId(@Bind("datasetId") String datasetId);
+
   /**
    * Find all draft/partial DataAccessRequests, sorted descending order
    *
--- a/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
@@ -59,18 +59,16 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
   List<DataAccessRequest> findAllDataAccessRequestsByDatasetId(@Bind("datasetId") String datasetId);
 
 
-  @UseRowReducer(DataAccessRequestReducer.class)
   @SqlQuery(
-          " SELECT dd.dataset_id, dar.id, dar.reference_id, dar.collection_id, dar.parent_id, dar.draft, dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date, "
-                  + "  (dar.data #>> '{}')::jsonb AS data FROM data_access_request dar "
-                  + "  LEFT JOIN dar_dataset dd ON dd.reference_id = dar.reference_id AND dd.dataset_id = :datasetId::int  "
-                  + "  WHERE dar.draft = false"
-                  + "  AND (EXISTS (SELECT 1 FROM election e"
-                    + "  INNER JOIN vote v on v.electionId = e.electionId and lower(v.type) = 'final'"
-                    + "  WHERE v.vote = true AND lower(e.electionType) = 'dataaccess'"
-                    + "  AND e.referenceId = dar.reference_id))"
-                  + "  AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL) ")
-  List<DataAccessRequest> findAllApprovedDataAccessRequestsByDatasetId(@Bind("datasetId") String datasetId);
+          " SELECT dar.user_id FROM data_access_request dar "
+          + "  LEFT JOIN dar_dataset dd ON dd.reference_id = dar.reference_id AND dd.dataset_id = :datasetId  "
+          + "  WHERE dar.draft = false"
+          + "  AND (EXISTS (SELECT 1 FROM election e"
+            + "  INNER JOIN vote v on v.electionId = e.electionId and lower(v.type) = 'final'"
+            + "  WHERE v.vote = true AND lower(e.electionType) = 'dataaccess'"
+            + "  AND e.referenceId = dar.reference_id))"
+          + "  AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL) ")
+  List<Integer> findAllUserIdsWithApprovedDARsByDatasetId(@Bind("datasetId") Integer datasetId);
 
   /**
    * Find all draft/partial DataAccessRequests, sorted descending order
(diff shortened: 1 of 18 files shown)


src/main/java/org/broadinstitute/consent/http/models/tdr/ApprovedUser.java
```diff
@@ -0,0 +1,20 @@
+package org.broadinstitute.consent.http.models.tdr;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class ApprovedUser {
+    @JsonProperty
```
rushtong: I'm 99% sure we don't need to use `@JsonProperty` if we serialize these objects in the response using `Gson`

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -592,6 +593,13 @@ public String getDatasetApprovedUsersContent(AuthUser authUser, Integer datasetI
         return builder.toString();
     }
 
+    public Collection<User> getUsersApprovedForDataset(Dataset dataset) {
+        List<DataAccessRequest> dars = this.dataAccessRequestDAO.findAllApprovedDataAccessRequestsByDatasetId(dataset.getDataSetId().toString());
+        Set<Integer> userIds = dars.stream().map(DataAccessRequest::getUserId).collect(Collectors.toSet());
+
+        return this.userDAO.findUsers(userIds);
```
rushtong: Safety note here, it's always good to make sure you're not passing an empty list into the find method. That ends up throwing a database error.

connorlbark: good catch! did not know that

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
```diff
@@ -36,7 +36,7 @@ public interface DatasetDAO extends Transactional<DatasetDAO> {
 
     String CHAIRPERSON = Resource.CHAIRPERSON;
 
-    @SqlUpdate("INSERT INTO dataset (name, createdate, create_user_id, update_date, update_user_id, objectId, active, alias) (SELECT :name, :createDate, :createUserId, :createDate, :createUserId, :objectId, :active, MAX(alias)+1 FROM dataset)")
+    @SqlUpdate("INSERT INTO dataset (name, createdate, create_user_id, update_date, update_user_id, objectId, active, alias) (SELECT :name, :createDate, :createUserId, :createDate, :createUserId, :objectId, :active, COALESCE(MAX(alias),0)+1 FROM dataset)")
```
connorlbark, line 39: made this change because if there are no datasets in the database `MAX(alias)` is null; this ensures that the first dataset has an alias of 1 if there are none.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -58,6 +58,20 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
...
+          " SELECT dd.dataset_id, dar.id, dar.reference_id, dar.collection_id, dar.parent_id, dar.draft, dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date, "
+                  + "  (dar.data #>> '{}')::jsonb AS data FROM data_access_request dar "
+                  + "  LEFT JOIN dar_dataset dd ON dd.reference_id = dar.reference_id AND dd.dataset_id = :datasetId::int  "
+                  + "  WHERE dar.draft = false"
+                  + "  AND (EXISTS (SELECT 1 FROM election e"
+                    + "  INNER JOIN vote v on v.electionId = e.electionId and lower(v.type) = 'final'"
+                    + "  WHERE v.vote = true AND lower(e.electionType) = 'dataaccess'"
+                    + "  AND e.referenceId = dar.reference_id))"
+                  + "  AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL) ")
+  List<DataAccessRequest> findAllApprovedDataAccessRequestsByDatasetId(@Bind("datasetId") String datasetId);
```
rushtong: Why not require `Integer datasetId` and then you wouldn't have to coerce it in the query. It would also make it easier to pass the actual datasetId down from the resource class.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -592,6 +593,15 @@ public String getDatasetApprovedUsersContent(AuthUser authUser, Integer datasetI
         return builder.toString();
     }
 
+    public Collection<User> getUsersApprovedForDataset(Dataset dataset) {
+        List<DataAccessRequest> dars = this.dataAccessRequestDAO.findAllApprovedDataAccessRequestsByDatasetId(dataset.getDataSetId().toString());
+        Set<Integer> userIds = dars.stream().map(DataAccessRequest::getUserId).collect(Collectors.toSet());
+        if (userIds.isEmpty()) {
+            return List.of();
+        }
```
rushtong: If instead you just queried for the `dar.user_id` instead of all of the related DAR properties, you could simplify this quite a bit.

src/main/java/org/broadinstitute/consent/http/service/TDRService.java
```diff
@@ -0,0 +1,32 @@
...
+    @Inject
+    public TDRService(DataAccessRequestService dataAccessRequestService) {
+        this.dataAccessRequestService = dataAccessRequestService;
+    }
+
+    public ApprovedUsers getApprovedUsersForDataset(Dataset dataset) {
+        Collection<User> users = this.dataAccessRequestService.getUsersApprovedForDataset(dataset);
+
+        List<ApprovedUser> approvedUsers =
+                users.stream().map((u) -> new ApprovedUser(u.getEmail())).collect(Collectors.toList());
```
rushtong: Minor suggestion. We could sort the list so the response is consistent across API calls. JSON doesn't enforce order, but we can make that happen if we want.

src/main/resources/assets/schemas/ApprovedUsers.yaml
```diff
@@ -0,0 +1,13 @@
+type: object
+properties:
+
+  ApprovedUsers:
```
rushtong:
```suggestion
  approvedUsers:
```

src/main/java/org/broadinstitute/consent/http/models/Dataset.java
```diff
@@ -243,6 +243,21 @@ public static String parseAliasToIdentifier(Integer alias) {
...
 
+    public static Integer parseIdentifierToAlias(String identifier) throws IllegalArgumentException {
+        try {
+            String givenPrefix = identifier.substring(0, PREFIX.length());
+            if (!givenPrefix.equals(PREFIX)) {
+                throw new IllegalArgumentException("Invalid prefix.");
+            }
+
+            String aliasAsString = identifier.substring(PREFIX.length()); // cut off DUOS-
+            return Integer.parseInt(aliasAsString); // parse remaining as integer
```
rushtong, line 254: One thing I see in testing, these two identifiers are in effect equal: `DUOS-5` and `DUOS-000005`. It doesn't make any practical difference, but perhaps we could do a check in the resource class to ensure that the dataset returned actually has the same `datasetIdentifier` that the user passed in.

rushtong (changes requested):
Absolutely love this ... you have all phases of the plumbing from endpoint to database. See feedback inline, I think we can make some improvements here.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
connorlbark: Good point - I think I saw another DAO call that used a string so I figured there was some reason strings were preferred, but this makes more sense.

src/main/java/org/broadinstitute/consent/http/models/tdr/ApprovedUser.java
```diff
@@ -0,0 +1,17 @@
+package org.broadinstitute.consent.http.models.tdr;
+
+public class ApprovedUser {
+    private String email;
```
rushtong, line 4: Thank you for not perpetuating the @JsonProperty pattern ... I'm intending on fixing this every time I touch the codebase

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @GET
+    @Produces("application/json")
+    @PermitAll
+    @Path("/{identifier}/approved/users")
+    public Response getApprovedUsers(@Auth AuthUser authUser, @PathParam("identifier") String identifier) {
+        try {
+            Dataset dataset = this.datasetService.findDatasetByIdentifier(identifier);
+            if (dataset == null) {
```
rushtong: Prefer `Objects.isNull`/`Objects.nonNull`/`Objects.equals` over raw equals

src/main/java/org/broadinstitute/consent/http/service/TDRService.java
```diff
@@ -0,0 +1,36 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class TDRService {
+    private final DataAccessRequestService dataAccessRequestService;
+    private final Logger logger = LoggerFactory.getLogger(this.getClass());
```
rushtong, line 18: Optional nit: we're not actually using logger here, but not a big issue.

src/main/java/org/broadinstitute/consent/http/service/TDRService.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Inject
+    public TDRService(DataAccessRequestService dataAccessRequestService) {
+        this.dataAccessRequestService = dataAccessRequestService;
+    }
+
+    public ApprovedUsers getApprovedUsersForDataset(Dataset dataset) {
+        Collection<User> users = this.dataAccessRequestService.getUsersApprovedForDataset(dataset);
```
rushtong: nit, don't need `this.`

rushtong (approved):
Looks good overall! A couple minor nits that shouldn't hold up the PR 👍🏽

shaemarks (approved):
Looks good to me!

PR #1687, opened 2022-08-04
[DUOS-1967][risk=no] SOs don't see users not in their institution

connorlbark:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1967

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/UserService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/UserService.java
@@ -208,13 +208,7 @@ public class UserService {
             case Resource.SIGNINGOFFICIAL :
                 Integer institutionId = user.getInstitutionId();
                 if (Objects.nonNull(user.getInstitutionId())) {
-                    List<User> institutionUsers = userDAO.getUsersFromInstitutionWithCards(institutionId);
-                    List<User> unregisteredUsers = userDAO.getCardsForUnregisteredUsers(institutionId);
-                    return Stream.of(
-                        institutionUsers,
-                        unregisteredUsers
-                    ).flatMap(Collection::stream)
-                    .collect(Collectors.toList());
+                    return userDAO.getUsersFromInstitutionWithCards(institutionId);
                 } else {
                     throw new NotFoundException("Signing Official (user: " + user.getDisplayName() + ") is not associated with an Institution.");
                 }
--- a/src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
@@ -423,8 +423,7 @@ public class UserServiceTest {
     public void testGetUsersByUserRole_SO() {
         User u = generateUser();
         u.setInstitutionId(1);
-        when(userDAO.getUsersFromInstitutionWithCards(anyInt())).thenReturn(List.of(new User()));
-        when(userDAO.getCardsForUnregisteredUsers(anyInt())).thenReturn(List.of(new User()));
+        when(userDAO.getUsersFromInstitutionWithCards(anyInt())).thenReturn(List.of(new User(), new User()));
         initService();
 
         List<User> users = service.getUsersAsRole(u, UserRoles.SIGNINGOFFICIAL.getRoleName());
@@ -437,7 +436,7 @@ public class UserServiceTest {
         User u = generateUser();
         u.setInstitutionId(null);
         initService();
-    service.getUsersAsRole(u, UserRoles.SIGNINGOFFICIAL.getRoleName());
+        service.getUsersAsRole(u, UserRoles.SIGNINGOFFICIAL.getRoleName());
     }
 
     @Test


rushtong (approved):
Nice and simple 👍🏽

shaemarks (approved):
Changes here look good, but I think some frontend work is necessary for this ticket as well. Under `SigningOfficialResearchers.js`, there is a call to `User.getUnassignedUsers()` to get users without an institution, and these users are passed to `LibraryCardFormModal` in the `SigningOfficialTable` for the users dropdown.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -208,13 +208,7 @@ public List<User> getUsersAsRole(User user, String roleName) {
...
                 Integer institutionId = user.getInstitutionId();
                 if (Objects.nonNull(user.getInstitutionId())) {
-                    List<User> institutionUsers = userDAO.getUsersFromInstitutionWithCards(institutionId);
-                    List<User> unregisteredUsers = userDAO.getCardsForUnregisteredUsers(institutionId);
-                    return Stream.of(
-                        institutionUsers,
-                        unregisteredUsers
-                    ).flatMap(Collection::stream)
-                    .collect(Collectors.toList());
+                    return userDAO.getUsersFromInstitutionWithCards(institutionId);
```
rushtong, line 211: I should have caught this before. This will only find users in the institution WITH LCs. I'm pretty sure we want users in the institution. I think we might need to revisit this entire code path to ensure we're getting the right thing.

connorlbark: on the query itself:
>            //This will pull in users tied to the institution
>            //Users will come with LCs issued by SOs institution (if any)
I think the `WithCards` is referring to the fact that if the user has any library cards it is populated when returned by the db, but it does not filter to only users with library cards.

rushtong: Ok, thanks, false alarm - love it when I'm wrong.

connorlbark:
> Changes here look good, but I think some frontend work is necessary for this ticket as well. Under `SigningOfficialResearchers.js`, there is a call to `User.getUnassignedUsers()` to get users without an institution, and these users are passed to `LibraryCardFormModal` in the `SigningOfficialTable` for the users dropdown.

At a cursory look, I believe the `LibraryCardFormModal` is to allow the `Add New Researcher` functionality, which I think is still ok ? The actual table of researchers itself does not show the unregistered users.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
connorlbark: When I test by looking at the UI,  researchers without cards in my local environment appear in the table.

shaemarks:
> At a cursory look, I believe the `LibraryCardFormModal` is to allow the `Add New Researcher` functionality, which I think is still ok ? The actual table of researchers itself does not show the unregistered users.

I may just be misunderstanding what “users dropdown" meant in the ticket (I interpreted it as the dropdown in the modal, but that may not be the correct assumption). However, the onClick functionality of the modal does issue a new library card, and I was able to do so for a researcher without an institution. 

Description of test I did:
In the dropdown of the modal, I found a different user of mine without an institution. When clicking "add", they received a library card and appeared in the table. After refreshing, they no longer appeared in the table because they still don't belong to my institution (just have a library card from there). When I tried to add them again, I got a conflict error because they already had a library card from the institution.

shaemarks:
> At a cursory look, I believe the `LibraryCardFormModal` is to allow the `Add New Researcher` functionality, which I think is still ok ? The actual table of researchers itself does not show the unregistered users.

I may just be misunderstanding what “users dropdown" meant in the ticket (I interpreted it as the dropdown in the modal, but that may not be the correct assumption). However, the onClick functionality of the modal does issue a new library card, and I was able to do so for a researcher without an institution. 

Description of test I did:
In the dropdown of the modal, I found a different user of mine without an institution. When clicking "add", they received a library card and appeared in the table. After refreshing, they no longer appeared in the table because they still don't belong to my institution (just have a library card from there). When I tried to add them again, I got a conflict error because they already had a library card from the institution.

connorlbark:
> > At a cursory look, I believe the `LibraryCardFormModal` is to allow the `Add New Researcher` functionality, which I think is still ok ? The actual table of researchers itself does not show the unregistered users.
> 
> I may just be misunderstanding what “users dropdown" meant in the ticket (I interpreted it as the dropdown in the modal, but that may not be the correct assumption). However, the onClick functionality of the modal does issue a new library card, and I was able to do so for a researcher without an institution.
> 
> Description of test I did: In the dropdown of the modal, I found a different user of mine without an institution. When clicking "add", they received a library card and appeared in the table. After refreshing, they no longer appeared in the table because they still don't belong to my institution (just have a library card from there). When I tried to add them again, I got a conflict error because they already had a library card from the institution.

Ahhh I think I see what you mean; I appreciate the thorough response / testing. I will look into this and add on a UI PR to this story.

connorlbark:
UI PR: https://github.com/DataBiosphere/duos-ui/pull/1737
Removes the ability to add unregistered users by removing the 'Add New Researcher' modal.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
rushtong: I still think we need `List<User> unregisteredUsers = userDAO.getCardsForUnregisteredUsers(institutionId);`. That finds all entries where a SO has created a "User" in their institution.

rushtong: >> UI PR: https://github.com/DataBiosphere/duos-ui/pull/1737

> Removes the ability to add unregistered users by removing the 'Add New Researcher' modal.

Ah, never mind, I missed that we were removing that feature.

rushtong (approved):

PR #1695, opened 2022-08-08
[DUOS-1988][risk=no] fix zendesk support request and expand functionality to selected institutions/SOs

shaemarks:
Addresses [DUOS-1988](https://broadworkbench.atlassian.net/browse/DUOS-1988)
Summary of changes:
- create new `HttpClientUtils` method to send POST request without authentication header (to resolve 401 Not Authorized error)
- use `ByteArrayContent` instead of `JsonHttpContent` to send ticket (workaround to resolve 422 Unprocessable Entity error)
- expand functionality of method that creates and sends a ticket for suggested institutions and signing officials to also describe the selection or existing institutions and signing officials 
- while based on the frontend implementation of `ResearcherProfile` I would expect at most one institution  (existing or new) and one signing official (existing or new) in the `UserUpdateFields`, the endpoint itself doesn't place any restrictions on which fields appear. To accommodate this, I tried to be more flexible with the grammar of the subject and description of the support ticket 

Note: To enable support request notifications reaching Zendesk, set `activateSupportNotifications: true` under `services:` in `consent.yaml`

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/service/SupportRequestService.java
```diff
@@ -133,40 +141,59 @@ public void handleSuggestedUserFieldsSupportRequest(UserUpdateFields userUpdateF
...
         }
 
+        //Append items for ticket subject and description
+        String subject = String.format("%s user updates: %s",
+                user.getDisplayName(),
+                String.join(", ", subjectItems));
+        String description = String.format("User %s [%s] has:\n%s",
+                user.getDisplayName(),
+                user.getEmail(),
+                String.join("\n", descriptionItems));
```
rushtong: There's a lot of code here in the service class that is very specific to ticket creation/population. An optional pattern would be to extract a chunk of that logic into a model class so it's more encapsulated. Not a requirement, just a style suggestion. See how we use generic utility methods on `UserUpdateFields` as a possible example.

src/main/java/org/broadinstitute/consent/http/util/HttpClientUtil.java
```diff
@@ -51,6 +51,13 @@ public HttpRequest buildPostRequest(GenericUrl genericUrl, HttpContent content,
     return request;
   }
 
+  public HttpRequest buildUnAuthedPostRequest(GenericUrl genericUrl, HttpContent content) throws Exception {
+    HttpTransport transport = new NetHttpTransport();
+    HttpRequest request = transport.createRequestFactory().buildPostRequest(genericUrl, content);
+    request.setHeaders(new HttpHeaders().set("X-App-ID", "DUOS"));
```
rushtong, line 57: No need to change this line - just an explanation. All Terra services communicate with each other and send this header. It's intended to identify the calling app. In this usage, we're talking to Zendesk, so it will likely be ignored.

rushtong (approved):
Looks good! 👍🏽 See optional suggestions inline.

connorlbark (approved):
looks really solid! i am not sure how to test tickets going through properly, but the tests look solid and give me confidence that it works properly (sends the right payload, etc.)

src/main/java/org/broadinstitute/consent/http/models/support/SupportTicket.java
```diff
@@ -9,10 +13,37 @@ public class SupportTicket {
...
+     * @param email       The email of the user requesting support
+     * @param subject     Subject line of the request
+     * @param description Description of the task or question
+     * @param url         The API url of this request
+     */
+    public SupportTicket(String name, SupportRequestType type, String email, String subject, String description, String url) {
+        if (Objects.isNull(name) || Objects.isNull(email)) {
+            throw new IllegalArgumentException("Name and email of user requesting support is required");
+        }
+        if (Objects.isNull(subject)) {
```
shaemarks, line 30: I carried these null checks over from `SupportRequestService.createSupportTicket`, but most existing model classes don't check for null values in the constructor. Thoughts on whether to keep or remove?

src/main/java/org/broadinstitute/consent/http/service/SupportRequestService.java
```diff
@@ -105,24 +68,25 @@ public void postTicketToSupport(SupportTicket ticket, AuthUser authUser) throws
...
+                    || Objects.nonNull(userUpdateFields.getSelectedSigningOfficialId());
 
-            // Only create and send ticket if either a suggestedInstitution or suggestedSigningOfficial has been provided
-            if (Objects.nonNull(suggestedInstitution) || Objects.nonNull(suggestedSigningOfficial)) {
+            //only send ticket if an institution or signing official is provided; ignore otherwise
+            if (updateFieldProvided) {
                 try {
-                    SupportTicket ticket = createSuggestedUserFieldsTicket(userUpdateFields, user);
-                    postTicketToSupport(ticket, authUser);
+                    SupportTicket ticket = supportTicketCreator.createInstitutionSOSupportTicket(userUpdateFields, user, configuration.postSupportRequestUrl());
```
shaemarks: The url is included in the comment body (based on the template from [DUOS-1890](https://broadworkbench.atlassian.net/browse/DUOS-1890)). Is there a preference between passing in the url here vs having a configuration field in `SupportTicketCreator`?

shaemarks:
Changes made since last approval:
- Refactor to move ticket creation/population out of `SupportRequestService` (`createSupportTicket` is now the `SupportTicket` constructor; `createInstitutionSOSupportTicket` is now located in `SupportTicketCreator`)
- Added `SupportTicketCreator` to handle creating certain types of tickets (in this case institution/SO request) since `InstitutionDAO` and `UserDAO` were needed to get names and emails for selected fields, and I didn't want the DAOs to be parameters in a SupportTicket constructor 
- moved tests into appropriate files

connorlbark (approved):
👍  changes make building a ticket much cleaner! big fan

src/main/java/org/broadinstitute/consent/http/models/support/SupportTicket.java
rushtong: I think we should keep these here, they are all required fields.

src/main/java/org/broadinstitute/consent/http/service/SupportRequestService.java
rushtong: If we're moving construction details to the creator class, then it does make sense to add a config there as well.

rushtong (approved):
👍🏽 Pending whatever choices you'd like to make. Looks good!

PR #1696, opened 2022-08-08
[DUOS-1969][risk=no] Copy data use from consents to datasets

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1969

- Copy all `DataUse` objects over to the `dataset` table.
- Refactor creation and querying from `consents` to `dataset`
- Refactor `Dac` to have a full Dataset object instead of the deprecated DTO.
- Add new `unmarshal` pattern for serializing returned API objects with `Gson` instead of `Jackson`

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
@@ -36,12 +36,18 @@ public interface DatasetDAO extends Transactional<DatasetDAO> {
 
     String CHAIRPERSON = Resource.CHAIRPERSON;
 
-    @SqlUpdate("INSERT INTO dataset (name, createdate, create_user_id, update_date, update_user_id, objectId, active, alias) (SELECT :name, :createDate, :createUserId, :createDate, :createUserId, :objectId, :active, COALESCE(MAX(alias),0)+1 FROM dataset)")
+    @SqlUpdate("INSERT INTO dataset (name, createdate, create_user_id, update_date, update_user_id, objectId, active, alias, data_use) (SELECT :name, :createDate, :createUserId, :createDate, :createUserId, :objectId, :active, COALESCE(MAX(alias),0)+1, :dataUse FROM dataset)")
     @GetGeneratedKeys
-    Integer insertDataset(@Bind("name") String name, @Bind("createDate") Timestamp createDate, @Bind("createUserId") Integer createUserId, @Bind("objectId") String objectId, @Bind("active") Boolean active);
+    Integer insertDataset(
+        @Bind("name") String name,
+        @Bind("createDate") Timestamp createDate,
+        @Bind("createUserId") Integer createUserId,
+        @Bind("objectId") String objectId,
+        @Bind("active") Boolean active,
+        @Bind("dataUse") String dataUse);
 
     @UseRowReducer(DatasetReducer.class)
-    @SqlQuery(" SELECT d.*, k.key, dp.propertyvalue, dp.propertykey, dp.propertyid, ca.consentid, c.dac_id, c.translateduserestriction, c.datause, dar_ds_ids.id as in_use " +
+    @SqlQuery(" SELECT d.*, k.key, dp.propertyvalue, dp.propertykey, dp.propertyid, ca.consentid, c.dac_id, c.translateduserestriction, dar_ds_ids.id as in_use " +
         " FROM dataset d " +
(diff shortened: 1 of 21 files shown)


src/main/resources/changelog-master.xml
```diff
@@ -88,4 +88,5 @@
     <include file="changesets/changelog-consent-83.0.xml" relativeToChangelogFile="true"/>
     <include file="changesets/changelog-consent-84.0.xml" relativeToChangelogFile="true"/>
     <include file="changesets/changelog-consent-85.0.xml" relativeToChangelogFile="true"/>
+    <include file="changesets/changelog-consent-87.0.xml" relativeToChangelogFile="true"/>
```
shaemarks: I think the new changelog should be 86.0

src/main/java/org/broadinstitute/consent/http/resources/Resource.java
```diff
@@ -214,4 +216,15 @@ void validateUserHasRoleName(User user, String roleName) {
...
 
+    /**
+     * Unmarshal/serialize an object using `Gson`. In general, we should prefer Gson
+     * over Jackson for ease of use and the need for far less boilerplate code.
+     *
+     * @param o The object to unmarshal
+     * @return String version of the object
+     */
+    protected String unmarshal(Object o) {
+        return new Gson().toJson(o);
```
shaemarks, line 227: Out of curiosity, does any resource method without  `new Gson.toJson` generate the response with Jackson by default?

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
```diff
@@ -174,7 +180,7 @@ public interface DatasetDAO extends Transactional<DatasetDAO> {
     Set<DatasetDTO> findAllDatasets();
 
     @UseRowMapper(DatasetPropertiesMapper.class)
-    @SqlQuery("SELECT d.*, k.key, dp.propertyvalue, ca.consentid, c.dac_id, c.translateduserestriction, c.datause " +
+    @SqlQuery("SELECT d.*, k.key, dp.propertyvalue, ca.consentid, c.dac_id, c.translateduserestriction " +
```
shaemarks, line 183: Does `DatasetPropertiesMapper` need `dataSetDTO.setDataUse(DataUse.parseDataUse(r.getString("datause")).orElse(null));` (line 40) updated as well? I know we're moving away from using DatasetDTO, but this DAO method, for example, is used for the `GET /api/dataset/{datasetId}` endpoint, where the dataUse is currently returned as null

shaemarks (commented):
Refactoring looks good overall, see comments inline

src/main/resources/changelog-master.xml
rushtong: I'm using `86` in https://github.com/DataBiosphere/consent/pull/1678, so if that lands first, this works, otherwise I'll update it.

src/main/java/org/broadinstitute/consent/http/resources/Resource.java
rushtong: Yes, jackson is the default in general for dropwizard. Anytime you send an Object to
```
Response.ok().entity(... some entity ...).build()
```
Jackson is called to do that entity conversion. If instead you pass it a string, Jackson is not called. You'll see this in some other cases like here ... https://github.com/DataBiosphere/consent/blob/develop/src/main/java/org/broadinstitute/consent/http/resources/DACUserResource.java#L121 

I created this method to make it easier to use in all resource classes.

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
rushtong: Great catch - will update that one.

src/main/resources/changelog-master.xml
shaemarks: gotcha 👍

src/main/java/org/broadinstitute/consent/http/resources/Resource.java
shaemarks: Ok, thanks for the explanation!

shaemarks (approved):
Looks good to me!

JVThomas (approved):
Code looks good. Tested locally, data_use column has been backfilled on existing datasets and newly registered datasets have the column populated.

PR #1699, opened 2022-08-10
[DUOS-1770][risk=no] Restrict manage v2 DARs to signing officials only

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1770

All usages of this endpoint have been removed aside from the SO console. Restrict this to the SO role and clean up unused code.

Removed an automation test that tried to hit this endpoint as an admin. These tests will be re-written in Java in https://broadworkbench.atlassian.net/browse/DUOS-1684 so we're not adding additional tests for the updated methods.

----

- Verify all tests go green

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsp/consent/chains/AdminChains.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsp/consent/chains/AdminChains.scala
@@ -4,12 +4,7 @@ import io.gatling.core.Predef._
 import io.gatling.core.structure.ChainBuilder
 import org.broadinstitute.dsp.consent.requests.Requests
 import org.broadinstitute.dsp.consent.TestConfig
-import spray.json._
-import DefaultJsonProtocol._
 import org.broadinstitute.dsp.consent.models.DataAccessRequestModels._
-import org.broadinstitute.dsp.consent.models.ElectionModels._
-import org.broadinstitute.dsp.consent.models.JsonProtocols
-import org.broadinstitute.dsp.consent.services._
 import io.netty.handler.codec.http.HttpResponseStatus._
 
 object AdminChains {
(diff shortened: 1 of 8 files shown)


shaemarks:
Side note: I think we should be able to remove the SO usage of this endpoint on the frontend as well. The endpoint is used on the `SigningOfficialConsole` component, with the `/signing_official_console` path, which we don't directly link to. With the new nav bar, we send users to `/signing_official_console/researchers` (`SigningOfficialResearcher` component) and `/signing_official_console/dar_requests` (`SigningOfficialDarRequests` component), neither of which call `DAR.getDataAccessManageV2`

rushtong:
> Side note: I think we should be able to remove the SO usage of this endpoint on the frontend as well. The endpoint is used on the `SigningOfficialConsole` component, with the `/signing_official_console` path, which we don't directly link to. With the new nav bar, we send users to `/signing_official_console/researchers` (`SigningOfficialResearcher` component) and `/signing_official_console/dar_requests` (`SigningOfficialDarRequests` component), neither of which call `DAR.getDataAccessManageV2`

`SigningOfficialConsole` currently calls `DAR.getDataAccessManageV2` ... when we end up removing that, we can finally remove the consent API altogether.

shaemarks:
> `SigningOfficialConsole` currently calls `DAR.getDataAccessManageV2` ... when we end up removing that, we can finally remove the consent API altogether.

Yeah, I was just mentioning that `SigningOfficialConsole` should be ready for removal since we no longer send users to it. As far as I know, we don't have a ticket to remove it yet

rushtong:
> Yeah, I was just mentioning that `SigningOfficialConsole` should be ready for removal since we no longer send users to it. As far as I know, we don't have a ticket to remove it yet

Aha, that is very true. I'll file a ticket for that and once that's finished, I'll update this PR.

shaemarks (approved):
Looks good to me! Signing official role is still supported, and other roles are deprecated for this endpoint

connorlbark (approved):
👍

PR #1709, opened 2022-08-18
[DUOS-1954][risk=no] add sharing_plan_document column to dataset table

shaemarks:
Addresses [DUOS-1954](https://broadworkbench.atlassian.net/browse/DUOS-1954)

- Updated dataset table, mapper, and model to include `sharing_plan_document`
- I didn't update the DatasetDTO code since we are trying to deprecate it, but I know it's used in some common places, so let me know if that is required as well

----
Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

PR #1711, opened 2022-08-19
DUOS-1983[risk=no] Added dac_approval column to Dataset table

JVThomas:
Addresses [DUOS-1983](https://broadworkbench.atlassian.net/browse/DUOS-1983)

PR adds `dac_approval` column to the `dataset` table. PR also backfills the `dac_approval` value to `TRUE` for datasets that are already associated to a DAC.

NOTE: Changelog is labeld as `92.0` intentionally to avoid conflicts with changelog updates in https://github.com/DataBiosphere/consent/pull/1709, which came earlier

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetMapper.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetMapper.java
@@ -14,6 +14,9 @@ public class DatasetMapper implements RowMapper<Dataset>, RowMapperHelper {
       dataset.setDataSetId(r.getInt("datasetid"));
       dataset.setObjectId(r.getString("objectid"));
       dataset.setName(r.getString("name"));
+      if(hasColumn(r, "dac_approval")) {
+        dataset.setDacApproval(r.getBoolean("dac_approval"));
+      }
       if (hasColumn(r, "createdate")) {
           dataset.setCreateDate(r.getDate("createdate"));
       }
--- a/src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetMapper.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetMapper.java
@@ -2,6 +2,8 @@ package org.broadinstitute.consent.http.db.mapper;
 
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.util.Objects;
+
 import org.broadinstitute.consent.http.models.Dataset;
 import org.broadinstitute.consent.http.models.DataUse;
 import org.jdbi.v3.core.mapper.RowMapper;
@@ -27,7 +29,9 @@ public class DatasetMapper implements RowMapper<Dataset>, RowMapperHelper {
           dataset.setUpdateDate(r.getTimestamp("update_date"));
       }
       if (hasColumn(r, "dac_approval")) {
-          dataset.setDacApproval(r.getBoolean("dac_approval"));
+        String boolString = r.getString("dac_approval");
+        Boolean value = Objects.isNull(boolString) ? null : Boolean.getBoolean(boolString);
+        dataset.setDacApproval(value);
       }
       if (hasColumn(r, "update_user_id")) {
           int userId = r.getInt("update_user_id");
--- a/src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetMapper.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetMapper.java
@@ -30,7 +30,7 @@ public class DatasetMapper implements RowMapper<Dataset>, RowMapperHelper {
       }
       if (hasColumn(r, "dac_approval")) {
         String boolString = r.getString("dac_approval");
-        Boolean value = Objects.isNull(boolString) ? null : Boolean.getBoolean(boolString);
+        Boolean value = Objects.isNull(boolString) ? null : r.getBoolean("dac_approval");
         dataset.setDacApproval(value);
       }
       if (hasColumn(r, "update_user_id")) {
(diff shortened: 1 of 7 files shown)


src/main/resources/changesets/changelog-consent-92.0.xml
```diff
@@ -0,0 +1,19 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+  <changeSet id="92.0" author="JVThomas">
+    <!--Add mapping and dataset property type columns -->
```
shaemarks: nit: comment looks leftover from a different changelog

src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetReducer.java
```diff
@@ -37,6 +37,9 @@ public void accumulate(Map<Integer, Dataset> map, RowView rowView) {
       Integer dsIdInUse = rowView.getColumn("in_use", Integer.class);
       dataset.setDeletable(Objects.isNull(dsIdInUse));
     }
+    if (hasColumn(rowView, "dac_approval", Boolean.class)) {
+      dataset.setDacApproval(rowView.getColumn("dac_approval", Boolean.class));
```
shaemarks, line 41: I feel like it shouldn't be necessary to set the `dacApproval `in both the `DatasetMapper` and `DatasetReducer`. I was under the impression that the mapper handles columns from the `dataset` table and the reducer handles columns from other tables. However, the  `r.getBoolean("dac_approval")` call used in `DatasetMapper` returns a `boolean`, and nulls are returned as false (whereas the call in `DatasetReducer` uses type `Boolean` and covers up this bug). The current code is working properly for me, but it might be worth addressing this, especially because if a query just used the `DatasetMapper`, the dataset could have the wrong value for `dacApproval`, and it might not be obvious that something is wrong

shaemarks (approved):
Tested locally and code looks good. My local instance of the db has dac_approval correctly backpopulated for datasets associated with a dac 👍

connorlbark (approved):
agree with shae's feedback, needs changelog update since master has a new one but looks good otherwise!

src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetReducer.java
JVThomas: I'm not a complete expert on Mappers and Reducers, but my understanding was that Reducers (specifically the ones we use that extend `LinkedHasMapRowReducer`) operate similarly to Mappers with one exception: it stores the row result as a key-value pair, with the key being a user designated identifier (usually an id). 

The difference kicks in when a model's value is repeated on multiple rows (usually on a `JOIN` for a one-to-many relationship). Both mappers and reducers will tackle their operations row by row, but a mapper is naive and will just create an instance and pass it to the resultant list. A reducer will check to see if the model is already saved in the map. If it's there it can just fetch the currently stored value and update it, otherwise it'll add it as a new key-value pair. The end result is that the mapper will have multiple instances of a model with the same id, whereas the reducer would not. That said, a reducer is going to be slower since it has to do things in the background (one being tracking insertion order).

The bug is a concern though, I keep forgetting that Java has both `boolean` and `Boolean`. I think a cast should solve the issue, but I'll add and/or update tests for the relevant queries on `DatasetDAO` to check on the null values just to be safe.

shaemarks: Makes sense, thanks for the explanation! Do you have any tips on determining if a particular column name should be checked in the mapper/reducer/both?

rushtong: I like to do column checks in both mappers and reducers. One more thing to remember, Reducers and mappers also go hand in hand. When a reducer tries to instantiate an object from a row, it will start with the existing mapper. We usually go a step further to populate additional info depending on the various logic conditions.

JVThomas:
@shaemarks @connorlbark Sorry about this, but would you two mind re-reviewing my PR? I've added additions to deal with the Boolean vs boolean issue that was mentioned earlier (alongside changelog updates).

src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetReducer.java
shaemarks: So for #1709 I only added 
```
if (hasColumn(r, "sharing_plan_document")) {
      dataset.setSharingPlanDocument(r.getString("sharing_plan_document"));
 }
```
in `DatasetMapper` and not `DatasetReducer` since the reducer calls the mapper to instantiate the object. Is this correct, or is additional code needed in the reducer?

rushtong: If the mapper reads it, then that will be enough to instantiate the field. Sometimes in the reducer, we need to do extra work to populate fields, and that's where I would want to double check that the columns I wanted existed.

src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetMapper.java
```diff
@@ -26,6 +28,11 @@ public Dataset map(ResultSet r, StatementContext ctx) throws SQLException {
       if (hasColumn(r, "update_date")) {
           dataset.setUpdateDate(r.getTimestamp("update_date"));
       }
+      if (hasColumn(r, "dac_approval")) {
+        String boolString = r.getString("dac_approval");
+        Boolean value = Objects.isNull(boolString) ? null : Boolean.getBoolean(boolString);
```
shaemarks:
```suggestion
        Boolean value = Objects.isNull(boolString) ? null : r.getBoolean("dac_approval");
```
The current code is now handling null columns properly, but when `dac_approval` is true, `boolString` is `"t"`, so `Boolean.getBoolean("t")` is returning false

shaemarks (commented):
Local database is populated with the correct dac_approval values for existing datasets, and tests were a nice addition. One suggestion inline on the DatasetMapper null bug and then 👍

shaemarks (approved):

connorlbark (approved):

PR #1721, opened 2022-08-25
DUOS-1984[risk=low] New API endpoint for DAC Chairs to Approve datasets

JVThomas:
Addresses [DUOS-1984](https://broadworkbench.atlassian.net/browse/DUOS-1984)

PR adds the DAC dataset approval flow, allowing DAC Chairs to approve/deny datasets for their DAC. This means a new endpoint on `DacResource`, new methods on `User` and `DatasetService`, and a new query method added on `DatasetDAO`. Tests for the new methods have been added alongside API documentation updates.

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
```diff
@@ -206,6 +213,36 @@ public Response filterUsers(@PathParam("term") String term) {
...
+    @RolesAllowed({CHAIRPERSON})
+    public Response approveDataset(@Auth AuthUser authUser, @PathParam("dacId") Integer dacId, @PathParam("datasetId") Integer datasetId, String json) {
+        try{
+            User user = userService.findUserByEmail(authUser.getEmail());
+            Dataset dataset = datasetService.findDatasetById(datasetId);
+            if(dataset.getDacId() != dacId) { 
+                //Vague message is intentional, don't want to reveal too much info
+                throw new NotFoundException("Dataset not found");
+            }
+            userService.checkIfUserHasRole(UserRoles.CHAIRPERSON.getRoleName(), user, dacId);
```
rushtong: Just an idea, since you've already looked up the user, you could also use `User.hasUserRole()` and look for the `dacId` after that check. Alternatively, the same logic could also go in `User` since we're doing similar role-based checks there.

src/main/java/org/broadinstitute/consent/http/models/User.java
```diff
@@ -349,4 +351,21 @@ public boolean doesUserHaveAnyRoleInSet(EnumSet<UserRoles> userRoles) {
...
+                .filter((r) -> {
+                    return r.getName().equals(role.getRoleName())
+                            && r.getRoleId().equals(role.getRoleId())
+                            && Objects.equals(r.getDacId(), dacId);
+                })
+                .collect(Collectors.toList());
+        if (targetRoles.isEmpty()) {
+            // Intentionally giving 404 to avoid giving permissions insights
+            throw new NotFoundException("User role not found");
+        }
```
rushtong: I think it makes more sense to keep the model class free of this kind of exception and throw it at the resource class level instead. This way, API specific errors are not propagated through the layers of the business logic.

src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
```diff
@@ -206,6 +213,36 @@ public Response filterUsers(@PathParam("term") String term) {
...
+            Dataset dataset = datasetService.findDatasetById(datasetId);
+            if(dataset.getDacId() != dacId) { 
+                //Vague message is intentional, don't want to reveal too much info
+                throw new NotFoundException("Dataset not found");
+            }
+            user.checkIfUserHasRole(UserRoles.CHAIRPERSON.getRoleName(), dacId);
+            JsonObject payload = new Gson().fromJson(json, JsonObject.class);
+            if(payload.size() == 0) {
+                throw new BadRequestException("Request body is empty");
+            }
```
rushtong: Super minor optimization - you can check the string length before coercing to a `JsonObject` which would allow you to fail faster in this case.

src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
```diff
@@ -206,6 +213,36 @@ public Response filterUsers(@PathParam("term") String term) {
...
+    public Response approveDataset(@Auth AuthUser authUser, @PathParam("dacId") Integer dacId, @PathParam("datasetId") Integer datasetId, String json) {
+        try{
+            User user = userService.findUserByEmail(authUser.getEmail());
+            Dataset dataset = datasetService.findDatasetById(datasetId);
+            if(dataset.getDacId() != dacId) { 
+                //Vague message is intentional, don't want to reveal too much info
+                throw new NotFoundException("Dataset not found");
+            }
+            user.checkIfUserHasRole(UserRoles.CHAIRPERSON.getRoleName(), dacId);
+            JsonObject payload = new Gson().fromJson(json, JsonObject.class);
```
rushtong: Another approach here is to create a very small model class (`DatasetApproval`?) specifically meant to capture the intended input argument. That would make it a tiny bit cleaner down below where we need to get the elements by string value and then coerce into a boolean.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -435,6 +437,23 @@ public List<Map<String, String>> autoCompleteDatasets(String partial, Integer da
         ).collect(Collectors.toList());
     }
 
+    public Dataset approveDataset(Dataset dataset, User user, Boolean approvalBool) {
```
rushtong: minor naming nit, `approvalBool` is a boolean, so the `Bool` part of it seems redundant.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -435,6 +437,23 @@ public List<Map<String, String>> autoCompleteDatasets(String partial, Integer da
...
+        Dataset datasetReturn = dataset;
+        //Only update and fetch the dataset if it hasn't already been approved
+        //If it has, simply returned the dataset in the argument (which was already queried for in the resource)
+        if(Objects.isNull(currentApprovalState) || !currentApprovalState) {
+            datasetDAO.updateDatasetApproval(approvalBool, Instant.now(), user.getUserId(), datasetId);
+            datasetReturn = datasetDAO.findDatasetById(datasetId);
+        } else {
+            if(Objects.isNull(approvalBool) || !approvalBool) {
+                throw new ForbiddenException("Dataset is already approved");
+            }
```
rushtong: I think we could use a different exception here. `ForbiddenException` implies the user has no permissions to perform this function, but they do, this is just a disallowed case. I think `IllegalArgumentException` might be a better fit since those end up being converted to `BadRequestException` at the resource level.

src/main/resources/assets/paths/approveDataset.yaml
```diff
@@ -0,0 +1,49 @@
...
+    200:
+      description: Returns the updated dataset
+      content:
+        application/json:
+          schema:
+            $ref: '../schemas/Dataset.yaml'
+    400:
+      description: Bad Request (missing or invalid request body)
+    403:
+      description: Forbidden (User cannot un-approve an already approved dataset)
```
rushtong: See comment above

src/main/resources/assets/paths/approveDataset.yaml
```diff
@@ -0,0 +1,49 @@
...
+      required: true
+      schema:
+        type: integer
+    - name: datasetId
+      in: path
+      required: true
+      schema:
+        type: integer
+  requestBody:
+      description: 'Payload that should contain the key-value pair of {"approval": Boolean}'
```
rushtong: minor doc nit:
```suggestion
      description: 'Payload that should contain the key-value pair of {"approval": true|false}'
```

rushtong: Alternatively, a new model class could be created for `DatasetApproval` with a single field, `approval`.

src/main/resources/assets/paths/approveDataset.yaml
```diff
@@ -0,0 +1,49 @@
...
+    404:
+      description: Not Found
+    500: 
+      description: Internal Server Error
+    
+    
+              
+
+      
+
```
rushtong: nit: rm extra white space

src/main/resources/assets/schemas/Dataset.yaml
```diff
@@ -66,3 +66,6 @@ properties:
   dataUse:
     type: object
     description: The Data Use question/answer set for this dataset (used for the associated consent).
+  dacApproval:
+    type: Boolean
```
rushtong:
```suggestion
    type: boolean
```

src/main/resources/assets/paths/approveDataset.yaml
```diff
@@ -0,0 +1,49 @@
+put:
+  summary: DAC Chair updates dataset approval status
+  description: Endpoint allows DAC Chairs to submit approval decisions for datasets under their purview (or requested purview)
```
rushtong: I think it would be nice to add some commentary about the allowed operations here, i.e. a dataset's approval can go from null -> true/false, or false -> true, but not true -> false

rushtong (commented):
Looking good. Love the use of `Instant`. See comments inline

src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
```diff
@@ -206,6 +212,37 @@ public Response filterUsers(@PathParam("term") String term) {
...
+    @PUT
+    @Consumes("application/json")
+    @Produces("application/json")
+    @Path("{dacId}/dataset/{datasetId}")
+    @RolesAllowed({CHAIRPERSON})
+    public Response approveDataset(@Auth AuthUser authUser, @PathParam("dacId") Integer dacId, @PathParam("datasetId") Integer datasetId, String json) {
+        try{
+            User user = userService.findUserByEmail(authUser.getEmail());
+            Dataset dataset = datasetService.findDatasetById(datasetId);
+            if(dataset.getDacId() != dacId) {
```
shaemarks:
```suggestion
            if(Objects.isNull(dataset) || !Objects.equals(dataset.getDacId(), dacId)) { 
```

Response message is different when the dataset actually doesn't exist since returned dataset is null and  `dataset.getDacId()` throws an error. 

Screenshot (102)

shaemarks (approved):
Code, tests, and swagger documentation look good 👍 
One small suggestion below

rushtong (approved):
Thank you for the changes! 👍🏽

PR #1722, opened 2022-08-25
[DUOS-2039][risk=no] Allow DataSubmitter role to be added/deleted through /api/user/{userId}/{roleId}

shaemarks:
Addresses [DUOS-2039](https://broadworkbench.atlassian.net/browse/DUOS-2039)

Found while reviewing [DUOS-1669](https://broadworkbench.atlassian.net/browse/DUOS-1669) and trying to give myself the DataSubmitter role. Both endpoints affected are admin-only, and the Swagger documentation lists DataSubmitter as a valid role to add/remove, so this should be ok to go through now 

----

- Verify all tests go green


rushtong (approved):
Nice catch! 👍🏽

quazi-h (approved):
Changes look good to me, tested this on a local instance of consent and was able to add the data submitter role for another user and then delete the role from the user.

PR #1724, opened 2022-08-29
[DUOS-1970][risk=no] Migrate DAC association to dataset

connorlbark:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1970

Removes `dacId` from Consent, migrates dacIds to dataset. Removes `ConsentManage`.

----
Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

PR #1728, opened 2022-08-31
[DUOS-1996][risk=no] DAC Chairs Submissions Console API

connorlbark:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1996

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
@@ -373,13 +373,13 @@ public interface DatasetDAO extends Transactional<DatasetDAO> {
      * @return all datasets associated with DAC
      */
     @UseRowReducer(DatasetReducer.class)
-    @SqlQuery("SELECT d.*, k.key, dp.property_value, dp.property_key, dp.property_type, dp.schema_property, dp.property_id, ca.consentid, c.translateduserestriction, dar_ds_ids.id as in_use " +
+    @SqlQuery("SELECT d.*, k.key, dp.property_value, dp.property_key, dp.property_type, dp.schema_property, dp.property_id, ca.consent_id, c.translated_use_restriction, dar_ds_ids.id as in_use " +
             " FROM dataset d " +
             " LEFT JOIN (SELECT DISTINCT dataset_id AS id FROM dar_dataset) dar_ds_ids ON dar_ds_ids.id = d.dataset_id " +
             " LEFT JOIN dataset_property dp ON dp.dataset_id = d.dataset_id " +
             " LEFT JOIN dictionary k ON k.key_id = dp.property_key " +
-            " LEFT JOIN consentassociations ca ON ca.datasetid = d.dataset_id " +
-            " LEFT JOIN consents c ON c.consentid = ca.consentid " +
+            " LEFT JOIN consent_associations ca ON ca.dataset_id = d.dataset_id " +
+            " LEFT JOIN consents c ON c.consent_id = ca.consent_id " +
             " WHERE d.dac_id = :dacId " +
             " OR (dp.schema_property = 'dataAccessCommitteeId' AND dp.property_value = :dacId::text)")
     List<Dataset> findDatasetsAssociatedWithDac(@Bind("dacId") Integer dacId);
--- a/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
@@ -366,6 +366,25 @@ public interface DatasetDAO extends Transactional<DatasetDAO> {
     Set<DatasetDTO> findDatasetsByDac(@Bind("dacId") Integer dacId);
 
     /**
+     * Finds all datasets which are assigned to this DAC and which
+     * have been requested for this DAC.
+     *
+     * @param dacId id
+     * @return all datasets associated with DAC
+     */
+    @UseRowReducer(DatasetReducer.class)
+    @SqlQuery("SELECT d.*, k.key, dp.property_value, dp.property_key, dp.property_type, dp.schema_property, dp.property_id, ca.consentid, c.translateduserestriction, dar_ds_ids.id as in_use " +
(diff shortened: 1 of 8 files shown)


src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
```diff
@@ -365,6 +365,25 @@ Integer insertDataset(
...
+     * @return all datasets associated with DAC
+     */
+    @UseRowReducer(DatasetReducer.class)
+    @SqlQuery("SELECT d.*, k.key, dp.property_value, dp.property_key, dp.property_type, dp.schema_property, dp.property_id, ca.consent_id, c.translated_use_restriction, dar_ds_ids.id as in_use " +
+            " FROM dataset d " +
+            " LEFT JOIN (SELECT DISTINCT dataset_id AS id FROM dar_dataset) dar_ds_ids ON dar_ds_ids.id = d.dataset_id " +
+            " LEFT JOIN dataset_property dp ON dp.dataset_id = d.dataset_id " +
+            " LEFT JOIN dictionary k ON k.key_id = dp.property_key " +
+            " LEFT JOIN consent_associations ca ON ca.dataset_id = d.dataset_id " +
+            " LEFT JOIN consents c ON c.consent_id = ca.consent_id " +
```
rushtong, line 383: As of https://github.com/DataBiosphere/consent/pull/1724 (which is merged), I believe you shouldn't have to join on these tables any longer.

rushtong: Ah ... I see `ca.consent_id, c.translated_use_restriction` in the query so we do need those columns. Those two values are of less value now that we store the data use on the dataset. `translated_use_restriction` is actually quite incorrect. It represents the translation of the V1 algorithm. `consent_id` is also useless since its only function was to tie the dataset to the use restriction on a consent. I will file tickets for deprecating these.

quazi-h (approved):
Changes here look good to me.

src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
```diff
@@ -192,9 +193,10 @@ public Response removeDacChair(@Auth AuthUser authUser, @PathParam("dacId") Inte
     @Produces("application/json")
     @RolesAllowed({ADMIN, MEMBER, CHAIRPERSON})
     public Response findAllDacDatasets(@Auth AuthUser user, @PathParam("dacId") Integer dacId) {
-        findDacById(dacId);
-        Set<DatasetDTO> datasets = dacService.findDatasetsByDacId(user, dacId);
-        return Response.ok().entity(unmarshal(datasets)).build();
+        Dac dac = findDacById(dacId);
+        checkUserRoleInDac(dac, user);
+        List<Dataset> datasets = dacService.findDatasetsByDacId(dacId);
+        return Response.ok().entity(datasets).build();
```
rushtong, line 202: Line 9 has an unused import, `import java.util.Set;`

rushtong: And line 31 has one too: `import org.broadinstitute.consent.http.models.dto.DatasetDTO;`

src/main/java/org/broadinstitute/consent/http/service/DacService.java
```diff
@@ -157,16 +157,8 @@ public User findUserById(Integer id) throws IllegalArgumentException {
...
-        if (isAuthUserAdmin(authUser)) {
-            return datasets;
-        }
-        List<Integer> dacIds = getDacIdsForUser(authUser);
-        if (dacIds.contains(dacId)) {
-            return datasets;
-        }
-        return Collections.emptySet();
+    public List<Dataset> findDatasetsByDacId(Integer dacId) {
+        return dataSetDAO.findDatasetsAssociatedWithDac(dacId);
```
rushtong, line 160: Line 35 has an unused import: `import org.broadinstitute.consent.http.models.dto.DatasetDTO;`

src/test/java/org/broadinstitute/consent/http/db/DacDAOTest.java
```diff
@@ -268,4 +272,49 @@ public void testFindDacsForDatasetIds() {
...
         assertEquals(1, dacs.size());
     }
+
+    @Test
+    public void testFindDatasetsAssociatedWithDac_AssignedDacId() {
+        Dataset datasetNonDac = createDataset();
+        Dac dac = createDac();
+
+        List<Dataset> results = datasetDAO.findDatasetsAssociatedWithDac(dac.getDacId());
+        assertEquals(0, results.size());
```
rushtong: This block could be its own test to ensure that we're not finding unassigned datasets. Since this is also repeated in the next test, it makes even more sense to extract it. Each test should be focused on testing a single condition. Adding multiple conditions in a test case is generally an anti-pattern (not always, but usually).

src/test/java/org/broadinstitute/consent/http/db/DacDAOTest.java
```diff
@@ -268,4 +272,49 @@ public void testFindDacsForDatasetIds() {
...
+                1,
+                "dataAccessCommitteeId",
+                dac.getDacId().toString(),
+                DatasetPropertyType.Number,
+                Date.from(Instant.now()))));
+
+        results = datasetDAO.findDatasetsAssociatedWithDac(dac.getDacId());
+        assertEquals(2, results.size());
+        assertTrue(results.contains(datasetAssignedDac));
+        assertTrue(results.contains(dataset));
```
rushtong: This seems like a separate test to me. There are three tests here, no datasets found, 1 dataset found via dataset association, and 1 dataset found via property association.

rushtong (commented):
This looks really good! See minor comments inline.

src/test/java/org/broadinstitute/consent/http/db/DacDAOTest.java
connorlbark: very good point - have updated all 3 tests to be more specific

rushtong (approved):
👍🏽

PR #1734, opened 2022-09-06
[DUOS-2040][risk=no] Add DAA Acceptance to user properties

connorlbark:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2040

----

- Verify all tests go green

### DIFF
--- a/src/main/resources/assets/paths/user.yaml
+++ b/src/main/resources/assets/paths/user.yaml
@@ -53,6 +53,7 @@ put:
               description: The suggested signing official for the user
               type: string
             daaAcceptance:
+              description: Will be "true" if the user has accepted the Broad and NIH's Data Access Agreement.
               type: string
   responses:
     200:
--- a/src/main/resources/assets/paths/user.yaml
+++ b/src/main/resources/assets/paths/user.yaml
@@ -52,6 +52,8 @@ put:
             suggestedSigningOfficial:
               description: The suggested signing official for the user
               type: string
+            daaAcceptance:
+              type: string
   responses:
     200:
       description: The user, along with researcher properties and library cards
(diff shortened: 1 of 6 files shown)


quazi-h (approved):
Changes to add a new DAA property look great!

src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
```diff
@@ -111,7 +111,8 @@ public void testUpdateUserFieldsById() {
             fields.setSelectedSigningOfficialId(1);
             fields.setSuggestedSigningOfficial(RandomStringUtils.random(10, true, false));
             fields.setSuggestedInstitution(RandomStringUtils.random(10, true, false));
-            assertEquals(3, fields.buildUserProperties(user.getUserId()).size());
+            fields.setDaaAcceptance(true);
+            assertEquals(4, fields.buildUserProperties(user.getUserId()).size());
```
quazi-h, line 115: The ticket calls for UserService.updateUserFieldsById() and UserResource.updateSelf() to be updated but I see that both of those end eventually call userUpdateFields.buildUserProperties which his where daaAcceptance is set, so that makes sense. Adding a test to check for daaAcceptance in  UserResourceTest would be redundant, right?

src/main/resources/assets/paths/user.yaml
```diff
@@ -52,6 +52,8 @@ put:
             suggestedSigningOfficial:
               description: The suggested signing official for the user
               type: string
+            daaAcceptance:
+              type: string
```
JVThomas, line 57: Nit: Add a description to the docs as well. For context, `daaAcceptance` represents whether or not a user has accepted the Broad and NIH's Data Access Agreement.

JVThomas (approved):
One nitpick, but other than that the code looks good!

src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
JVThomas: Correct. There's no reason to re-test Service class logic in the Resource class test(s).

PR #1737, opened 2022-09-08
DUOS-2071[risk=no] Updated DAR draft update endpoint to set (rather than add) dataset ids when processing payload

JVThomas:
Addresses [DUOS-2071](https://broadworkbench.atlassian.net/browse/DUOS-2071)

PR addresses bug where dataset ids were added rather than set on intermediate DAR object on the DAR draft update endpoint. The result of the above made it so that datasets could never be removed from a draft. 

PR also creates a new `DataAccessRequestServiceDAO` class with an `updateByReferenceId` to ensure that DAR, Collection, and DarDataset records are all updated within a single transaction rather than three separate ones. 

With the above addition comes updates to method references as well as new/updated tests. PR also includes misc. cleanup of unused imports.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
@@ -64,7 +64,8 @@ public class DataAccessRequestResourceVersion2 extends Resource {
       EmailNotifierService emailNotifierService,
       GCSService gcsService,
       UserService userService,
-      MatchService matchService) {
+      MatchService matchService
+    ) {
     this.dataAccessRequestService = dataAccessRequestService;
     this.emailNotifierService = emailNotifierService;
     this.gcsService = gcsService;
@@ -239,7 +240,7 @@ public class DataAccessRequestResourceVersion2 extends Resource {
       // it in dar data.
       data.setReferenceId(originalDar.getReferenceId());
       originalDar.setData(data);
-      originalDar.addDatasetIds(data.getDatasetIds());
+      originalDar.setDatasetIds(data.getDatasetIds());
       DataAccessRequest updatedDar =
           dataAccessRequestService.updateByReferenceId(user, originalDar);
       return Response.ok().entity(updatedDar.convertToSimplifiedDar()).build();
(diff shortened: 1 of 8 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
```diff
@@ -239,7 +240,7 @@ public Response updatePartialDataAccessRequest(
       // it in dar data.
       data.setReferenceId(originalDar.getReferenceId());
       originalDar.setData(data);
-      originalDar.addDatasetIds(data.getDatasetIds());
+      originalDar.setDatasetIds(data.getDatasetIds());
```
rushtong, line 243: Nice!

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -451,20 +455,11 @@ public DataAccessRequest insertSubmittedDataAccessRequest(User user, String refe
...
-        }
-        // Update the dar_dataset collection
-        syncDataAccessRequestDatasets(dar.getDatasetIds(), dar.getReferenceId());
-
-        return findByReferenceId(dar.getReferenceId());
+      try {
+        return dataAccessRequestServiceDAO.updateByReferenceId(user, dar);
+      } catch(SQLException e) {
+        throw new IllegalArgumentException("Update failed due to invalid parameters");
+      }
```
rushtong: If we hit any sql exception, we should definitely log that here as an error.
You should be able to re-throw the sql exception - those should be handled nicely in the resource layer. We generally show those to clients as an internal server error.

rushtong (commented):
Looks great! I just have one suggestion for exception handling - see comments inline.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
JVThomas: Is there a clean way of re-throwing that doesn't involve having to add `throws SQLException` to functions that call `updateByReferenceId`?

rushtong: If you look at other usages in this pattern, we generally catch the sql error in the service method and do the logging there, so we could do that instead of logging it here. Then you could decide at the service layer what makes the most sense. However, no, if this method throws sql exceptions and passes them up the stack, then the stack has to deal with them.

rushtong (approved):
👍🏽

connorlbark (commented):
👍 tests are solid!

connorlbark (approved):

PR #1741, opened 2022-09-09
[DUOS-2082][risk=no] Update random string generation in tests

rushtong:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2082

Usages of `RandomStringUtils.random(10)` can generate strings that our database cannot store. Use `RandomStringUtils.randomAlphabetic(10)` instead which restricts the string values to the `[a-A,z-Z]` range.

----

- Verify all tests go green

### DIFF
--- a/src/test/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAOTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAOTest.java
@@ -44,7 +44,7 @@ public class DarCollectionSummaryDAOTest extends DAOTestHelper {
   }
 
   private User createUserForTest() {
-    Integer userId = userDAO.insertUser(RandomStringUtils.random(10), RandomStringUtils.randomAlphabetic(10), new Date());
+    Integer userId = userDAO.insertUser(RandomStringUtils.randomAlphabetic(10), RandomStringUtils.randomAlphabetic(10), new Date());
     return userDAO.findUserById(userId);
   }
 
--- a/src/test/java/org/broadinstitute/consent/http/db/ConsentDAOTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/db/ConsentDAOTest.java
@@ -129,7 +129,7 @@ public class ConsentDAOTest extends DAOTestHelper {
     public void testUpdateConsentTranslatedUseRestriction() {
         Consent consent = createConsent();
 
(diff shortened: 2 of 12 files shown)


src/test/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAOTest.java
```diff
@@ -44,7 +44,7 @@ private Institution createInstitution(Integer userId) {
   }
 
   private User createUserForTest() {
-    Integer userId = userDAO.insertUser(RandomStringUtils.random(10), RandomStringUtils.randomAlphabetic(10), new Date());
+    Integer userId = userDAO.insertUser(RandomStringUtils.randomAlphabetic(10), RandomStringUtils.randomAlphabetic(10), new Date());
```
rushtong, line 47: See example error case from this test method: https://github.com/DataBiosphere/consent/runs/8277876664?check_suite_focus=true

```
org.jdbi.v3.core.statement.UnableToExecuteStatementException: org.postgresql.util.PSQLException: 
ERROR: invalid byte sequence for encoding "UTF8": 0x00 
[statement:"/* UserDAO.insertUser */ 
    INSERT INTO users (email, display_name, create_date) values (:email, :displayName, :createDate)", 
    arguments:{positional:{0:콑鉲𓋩🏐𒔏瑲,1:kXMWMacZVu,2:Fri Sep 09 20:12:49 UTC 2022}, named:{displayName:kXMWMacZVu,email:콑鉲𓋩🏐𒔏瑲,createDate:Fri Sep 09 20:12:49 UTC 2022}, finder:[]}]
```

JVThomas (approved):
LGTM

quazi-h (approved):
Changes look good and make sense to me. Just to be thorough I searched the consent repo to see if there are other places we were using RandomStringUtils.random and it looks like theres a lot of usages in many other tests. I see that the scope of this ticket is to fix the flaky user tests so it makes sense to not address them in this PR but should we make a ticket to address all the other places we are using that method?

rushtong:
> Just to be thorough I searched the consent repo to see if there are other places we were using RandomStringUtils.random and it looks like theres a lot of usages in many other tests. I see that the scope of this ticket is to fix the flaky user tests so it makes sense to not address them in this PR but should we make a ticket to address all the other places we are using that method?

I did look through them and all of the other cases look harmless to me. I think we can deal with future cases of these as they come up.

PR #1744, opened 2022-09-13
[DUOS-2045][risk=no] New TDR endpoint to draft a DAR from aliases

quazi-h:
[DUOS-2045](https://broadworkbench.atlassian.net/browse/DUOS-2045)

- Added a new POST endpoint under TDRResource to create a draftDAR when given a list of dataset identifiers.
- Added a new TDRService method that takes a list of identifiers and returns the dataset ids.
- Added a new UserService method to ensure that the currently authenticated user is a valid DUOS and SAM user and adds them if they are not.
- Added the new endpoint and a new yaml config for it to api-docs.
- Added tests for the new endpoint and service methods.

**Testing**
- Test this with dataset identifiers (DUOS-XXXX). 
- Existing valid dataset identifiers in dev: DUOS-000003, DUOS-000004, DUOS-000005
- You can find more by going to the Researcher Console -> Dataset Catalog -> Dataset Id

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -64,4 +81,60 @@ public Response getDatasetByIdentifier(@Auth AuthUser authUser, @PathParam("iden
             return createExceptionResponse(e);
         }
     }
+
+    @POST
+    @Consumes("application/json")
+    @Produces("application/json")
+    @Path("/dar/draft")
+    @RolesAllowed({ADMIN})
```
rushtong: This should be `@PermitAll` for now.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -64,4 +81,60 @@ public Response getDatasetByIdentifier(@Auth AuthUser authUser, @PathParam("iden
...
+        @Context UriInfo info,
+        @QueryParam("identifiers") String identifiers,
+        @QueryParam("projectTitle") String projectTitle) {
+      try {
+        // Ensure that the user is a registered DUOS and SAM user
+        User user = findOrCreateUser(authUser);
+        if (Objects.isNull(identifiers) || identifiers.isBlank()) {
+          throw new BadRequestException("No dataset identifiers were provided");
+        } else {
+          List<Integer> datasetIds = Arrays.asList(identifiers.split(","))
```
rushtong: ~nit: naming. These are really datasetIdentifiers ... datasetId is the assigned database id for a datset.~
Misread this loop ... see below.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -64,4 +81,60 @@ public Response getDatasetByIdentifier(@Auth AuthUser authUser, @PathParam("iden
...
+              .stream()
+              .filter(identifier -> !identifier.isBlank())
+              .map(identifier -> this.datasetService.findDatasetByIdentifier(identifier).getDataSetId())
```
rushtong: I prefer not to put db calls in a loop like this. It's better to get all of the ids, then pass that list value in to a service method that can do this work more efficiently.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -64,4 +81,60 @@ public Response getDatasetByIdentifier(@Auth AuthUser authUser, @PathParam("iden
...
+              .collect(Collectors.toList());
+        if (datasetIds.isEmpty()) {
+            throw new BadRequestException("No datasets associated with identifiers: " + identifiers);
+        }
```
rushtong: There is an error case here that we need to somehow deal with. If the user passed in 3 identifiers, and we only found 2 of them, we should error out in some way with a useful error message.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -64,4 +81,60 @@ public Response getDatasetByIdentifier(@Auth AuthUser authUser, @PathParam("iden
...
+    User findOrCreateUser(AuthUser authUser) {
+        // Ensure that the user is a registered DUOS and SAM user
+        User tdrUser;
+        try {
+            tdrUser = this.userService.findUserByEmail(authUser.getEmail());
+        } catch (NotFoundException nfe) {
+            User newTdrUser = new User();
+            newTdrUser.setEmail(authUser.getEmail());
+            newTdrUser.setDisplayName(authUser.getName());
+            tdrUser = this.userService.createUser(newTdrUser);
```
rushtong: This method doesn't actually register them in Sam, that happens in `SamService.postRegistrationInfo(AuthUser authUser)`. You can see how we do this in `SamResource.postSelfTos(@Auth AuthUser authUser)`.

rushtong: Thinking more about this. It might be time to consolidate the create user/register in sam paths. Not for this PR, but for a cleanup story.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -64,4 +81,60 @@ public Response getDatasetByIdentifier(@Auth AuthUser authUser, @PathParam("iden
...
+        DataAccessRequest result = this.darService.insertDraftDataAccessRequest(user, newDar);
+        URI uri = info.getRequestUriBuilder().path("/" + result.getReferenceId()).build();
+        return Response.created(uri).entity(result.convertToSimplifiedDar()).build();
+        }
+      } catch (Exception e) {
+          return createExceptionResponse(e);
+      }
+    }
+
+    // should this be private? otherwise can't test
```
connorlbark: could make this a service method in TDRService ?

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -1,33 +1,50 @@
...
+
 
 import javax.annotation.security.PermitAll;
-import javax.ws.rs.GET;
-import javax.ws.rs.NotFoundException;
-import javax.ws.rs.Path;
-import javax.ws.rs.PathParam;
-import javax.ws.rs.Produces;
+import javax.annotation.security.RolesAllowed;
+import javax.ws.rs.*;
```
rushtong: Avoid * imports

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
rushtong: Agreed, this should probably be a service method. I think it's better handled in `UserService` - might need some more work there to make that happen. Sam utilities are in their own service, not a DAO, so making `UserService` dependent on `SamService` could introduce a circular dependency. Have not checked, but that is a potential issue. On the other hand, if we moved this to `SamService`, we could add `userDAO` there and avoid any dependency problems.

src/main/resources/assets/paths/tdrDraftDAR.yaml
```diff
@@ -0,0 +1,25 @@
...
+      required: false
+      schema:
+        type: string
+  tags:
+    - TDR
+  responses:
+    201:
+      description: Returns the created draft Data Access Request
+    400:
+      description: Bad request, no valid dataset identifiers were provided
```
rushtong: I think we could return a 404 error if there were any dataset identifiers that we could not find.

src/test/java/org/broadinstitute/consent/http/resources/TDRResourceTest.java
```diff
@@ -98,4 +115,51 @@ public void testGetDatasetByIdentifier404() {
...
+    public void testFindOrCreateUser() {
+        initResource();
+
+        User tdrUser = resource.findOrCreateUser(new AuthUser());
+        assertNotNull(tdrUser);
+
+        doThrow(new NotFoundException()).when(userService).findUserByEmail(any());
+        User newTdrUser = resource.findOrCreateUser(any());
+        assertNotNull(newTdrUser);
+    }
```
rushtong: As `resource.findOrCreateUser` currently stands, there are two service class methods you would need to mock out, `userService.findUserByEmail` and `userService.createUser`. You would also want to test out the two cases, one where the user is found, and one where the user isn't found. To do that, look at usages of mockito spy and verify in the current codebase.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
rushtong: nit: `this.` is generally unnecessary in these kinds of cases. See https://stackoverflow.com/questions/2411270/when-should-i-use-this-in-a-class

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
rushtong: Similar comment to above, avoid `this` unless necessary.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -64,4 +81,60 @@ public Response getDatasetByIdentifier(@Auth AuthUser authUser, @PathParam("iden
...
+        // Ensure that the user is a registered DUOS and SAM user
+        User user = findOrCreateUser(authUser);
+        if (Objects.isNull(identifiers) || identifiers.isBlank()) {
+          throw new BadRequestException("No dataset identifiers were provided");
+        } else {
+          List<Integer> datasetIds = Arrays.asList(identifiers.split(","))
+              .stream()
+              .filter(identifier -> !identifier.isBlank())
+              .map(identifier -> this.datasetService.findDatasetByIdentifier(identifier).getDataSetId())
+              .collect(Collectors.toList());
```
rushtong: `.collect(Collectors.toList())` -> `.toList()`

src/test/java/org/broadinstitute/consent/http/resources/TDRResourceTest.java
```diff
@@ -98,4 +115,51 @@ public void testGetDatasetByIdentifier404() {
         assertEquals(404, r.getStatus());
     }
 
+    @Test
+    public void testCreateDraftDataAccessRequest() {
+
+        String identifiers = "DUOS-00001, DUOS-00002";
```
rushtong, line 134: I'm not quite sure how this test is failing, but one thing I see is that you're splitting on the `,` but this string has an extra space after the comma, so you'd have two elements in your split array: `["DUOS-00001", " DUOS-00002"]` ... notice the extra space. That's not the source of the failure, but it is a potential bug. We should probably be trimming those values in the resource class.

src/test/java/org/broadinstitute/consent/http/resources/TDRResourceTest.java
```diff
@@ -98,4 +115,51 @@ public void testGetDatasetByIdentifier404() {
...
+
+        Dataset d1 = new Dataset();
+        d1.setDataSetId(1);
+
+        Dataset d2 = new Dataset();
+        d2.setDataSetId(2);
+
+        when(datasetService.findDatasetByIdentifier("DUOS-00001")).thenReturn(d1);
+        when(datasetService.findDatasetByIdentifier("DUOS-00002")).thenReturn(d2);
```
rushtong: You'll also need to mock out `darService.insertDraftDataAccessRequest`

rushtong: I was able to get this test to work with a few changes. 

1. Update the identifiers string to remove the extra space
2. Mock out `darService.insertDraftDataAccessRequest` to return a Dar with a darData
3. Mock out `uriInfo` since we're using that in the response. Need to do this twice, once for `path` and once for `build`.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
rushtong: The `asList` is unnecessary here, you can go straight to `Arrays.stream(<list items?)`

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
rushtong: Another option here is to introduce another layer between `SamService` and a new `SamDAO`. In effect, we would:

1. rename `SamService` to `SamDAO`
2. create a new `SamService` that has a `SamDAO`
3. re-implement the same methods in the new version of the service with passthroughs to the dao

That way, any service could then pull in a `SamDAO` and not have to worry about circular dependencies. I wish I had thought of this in the original implementation - would have made this kind of work a lot easier.

rushtong: See https://github.com/DataBiosphere/consent/pull/1745 for a potential solution

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
rushtong: Reflecting on this, it should be `@PermitAll` for sure. Any user can call this endpoint and we'll ensure that they are, by default, a researcher in DUOS if they don't already exist.

quazi-h: Updated to @PermitAll.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
quazi-h: My IntelliJ has been doing that automatically, I added the individual imports back in and updated my settings so it stops doing that.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
quazi-h: Refactored this a little - created a service method "getDatasetIdsByIdentifier" in TDRService that takes in a List<String>.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
quazi-h: Removed instances of "this." and passing a list of identifiers to a new service method.
Thank you for the link, that was actually very helpful in explaining when to use it.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
quazi-h: Updated the failure logic to check that the length of the identifier list is equal to the length of the datasetId list we get back from the service method.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
quazi-h: Merged changes from [1745](https://github.com/DataBiosphere/consent/pull/1745) and implemented a new UserService method to handle looking up and creating a user with the newly available SamDAO.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
quazi-h: Updated this method which I moved to UserService to call SamDAO.postRegistrationInfo. Does that endpoint actually create the SAM user, too? I didn't see any "create" methods in the Sam classes. Created a stub ticket for the cleanup work: [DUOS-2097](https://broadworkbench.atlassian.net/browse/DUOS-2097).

src/main/resources/assets/paths/tdrDraftDAR.yaml
quazi-h: Updated response to 404.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
rushtong: > Does that endpoint actually create the Sam user, too?

Yes, that sam endpoint will create a Sam user. A user in Sam is visible to all Terra services. More details available in Sam itself: https://sam.dsde-prod.broadinstitute.org/#/Users/createUserV2

The urls we use in Consent to send to Sam are in `ServicesConfiguration`. This is the one we use for the endpoint under discussion:

```
  public String postRegisterUserV2SelfUrl() {
    return getSamUrl() + "register/user/v2/self";
  }
```

src/test/java/org/broadinstitute/consent/http/resources/TDRResourceTest.java
quazi-h: Created a new UserService method to handle this. Added tests for both user found, user not found cases and using spy/verify to check. That was a great exercise for me to learn about those Mockito methods, thanks!

src/test/java/org/broadinstitute/consent/http/resources/TDRResourceTest.java
quazi-h: Got this test working and added a trim() when taking in the list of identifiers.

src/test/java/org/broadinstitute/consent/http/resources/TDRResourceTest.java
quazi-h: Mocking the darService and uriInfo helped get this test working.

connorlbark (approved):
👍 looks good to me

src/main/java/org/broadinstitute/consent/http/service/TDRService.java
```diff
@@ -15,11 +16,15 @@
 
 public class TDRService {
     private final DataAccessRequestService dataAccessRequestService;
+    private final DatasetService datasetService;
+    private final DatasetDAO datasetDAO;
     private final Logger logger = LoggerFactory.getLogger(this.getClass());
 
     @Inject
-    public TDRService(DataAccessRequestService dataAccessRequestService) {
+    public TDRService(DataAccessRequestService dataAccessRequestService, DatasetService datasetService, DatasetDAO datasetDAO) {
```
JVThomas: Remove the DatasetService dependency since you're no longer using it in the Service class.

src/main/java/org/broadinstitute/consent/http/service/TDRService.java
```diff
@@ -15,11 +16,15 @@
...
         this.dataAccessRequestService = dataAccessRequestService;
+        this.datasetService = datasetService;
```
JVThomas: Similar to the above comment, remove the local assignment.

src/main/java/org/broadinstitute/consent/http/service/TDRService.java
```diff
@@ -33,4 +38,18 @@ public ApprovedUsers getApprovedUsersForDataset(Dataset dataset) {
 
         return new ApprovedUsers(approvedUsers);
     }
+
+    public List<Integer> getDatasetIdsByIdentifier(List<String> identifiers) {
+        List<Integer> datasetIds = identifiers
+                .stream()
+                .filter(identifier -> !identifier.isBlank())
+                .map(identifier -> datasetDAO.findDatasetByAlias(Dataset.parseIdentifierToAlias(identifier)))
```
JVThomas: We want to avoid making individual DAO calls within a map. If a user provides, let say 30, identifiers, then we'd be making 30 individual calls to the database for every one user request.

What should happen is that you need to make a new DAO method that takes an ArrayList of identifiers as a parameter and perform a bulk fetch. This way, regardless of the number of identifiers, you'd only be making a single database call every time.

Look at `findVotesByElectionIds` in `VoteDAO.java` for an example on how to write up these types of DAO methods.

JVThomas (changes requested):
Code's looking good so far, left some comments for you to look over

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -64,4 +86,45 @@ public Response getDatasetByIdentifier(@Auth AuthUser authUser, @PathParam("iden
...
+          throw new BadRequestException("No dataset identifiers were provided");
+        } else {
+          List<String> identifierList = Arrays.stream(identifiers.split(","))
+                  .map(String::trim)
+                  .toList();
+          List<Integer> datasetIds = tdrService.getDatasetIdsByIdentifier(identifierList);
+        // Check that we were able to find a dataset id for all identifiers provided
+        if (identifierList.size() != datasetIds.size()) {
+            throw new BadRequestException("Invalid dataset identifiers were provided.");
+        }
```
rushtong: I think a more appropriate error here would be a not found exception for the identifiers that we could not find as opposed to a BRE. It would be nice to let the client know which identifiers were NOT found so they can rectify their request.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -64,4 +86,45 @@ public Response getDatasetByIdentifier(@Auth AuthUser authUser, @PathParam("iden
...
+    public Response createDraftDataAccessRequest(
+        @Auth AuthUser authUser,
+        @Context UriInfo info,
+        @QueryParam("identifiers") String identifiers,
+        @QueryParam("projectTitle") String projectTitle) {
+      try {
+        // Ensure that the user is a registered DUOS and SAM user
+        User user = userService.findOrCreateUser(authUser);
+        if (Objects.isNull(identifiers) || identifiers.isBlank()) {
+          throw new BadRequestException("No dataset identifiers were provided");
```
rushtong, line 104: Since we're checking on the validity of `identifiers`, we should also ensure that `projectTitle` isn't null or blank as well.

src/main/java/org/broadinstitute/consent/http/service/TDRService.java
```diff
@@ -33,4 +38,18 @@ public ApprovedUsers getApprovedUsersForDataset(Dataset dataset) {
...
+                // technically, it is possible to have two dataset identifiers which
+                // have the same alias but are not the same: e.g., DUOS-5 and DUOS-00005
+                .filter(d -> !identifiers.contains(d.getDatasetIdentifier()))
```
rushtong: In practice, this is filtering out the dataset that was found in the map step. I'm currently looking into why.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -391,4 +395,26 @@ private Boolean checkForValidInstitution(Integer institutionId) {
...
+            newUser.setDisplayName(authUser.getName());
+            user = createUser(newUser);
+        }
+        // Ensure that the user is a registered SAM user
+        try {
+            samDAO.postRegistrationInfo(authUser);
+        } catch (ConsentConflictException cce) {
+          // no-op in the case of conflicts.
+        } catch (Exception e) {
+          throw new NotFoundException();
```
rushtong: I don't think we should be swallowing the exception and re-throwing as a not-found here. There could be network errors, or server exceptions that are best left to the resource class to handle. This pattern was originally in a resource class to begin with (see `SamResource`) so that layer figure out what kind of error to return to the client based on the type of exception thrown.

rushtong (changes requested):
This is currently non-functional for me when I test with a valid dataset identifier, `DUOS-000593`, and `TDR Testing` as a project title. See other feedback inline - I'll dig into the current non-functionality.

src/main/java/org/broadinstitute/consent/http/service/TDRService.java
rushtong: The conditional is backwards here, we need to know that `identifiers` DOES contain the dataset identifier.
```
.filter(d -> identifiers.contains(d.getDatasetIdentifier()))
```

src/main/java/org/broadinstitute/consent/http/service/TDRService.java
```diff
@@ -33,4 +38,18 @@ public ApprovedUsers getApprovedUsersForDataset(Dataset dataset) {
...
+                .map(d -> d.getDataSetId())
```
rushtong:
```suggestion
                .map(Dataset::getDataSetId)
```

src/main/java/org/broadinstitute/consent/http/service/TDRService.java
```diff
@@ -33,4 +38,18 @@ public ApprovedUsers getApprovedUsersForDataset(Dataset dataset) {
 
         return new ApprovedUsers(approvedUsers);
     }
+
+    public List<Integer> getDatasetIdsByIdentifier(List<String> identifiers) {
+        List<Integer> datasetIds = identifiers
```
rushtong: You can inline this too:
```suggestion
        return identifiers
```

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -64,4 +86,45 @@ public Response getDatasetByIdentifier(@Auth AuthUser authUser, @PathParam("iden
...
+        DataAccessRequest newDar = new DataAccessRequest();
+        DataAccessRequestData data = new DataAccessRequestData();
+        String referenceId = UUID.randomUUID().toString();
+        newDar.setReferenceId(referenceId);
+        data.setReferenceId(referenceId);
+        data.setProjectTitle(projectTitle);
+        newDar.setData(data);
+        newDar.setDatasetIds(datasetIds);
+        DataAccessRequest result = darService.insertDraftDataAccessRequest(user, newDar);
+        URI uri = info.getRequestUriBuilder().path("/" + result.getReferenceId()).build();
```
rushtong: The URI here is a little confusing, but we need one to return a `created` response object. This builds one based on the current URL that the client used, which includes url parameters. This is used in the response headers like so:

```
 access-control-allow-credentials: true 
 access-control-allow-headers: authorization,content-type,accept,origin,x-app-id 
 access-control-allow-methods: GET,POST,PUT,PATCH,DELETE,OPTIONS,HEAD 
 access-control-allow-origin: * 
 access-control-max-age: 1728000 
 connection: Keep-Alive 
 content-length: 246 
 content-type: application/json 
 date: Mon,26 Sep 2022 18:21:47 GMT 
 keep-alive: timeout=5,max=100 
 location: https://local.broadinstitute.org:27443/api/tdr/dar/draft/6068d46c-2e66-4102-b8af-91a4ccc24643?identifiers=DUOS-000593%2C%20DUOS-000594&projectTitle=TDR%20Testing%204 
 server: Apache 
 strict-transport-security: max-age=86400; includeSubDomains 
 vary: Origin 
 x-content-type-options: nosniff 
 x-frame-options: SAMEORIGIN
```

Note the `location` header. In principle, what this should return is the actual DAR url. There should be some standardized code that we can write to do this sort of thing, but that's way outside the scope of this PR. Instead, we should manually reconstruct the correct dar url like so:

```
URI uri = info.getRequestUriBuilder().replacePath("").path("api/dar/v2/" + result.getReferenceId()).replaceQuery(null).build()
```

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
quazi-h: Added conditional logic to check that projectTitle is not null and not blank before attempting to set it for the DAR data. No error thrown because it is an optional argument.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
quazi-h: Refactored the endpoint logic to produce a list of the identifiers that did not match to a dataset when the number of datasets do not match the number of identifiers provided. Also added a specific test case for the NFE.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
quazi-h: Updated the URI being built here to return the url for the actual DAR being created instead of the URL of this endpoint being called.

src/main/java/org/broadinstitute/consent/http/service/TDRService.java
quazi-h: Removed local DatasetService references.

src/main/java/org/broadinstitute/consent/http/service/TDRService.java
quazi-h: Removed local DatasetService references.

src/main/java/org/broadinstitute/consent/http/service/TDRService.java
quazi-h: Added a new DatasetDAO method that takes in a list of identifiers and returns a list of Datasets.

src/main/java/org/broadinstitute/consent/http/service/TDRService.java
quazi-h: Flipped the conditional logic here to ensure we are checking that the identifiers contains the dataset identifier.

src/main/java/org/broadinstitute/consent/http/service/TDRService.java
quazi-h: Refactored all new map calls to this syntax.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
quazi-h: Updated this logic to catch and log the exception instead of throwing a NotFoundException.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -64,4 +86,64 @@ public Response getDatasetByIdentifier(@Auth AuthUser authUser, @PathParam("iden
...
+        newDar.setReferenceId(referenceId);
+        data.setReferenceId(referenceId);
+        if (!Objects.isNull(projectTitle) && !projectTitle.isBlank()) {
+          data.setProjectTitle(projectTitle);
+        }
+        newDar.setData(data);
+        newDar.setDatasetIds(datasetIds);
+        DataAccessRequest result = darService.insertDraftDataAccessRequest(user, newDar);
+        // URI should return the new DAR url
+        URI uri = info.getRequestUriBuilder().replacePath("").replacePath("api/dar/v2/" + result.getReferenceId()).replaceQuery(null).build();
```
rushtong: I was looking into this some more, I think an easier way to do this is to use the `UriBuilder.fromPath()` method instead:
```suggestion
        URI uri = UriBuilder.fromPath("api/dar/v2/" + result.getReferenceId()).build();
```
This also means you don't need the `@Context UriInfo info` in the method signature.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -391,4 +395,26 @@ private Boolean checkForValidInstitution(Integer institutionId) {
...
+            user = createUser(newUser);
+        }
+        // Ensure that the user is a registered SAM user
+        try {
+            samDAO.postRegistrationInfo(authUser);
+        } catch (ConsentConflictException cce) {
+          // no-op in the case of conflicts.
+        } catch (Exception e) {
+            logger.error("Error posting user registration information to Sam for user " + authUser.getName() + ". Error: " + e.getMessage());
+        }
```
rushtong: We're still swallowing the exception when it would be better handled in the resource class. I would suggest we let all other exceptions bubble up here.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -64,4 +86,64 @@ public Response getDatasetByIdentifier(@Auth AuthUser authUser, @PathParam("iden
...
+                  .toList();
+        // Check that we were able to find a dataset id for all identifiers provided
+        if (identifierList.size() != datasets.size()) {
+            // isolate a list of identifiers that were NOT found
+            List<Integer> notFoundIdentifiers = identifierList
+                    .stream()
+                    .filter(identifier -> !datasetAliases.contains(identifier))
+                    .toList();
+            // throw a NFE to let the client know which identifiers were NOT found so they can rectify their request
+            throw new NotFoundException("Invalid dataset identifiers were provided: " + notFoundIdentifiers);
```
rushtong: When I try this with the following dataset, `DUOS-01234`, I get the following response:
```
{
  "message": "Invalid dataset identifiers were provided: [1234]",
  "code": 404
}
```
It looks like you're using the trimmed version of the alias instead of the values provided by the user for the error message.

rushtong: I found another edge case here. When I used the following: `DUOS-000594, DUOS-000594` (note that they're the same and valid identifiers), I got a confusing error message:
```
{
  "message": "Invalid dataset identifiers were provided: []",
  "code": 404
}
```
I'm split on the right response for this case. On one hand, it's clearly a bad request. However, it's a harmless bad request and one we can easily fix. I would suggest we call distinct the input list, then do the dataset lookups.

rushtong (changes requested):
See comments inline, I think we're making great progress here. Some new error cases I hadn't considered and some cleanup.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -64,4 +87,68 @@ public Response getDatasetByIdentifier(@Auth AuthUser authUser, @PathParam("iden
...
+          List<String> identifierList = Arrays.stream(identifiers.split(","))
+                  .map(String::trim)
+                  .filter(identifier -> !identifier.isBlank())
+                  // this will filter duplicate identifier strings, ex. "DUOS-000594, DUOS-000594"
+                  .distinct()
+                  .toList();
+          List<Integer> aliasList = identifierList
+                  .stream()
+                  .map(Dataset::parseIdentifierToAlias)
+                  // this will filter duplicate aliases, ex. "DUOS-00593, DUOS-000593"
```
rushtong: minor nit, the aliases are the trimmed integer values
```suggestion
                  // this will filter duplicate aliases, ex. "593, 593"
```

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -64,4 +87,68 @@ public Response getDatasetByIdentifier(@Auth AuthUser authUser, @PathParam("iden
...
+                  .distinct()
+                  .toList();
+          List<Dataset> datasets = tdrService.getDatasetsByIdentifier(aliasList);
+          List<Integer> datasetAliases = datasets.stream().map(Dataset::getAlias).toList();
+        // Check that we were able to find a dataset id for all identifiers provided
+        if (aliasList.size() != datasets.size()) {
```
rushtong: minor spacing nit, these two lines are not indented at the same level as the previous lines.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
rushtong: This looks like it's still an issue after the latest commit.

rushtong:
I'm seeing an issue with processing draft summaries. [On line 145](https://github.com/DataBiosphere/consent/blob/develop/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java#L145), we look for the create date in the data object instead of on the dar itself. This throws an error when processing it so the dar we create here never shows up in the user's list of dar summaries. We should add the create date to the data object in this PR just to get past that, but I'll file a ticket for updating the summary processing to use a real create date instead of the one on data which is a deprecated field.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -64,4 +87,68 @@ public Response getDatasetByIdentifier(@Auth AuthUser authUser, @PathParam("iden
...
+        if (!Objects.isNull(projectTitle) && !projectTitle.isBlank()) {
+          data.setProjectTitle(projectTitle);
+        }
+        newDar.setData(data);
+        newDar.setDatasetIds(datasetIds);
+        DataAccessRequest result = darService.insertDraftDataAccessRequest(user, newDar);
+        // URI should return the new DAR url
+        URI uri = info.getRequestUriBuilder().replacePath("").replacePath("api/dar/v2/" + result.getReferenceId()).replaceQuery(null).build();
+        return Response.created(uri).entity(result.convertToSimplifiedDar()).build();
+        }
```
rushtong: nit: there are some alignment issues here.

rushtong (commented):
Functionally, this is working well 👍🏽 
There are some minor nits listed inline - please address as you see fit.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
quazi-h: Updated the not found exception logic to filter down the list of identifier strings instead of integer aliases. I considered using Dataset.parseAliasToIdentifier to rebuild the identifiers from the list of aliases but that would make it possible to return a not found identifier that does not match to the original identifier provided if the one provided does not follow the exact alias format (padded with 6 zeros). That includes the example mentioned in your comment (DUOS-01234).

Also updated the identifier logic to call distinct on the input list as well as the list of alias integers after the strings have been processed. This handles two edge cases: when multiples of the same identifier are provided, and when variations of the identifier string that resolve to the same alias are provided:
examples:
"DUOS-000594, DUOS-000594"
"DUOS-00593, DUOS-000593"
Both of these cases were returning the 404 error with no actual invalid identifiers returned. With the changes both of those inputs would be accepted.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
quazi-h: Updated this so that we do not catch this exception and instead let this method throw an exception so it is thrown within the resource class.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
quazi-h: Updating the URI to the suggested logic returns the correct sub path but not the main part of the URL (at least according to my unit tests when debugging).
With current changes, URI is:
`https://test.domain.org/api/dar/v2/<reference_id>`
With the suggested changes, URI would be:
`api/dar/v2/<reference_id>`
Is that intended?

rushtong (approved):

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
quazi-h: Refactored the URI with the suggested change.

rushtong (approved):
👍🏽

JVThomas (approved):
Code looks good to me

PR #1745, opened 2022-09-14
[DUOS-2095][risk=no] Refactor sam service into service/dao

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2095

This PR separates Sam into a service and a lower level DAO class. The DAO makes http requests to Sam, the service is essentially passthroughs. This allows for other services to import the DAO without intermixing service layer classes.

----

- Verify all tests go green

### DIFF
--- a/src/test/java/org/broadinstitute/consent/http/service/dao/SamDAOTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/service/dao/SamDAOTest.java
@@ -1,4 +1,4 @@
-package org.broadinstitute.consent.http.service;
+package org.broadinstitute.consent.http.service.dao;
 
 import com.google.api.client.http.HttpStatusCodes;
 import com.google.gson.Gson;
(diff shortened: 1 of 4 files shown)


quazi-h (approved):
The refactor looks good, I think this will help with the issue you mentioned in my open PR - avoiding circular references between sam and user services.

src/test/java/org/broadinstitute/consent/http/service/dao/SamDAOTest.java
```diff
@@ -39,9 +39,9 @@
 import static org.mockserver.model.HttpRequest.request;
 import static org.mockserver.model.HttpResponse.response;
 
-public class SamServiceTest implements WithMockServer {
+public class SamDAOTest implements WithMockServer {
 
-  private SamService service;
+  private SamDAO service;
```
connorlbark: since it's blocking quazi, we can merge it as is, but nit: rename this to `samDao`? perhaps quazi you could do this in your ticket.

connorlbark (approved):
👍

src/test/java/org/broadinstitute/consent/http/service/dao/SamDAOTest.java
rushtong: Just getting to this, sorry. Merging after tests pass 👍🏽

PR #1750, opened 2022-09-16
[DUOS-2055][risk=no] Add ITDirector role

connorlbark:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2055

Changed all of the role endpoints so that we can add ITDirectors. Based upon the language of the ticket, I wasn't sure if this was all I need to do or if I need to change other endpoints, but at the very least the ITDirector role now exists and is able to be assigned to users.

----

- Verify all tests go green

### DIFF
--- a/src/main/resources/changesets/changelog-consent-100.0.xml
+++ b/src/main/resources/changesets/changelog-consent-100.0.xml
@@ -1,6 +1,6 @@
 <databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
          http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
-  <changeSet id="99.0" author="connorlbark">
+  <changeSet id="99.0" author="connorlbari=k">
       <insert tableName="roles">
           <column name="roleid" value="9"/>
           <column name="name" value="ITDirector"/>
--- a/src/main/resources/changesets/changelog-consent-100.0.xml
+++ b/src/main/resources/changesets/changelog-consent-100.0.xml
@@ -1,6 +1,6 @@
 <databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
          http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
-  <changeSet id="99.0" author="connorlbari=k">
+  <changeSet id="99.0" author="connorlbark">
       <insert tableName="roles">
           <column name="roleid" value="9"/>
           <column name="name" value="ITDirector"/>
--- a/src/main/resources/changesets/changelog-consent-100.0.xml
+++ b/src/main/resources/changesets/changelog-consent-100.0.xml
@@ -1,0 +1,9 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+  <changeSet id="99.0" author="connorlbark">
+      <insert tableName="roles">
+          <column name="roleid" value="9"/>
+          <column name="name" value="ITDirector"/>
+      </insert>
+  </changeSet>
+</databaseChangeLog>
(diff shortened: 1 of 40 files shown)


quazi-h (approved):
Changes look good and are working correctly. I see IT Director listed under the two main role endpoints. I am able to add the ITDirector role to a user as well as delete the role using those existing PUT/DELETE endpoints.

I'm not sure if it's in the scope of this ticket, but do we need to update the DarCollection endpoints that use {roleName} as a parameter? Most of them are GET endpoints that filter by role:
DarCollectionResource
	getCollectionsForUserByRole
	getCollectionSummariesForUserByRole
	getCollectionSummaryForRoleById
	queryCollectionsByFiltersAndUserRoles
	getCollectionsByToken	
IT Director is not available in the drop down menus for selecting a role for those endpoints.

Also, I was trying to test the updateSelf and update endpoints under UserResource and was able to remove a ITDirector role but have been unable to add one but I think I'm just doing it incorrectly.

JVThomas:
The ticket states the following
`Update Roles, and the various PUT role apis that follows the same restrictions as Signing Official. In future stories, we will restrict an IT Director user from modifying their institution in the UI, similarly to what we do for SOs.`

In the case of an SO or IT Director, user's should not be able to update their own institutions once the role has been assigned to them. In previous endpoints this was enforced, however it seems that the current `updateSelf` endpoint does not, so this functionality needs to be added back in.

The ticket also introduces a new functionality to the user update flow. Previously only Admins were allowed to add and remove roles, but now we're extending it to SO's, but in a very limited sense. SO's should now be able to add or remove the SO or IT Director role for any user within their current institution.

Regarding this line...
`SOs can access this API to add remove other SO and IT Director roles from their institution. On add, we add the SO’s institution to the user`

This line basically means that SOs can assign SO and IT Director roles to users that DO NOT have an institution assigned. This allows SOs to be proactive and bring in newly registered SOs and IT directors into their institution by simply assigning roles.

In the case of removing the roles, you do not need to remove the institution, just the role. This will allow us to accommodate a scenario where a user has the role removed but is still in the institution as well as a scenario where the user has left the institution.

This whole spec is a lot, which is why I recommend splitting the tasks into separate tickets. This PR should cover the role addition ~~by Admins and SO~~ to the table and ensure that users cannot change their own institutions if they are an IT Director or SO. The rest of the functionality should be tackled in a different ticket (I'll spec it out and write up a ticket myself).

connorlbark:
ok!! so i changed the update self endpoint to only let you change your institution id if it's not set already if you're an SO or an ITD

src/main/resources/changesets/changelog-consent-100.0.xml
```diff
@@ -0,0 +1,9 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+  <changeSet id="99.0" author="connorlbark">
```
JVThomas: Update the id to 100.0 and add the file to the changelog-master

src/main/resources/changesets/changelog-consent-100.0.xml
```diff
@@ -0,0 +1,9 @@
...
+      <insert tableName="roles">
+          <column name="roleid" value="9"/>
```
JVThomas: Normally you shouldn't have to include the `roleid` on an insert since it's supposed to be an auto-incrementing value. However I see that there have been earlier changelog files that manually added the id for older roles, and as a result the auto-increment never updated properly (its at 7 when it should be at 9).

We can fix this behavior by using an sql script to reset the auto-increment value to 9. So right before the insert tag, add this...

```
<sql>
    ALTER SEQUENCE roles_roleid_seq RESTART WITH 9
</sql>
 ```
 
This will reset the auto-increment tag to 9. Then, in your insert tag, remove the id assignment and the auto-increment should work as expected.
 
 NOTE: Generally the auto-increment value is saved internally in PostgreSQL in the form `{tablename}_{incrementing column name}_seq`, but you can always see for yourself in Postico by looking at the Structure tab of a table and looking at the `Sequence` column.
 
Screen Shot 2022-09-28 at 7 48 14 AM

JVThomas (changes requested):
Functionally the code is there, there's just some hiccups with the changelog and database that needs to be addressed.

src/main/resources/changesets/changelog-consent-100.0.xml
connorlbark: ahhhhh, this is a very good catch, much much cleaner this way. thanks!!

connorlbark:
I also added a test which makes sure all roles in UserRoles can be added to a user (making sure that each role actually exists in the Roles db). This way, there's some way to catch that even if the role isn't added in any other tests.

JVThomas:
> I also added a test which makes sure all roles in UserRoles can be added to a user (making sure that each role actually exists in the Roles db). This way, there's some way to catch that even if the role isn't added in any other tests.

Good thinking. It's a simple test that ensures that the roles saved in the database line up with the roles declared in the `UserRoles` enum

JVThomas (approved):
Code looks good to me, though I suggest getting a re-review from @quazi-broad or anyone else for an updated second thumb

quazi-h (approved):
New changes and tests look good to me!

PR #1751, opened 2022-09-19
[DUOS-2054][risk=no] Update Institution Fields

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2054

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/mapper/InstitutionMapper.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/mapper/InstitutionMapper.java
@@ -52,7 +52,8 @@ public class InstitutionMapper implements RowMapper<Institution>, RowMapperHelpe
       institution.setVerificationFilename(resultSet.getString("verification_filename"));
     }
     if (hasColumn(resultSet, "organization_type")) {
-      institution.setOrganizationType(OrganizationType.getOrganizationTypeFromString(resultSet.getString("organization_type")));
+      OrganizationType type = OrganizationType.getOrganizationTypeFromString(resultSet.getString("organization_type"));
+      institution.setOrganizationType(type);
     }
     if (hasColumn(resultSet, "create_user")) {
       institution.setCreateUserId(resultSet.getInt("create_user"));
--- a/src/main/java/org/broadinstitute/consent/http/db/mapper/InstitutionMapper.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/mapper/InstitutionMapper.java
@@ -1,5 +1,6 @@
 package org.broadinstitute.consent.http.db.mapper;
 
+import org.broadinstitute.consent.http.enumeration.OrganizationType;
 import org.broadinstitute.consent.http.models.Institution;
 import org.jdbi.v3.core.mapper.RowMapper;
 import org.jdbi.v3.core.statement.StatementContext;
@@ -34,6 +35,24 @@ public class InstitutionMapper implements RowMapper<Institution>, RowMapperHelpe
     }
     if (hasColumn(resultSet, "it_director_email")) {
       institution.setItDirectorEmail(resultSet.getString("it_director_email"));
+    }
+    if (hasColumn(resultSet, "institution_url")) {
+      institution.setInstitutionUrl(resultSet.getString("institution_url"));
+    }
+    if (hasColumn(resultSet, "duns_number")) {
+      institution.setDunsNumber(resultSet.getInt("duns_number"));
+    }
+    if (hasColumn(resultSet, "org_chart_url")) {
+      institution.setOrgChartUrl(resultSet.getString("org_chart_url"));
+    }
+    if (hasColumn(resultSet, "verification_url")) {
+      institution.setVerificationUrl(resultSet.getString("verification_url"));
+    }
+    if (hasColumn(resultSet, "verification_filename")) {
+      institution.setVerificationFilename(resultSet.getString("verification_filename"));
+    }
+    if (hasColumn(resultSet, "organization_type")) {
+      institution.setOrganizationType(OrganizationType.getOrganizationTypeFromString(resultSet.getString("organization_type")));
     }
     if (hasColumn(resultSet, "create_user")) {
       institution.setCreateUserId(resultSet.getInt("create_user"));
--- a/src/main/java/org/broadinstitute/consent/http/enumeration/OrganizationType.java
+++ b/src/main/java/org/broadinstitute/consent/http/enumeration/OrganizationType.java
@@ -1,0 +1,28 @@
+package org.broadinstitute.consent.http.enumeration;
+
+import liquibase.hub.model.Organization;
+
+public enum OrganizationType {
+    FOR_PROFIT("for_profit"),
+    NON_PROFIT("non_profit");
+
+    final String value;
+
+    OrganizationType(String value) {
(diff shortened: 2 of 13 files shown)


src/main/resources/changesets/changelog-consent-99.0.xml
```diff
@@ -0,0 +1,22 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <changeSet id="99.0" author="connorlbari=k">
+        <addColumn tableName="institution">
+            <column name="institution_url" type="text"/>
+            <column name="duns_number" type="integer" defaultValue="null"/>
```
connorlbark, line 7: assumption: duns_number is an integer - can change if this gets clarified to be something different.

src/main/resources/assets/schemas/Institution.yaml
```diff
@@ -6,6 +6,31 @@ properties:
...
+    description: Name of IT Director
+  itDirectorEmail:
+    type: string
+    description: Email of IT Director
+  institutionUrl:
+    type: string
+    description: Link to main page of institution
+  dunsNumber:
+    type: string
+    description: '?'
```
connorlbark: unsure what this should be

src/main/resources/changesets/changelog-consent-99.0.xml
connorlbark: brief research indicates it is a simple 9-digit number. this integer field should work

src/main/java/org/broadinstitute/consent/http/resources/Resource.java
```diff
@@ -121,6 +122,8 @@ private interface ExceptionHandler {
                 Response.status(Response.Status.BAD_REQUEST).type(MediaType.APPLICATION_JSON).entity(new Error(e.getMessage(), Response.Status.BAD_REQUEST.getStatusCode())).build());
         dispatch.put(BadRequestException.class, e ->
                 Response.status(Response.Status.BAD_REQUEST).type(MediaType.APPLICATION_JSON).entity(new Error(e.getMessage(), Response.Status.BAD_REQUEST.getStatusCode())).build());
+        dispatch.put(MalformedJsonException.class, e ->
+                Response.status(Response.Status.BAD_REQUEST).type(MediaType.APPLICATION_JSON).entity(new Error(e.getMessage(), Response.Status.BAD_REQUEST.getStatusCode())).build());
```
connorlbark, line 126: got this exception, and it sent back as a 500 - added this here since it seems to be an indicator that the user sent bad Json.

src/main/java/org/broadinstitute/consent/http/db/mapper/InstitutionMapper.java
```diff
@@ -1,5 +1,6 @@
 package org.broadinstitute.consent.http.db.mapper;
```
JVThomas, line 1: There may be other mappers/reducers that need to be updated outside of `InstitutionMapper`. One example is `InstitutionWithUsersReducer`

src/main/java/org/broadinstitute/consent/http/db/mapper/InstitutionMapper.java
```diff
@@ -35,6 +36,24 @@ public Institution map(ResultSet resultSet, StatementContext statementContext) t
...
+      institution.setOrgChartUrl(resultSet.getString("org_chart_url"));
+    }
+    if (hasColumn(resultSet, "verification_url")) {
+      institution.setVerificationUrl(resultSet.getString("verification_url"));
+    }
+    if (hasColumn(resultSet, "verification_filename")) {
+      institution.setVerificationFilename(resultSet.getString("verification_filename"));
+    }
+    if (hasColumn(resultSet, "organization_type")) {
+      institution.setOrganizationType(OrganizationType.getOrganizationTypeFromString(resultSet.getString("organization_type")));
```
JVThomas: Nit: Split this line up to make it a bit more readable.

src/main/java/org/broadinstitute/consent/http/enumeration/OrganizationType.java
```diff
@@ -0,0 +1,28 @@
+package org.broadinstitute.consent.http.enumeration;
+
+import liquibase.hub.model.Organization;
+
+public enum OrganizationType {
+    FOR_PROFIT("for_profit"),
```
JVThomas: The String values should be human readable (For-profit, nonprofit) so that the front end does not have to do any additional string manipulation should it have to render it on the UI.

src/main/resources/assets/schemas/Institution.yaml
```diff
@@ -6,6 +6,31 @@ properties:
...
+    description: Link to institution's staff directory/chart
+  verificationUrl:
+    type: string
+    description: Link to GCP Bucket of verification file
+  verificationFilename:
+    type: string
+    description: Name of verification file
+  organizationType:
+    type: string
+    enum: [ for_profit, non_profit ]
```
JVThomas: Similar comment to the enum version above

src/main/resources/changesets/changelog-consent-99.0.xml
```diff
@@ -0,0 +1,22 @@
...
+            <column name="org_chart_url" type="text"/>
+            <column name="verification_url" type="text"/>
+            <column name="verification_filename" type="varchar(255)"/>
+            <column name="organization_type" type="text"/>
+        </addColumn>
+        <rollback>
```
JVThomas: Liquibase supports auto-rollback for `addColumn`. You don't need to manually define the rollback here.

https://docs.liquibase.com/workflows/liquibase-community/liquibase-auto-rollback.html

JVThomas (changes requested):
Code's looking good, left some comments that need to be addressed.

src/main/java/org/broadinstitute/consent/http/db/mapper/InstitutionMapper.java
connorlbark: I think this is the only place it needs to be? I could be wrong, but from my understanding, the mapper is called before any reducers, and the InstitutionWithUsersReducer only adds user-related information, since the mapper already caught the institution's fields.

src/main/resources/assets/schemas/Institution.yaml
JVThomas: This can be something as simple as '"9-digit identification number provided by Duns & Bradstreet"

JVThomas (approved):
Added a comment on what to add for the dunsNumber description. Other than that the code looks good!

src/main/resources/assets/schemas/Institution.yaml
quazi-h: Yeah, I think keeping it simple would be good "A unique nine-digit
identification number for businesses provided by Dun & Bradstreet."
Here's D&B's own explanation of it:
https://www.dnb.com/ca-en/duns-number.html

quazi-h (approved):
Changes look good, should be all set once you update the DUNS number description.

PR #1752, opened 2022-09-20
[DUOS-2092][risk=no] Return Library Cards + Institution on Find All Users Endpoint

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2092

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
@@ -155,7 +155,7 @@ public interface UserDAO extends Transactional<UserDAO> {
                     " FROM users u " +
                     " LEFT JOIN user_role ur ON ur.user_id = u.user_id " +
                     " LEFT JOIN roles r ON r.roleid = ur.role_id " +
-                    " LEFT JOIN library_card lc ON lc.user_id = u.user_id AND lc.institution_id = :institutionId " +
+                    " LEFT JOIN library_card lc ON lc.user_id = u.user_id " +
                     " LEFT JOIN institution lci ON lc.institution_id = lci.institution_id" +
                     " LEFT JOIN institution i ON u.institution_id = i.institution_id")
     Set<User> findUsersWithLCsAndInstitution();
--- a/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
@@ -134,6 +134,32 @@ public interface UserDAO extends Transactional<UserDAO> {
         + " ORDER BY u.create_date DESC ")
     Set<User> findUsers();
 
+    @RegisterBeanMapper(value = User.class, prefix = "u")
+    @RegisterBeanMapper(value = UserRole.class)
+    @RegisterBeanMapper(value = LibraryCard.class, prefix = "lc")
+    @RegisterBeanMapper(value = Institution.class, prefix = "lci")
+    @RegisterBeanMapper(value = Institution.class, prefix = "i")
+    @UseRowReducer(UserWithRolesReducer.class)
+    @SqlQuery(
+            //This will pull in users tied to the institution
+            //Users will come with LCs issued by SOs institution (if any)
+            " SELECT " +
+                    User.QUERY_FIELDS_WITH_U_PREFIX + QUERY_FIELD_SEPARATOR +
+                    " r.name, ur.role_id, ur.user_role_id, ur.dac_id, ur.user_id, " +
+                    " lc.id AS lc_id , lc.user_id AS lc_user_id, lc.institution_id AS lc_institution_id, " +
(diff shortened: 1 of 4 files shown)


quazi-h (approved):
Code looks good to me and when testing locally I do see "institutionId", "institution", and "libraryCards" in the response.

src/test/java/org/broadinstitute/consent/http/db/UserDAOTest.java
```diff
@@ -198,6 +199,22 @@ public void testFindUsers_noArgs() {
         assertFalse(users.get(0).getRoles().isEmpty());
     }
 
+    @Test
+    public void testFindUsersWithLCsAndInstitution() {
+        User user = createUserWithInstitution();
+        Integer lcId = libraryCardDAO.insertLibraryCard(user.getUserId(), user.getInstitutionId(), "asdf", user.getDisplayName(), user.getEmail(), user.getUserId(), new Date());
```
JVThomas: For this test can you create a second user with a Library Card assigned to them as well? 

Right now the test checks against a database where there is only one user and one Library Card. Assuming I know nothing about the code and am only looking at the test I can say that either

1. The code is working as expected
2. The code is providing a false positive and is actually fetching all LCs. 

By adding unassociated users and unassociated LCs you can test that assumption 2 isn't happening.

JVThomas (changes requested):
Code looks good. I left one comment on your test but after that it'll be good to go.

src/test/java/org/broadinstitute/consent/http/db/UserDAOTest.java
connorlbark: good point! i have made a test which creates two users now.

src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
```diff
@@ -134,6 +134,32 @@ void updateUser(@Bind("displayName") String displayName,
...
+            //This will pull in users tied to the institution
+            //Users will come with LCs issued by SOs institution (if any)
+            " SELECT DISTINCT " +
+                    User.QUERY_FIELDS_WITH_U_PREFIX + QUERY_FIELD_SEPARATOR +
+                    " r.name, ur.role_id, ur.user_role_id, ur.dac_id, ur.user_id, " +
+                    " lc.id AS lc_id , lc.user_id AS lc_user_id, lc.institution_id AS lc_institution_id, " +
+                    " lc.era_commons_id AS lc_era_commons_id, lc.user_name AS lc_user_name, lc.user_email AS lc_user_email, " +
+                    " lc.create_user_id AS lc_create_user_id, lc.create_date AS lc_create_date, " +
+                    " lc.update_user_id AS lc_update_user_id, " +
+                    Institution.QUERY_FIELDS_WITH_LCI_PREFIX + ", " +
```
rushtong: nit, you can use `QUERY_FIELD_SEPARATOR` in place of `", "`

src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
```diff
@@ -134,6 +134,32 @@ void updateUser(@Bind("displayName") String displayName,
...
+                    Institution.QUERY_FIELDS_WITH_I_PREFIX +
+                    " FROM users u " +
+                    " LEFT JOIN user_role ur ON ur.user_id = u.user_id " +
+                    " LEFT JOIN roles r ON r.roleid = ur.role_id " +
+                    " LEFT JOIN library_card lc ON lc.user_id = u.user_id " +
+                    " LEFT JOIN institution lci ON lc.institution_id = lci.institution_id" +
+                    " LEFT JOIN institution i ON u.institution_id = i.institution_id")
+    Set<User> findUsersWithLCsAndInstitution();
```
rushtong: Curious - why `Set` as opposed to `List`? I usually go with lists since they retain ordering, even if we're not immediately ordering the results. If we ever need to add ordering, the method is ready. With `UseRowReducer`, that handles distinct entities so uniqueness is already handled.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -211,7 +211,7 @@ public List<User> getUsersAsRole(User user, String roleName) {
                     throw new NotFoundException("Signing Official (user: " + user.getDisplayName() + ") is not associated with an Institution.");
                 }
             case Resource.ADMIN :
-                return new ArrayList<>(userDAO.findUsers());
+                return new ArrayList<>(userDAO.findUsersWithLCsAndInstitution());
```
rushtong: If `userDAO.findUsersWithLCsAndInstitution()` returned a list, we wouldn't need this construction.

rushtong (commented):
Minor comments inline are optional - overall it looks good 👍🏽

JVThomas (approved):
Looks good to me!

PR #1755, opened 2022-09-23
Add new fields to email_entity

quazi-h:
[DUOS-2060](https://broadworkbench.atlassian.net/browse/DUOS-2060)
- Adding fields to email_entity with a changelog
- Expanded MailMessage methods
- Updated MailMessageMapper
----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/MailMessageDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/MailMessageDAO.java
@@ -20,8 +20,8 @@ public interface MailMessageDAO extends Transactional<MailMessageDAO> {
     Integer existsCollectDAREmail(@Bind("darReferenceId") String darReferenceId, @Bind("rpReferenceId") String rpReferenceId);
 
     @SqlUpdate("INSERT INTO email_entity " +
-            "(vote_id, entity_reference_id, user_id, email_type, date_sent, email_text) VALUES " +
-            "(:voteId, :entityReferenceId, :userId, :emailType, :dateSent, :emailText)")
+            "(vote_id, entity_reference_id, user_id, email_type, date_sent, email_text, create_date) VALUES " +
+            "(:voteId, :entityReferenceId, :userId, :emailType, :dateSent, :emailText, :dateSent)")
     void insertEmail(@Bind("voteId") Integer voteId,
                      @Bind("entityReferenceId") String entityReferenceId,
                      @Bind("userId") Integer userId,
@@ -30,8 +30,8 @@ public interface MailMessageDAO extends Transactional<MailMessageDAO> {
                      @Bind("emailText") String emailText);
 
     @SqlBatch("INSERT INTO email_entity " +
-            "(entity_reference_id, user_id, email_type, date_sent, email_text) VALUES " +
-            "(:entityReferenceId, :userId, :emailType, :dateSent, :emailText)")
+            "(entity_reference_id, user_id, email_type, date_sent, email_text, create_date) VALUES " +
+            "(:entityReferenceId, :userId, :emailType, :dateSent, :emailText, :dateSent)")
     void insertBulkEmailNoVotes(@Bind("userId") List<Integer> userIds,
                          @Bind("entityReferenceId") String entityReferenceId,
                          @Bind("emailType") Integer emailType,
--- a/src/main/resources/changesets/changelog-consent-100.0.xml
+++ b/src/main/resources/changesets/changelog-consent-100.0.xml
@@ -5,13 +5,13 @@
         <addColumn tableName="email_entity">
             <column name="sendgrid_response" type="text"/>
             <column name="sendgrid_status" type="integer" defaultValue="null"/>
-            // todo: default value of date_sent for existing records
-            <column name="create_date" type="date" defaultValue="date_sent"??/>
+            <column name="create_date" type="date"/>
         </addColumn>
         <sql>
-        <!-- Populate date_sent column on email_entity table
-                 with date_sent value for existing records -->
-            UPDATE email_entity e SET create_date = date_sent;
+            UPDATE email_entity e
+            SET create_date = date_sent
+            WHERE create_date IS NULL
+            AND date_sent IS NOT NULL;
         </sql>
     </changeSet>
 </databaseChangeLog>
--- a/src/main/resources/changesets/changelog-consent-100.0.xml
+++ b/src/main/resources/changesets/changelog-consent-100.0.xml
@@ -1,0 +1,39 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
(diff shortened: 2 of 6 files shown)


src/main/resources/changesets/changelog-consent-99.0.xml
```diff
@@ -1,7 +1,7 @@
 <databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
                    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
          http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
-    <changeSet id="99.0" author="connorlbari=k">
+    <changeSet id="99.0" author="connorlbark">
```
rushtong, line 4: Changes to existing change sets that have been run in an environment will throw checksum errors. Liquibase hashes the file and stores the result in a table. If they don't match, LB chokes. To get around this, you have to mark changed files with:
```
<validCheckSum>ANY</validCheckSum>
```
There are quite a few that I modified when migrating from mysql to postgres. See https://www.liquibase.com/blog/what-affects-changeset-checksums for more info.

src/main/resources/changesets/changelog-consent-100.0.xml
```diff
@@ -0,0 +1,17 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <changeSet id="100.0" author="qhoque">
+        <addColumn tableName="email_entity">
+            <column name="sendgrid_response" type="text"/>
+            <column name="sendgrid_status" type="integer" defaultValue="null"/>
+            // todo: default value of date_sent for existing records
+            <column name="create_date" type="date" defaultValue="date_sent"??/>
```
rushtong: The comment and `??` are not valid xml. Since we are populating the create date in the sql, we don't need a default value here. DAO calls should be populating this value on insert. Postgres has a `CURRENT_DATE` function, and [liquibase might also be able to handle this](https://stackoverflow.com/questions/9089678/liquibase-insert-insert-current-date) (caveat, have not tried this so it could be incorrect).

src/main/resources/changesets/changelog-consent-100.0.xml
```diff
@@ -0,0 +1,17 @@
...
+        </addColumn>
+        <sql>
+        <!-- Populate date_sent column on email_entity table
+                 with date_sent value for existing records -->
```
rushtong: nit: unnecessary comment

src/main/resources/changesets/changelog-consent-100.0.xml
```diff
@@ -0,0 +1,17 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <changeSet id="100.0" author="qhoque">
```
rushtong: This changeset isn't actually run. It needs to be included in `changelog-master.xml`

rushtong (changes requested):
There are two insert statements in `MailMessageDAO` that we should add create dates for so that we don't have an interim period where existing messages have the value (from the changeset) and new entries do not (from inserts). Just set the value to the same value as `dateSent`.

src/main/resources/changesets/changelog-consent-100.0.xml
```diff
@@ -0,0 +1,17 @@
...
+        <addColumn tableName="email_entity">
+            <column name="sendgrid_response" type="text"/>
+            <column name="sendgrid_status" type="integer" defaultValue="null"/>
+            // todo: default value of date_sent for existing records
+            <column name="create_date" type="date" defaultValue="date_sent"??/>
+        </addColumn>
+        <sql>
+        <!-- Populate date_sent column on email_entity table
+                 with date_sent value for existing records -->
+            UPDATE email_entity e SET create_date = date_sent;
```
rushtong: Had another thought here ... as a general rule, whenever we're updating a value, we should set a where condition so that we're only updating columns that are null. That prevents accidental overwrites. And we should also check that we're not putting a null into the create_date field. From looking at the databases, all values are populated, so there's no practical risk, but it is a good habit to get into.

rushtong (approved):
Thanks for the changes! 👍🏽

src/main/java/org/broadinstitute/consent/http/db/MailMessageDAO.java
```diff
@@ -20,8 +20,8 @@ public interface MailMessageDAO extends Transactional<MailMessageDAO> {
     Integer existsCollectDAREmail(@Bind("darReferenceId") String darReferenceId, @Bind("rpReferenceId") String rpReferenceId);
 
     @SqlUpdate("INSERT INTO email_entity " +
-            "(vote_id, entity_reference_id, user_id, email_type, date_sent, email_text) VALUES " +
-            "(:voteId, :entityReferenceId, :userId, :emailType, :dateSent, :emailText)")
+            "(vote_id, entity_reference_id, user_id, email_type, date_sent, email_text, create_date) VALUES " +
+            "(:voteId, :entityReferenceId, :userId, :emailType, :dateSent, :emailText, :dateSent)")
```
connorlbark, line 24: perhaps I am misunderstanding, but when would `create_date` ever be anything different than `date_sent`? Should these values be separately inserted?

connorlbark (approved):
changes look simple - one quick clarification then it looks good to go.

src/main/java/org/broadinstitute/consent/http/db/MailMessageDAO.java
rushtong: `date_sent` could be null in the event that it was not sent.

connorlbark: Right, but on insert they are always the same. Would we need to set them separately instead of insterting both of them with the same value always?

connorlbark: so instead of `create_date` being populated with `:dateSent`, it would be populated with a new argument to the DAO call, `:createDate`. So that it was possible for `create_date` and `date_sent` to have different values on insert.

rushtong: > so instead of `create_date` being populated with `:dateSent`, it would be populated with a new argument to the DAO call, `:createDate`. So that it was possible for `create_date` and `date_sent` to have different values on insert.

That's for a future story where we would distinguish between messages sent and not sent and therefore have different values. Today, all messages really are sent so they are, in fact, the same value. 

The reason we're populating the create date here is that we're adding that as a new column and we don't want to have a mix of entries that have create dates and some that do not. If we didn't put them here, current code paths would populate new entries with no creation date alongside back-populated ones that do have a creation date and we'd have a bit of a mess with inconsistent data.

PR #1771, opened 2022-10-06
[DUOS-2129][risk=low] New Search Datasets Endpoint

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2129

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
@@ -328,12 +328,18 @@ public class Dataset {
         }
 
         if (Objects.nonNull(this.dataUse)) {
-            if (this.dataUse.getEthicsApprovalRequired()) {
+            if (Objects.nonNull(this.dataUse.getEthicsApprovalRequired())
+                    && this.dataUse.getEthicsApprovalRequired()) {
                 matchTerms.add("irb");
             }
 
-            if (this.dataUse.getCommercialUse()) {
-                matchTerms.add("commercialUse");
+            if (Objects.nonNull(this.dataUse.getCollaboratorRequired())
+                    && this.dataUse.getCollaboratorRequired()) {
+                matchTerms.add("collaborator");
+            }
+            
+            if (Objects.nonNull(this.dataUse.getDiseaseRestrictions())) {
+                matchTerms.addAll(this.dataUse.getDiseaseRestrictions());
             }
         }
 
--- a/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
@@ -337,7 +337,7 @@ public class Dataset {
                     && this.dataUse.getCollaboratorRequired()) {
                 matchTerms.add("collaborator");
             }
-            
+
             if (Objects.nonNull(this.dataUse.getDiseaseRestrictions())) {
                 matchTerms.addAll(this.dataUse.getDiseaseRestrictions());
             }
--- a/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
@@ -1,9 +1,12 @@
 package org.broadinstitute.consent.http.models;
 
 import org.apache.commons.lang3.StringUtils;
+import org.checkerframework.checker.nullness.qual.NonNull;
 
+import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 
@@ -299,6 +302,53 @@ public class Dataset {
 
     public void setSharingPlanDocumentName(String sharingPlanDocumentName) {
         this.sharingPlanDocumentName = sharingPlanDocumentName;
+    }
+
+    /**
+     * Checks if the Dataset matches a raw search query. Searches on all dataset properties
+     * and some data use properties.
+     *
+     * @param query Raw string query
+     * @return if the Dataset matched query
+     */
+    public boolean isStringMatch(@NonNull String query) {
+        String lowerCaseQuery = query.toLowerCase();
+        List<String> queryTerms = new ArrayList<>(List.of(lowerCaseQuery.split("\\s+")));
+
+        List<String> matchTerms = new ArrayList<>();
+        matchTerms.add(this.getName());
+        matchTerms.add(this.getDatasetIdentifier());
+
+        if (Objects.nonNull(this.getProperties())) {
+            this.getProperties().forEach(prop -> {
+                if (Objects.nonNull(prop.getPropertyValue())) {
+                    matchTerms.add(prop.getPropertyValueAsString().toLowerCase());
+                }
+            });
+        }
+
+        if (Objects.nonNull(this.dataUse)) {
+            if (this.dataUse.getEthicsApprovalRequired()) {
+                matchTerms.add("irb");
+            }
+
+            if (this.dataUse.getCommercialUse()) {
+                matchTerms.add("commercialUse");
+            }
+        }
+
+        return queryTerms
+                .stream()
+                .filter(Objects::nonNull)
+                // all terms must match at least one thing
+                .allMatch((q) ->
+                        matchTerms
+                                .stream()
+                                .filter(Objects::nonNull)
+                                .map(String::toLowerCase)
+                                .anyMatch(
+                                        (t) -> t.contains(q)
+                                ));
     }
 
     @Override
--- a/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
@@ -227,6 +227,10 @@ public class Dataset {
     }
 
     public String getDatasetIdentifier() {
+        if (Objects.isNull(this.getAlias())) {
+            return null;
+        }
+
         return parseAliasToIdentifier(this.getAlias());
     }
 
--- a/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
@@ -313,33 +313,35 @@ public class Dataset {
      */
     public boolean isStringMatch(@NonNull String query) {
         String lowerCaseQuery = query.toLowerCase();
-        List<String> queryTerms = new ArrayList<>(List.of(lowerCaseQuery.split("\\s+")));
+        List<String> queryTerms = List.of(lowerCaseQuery.split("\\s+"));
 
         List<String> matchTerms = new ArrayList<>();
         matchTerms.add(this.getName());
         matchTerms.add(this.getDatasetIdentifier());
 
-        if (Objects.nonNull(this.getProperties())) {
-            this.getProperties().forEach(prop -> {
-                if (Objects.nonNull(prop.getPropertyValue())) {
-                    matchTerms.add(prop.getPropertyValueAsString().toLowerCase());
-                }
-            });
+        if (Objects.nonNull(getProperties()) && !getProperties().isEmpty()) {
+            List<String> propVals = getProperties()
+                    .stream()
+                    .filter((dp) -> Objects.nonNull(dp.getPropertyValue()))
+                    .map(DatasetProperty::getPropertyValueAsString)
+                    .map(String::toLowerCase)
+                    .toList();
+            matchTerms.addAll(propVals);
         }
 
-        if (Objects.nonNull(this.dataUse)) {
-            if (Objects.nonNull(this.dataUse.getEthicsApprovalRequired())
-                    && this.dataUse.getEthicsApprovalRequired()) {
+        if (Objects.nonNull(dataUse)) {
+            if (Objects.nonNull(dataUse.getEthicsApprovalRequired())
+                    && dataUse.getEthicsApprovalRequired()) {
                 matchTerms.add("irb");
             }
 
-            if (Objects.nonNull(this.dataUse.getCollaboratorRequired())
-                    && this.dataUse.getCollaboratorRequired()) {
+            if (Objects.nonNull(dataUse.getCollaboratorRequired())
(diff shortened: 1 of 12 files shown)


src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -391,42 +391,79 @@ public Set<DatasetDTO> describeDatasets(Integer userId) {
...
+        List<Dataset> datasets = this.findAllDatasetsByUser(user);
+
+        return datasets.stream().filter(ds ->this.filterDatasetByQuery(ds, query)).toList();
+    }
+
+
+    // dataset properties which can be searched by
+    private static final List<String> filterableDatasetProperties = List.of(
+            "Principal Investigator(PI)"
+    );
```
connorlbark: should we add more properties which can be filtered/searched upon?

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -391,42 +391,79 @@ public Set<DatasetDTO> describeDatasets(Integer userId) {
...
                 datasets.addAll(chairSpecificDatasets);
             }
         }
         datasets.forEach(d -> d.setDeletable(!datasetIdsInUse.contains(d.getDataSetId())));
         return datasets;
     }
 
+
+    public List<Dataset> searchDatasets(String query, User user) {
+        List<Dataset> datasets = this.findAllDatasetsByUser(user);
```
connorlbark, line 414: this call is very slow locally (~17 seconds usually on my machine). i think this is just a local setup issue, though; if you all think not, I can investigate modifying the method to be faster.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
connorlbark: should we make it searchable on `dataUse` fields? for example, 'irb' becomes a searchable term if `ethicsApprovalRequired == true`, or, e.g, diseases?

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
rushtong: My local setup shows `3825 ms` for getting all datasets as an admin and `1780 ms` as a chair so I do think this might be a non-issue in deployed environments.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
rushtong: I'm a fan of searching on anything in the dataset. This is an area where we have not put much effort into because the Data Catalog team is doing something much more comprehensive and repeating some of that work in DUOS is not a priority. However, since we're here, I do think we can expand our search fields to be whatever makes the most sense.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -391,42 +391,79 @@ public Set<DatasetDTO> describeDatasets(Integer userId) {
...
+
+        return datasets.stream().filter(ds ->this.filterDatasetByQuery(ds, query)).toList();
+    }
+
+
+    // dataset properties which can be searched by
+    private static final List<String> filterableDatasetProperties = List.of(
+            "Principal Investigator(PI)"
+    );
+    private boolean filterDatasetByQuery(Dataset ds, String query) {
```
rushtong: Some method comments here would be nice so the search approach is clear. Two 👍🏽 on this approach, by the way.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -391,42 +391,79 @@ public Set<DatasetDTO> describeDatasets(Integer userId) {
...
+        terms.add(ds.getDatasetName());
+
+        if (Objects.nonNull(ds.getProperties())) {
+            ds.getProperties().forEach(prop -> {
+                if (Objects.nonNull(prop.getPropertyName()) && Objects.nonNull(prop.getPropertyValue())
+                        && filterableDatasetProperties.contains(prop.getPropertyName())) {
+                    terms.add(prop.getPropertyValueAsString());
+                }
+            });
+        }
```
rushtong: If we're expanding the search to any field, we can simplify and speed this up a bit. Stream over all props for non-null values, map to lower case, collect to a list, then add the names and identifier after the list is constructed.

src/main/resources/assets/paths/datasetSearch.yaml
```diff
@@ -0,0 +1,21 @@
...
+  description: Returns all datasets matching a search query.
+  parameters:
+    - name: query
+      in: query
+      description: Search Query
+      required: true
+      schema:
+        type: string
+  tags:
+    - Datasets
```
rushtong:
```suggestion
    - Dataset
```

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
rushtong: Another thought here is that you could extract this logic to `Dataset` itself, i.e. 
```
public boolean isStringMatch(String query) { ... }
```
which would make streaming over a list of datasets a lot cleaner here in the service class.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -365,6 +366,20 @@ public Response datasetAutocomplete(@Auth AuthUser authUser, @PathParam("partial
...
 
+    @GET
+    @Path("/search")
+    @Produces("application/json")
+    @PermitAll
+    public Response searchDatasets(@Auth AuthUser authUser, @QueryParam("query") String query){
+        try {
+            User user = userService.findUserByEmail(authUser.getEmail());
+            List<Dataset> datasets = datasetService.searchDatasets(query, user);
+            return Response.ok(datasets, MediaType.APPLICATION_JSON).build();
```
rushtong: The common approach we use is to use the entity method for the object returned. This is a little cleaner in that you don't have to specify the media type.
```suggestion
            return Response.ok().entity(datasets).build();
```

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -365,6 +366,20 @@ public Response datasetAutocomplete(@Auth AuthUser authUser, @PathParam("partial
         }
     }
 
+    @GET
+    @Path("/search")
+    @Produces("application/json")
```
rushtong, line 413: Not for this PR, but probably a tech debt PR. We really should use a constant here throughout the app.

quazi-h (approved):
Changes look good to me. The new endpoint was returning the full dataset object when searching keywords. Tried a few cases where multiple datasets matched to a query and saw all expected datasets.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
connorlbark: Hmm, curious. Maybe I should investigate my local setup's performance - it's an issue on most endpoints that they take a very long time. Using the UI on a local consent is a bit ... slow ....

rushtong: The numbers I quoted are the response times from the swagger page. If you're looking at the UI, there's a whole host of other things going on.

connorlbark:
@rushtong this should be ready for a re-review - expanded the logic to be more intuitive (imo), used all dataset properties, and also added a couple DAR-relevant data-use terms (though incomplete - I can add more if desired)

src/main/java/org/broadinstitute/consent/http/models/Dataset.java
```diff
@@ -297,6 +304,59 @@ public void setSharingPlanDocumentName(String sharingPlanDocumentName) {
...
+
+            if (Objects.nonNull(this.dataUse.getCollaboratorRequired())
+                    && this.dataUse.getCollaboratorRequired()) {
+                matchTerms.add("collaborator");
+            }
+
+            if (Objects.nonNull(this.dataUse.getDiseaseRestrictions())) {
+                matchTerms.addAll(this.dataUse.getDiseaseRestrictions());
+            }
+        }
```
rushtong: What would be really nice here is the ability to pull a full string from a `DataUse` that has all of these values. We do have that, but not here - it's in Ontology:
```
curl -X 'POST' 'https://consent-ontology.dsde-prod.broadinstitute.org/translate?for=dataset' -H 'accept: text/plain' -H 'Content-Type: application/json' -d '{ "generalUse": false, "hmbResearch": true, "diseaseRestrictions": [ "http://purl.obolibrary.org/obo/DOID_602" ] }'

Samples are restricted for use under the following conditions:
Data is limited for health/medical/biomedical research. [HMB]
Data use is limited for studying: cancerophobia [DS]
Commercial use is not prohibited.
Data use for methods development research irrespective of the specified data use limitations is not prohibited.
Restrictions for use as a control set for diseases other than those defined were not specified.
```
Maybe as a future feature, we might have this saved as part of the dataset itself which would facilitate this matching better.

src/main/java/org/broadinstitute/consent/http/models/Dataset.java
```diff
@@ -297,6 +304,59 @@ public void setSharingPlanDocumentName(String sharingPlanDocumentName) {
...
+    /**
+     * Checks if the Dataset matches a raw search query. Searches on all dataset properties
+     * and some data use properties.
+     *
+     * @param query Raw string query
+     * @return if the Dataset matched query
+     */
+    public boolean isStringMatch(@NonNull String query) {
+        String lowerCaseQuery = query.toLowerCase();
+        List<String> queryTerms = new ArrayList<>(List.of(lowerCaseQuery.split("\\s+")));
```
rushtong: No need for the extra ArrayList construction ... 
```suggestion
        List<String> queryTerms = List.of(lowerCaseQuery.split("\\s+"));
```

src/main/java/org/broadinstitute/consent/http/models/Dataset.java
```diff
@@ -297,6 +304,59 @@ public void setSharingPlanDocumentName(String sharingPlanDocumentName) {
...
+        matchTerms.add(this.getName());
+        matchTerms.add(this.getDatasetIdentifier());
+
+        if (Objects.nonNull(this.getProperties())) {
+            this.getProperties().forEach(prop -> {
+                if (Objects.nonNull(prop.getPropertyValue())) {
+                    matchTerms.add(prop.getPropertyValueAsString().toLowerCase());
+                }
+            });
+        }
```
rushtong: Another approach to avoid N list mutations is to collect the list up then add them all at once
```
        if (Objects.nonNull(getProperties()) && !getProperties().isEmpty()) {
            List<String> propVals = getProperties()
                .stream()
                .map(DatasetProperty::getPropertyValueAsString)
                .map(String::toLowerCase)
                .toList();
            matchTerms.addAll(propVals);
        }
```

rushtong: Side note, there's a lot of unnecessary `this.` in these methods.

src/main/java/org/broadinstitute/consent/http/models/Dataset.java
```diff
@@ -297,6 +304,59 @@ public void setSharingPlanDocumentName(String sharingPlanDocumentName) {
...
+     * @param query Raw string query
+     * @return if the Dataset matched query
+     */
+    public boolean isStringMatch(@NonNull String query) {
+        String lowerCaseQuery = query.toLowerCase();
+        List<String> queryTerms = new ArrayList<>(List.of(lowerCaseQuery.split("\\s+")));
+
+        List<String> matchTerms = new ArrayList<>();
+        matchTerms.add(this.getName());
+        matchTerms.add(this.getDatasetIdentifier());
```
rushtong, line 320:
```suggestion
        matchTerms.add(getName());
        matchTerms.add(getDatasetIdentifier());
```

src/test/java/org/broadinstitute/consent/http/models/DatasetTests.java
```diff
@@ -18,4 +26,117 @@ public void testParseIdentifierToAlias() {
...
+    }
+
+    @Test
+    public void testIsStringMatchDataUseCommercial() {
+        Dataset ds = new Dataset();
+
+        assertFalse(ds.isStringMatch("collaborator"));
+
+        DataUse du = new DataUse();
+        du.setCollaboratorRequired(true);
```
rushtong: We have a handy builder object for this:
```suggestion
        DataUse du = new DataUseBuilder().setCollaboratorRequired(true).build();
```

rushtong:
@connorlbark - Couple more bits of cleanup and I think this is pretty close to finished!

src/main/java/org/broadinstitute/consent/http/models/Dataset.java
connorlbark: good catch - originally had this because `List.of` is immutable & I was adding things later, but I removed that 👍

src/main/java/org/broadinstitute/consent/http/models/Dataset.java
connorlbark: Oh hmm is it stylistically bad to use `this`? I always use it when it's optional just for clarity reasons. I am more than happy to not do that if not doing is preferred

rushtong: [Primarily used in setters](https://stackoverflow.com/questions/2411270/when-should-i-use-this-in-a-class). The other cases are fairly rare. Personally, I find `this.getProp()` less readable than `getProp()` ... it's just extra text in the code.

src/main/java/org/broadinstitute/consent/http/models/Dataset.java
connorlbark: I agree with this, I was searching for something like this originally and since I couldn't find it ultimately abandoned the idea of searching on all data use terms (since it would require many lines of code which would require frequent updating). Saving this would be very helpful I think.

rushtong (approved):
Looks great, thanks for the changes!

PR #1780, opened 2022-10-17
[DUOS-2157][risk=low] Add Dataset Batch GET Endpoint

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2157

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
@@ -53,6 +53,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 @Path("api/dataset")
@@ -289,7 +290,7 @@ public class DatasetResource extends Resource {
         try {
             List<Dataset> datasets = datasetService.getDatasets(datasetIds);
 
-            List<Integer> foundIds = datasets.stream().map(Dataset::getDataSetId).toList();
+            Set<Integer> foundIds = datasets.stream().map(Dataset::getDataSetId).collect(Collectors.toSet());
             if (!foundIds.containsAll(datasetIds)) {
                 // find the differences
                 List<Integer> differences = new ArrayList<>(datasetIds);
@@ -299,7 +300,7 @@ public class DatasetResource extends Resource {
                                 + String.join(",", differences.stream().map((i) -> i.toString()).collect(Collectors.toSet())));
 
             }
-            return Response.ok(datasets, MediaType.APPLICATION_JSON).build();
+            return Response.ok(datasets).build();
         } catch (Exception e){
             return createExceptionResponse(e);
         }
--- a/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
@@ -281,6 +281,35 @@ public class DatasetResource extends Resource {
     }
 
     @GET
+    @Path("/batch")
+    @Produces("application/json")
+    @PermitAll
+    public Response getDatasets(@QueryParam("ids") List<Integer> datasetIds){
+        try {
+            List<Dataset> datasets = datasetService.getDatasets(datasetIds);
+            if (datasets.size() != datasetIds.size()) {
+                List<Integer> idsNotFound =
+                        datasetIds
+                                .stream()
+                                .filter(
+                                        (id) -> !datasets
+                                                .stream()
+                                                .map(Dataset::getDataSetId)
+                                                .toList()
+                                                .contains(id))
+                                .toList();
+
+                throw new NotFoundException(
+                        "Could not find datasets with ids: "
+                                + String.join(",", idsNotFound.stream().map((i) -> i.toString()).toList()));
+            }
+            return Response.ok(datasets, MediaType.APPLICATION_JSON).build();
+        } catch (Exception e){
+            return createExceptionResponse(e);
+        }
+    }
+
+    @GET
     @Consumes("application/json")
     @Produces("application/json")
     @Path("/validate")
--- a/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
@@ -47,6 +47,7 @@ import javax.ws.rs.core.UriInfo;
 import java.io.ByteArrayInputStream;
 import java.io.InputStream;
 import java.net.URI;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -673,6 +674,46 @@ public void testGetDatasetNotFound() {
...
+
+        when(datasetService.getDatasets(List.of(1,2,3,4))).thenReturn(List.of(
+                ds1,
+                ds3
+        ));
+
+        initResource();
+        Response response = resource.getDatasets(List.of(1,2,3,4));
+        assertEquals(404, response.getStatus());
+        assertEquals("Could not find datasets with ids: 2,4", ((Error)response.getEntity()).getMessage());
```
connorlbark: is there a more efficient way to check error message?

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -280,6 +280,35 @@ public Response getDataset(@PathParam("datasetId") Integer datasetId){
...
     }
 
+    @GET
+    @Path("/batch")
+    @Produces("application/json")
+    @PermitAll
+    public Response getDatasets(@QueryParam("ids") List<Integer> datasetIds){
+        try {
+            List<Dataset> datasets = datasetService.getDatasets(datasetIds);
+            if (datasets.size() != datasetIds.size()) {
```
rushtong: Size is one way to do this, but can be an error in the case where a user provides the same id more than once (i.e. ` 1, 1, 2, 3` -> list of datasets will be 3). 

One thing to do would be to ensure distinctness on the provided list.

Another option might look something like this:
```
List<Integer> foundIds = datasets.stream().map(Dataset::getDataSetId).toList();
if (!foundIds.containsAll(datasetIds)) { 
  // find the differences
  List<Integer> differences = new ArrayList<>(datasetIds);
  differences.removeAll(foundIds);
  ... throw error with differences ... 
}
```

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
rushtong: IMHO, message testing is not very helpful and often brittle so I discourage it. But a direct answer is no, reading the entity is the only way to get the content out of the response.

src/main/resources/assets/paths/datasetBatch.yaml
```diff
@@ -0,0 +1,27 @@
+get:
+  summary: Get Dataset by Id
```
rushtong:
```suggestion
  summary: Get Dataset by list of Ids
```

src/main/resources/assets/paths/datasetBatch.yaml
```diff
@@ -0,0 +1,27 @@
...
+  description: Finds Dataset and DatasetProperties of the Dataset specified by id.
```
rushtong, line 3:
```suggestion
  description: Finds Dataset and DatasetProperties of the Dataset specified by list of ids.
```

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
rushtong: Testing with a duplicate ID that is valid shows this response:
```
{
  "message": "Could not find datasets with ids: ",
  "code": 404
}
```

src/main/resources/assets/paths/datasetBatch.yaml
```diff
@@ -0,0 +1,27 @@
...
+  parameters:
+    - name: ids
+      in: query
+      description: Comma separate dataset ids, e.g. "1,2,3,4,5"
+      required: true
+      schema:
+        type: array
+        items:
+          type: integer
```
rushtong: The curl call shows a different format: `/api/dataset/batch?ids=2&ids=369`

If, instead, I use a comma separated list this way, Jersey doesn't actually parse the list in the way that is described in the docs:
```
curl -X 'GET' 'https://local.broadinstitute.org:27443/api/dataset/batch?ids=2,369'
```
yields
```
{"code":404,"message":"HTTP 404 Not Found"}
```
since that endpoint doesn't exist

src/main/resources/assets/paths/datasetBatch.yaml
```diff
@@ -0,0 +1,27 @@
...
+  tags:
+    - Dataset
+  responses:
+    200:
+      description: The Dataset with properties
```
rushtong:
```suggestion
      description: The list of Datasets with properties
```

rushtong (changes requested):
Looking good! Some doc updates and an edge case to consider inline.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
connorlbark: Hmm, I totally feel you which is why I was unsure of this test. However, I wanted to make sure the logic to return just the datasets not found worked properly. I considered making it a helper function, but I couldn't figure out a logical place to make that a public/testable function. So I leaned on testing the message directly instead.

connorlbark:
@rushtong thanks for the helpful feedback !! I think I addressed it all

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -280,6 +281,30 @@ public Response getDataset(@PathParam("datasetId") Integer datasetId){
...
+    @GET
+    @Path("/batch")
+    @Produces("application/json")
+    @PermitAll
+    public Response getDatasets(@QueryParam("ids") List<Integer> datasetIds){
+        try {
+            List<Dataset> datasets = datasetService.getDatasets(datasetIds);
+
+            List<Integer> foundIds = datasets.stream().map(Dataset::getDataSetId).toList();
+            if (!foundIds.containsAll(datasetIds)) {
```
rushtong, line 294: So, I may have steered you poorly - I didn't actually test the code I used as an example. Turns out `List. containsAll()` [has poor performance](https://stackoverflow.com/questions/10199772/what-is-the-cost-of-containsall-in-java). The suggestion is to use a hash instead,
```
if (!new HashSet<>(foundIds).containsAll(datasetIds)) {
```

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -280,6 +281,30 @@ public Response getDataset(@PathParam("datasetId") Integer datasetId){
...
+                // find the differences
+                List<Integer> differences = new ArrayList<>(datasetIds);
+                differences.removeAll(foundIds);
+                throw new NotFoundException(
+                        "Could not find datasets with ids: "
+                                + String.join(",", differences.stream().map((i) -> i.toString()).collect(Collectors.toSet())));
+
+            }
+            return Response.ok(datasets, MediaType.APPLICATION_JSON).build();
```
rushtong: One more thing, this can be simpler, `MediaType.APPLICATION_JSON` is already specified on line 286.

```
return Response.ok(datasets).build();
```
or 
```
return Response.ok().entity(datasets).build();
```
In the above cases, Jackson will coerce the objects to JSON format based on `@Produces`

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
rushtong: One thing you could do is test for the existence/non-existence of the IDs in the string instead of looking at complete string equality.

rushtong (commented):
The updates look good, please see additional (optional) comments inline.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
connorlbark: ahh good idea, just implemented I believe a decent way of doing this.

connorlbark:
@rushtong should be ready for review again !

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -695,6 +697,48 @@ public void testGetDatasets() {
...
+                ds1,
+                ds2
+        ));
+
+        initResource();
+        Response response = resource.getDatasets(List.of(1,1,2,2,3,3));
+        assertEquals(404, response.getStatus());
+        assertTrue(((Error)response.getEntity()).getMessage().contains("3"));
+        assertFalse(((Error)response.getEntity()).getMessage().contains("2"));
+        assertFalse(((Error)response.getEntity()).getMessage().contains("1"));
```
rushtong, line 738: Nice 👍🏽

rushtong (approved):
Thanks for the changes! 👍🏽

JVThomas (approved):
Tested locally, batch endpoint is working as intended.

PR #1783, opened 2022-10-19
[DUOS-2091][risk=medium] Update Dataset Registration Schema and API

rushtong:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2091

Includes schema, model, resource, and test method updates to accommodate the updates to the dataset registration schema. Most of the domain changes were generated using `jsonschema2pojo`

----

- Verify all tests go green


src/main/resources/dataset-registration-schema_v1.json
```diff
@@ -151,6 +148,11 @@
       "description": "NIH Institution/Center for Submission",
       "enum": ["NCI", "NEI", "NHLBI", "NHGRI", "NIA", "NIAAA", "NIAID", "NIAMS", "NIBIB", "NICHD", "NIDCD", "NIDCR", "NIDDK", "NIDA", "NIEHS", "NIGMS", "NIMH", "NIMHD", "NINDS", "NINR", "NLM", "CC", "CIT", "CSR", "FIC", "NCATS", "NCCIH"]
     },
+    "nihInstitutionalCertificationFileName": {
+      "type": "string",
+      "label": "If an Institutional Certification for this consent group exists, please upload it here (file upload)",
```
quazi-h:
```suggestion
      "label": "NIH Institutional Certification",
```
The ticket mentioned that should be the label but wasn't sure if that's correct because I see a lot of the fields here have the same label / description.

quazi-h (approved):
Overall changes look good, confirmed file type is part of the consent group object, NIH supporting ICs is now an array, and the new NIH optional file name looks good, just left one minor suggestion in the code to update the label for the new field.

src/main/resources/dataset-registration-schema_v1.json
rushtong: We're using both label and description here as the same thing. The label is meant to be the thing that the UI can use to label the field. The description is more of a documentation thing meant to describe in more general terms what the field is about. In practice, the UI is self-defining the labels in UI code, so they're a little extraneous here.

rushtong: That being said, the ticket does give us two different values, so thanks, will apply that suggestion.

JVThomas (approved):
Code looks good to me (tested locally).

PR #1794, opened 2022-10-27
[DUOS-2059][risk=no] SendGrid Refactor

rushtong:
## Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DUOS-2059

This PR makes some minor, but extensive, changes to how we send messages via SendGrid. This is already a large PR, so I am hoping to move this through before tackling the original goal of capturing SendGrid response info per email we send.

### Changes:
* Renamed the `MailService` class to `SendGridAPI` for clarity. Unfortunately, git didn't retain the rename and considers it a +/-
* Migrated all uses of sending multiple messages to sending a single message
  * I want to do this so that we can more easily track what was sent via SendGrid per message
* Lots of message and test refactoring to accomplish ☝🏽 

---
A future PR will:
* Make use of the returned SendGrid `Response` 
* Improve testing of `SendGridAPI` (the actual `SendGrid` object is not mockable in the current implementation)
* Save a SendGrid response per email entity

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/mail/SendGridAPI.java
+++ b/src/main/java/org/broadinstitute/consent/http/mail/SendGridAPI.java
@@ -1,5 +1,6 @@
 package org.broadinstitute.consent.http.mail;
 
+import com.google.api.client.http.HttpStatusCodes;
 import com.sendgrid.Method;
 import com.sendgrid.Request;
 import com.sendgrid.Response;
@@ -23,6 +24,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.io.Writer;
+import java.util.Map;
 import java.util.Optional;
 
 public class SendGridAPI {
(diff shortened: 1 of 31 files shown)


src/main/java/org/broadinstitute/consent/http/mail/SendGridAPI.java
```diff
@@ -0,0 +1,144 @@
...
+    public SendGridAPI(MailConfiguration config) {
+        this.fromAccount = config.getGoogleAccount();
+        this.sendGrid = new SendGrid(config.getSendGridApiKey());
+        this.activateEmailNotifications = config.isActivateEmailNotifications();
+    }
+
+    private Optional<Response> sendMessage(Mail message) {
+        if (activateEmailNotifications) {
+            try {
+                // See https://github.com/sendgrid/sendgrid-java/issues/163
```
JVThomas, line 58: Getting a `Page Not Found` when trying to access this link

rushtong: Yeah, I think they disabled public view of their issues page. This is copied from the previous class and unrelated to the refactor, so I left it in place until I can find a solution.

JVThomas (approved):
Code looks good to me

connorlbark (approved):
👍 I like this refactor a lot, as someone who recently dove in to learn about this side of the codebase, this is much less confusing and more straightforward now.

PR #1797, opened 2022-10-29
[DUOS-2059][risk=no] Update email handling for SendGrid Responses

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2059

This PR centralizes all email sending and capturing through a single method so we can easily capture the content of the email sent as well as the response from SendGrid. In practice, I'm finding that there are silent errors we weren't capturing which is very likely the source of sporadic user complaints about emails not being received.

## Changes
* Renamed `EmailNotifierService` to `EmailService` but github shows them as a separate add/remove. Git blame, however, shows the files as correctly renamed.
* Added enum to track the types of email
  * Note that we have a table for `email_type` but have never used it. Dropping that in favor of a code-based enum.
* Reworked the `email_entity` table a bit so it is more consistent with other fields and has the right nullability settings
* Removed the `election_id` field that doesn't exist as a column in `email_entity`
* Reworked `SendGridAPI` constructor so we can better test with a mock `SendGrid` object.
* Check for disabled email preference before making the call to `SendGrid`

## Examples
This screen shows the capture of both successful and unsuccessful SendGrid API calls.
Screen Shot 2022-10-29 at 11 25 19 AM

This screen shows the capture when email is enabled/disabled. 707 is enabled, 708 is disabled:
Screen Shot 2022-11-04 at 9 44 03 AM

----

- Verify all tests go green


quazi-h (approved):
These changes look good to me.

I tried testing this locally by setting activateEmailNotifications to true under mailConfiguration in my consent.yaml and then using the Notifier service to send reminder emails to my user for some open votes using specific voteIds. I'm not seeing any errors when using the notifier endpoint, saw the successful 200 response in my console, and confirmed that the voteId isReminderSent attributes were updated to "true" after hitting the endpoint.

I'm having issues connecting to my local DB to confirm the new entries in the email_entity table but will work on troubleshooting that.

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
```diff
@@ -0,0 +1,446 @@
...
+            response.orElse(null),
+            data.get("electionId"),
+            voteId,
+            Integer.valueOf(data.get("dacUserId")),
+            EmailType.REMINDER,
+            template
+        );
+    }
+
+    public void sendDarNewCollectionElectionMessage(List<User> users, DarCollection darCollection) throws IOException, TemplateException {
```
JVThomas, line 191: Question about `email_preference`: When should it be taken into account? In this function it takes a list of users as an argument. With its current usage in `DarCollectionService.java ln:824`, that list of users is obtained by a query that does not take `email_preference` into account.

In contrast, `sendClosedDataSetElectionsMessage` queries users within the function itself with a DAO method that does take `email_preference` into account.

rushtong: IIRC, that preference is an admin only feature. I think we should honor it regardless, so I'll dig into this detail and see if we can treat it more consistently here as opposed to elsewhere. Looks like there are some conflicts too - another reason to dig back in.

rushtong:
@JVThomas, @quazi-broad, I added some more logic to the call to SendGrid to ensure we're not sending email to users with a disabled email preference. Please take another look, thank you!

src/main/java/org/broadinstitute/consent/http/mail/SendGridAPI.java
```diff
@@ -69,7 +122,13 @@ private Optional<Response> sendMessage(Mail message) {
                 for (String key : sendGrid.getRequestHeaders().keySet())
                     request.addHeader(key, sendGrid.getRequestHeaders().get(key));
                 // send
-                return Optional.of(sendGrid.makeCall(request));
+                Response response = sendGrid.makeCall(request);
+                if (response.getStatusCode() > 202) {
```
otchet-broad, line 131: Silly question, but why not log all messages sent?  I could see where you might not want them all as errors, but I thought I read that one of the concerns that there were missing messages.  By logging the response for all messages might give us insight into if DUOS sent the message but SendGrid lost it.

--Alright.  I see we're storing the response in the DB.  Will those get pruned?  If so, I'd still consider logging it.

src/main/java/org/broadinstitute/consent/http/mail/SendGridAPI.java
```diff
@@ -48,15 +55,61 @@ public class SendGridAPI {
...
+                }
+                return user.getEmailPreference();
+            }
+        }
+        return false;
+    }
+
+    public Optional<Response> sendMessage(Mail message) {
+        boolean userEmailPreference = findUserEmailPreference(message);
+        if (activateEmailNotifications && userEmailPreference) {
```
otchet-broad, line 117: Would you want to log the case where the email would have been sent but the user's preference to suppress was set at the time?

src/main/java/org/broadinstitute/consent/http/mail/SendGridAPI.java
rushtong: For successful requests, SendGrid doesn't send a body response, just a header, so the only responses we can do anything with are error cases.

rushtong: In general, we don't do db pruning. I can envision a targeted event where we might need to do that, but as a matter of application functionality, we tend not to go down that road.

src/main/java/org/broadinstitute/consent/http/mail/SendGridAPI.java
rushtong: Good point, that's likely worth an `info` log, will do.

src/main/java/org/broadinstitute/consent/http/mail/SendGridAPI.java
rushtong: Docs for what I asserted above: https://docs.sendgrid.com/api-reference/mail-send/mail-send#responses

src/main/java/org/broadinstitute/consent/http/mail/SendGridAPI.java
```diff
@@ -138,14 +197,14 @@ public Optional<Response> sendDataCustodianApprovalMessage(String toAddress, Str
...
     }
 
-    public void sendDatasetDeniedMessage(String toAddress, Writer template) throws MessagingException {
+    public Optional<Response> sendDatasetDeniedMessage(String toAddress, Writer template) throws MessagingException {
         Mail message = datasetDeniedMessage.datasetDeniedMessage(toAddress, fromAccount, template);
-        sendMessage(message);
+        return sendMessage(message);
     }
 
 }
```
otchet-broad, line 215: Other questions in the back of my mind that shouldn't gate this pull request:
Do we need to facilitate resending messages that fail?
Should we build a message sending queue that backs off delivery attempts when the service returns [429](https://docs.sendgrid.com/api-reference/how-to-use-the-sendgrid-v3-api/rate-limits#rate-limit-response-header)'s  
How can we leverage the SendGrid API to determine when a mailbox no longer is accepting mail and what (if anything) should we do in DUOS to deal with that.

rushtong: All good questions. I am drafting a ticket for retries, but had not considered making use of other sendgrid functionalities. I'll dig more into that too.

otchet-broad (approved):
This seems like a positive change to give us better visibility into SendGrid and sets us up nicely for some follow-up concepts in the future.

JVThomas (approved):
Code looks good. `email_preferences` is consistently taken into account and logging successfully keeps track of email messaging outcomes.

quazi-h (approved):
New changes since my last review look good to me, good call on checking the email preference, we wouldn't want to spam users with emails if they have that disabled. New tests for this additional functionality are looking good as well.

PR #1801, opened 2022-11-02
[DUOS-2176][risk=no] Mark unused APIs as deprecated

rushtong:
## Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DUOS-2176

Once these are deprecated for a release cycle, will remove them in a future PR.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataRequestElectionResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataRequestElectionResource.java
@@ -84,6 +84,7 @@ public class DataRequestElectionResource extends Resource {
     }
 
 
+    @Deprecated // See DUOS-2176
     @GET
     @Produces("application/json")
     @PermitAll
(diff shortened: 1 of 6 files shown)


otchet-broad:
I noticed that reprocessPurposeMatches is the only method in the class that doesn't seem to be mark deprecated.  I didn't see where the front end used that resource.  Should that one be deprecated as well?  If so, should the ref'ed model also be deprecated?

rushtong:
> I noticed that reprocessPurposeMatches is the only method in the class that doesn't seem to be mark deprecated. I didn't see where the front end used that resource. Should that one be deprecated as well? If so, should the ref'ed model also be deprecated?

Great question. This is not used by the front end, only for administrative purposes. The case where we want to update a particular DAR's match status happens if a dataset has (in some non-standard way) changed its data use. For example, we currently have a group of datasets in prod that are incorrectly categorized (see https://broadworkbench.atlassian.net/browse/DUOS-2138). We will need to modify those and reprocess any DARs that have requested access to them.

otchet-broad (approved):

quazi-h (approved):
Changes are straightforward, looks good to me.

PR #1809, opened 2022-11-08
[DUOS-2189][risk=no] Remove nih certification file from schema

rushtong:
## Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DUOS-2189

----

- Verify all tests go green


quazi-h (approved):
Removal of NIH certification from the backend here looks good to me.

JVThomas:
Quick question: Is the `nihInstitutionalCertificationFileName` to be removed completely from the schema? The schema endpoint still has it in its response because it reads the schema off of `dataset-registration-schema_v1.json`, which still has the key listed.

rushtong:
> Quick question: Is the `nihInstitutionalCertificationFileName` to be removed completely from the schema? The schema endpoint still has it in its response because it reads the schema off of `dataset-registration-schema_v1.json`, which still has the key listed.

That should be gone ... looks like I missed it in the chaos ... will address.

otchet-broad (approved):
Model and resource changes all look good.

PR #1811, opened 2022-11-08
[DUOS-2182][risk=medium] - Ensure results are filtered to appropriate objects.

otchet-broad:
[DUOS-2182](https://broadworkbench.atlassian.net/browse/DUOS-2182) is a production bug that returns unintended information in a response.  
The underlying queries in two code paths were performing `LEFT JOIN` operations with terms intending to reduce the result set. LEFT JOIN will enrich, not reduce.

Tests and assertions were added to confirm the issues in` DataAccessRequestDAOTest.java` and updated queries were developed for `DataAccessRequestDAO.java` that reduced the results to only include responses with the specific dataset ID.

The impact to the MetricsService was considered, but the team indicated they're not being used in a meaningful way at this time.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
@@ -67,11 +67,12 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
 
   @SqlQuery(
         " SELECT dar.user_id FROM data_access_request dar "
-          + "  LEFT JOIN dar_dataset dd ON dd.reference_id = dar.reference_id AND dd.dataset_id = :datasetId  "
+          + "  LEFT JOIN dar_dataset dd ON dd.reference_id = dar.reference_id "
           + "  WHERE dar.draft = false"
             + "  AND (EXISTS (SELECT 1 FROM election e"
           + "  INNER JOIN vote v on v.electionId = e.election_id AND LOWER(v.type) = 'final'"
           + "  WHERE v.vote = true AND LOWER(e.election_type) = 'dataaccess'"
+          + "  AND e.dataset_id = :datasetId"
             + "  AND e.reference_id = dar.reference_id))"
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -51,21 +51,28 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
...
+              + " (dar.data #>> '{}')::jsonb AS data "
+              + " FROM data_access_request dar "
+              + " INNER JOIN election e "
+              + " ON e.reference_id = dar.reference_id "
+              + " INNER JOIN vote v "
+              + " ON e.election_id = v.electionid "
+              + " WHERE lower(v.type) = 'final' AND v.vote = true AND e.dataset_id = :datasetId "
+              + " AND e.archived IS NOT true AND lower(e.election_type) = 'dataaccess'"
+  )
   List<DataAccessRequest> findAllDataAccessRequestsByDatasetId(@Bind("datasetId") Integer datasetId);
```
rushtong: This method is only used for two purposes, one to get all approved DARs and the other for a metrics call that we're not considering. This method is no longer named correctly based on the updated query. I'd suggest something like `findAllApprovedDataAccessRequestsByDatasetId` or something similar.

rushtong (commented):
One minor naming suggestion, then 👍🏽 
Again, great tests!

connorlbark (approved):
👍  agree with gregs feedback but other than that it looks great

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
otchet-broad: Agreed.  I had made that exact change, stashed it away and forgot to commit it.

rushtong (approved):
Thanks for the changes 👍🏽

PR #1815, opened 2022-11-15
[DUOS-2183] Add DB Table for Files

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2183

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/cloudstore/GCSService.java
```diff
@@ -67,23 +67,24 @@ public Bucket getRootBucketWithMetadata() {
...
      *
      * @param content InputStream content
      * @param mediaType String media type
-     * @param fileName String file name
+     * @param bucketName String unique name of the bucket
      * @return BlobId of the stored document
      * @throws IOException Exception when storing document
      */
-    public BlobId storeDocument(InputStream content, String mediaType, String fileName)
+    public BlobId storeDocument(InputStream content, String mediaType, String bucketName)
```
connorlbark: wanted to make it clearer that the fileName was just what is used as the name of the bucket. This change was made because I was passing in UUIDs instead of the file name. I pass in UUIDs so that there are no conflicts if users upload two files with the exact same name.

src/main/java/org/broadinstitute/consent/http/service/UserFileService.java
```diff
@@ -0,0 +1,90 @@
...
+import org.broadinstitute.consent.http.models.UserFile;
+
+import javax.ws.rs.NotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Date;
+import java.util.List;
+import java.util.UUID;
+
+public class UserFileService {
```
connorlbark: these service classes are not yet used (except in tests) but I logically assumed these would be the minimum necessary functions for UserFile to be a useful abstraction.

src/main/resources/changesets/changelog-consent-103.0.xml
```diff
@@ -0,0 +1,28 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <changeSet id="103.0" author="connorlbark">
```
otchet-broad: Huzzah!  We now have two PRs with the same id.

From Liquibase's [documentation](https://docs.liquibase.com/concepts/changelogs/changelog-formats.html): ...
The id tag is only used as an identifier, it does not direct the order that changes are run and does not have to be an integer.

Should we move to a different id value?  Maybe a ticket? "DUOS-2183" or date and ticket "20221115_DUOS-2183"

connorlbark: Usually we just update whichever merges in second; so mine will be 104.0 if yours goes in first.

rushtong: I don't have a strong preference for file naming aside from the nicety of seeing all changelogs sequentially ordered. In terms of ID naming options, I lean towards the date-based ids over others.

src/main/java/org/broadinstitute/consent/http/cloudstore/GCSService.java
otchet-broad: Would it be better to change the type to UUID and the name to bucketUUID?

connorlbark: Right now, it is used in other places. There is another ticket in flight which would make that change possible - https://broadworkbench.atlassian.net/browse/DUOS-2184 - I believe within this ticket it would be possible to make that a UUID type.

src/main/java/org/broadinstitute/consent/http/models/UserFile.java
```diff
@@ -0,0 +1,136 @@
...
+            + " uf.category as uf_file_name, "
+            + " uf.media_type as uf_file_name, "
+            + " uf.create_user_id as uf_file_name, "
+            + " uf.create_date as uf_create_date, "
+            + " uf.deleted as uf_deleted ";
+
+    private Integer userFileId;
+    private String entityId;
+    private String fileName;
+    private String bucketName;
```
otchet-broad: Would it be better to type this as a UUID?

src/main/resources/changesets/changelog-consent-103.0.xml
```diff
@@ -0,0 +1,28 @@
...
+        <createTable tableName="user_file">
+            <column name="user_file_id" autoIncrement="true" type="int">
+                <constraints primaryKey="true"/>
+            </column>
+            <!-- Could be: Reference ID, Dataset ID column, etc. -->
+            <column name="entity_id" type="text"/>
+            <column name="file_name" type="text" >
+                <constraints nullable="false"/>
+            </column>
+            <column name="bucket_name" type="text">
```
otchet-broad, line 14: Should this be text or a UUID?

src/main/java/org/broadinstitute/consent/http/models/UserFile.java
```diff
@@ -0,0 +1,136 @@
+package org.broadinstitute.consent.http.models;
+
+import org.broadinstitute.consent.http.enumeration.UserFileCategory;
+
+import java.io.InputStream;
+import java.util.Date;
+import java.util.Objects;
+
+public class UserFile {
```
rushtong: Nit, naming (naming is hard!). These are not necessarily `User` `Files`. Files could exist for any conceivable entity. Could we name this something more generic such as `FileObject` or `FileStorageObject` or anything in that genre.

src/main/java/org/broadinstitute/consent/http/db/UserFileDAO.java
```diff
@@ -0,0 +1,67 @@
...
+            @Bind("bucketName") String bucketName,
+            @Bind("blobName") String blobName,
+            @Bind("mediaType") String mediaType,
+            @Bind("entityId") String entityId,
+            @Bind("createUserId") Integer createUserId,
+            @Bind("createDate") Date createDate
+    );
+
+    @SqlUpdate(
+            "UPDATE user_file SET deleted=true WHERE user_file_id = :userFileId"
```
otchet-broad: Should we store the user and when the soft delete was made?

src/main/java/org/broadinstitute/consent/http/models/UserFile.java
rushtong: Do we even need a bucket name? The general bucket we use is an environment-specific location, so we have separate ones for dev, staging, prod, etc. A `BlobId` should be enough.

src/main/java/org/broadinstitute/consent/http/db/UserFileDAO.java
rushtong: Agreed, we should track update user and update date.

src/main/java/org/broadinstitute/consent/http/models/UserFile.java
connorlbark: good point, I was trying to avoid just `File` because I felt a class with the name `File` has a specific meaning. `FileStorageObject` seems apt.

src/main/java/org/broadinstitute/consent/http/cloudstore/GCSService.java
otchet-broad: This is a great find, by the way.  I'm digging in a bit further to make sure I understand the full repercussions of what " I pass in UUIDs so that there are no conflicts if users upload two files with the exact same name" implies about the existing codebase.

rushtong: IIUC, this is intended to store the parent bucket that the file is stored in. That's not a UUID presently, it's a plain string. Look at your local dev config for an example:
```
googleStore:
  password: /etc/consent-account.json
  endpoint: https://storage.googleapis.com/
  bucket: broad-dsde-dev-consent
```

src/main/java/org/broadinstitute/consent/http/enumeration/UserFileCategory.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public enum UserFileCategory {
+    IRB_COLLABORATION_LETTER("irbCollaborationLetter"),
+    DATA_USE_LETTER("dataUseLetter"),
+    ALTERNATIVE_DATA_SHARING_PLAN("alternativeDataSharingPlan");
```
rushtong: We also need an NIH Institutional Certification category as well

src/main/java/org/broadinstitute/consent/http/service/UserFileService.java
```diff
@@ -0,0 +1,91 @@
...
+            String mediaType,
+            UserFileCategory category,
+            String entityId,
+            Integer createUserId
+    ) throws IOException {
+        // upload to GCS
+        BlobId blobId = gcsService.storeDocument(
+                content,
+                mediaType,
+                UUID.randomUUID().toString());
```
rushtong: We should wrap this in a try and log any errors

src/main/java/org/broadinstitute/consent/http/service/UserFileService.java
```diff
@@ -0,0 +1,91 @@
...
+
+        userFile.setUploadedFile(document);
+    }
+
+    public UserFile fetchById(
+            Integer userFileId
+    ) throws NotFoundException {
+        UserFile userFile = userFileDAO.findFileById(userFileId);
+        // download file from GCS
+        fetchAndPopulateUploadedFile(userFile);
```
rushtong: I'm not sold on fetching stream content _ALL THE TIME_. We should really only be pulling GCS content when a user specifically requests it. Imagine this user flow ... 

1. User navigates to an entity.
2. User should see that entity has associated files
  a. User then navigates away from page (no need to ever have queried for the IS content)
  b. User clicks on an associated file entity (now we need to pull the IS content)

We should have two methods, one to get the file as metadata and a separate one to get the content.

src/main/java/org/broadinstitute/consent/http/service/UserFileService.java
```diff
@@ -0,0 +1,91 @@
...
+        return userFile;
+    }
+
+    public List<UserFile> fetchAllByEntityId(
+            String entityId
+    ) throws NotFoundException {
+        List<UserFile> userFiles = userFileDAO.findFilesByEntityId(entityId);
+        // download all files from GCS
+        userFiles.forEach(this::fetchAndPopulateUploadedFile);
```
rushtong: Be wary of external api or dao calls in loops. I generally do not try to prematurely optimize code, but this is one exception since it almost always leads to real performance problems.

connorlbark: I agree, but I assumed there wasn't a way to fetch multiple files from GCS at once. I will investigate this

src/main/java/org/broadinstitute/consent/http/service/UserFileService.java
connorlbark: Yes I agree. I envisioned a `findFile` method as well. I can implement it as well for completeness. The thought was to get all of the 'non-trivial' service methods out of the way.

src/main/java/org/broadinstitute/consent/http/cloudstore/GCSService.java
rushtong: > wanted to make it clearer that the fileName was just what is used as the name of the bucket.

I don't think this is actually true ... bucket location is configured separately from the files that are named inside the bucket. Partial screen cap of our current dev bucket:

Screen Shot 2022-11-16 at 12 47 42 PM

src/main/java/org/broadinstitute/consent/http/models/UserFile.java
connorlbark: wasn't sure if it would be possible for the bucket to change in the future, so I was tracking it just to be safe

src/main/java/org/broadinstitute/consent/http/cloudstore/GCSService.java
connorlbark: I just realized some point of confusion here: I meant to name this variable `blobName`. The variable is the name of the blob, which is why it's a UUID. The bucket name itself is `broad-dsde-dev-consent` always.

~However, right now, there are many files which already exist which are uploaded based on filename, not uuid. They cannot be coerced to UUIDs.~ (this is incorrect, they should all exist as UUIDs. thus, changing type to UUID makes sense.)

src/main/java/org/broadinstitute/consent/http/service/UserFileService.java
rushtong: > I agree, but I assumed there wasn't a way to fetch multiple files from GCS at once. I will investigate this

I'm not aware of a convenient way of doing this either.

src/main/java/org/broadinstitute/consent/http/cloudstore/GCSService.java
otchet-broad: Did a little test to put my mind at ease on this one.  I created two DARs, both requiring a COL agreement and uploaded two files (same name, different contents) and saw that the files are properly differentiated for return.

otchet-broad: DataAccessRequestResourceVersion2.java line #428 does the lifting to keep files uniquely named.  Now that I better understand things, do you really want to name this "blobId"?

src/main/java/org/broadinstitute/consent/http/models/UserFile.java
otchet-broad: Ok, now that I'm tracking, could we refer to blob_id as blobId?  I'm having a hard time reading this and keeping the different elements being tracked.

src/main/java/org/broadinstitute/consent/http/cloudstore/GCSService.java
connorlbark: Oh, I see. I had gotten confused with the way the pathing had worked. Sure, we can make this a UUID type.

src/main/java/org/broadinstitute/consent/http/service/UserFileService.java
otchet-broad: @rushtong -could it be even simpler as a passthrough to GCSService.getDocument and return the InputStream?

src/main/resources/changesets/changelog-consent-103.0.xml
```diff
@@ -0,0 +1,28 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="blob_name" type="text">
```
otchet-broad: Should this really just be blobId?

BlobId encapsulates both the bucket, name, and generation all at once.

src/main/java/org/broadinstitute/consent/http/service/UserFileService.java
rushtong: > could it be even simpler as a passthrough to GCSService.getDocument and return the InputStream?

Yes, that is the intention of that service call.

src/main/resources/changesets/changelog-consent-103.0.xml
connorlbark: right, that's why I didn't call it blob_id; I saved them separately and called it `blob_name` to make it less ambigious given [the `BlobId` class](https://googleapis.dev/java/google-cloud-clients/0.97.0-alpha/com/google/cloud/storage/BlobId.html), which takes in `BlobId.of(bucketName, blobName)` and is used by GCSService. however, it seems that this naming is confusing, and if we're going to remove the bucket since it is static, then I can rename it to something a bit clearer.  I'm still not convinced on `blob_id` since it is a loaded term.

connorlbark: ohhh actually this documentation is out of date. I think it make make sense to instead save everything as a gsUtilUri: https://cloud.google.com/java/docs/reference/google-cloud-storage/latest/com.google.cloud.storage.BlobId#inherited-members . I didn't see this function in the old documentation. This way, there's only one string to keep track of. thus, it could be saved as `gcs_file_uri`, and hopefully be a lot less confusing.

connorlbark:
Just pushed up a big update. Major changes, aside from smaller feedback items:
- Renamed to `FileStorageObject`
- Database now stores the GCS URI (e.g. `gs://bucket/blob`, and in the mapper converts this to the `BlobId` type on the model. Should be very straightforward to use/understand now.
- I found that you can actually get a list of files from GCS in one batch call. I made a new GCS service function using this batch functionality and reimplemented fetching for multiple files to use this service function.
- Added delete id, delete user (populated on deletion). Added update user and update id columns, but since no update functions exist right now they are unused.

src/main/resources/changesets/changelog-consent-104.0.xml
```diff
@@ -0,0 +1,31 @@
...
+            <column name="file_name" type="text" >
+                <constraints nullable="false"/>
+            </column>
+            <column name="gcs_file_uri" type="text">
+                <constraints nullable="false"/>
+            </column>
+            <column name="category" type="text"/>
+            <column name="media_type" type="text"/>
+            <column name="create_user_id" type="integer"/>
+            <column name="create_date" type="date"/>
```
rushtong: I should have commented on this earlier. We should be working towards using timestamps instead of dates. In Postgres, a date doesn't store a time, just the raw date (i.e. day, month, year). (see https://www.postgresql.org/docs/11/datatype-datetime.html)

This also leads toward another java convention that we should be working towards which is using `Instant` (which is really a timestamp) over `Date` objects. (see https://www.baeldung.com/migrating-to-java-8-date-time-api)

src/main/java/org/broadinstitute/consent/http/service/FileStorageObjectService.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Map;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+public class FileStorageObjectService {
+
+    GCSService gcsService;
+    FileStorageObjectDAO fileStorageObjectDAO;
+
+    Logger log = LoggerFactory.getLogger(FileStorageObjectService.class);
```
rushtong: See the new-ish `ConsentLogger` for a better way of logging things. We need to be exceptionally careful about logging exceptions as they are sent to Sentry and reported out via slack channels.

src/main/java/org/broadinstitute/consent/http/service/FileStorageObjectService.java
```diff
@@ -0,0 +1,118 @@
...
+
+        BlobId blobId;
+        try {
+            // upload to GCS
+            blobId = gcsService.storeDocument(
+                    content,
+                    mediaType,
+                    UUID.randomUUID());
+        } catch (Exception e) {
+            log.error("Failed to upload user file", e);
```
rushtong: Good example of where we should add some more user-specific info (`createUserId`) so we can trace the error.

src/main/java/org/broadinstitute/consent/http/service/FileStorageObjectService.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    // fetches file from GCS and adds it to the fileStorageObject
+    private void fetchAndPopulateUploadedFile(FileStorageObject fileStorageObject) throws NotFoundException  {
+        try {
+            InputStream document = gcsService.getDocument(fileStorageObject.getBlobId());
+            fileStorageObject.setUploadedFile(document);
+        } catch (Exception e) {
+            log.error("Could not get file from GCS", e);
+            throw e;
```
rushtong: The `NotFoundException` is a case where we would **not** want to capture a sentry logging message. I think that case would qualify for log level warn.

src/main/java/org/broadinstitute/consent/http/models/FileStorageObject.java
```diff
@@ -0,0 +1,164 @@
...
+            + " fso.deleted as fso_deleted ";
+
+    private Integer fileStorageObjectId;
+    private String entityId;
+    private String fileName;
+    private BlobId blobId;
+    private FileCategory category;
+    private String mediaType;
+    private Integer createUserId;
+    private Date createDate;
```
rushtong: Suggest that we make this an `Instant`

src/main/java/org/broadinstitute/consent/http/models/FileStorageObject.java
```diff
@@ -0,0 +1,164 @@
...
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        FileStorageObject fileStorageObject = (FileStorageObject) o;
+        return Objects.equals(fileStorageObjectId, fileStorageObject.fileStorageObjectId) && Objects.equals(entityId, fileStorageObject.entityId) && Objects.equals(fileName, fileStorageObject.fileName) && Objects.equals(blobId, fileStorageObject.blobId) && category == fileStorageObject.category && Objects.equals(mediaType, fileStorageObject.mediaType) && Objects.equals(createUserId, fileStorageObject.createUserId) && Objects.equals(createDate, fileStorageObject.createDate) && Objects.equals(deleted, fileStorageObject.deleted);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(fileStorageObjectId, entityId, fileName, blobId, category, mediaType, createUserId, createDate, deleted);
+    }
```
rushtong: For equals and hashcode, I generally like to keep them simple. The `fileStorageObjectId` should be enough to guarantee equality/uniqueness.

rushtong: I also like apache `EqualsBuilder` and `HashCodeBuilder` for these.

rushtong: Lastly, see the docs on `Objects.hash` ... I think you want `Objects.hashCode`

src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
```diff
@@ -612,4 +617,31 @@ protected DataAccessRequest createDarForCollection(User user, Integer collection
...
+                fileName,
+                category.getValue(),
+                bucketName,
+                gcsFileUri,
+                entityId,
+                createUserId,
+                createDate
+        );
+        return fileStorageObjectDAO.findFileById(newFileStorageObjectId);
+    }
```
rushtong: This is a pattern I want to avoid ... this class should be MUCH smaller than it is. We should be moving most object construction to the actual test class as much as possible.

src/test/java/org/broadinstitute/consent/http/db/FileStorageObjectDAOTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+        assertEquals(2, irbFiles.size());
+        assertTrue(irbFiles.contains(file1));
+        assertTrue(irbFiles.contains(file2));
+
+        assertEquals(1, altDataSharingFiles.size());
+        assertTrue(altDataSharingFiles.contains(file3));
+    }
+
+
```
rushtong: Move those object creation methods here!

src/test/java/org/broadinstitute/consent/http/service/FileStorageObjectServiceTest.java
```diff
@@ -0,0 +1,241 @@
...
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.mockito.MockitoAnnotations.openMocks;
+
+public class FileStorageObjectServiceTest {
+
+    @Mock
+    FileStorageObjectDAO fileStorageObjectDAO;
```
rushtong: These mocks can be private

src/test/java/org/broadinstitute/consent/http/db/FileStorageObjectDAOTest.java
```diff
@@ -0,0 +1,182 @@
...
+        assertEquals(fileName, newFileStorageObject.getFileName());
+        assertEquals(category, newFileStorageObject.getCategory().getValue());
+        assertEquals(BlobId.fromGsUtilUri(gcsFileUri), newFileStorageObject.getBlobId());
+        assertEquals(mediaType, newFileStorageObject.getMediaType());
+        assertEquals(entityId, newFileStorageObject.getEntityId());
+        assertEquals(createUserId, newFileStorageObject.getCreateUserId());
+        assertEquals(
+                createDate.getTime(),
+                newFileStorageObject.getCreateDate().getTime(),
+                86400000); // delta: within one day (86,400,000 ms)
```
rushtong: See earlier comments re: `Date`

src/test/java/org/broadinstitute/consent/http/service/FileStorageObjectServiceTest.java
```diff
@@ -0,0 +1,241 @@
...
+
+    @Test
+    public void testFetchById() throws IOException {
+        String bucket = RandomStringUtils.randomAlphabetic(10);
+        String blob = RandomStringUtils.randomAlphabetic(10);
+
+        FileStorageObject file = new FileStorageObject();
+        file.setBlobId(BlobId.of(bucket, blob));
+
+        String content = RandomStringUtils.randomAlphabetic(10);
```
rushtong: Thinking out loud - this could be a case where we want to test content that would NOT be db appropriate (i.e. `RandomStringUtils.random`) since it is file content, not stored metadata.

rushtong (commented):
Fantastic work! Lots of good thinking here. See feedback inline!

src/main/java/org/broadinstitute/consent/http/models/FileStorageObject.java
connorlbark: Yep! These were autogenerated by intellij so I can change them.

connorlbark: > Objects.hashCode

I think that cannot take in an array of items? It seems like `hash` is appropriate here.

src/main/java/org/broadinstitute/consent/http/service/FileStorageObjectService.java
connorlbark: Hmmm ConsentLogger seems to only be available for the Resource class?

connorlbark:
@rushtong so it seems that you cannot get an instant from the database.... `getObject("...", Instant.class)` throws an exception. I can however, do `getTimestamp("...").toInstant()`. Locally that works fine, but it seems like there are some weird truncation issues when I push it up which make the tests fail. Going to add some delta so that they will be resistant to the truncation... but not feeling great about this approach, if you have any thoughts.

src/main/java/org/broadinstitute/consent/http/service/FileStorageObjectService.java
rushtong: ConsentLogger is an interface, any class can implement it. In this instance, it's more of a Trait than anything else - which is my preference for interfaces.

src/main/java/org/broadinstitute/consent/http/models/FileStorageObject.java
rushtong: > I think that cannot take in an array of items? It seems like `hash` is appropriate here.

Good point, I missed that.

rushtong:
> I can however, do `getTimestamp("...").toInstant()`

That is the right approach

connorlbark:
> > I can however, do `getTimestamp("...").toInstant()`
> 
> That is the right approach

Ok coool I was more unhappy with adding a delta to the comparison, but I changed to comparing at the seconds level and that seems to work fine.

src/test/java/org/broadinstitute/consent/http/db/FileStorageObjectDAOTest.java
```diff
@@ -0,0 +1,202 @@
...
+                fileName,
+                category,
+                gcsFileUri,
+                mediaType,
+                entityId,
+                createUserId,
+                createDate
+        );
+
+        FileStorageObject newFileStorageObject = fileStorageObjectDAO.findFileById(newFileStorageObjectId);
```
otchet-broad, line 50: Excellent work on these tests!  Thank you!

One thing you'll see me do occasionally is inert more than one item because I want to discern if am I finding the object I expect because it's the _only object_ or is the input (_newFileStorageObjectId_) really affecting the outcome.  For this case, I like to think about the constraints I expect the database to apply and how to create coverage for that too.

src/main/resources/changesets/changelog-consent-104.0.xml
```diff
@@ -0,0 +1,31 @@
...
+            <column name="file_name" type="text" >
+                <constraints nullable="false"/>
+            </column>
+            <column name="gcs_file_uri" type="text">
+                <constraints nullable="false"/>
+            </column>
+            <column name="category" type="text"/>
+            <column name="media_type" type="text"/>
+            <column name="create_user_id" type="integer"/>
+            <column name="create_date" type="timestamp"/>
```
otchet-broad: Two notes on this:
1) There's a generic Liquibase "DATETIME" type that will be [automagically converted](https://docs.liquibase.com/concepts/changelogs/attributes/column.html#:~:text=DATE-,DATETIME,-TIME) to the correct type for the database. Not suggesting you make a change here, but it looks like we use both in our changesets.
2) Should the created date ever be null?  Maybe consider a non-null constraint and a default value.

src/main/resources/changesets/changelog-consent-104.0.xml
```diff
@@ -0,0 +1,31 @@
...
+            <column name="entity_id" type="text"/>
+            <column name="file_name" type="text" >
+                <constraints nullable="false"/>
+            </column>
+            <column name="gcs_file_uri" type="text">
+                <constraints nullable="false"/>
+            </column>
+            <column name="category" type="text"/>
+            <column name="media_type" type="text"/>
+            <column name="create_user_id" type="integer"/>
```
otchet-broad: Do we always require the user?  If so, nullable="false" should be a constraint.
Any thoughts on a foreign key constraint like you suggested in my changeset 103?

src/main/resources/changesets/changelog-consent-104.0.xml
```diff
@@ -0,0 +1,31 @@
...
+        <createTable tableName="file_storage_object">
+            <column name="file_storage_object_id" autoIncrement="true" type="int">
+                <constraints primaryKey="true"/>
+            </column>
+            <!-- Could be: Reference ID, Dataset ID column, etc. -->
+            <column name="entity_id" type="text"/>
+            <column name="file_name" type="text" >
+                <constraints nullable="false"/>
+            </column>
+            <column name="gcs_file_uri" type="text">
```
otchet-broad, line 14: Nailed it!  I really like this and all of the changes this brought about.  Well done.

src/main/java/org/broadinstitute/consent/http/db/mapper/FileStorageObjectMapper.java
```diff
@@ -0,0 +1,87 @@
...
+        if (hasColumn(r, "media_type")) {
+            file.setMediaType(r.getString("media_type"));
+        }
+
+        if (hasColumn(r, "create_user_id")) {
+            file.setCreateUserId(r.getInt("create_user_id"));
+        }
+
+        if (hasColumn(r, "create_date")) {
+            Timestamp createDate = r.getTimestamp("create_Date");
```
otchet-broad: Nit. create_Date -> create_date

otchet-broad (commented):
Outstanding improvements to readability.
I've left a few minor comments.

src/test/java/org/broadinstitute/consent/http/db/FileStorageObjectDAOTest.java
connorlbark: solid point!! adding this

src/main/resources/changesets/changelog-consent-104.0.xml
connorlbark: def agree with non-null! adding

src/main/resources/changesets/changelog-consent-104.0.xml
connorlbark: added foreign keys to all of the user_ids!

src/main/java/org/broadinstitute/consent/http/models/FileStorageObject.java
```diff
@@ -0,0 +1,164 @@
...
+import com.google.cloud.storage.BlobId;
+import org.broadinstitute.consent.http.enumeration.FileCategory;
+
+import java.io.InputStream;
+import java.time.Instant;
+import java.util.Objects;
+
+public class FileStorageObject {
+
+    public static final String FILE_STORAGE_OBJECT_QUERY_WITH_FSO_PREFIX =
```
otchet-broad: Is this used?

connorlbark: Oh oops, I added that for future users of the class but now it's outdated. I'll just remove it.

otchet-broad (approved):
Nice work, Connor!

quazi-h (approved):
These changes look good to me. I tested the functionality of the changes by trying to store and retrieve IRB and collaboration documents for different DARs and everything seemed to work as expected. Still having trouble connecting my Postico to my locally running consent DB instance so I'll try to pair with someone to get that sorted and can take a closer look at DB changes in future PRs but so far this is looking great to me.

PR #1829, opened 2022-11-28
[DUOS-2185][risk=no] Add NIH Certification File to Dataset

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2185

Fairly large lift here. FSO Mapper needed to be able to conditionally use `fso_` prefix in order to add FSO objects to the Dataset object, so a separate subclass was made for that. Marshaling/Unmarshaling `Instant` objects (or any new DateTime object) requires custom Gson adapters. Refactored the way we create the `Gson()` object to easily add custom type adapters, and put the new Gson object in locations which otherwise broke without it. More details / thoughts below in comments.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/ConsentModule.java
+++ b/src/main/java/org/broadinstitute/consent/http/ConsentModule.java
@@ -1,8 +1,5 @@
 package org.broadinstitute.consent.http;
 
-import com.google.gson.GsonBuilder;
-import com.google.gson.JsonPrimitive;
-import com.google.gson.JsonSerializer;
 import com.google.inject.AbstractModule;
 import com.google.inject.Inject;
 import com.google.inject.Provides;
@@ -38,7 +35,6 @@ import org.broadinstitute.consent.http.db.UserRoleDAO;
 import org.broadinstitute.consent.http.db.VoteDAO;
 import org.broadinstitute.consent.http.mail.SendGridAPI;
 import org.broadinstitute.consent.http.mail.freemarker.FreeMarkerTemplateHelper;
-import org.broadinstitute.consent.http.models.Institution;
 import org.broadinstitute.consent.http.service.AcknowledgementService;
 import org.broadinstitute.consent.http.service.AuditService;
 import org.broadinstitute.consent.http.service.ConsentService;
@@ -70,7 +66,6 @@ import org.broadinstitute.consent.http.service.dao.DatasetServiceDAO;
 import org.broadinstitute.consent.http.service.dao.VoteServiceDAO;
 import org.broadinstitute.consent.http.service.sam.SamService;
 import org.broadinstitute.consent.http.util.gson.GsonUtil;
-import org.broadinstitute.consent.http.util.gson.InstantTypeAdapter;
 import org.jdbi.v3.core.Jdbi;
 import org.jdbi.v3.gson2.Gson2Config;
 import org.jdbi.v3.gson2.Gson2Plugin;
@@ -78,8 +73,6 @@ import org.jdbi.v3.guava.GuavaPlugin;
 import org.jdbi.v3.sqlobject.SqlObjectPlugin;
 
 import javax.ws.rs.client.Client;
-import java.sql.Timestamp;
-import java.time.Instant;
 
 public class ConsentModule extends AbstractModule {
 
--- a/src/main/java/org/broadinstitute/consent/http/ConsentModule.java
+++ b/src/main/java/org/broadinstitute/consent/http/ConsentModule.java
@@ -1,5 +1,8 @@
 package org.broadinstitute.consent.http;
 
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonPrimitive;
+import com.google.gson.JsonSerializer;
 import com.google.inject.AbstractModule;
 import com.google.inject.Inject;
 import com.google.inject.Provides;
@@ -35,6 +38,7 @@ import org.broadinstitute.consent.http.db.UserRoleDAO;
 import org.broadinstitute.consent.http.db.VoteDAO;
 import org.broadinstitute.consent.http.mail.SendGridAPI;
 import org.broadinstitute.consent.http.mail.freemarker.FreeMarkerTemplateHelper;
+import org.broadinstitute.consent.http.models.Institution;
 import org.broadinstitute.consent.http.service.AcknowledgementService;
 import org.broadinstitute.consent.http.service.AuditService;
 import org.broadinstitute.consent.http.service.ConsentService;
@@ -65,12 +69,17 @@ import org.broadinstitute.consent.http.service.dao.DataAccessRequestServiceDAO;
 import org.broadinstitute.consent.http.service.dao.DatasetServiceDAO;
 import org.broadinstitute.consent.http.service.dao.VoteServiceDAO;
 import org.broadinstitute.consent.http.service.sam.SamService;
+import org.broadinstitute.consent.http.util.gson.GsonUtil;
+import org.broadinstitute.consent.http.util.gson.InstantTypeAdapter;
 import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.gson2.Gson2Config;
 import org.jdbi.v3.gson2.Gson2Plugin;
 import org.jdbi.v3.guava.GuavaPlugin;
 import org.jdbi.v3.sqlobject.SqlObjectPlugin;
 
 import javax.ws.rs.client.Client;
+import java.sql.Timestamp;
+import java.time.Instant;
 
 public class ConsentModule extends AbstractModule {
 
@@ -115,6 +124,7 @@ public class ConsentModule extends AbstractModule {
         jdbi.installPlugin(new SqlObjectPlugin());
         jdbi.installPlugin(new Gson2Plugin());
         jdbi.installPlugin(new GuavaPlugin());
+        jdbi.getConfig().get(Gson2Config.class).setGson(GsonUtil.buildGson());
 
         this.consentDAO = this.jdbi.onDemand(ConsentDAO.class);
         this.counterDAO = this.jdbi.onDemand(CounterDAO.class);
--- a/src/main/java/org/broadinstitute/consent/http/db/FileStorageObjectDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/FileStorageObjectDAO.java
@@ -16,14 +16,16 @@ import java.util.List;
 public interface FileStorageObjectDAO extends Transactional<InstitutionDAO> {
 
     @SqlUpdate(
-        "INSERT INTO file_storage_object "
-            + " (file_name, category, gcs_file_uri, "
-            + " media_type, entity_id, create_user_id,"
-            + " create_date, deleted) "
-            + " VALUES "
-            + " (:fileName, :category, :gcsFileUri, "
-            + " :mediaType, :entityId, :createUserId, "
-            + " :createDate, false) "
+            """
+            INSERT INTO file_storage_object
+            (file_name, category, gcs_file_uri,
+            media_type, entity_id, create_user_id,
+            create_date, deleted)
+            VALUES
+            (:fileName, :category, :gcsFileUri,
+            :mediaType, :entityId, :createUserId,
+            :createDate, false)
+            """
     )
     @GetGeneratedKeys
     Integer insertNewFile(
@@ -37,7 +39,13 @@ public interface FileStorageObjectDAO extends Transactional<InstitutionDAO> {
     );
 
     @SqlUpdate(
-            "UPDATE file_storage_object SET deleted=true, delete_user_id=:deleteUserId, delete_date=:deleteDate WHERE file_storage_object_id = :fileStorageObjectId"
+            """
+            UPDATE file_storage_object
+            SET deleted=true,
+                delete_user_id=:deleteUserId,
+                delete_date=:deleteDate
+            WHERE file_storage_object_id = :fileStorageObjectId
+            """
     )
     void deleteFileById(@Bind("fileStorageObjectId") Integer fileStorageObjectId,
                         @Bind("deleteUserId") Integer deleteUserId,
@@ -45,7 +53,14 @@ public interface FileStorageObjectDAO extends Transactional<InstitutionDAO> {
 
 
     @SqlUpdate(
-            "UPDATE file_storage_object SET gcs_file_uri=:gcsFileUri, media_type=:mediaType, update_user_id=:updateUserId, update_date=:updateDate WHERE file_storage_object_id = :fileStorageObjectId"
+            """
+            UPDATE file_storage_object
+            SET gcs_file_uri=:gcsFileUri,
+                media_type=:mediaType,
+                update_user_id=:updateUserId,
+                update_date=:updateDate
+            WHERE file_storage_object_id = :fileStorageObjectId
+            """
     )
     void updateFileById(
             @Bind("fileStorageObjectId") Integer fileStorageObjectId,
@@ -55,7 +70,13 @@ public interface FileStorageObjectDAO extends Transactional<InstitutionDAO> {
             @Bind("updateDate") Instant updateDate);
 
     @SqlUpdate(
-            "UPDATE file_storage_object SET deleted=true, delete_user_id=:deleteUserId, delete_date=:deleteDate WHERE entity_id = :entityId"
+            """
+            UPDATE file_storage_object
+            SET deleted=true,
+                delete_user_id=:deleteUserId,
+                delete_date=:deleteDate
+            WHERE entity_id = :entityId
+            """
     )
     void deleteFilesByEntityId(@Bind("entityId") String entityId,
                                @Bind("deleteUserId") Integer deleteUserId,
@@ -63,17 +84,32 @@ public interface FileStorageObjectDAO extends Transactional<InstitutionDAO> {
 
 
     @SqlQuery(
-            "SELECT * FROM file_storage_object WHERE file_storage_object_id = :fileStorageObjectId"
+            """
+            SELECT *
+            FROM file_storage_object
+            WHERE file_storage_object_id = :fileStorageObjectId
+            """
     )
     FileStorageObject findFileById(@Bind("fileStorageObjectId") Integer fileStorageObjectId);
 
     @SqlQuery(
-            "SELECT * FROM file_storage_object WHERE entity_id = :entityId AND deleted != true"
+            """
+            SELECT *
+            FROM file_storage_object
+            WHERE entity_id = :entityId AND
+                  deleted != true
+            """
     )
     List<FileStorageObject> findFilesByEntityId(@Bind("entityId") String entityId);
 
     @SqlQuery(
-            "SELECT * FROM file_storage_object WHERE entity_id = :entityId AND category = :category AND deleted != true"
+            """
+            SELECT *
+            FROM file_storage_object
+            WHERE entity_id = :entityId
+                  AND category = :category
+                  AND deleted != true
+            """
     )
     List<FileStorageObject> findFilesByEntityIdAndCategory(@Bind("entityId") String entityId, @Bind("category") String category);
 }
--- a/src/main/java/org/broadinstitute/consent/http/db/FileStorageObjectDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/FileStorageObjectDAO.java
@@ -43,6 +43,17 @@ public interface FileStorageObjectDAO extends Transactional<InstitutionDAO> {
                         @Bind("deleteUserId") Integer deleteUserId,
                         @Bind("deleteDate") Instant deleteDate);
 
+
+    @SqlUpdate(
+            "UPDATE file_storage_object SET gcs_file_uri=:gcsFileUri, media_type=:mediaType, update_user_id=:updateUserId, update_date=:updateDate WHERE file_storage_object_id = :fileStorageObjectId"
+    )
+    void updateFileById(
+            @Bind("fileStorageObjectId") Integer fileStorageObjectId,
+            @Bind("gcsFileUri") String gcsFileUri,
+            @Bind("mediaType") String mediaType,
+            @Bind("updateUserId") Integer updateUserId,
+            @Bind("updateDate") Instant updateDate);
+
     @SqlUpdate(
             "UPDATE file_storage_object SET deleted=true, delete_user_id=:deleteUserId, delete_date=:deleteDate WHERE entity_id = :entityId"
     )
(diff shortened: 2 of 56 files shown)


src/main/java/org/broadinstitute/consent/http/db/mapper/FileStorageObjectMapper.java
```diff
@@ -16,70 +16,82 @@ public class FileStorageObjectMapper implements RowMapper<FileStorageObject>, Ro
...
+            Timestamp deleteDate = r.getTimestamp(addPrefix("delete_date"));
             file.setDeleteDate((Objects.nonNull(deleteDate) ? deleteDate.toInstant() : null));
         }
 
-        if (hasColumn(r, "update_user_id")) {
-            file.setUpdateUserId(r.getObject("update_user_id", Integer.class));
-
+        if (hasColumn(r, addPrefix("update_user_id"))) {
+            Integer id = r.getInt(addPrefix("update_user_id"));
+            file.setUpdateUserId(id.equals(0) ? null : id);
```
connorlbark, line 79: So, the issue here is that, since the value is coerced to an `int` primitive, it cannot be null, so postrgres makes `null -> 0`. Logically, it would make sense to do `getClass("asdf", Integer.class)`... however, for some reason, JDBI didn't like the idea of converting this `int8` (I think) type to an `Integer` and only allowed `Long`. But I kept running into weird issues... and not all of the `user_id` values seemed to have issues mapping to Integer, and instead didn't like being mapped to Long. To avoid mixing user ids of `Long` and `Integer`, I decided doing this would be simpler...

src/main/java/org/broadinstitute/consent/http/db/mapper/FileStorageObjectMapperWithFSOPrefix.java
```diff
@@ -0,0 +1,11 @@
+package org.broadinstitute.consent.http.db.mapper;
+
+import org.broadinstitute.consent.http.models.FileStorageObject;
+import org.jdbi.v3.core.mapper.RowMapper;
+
+public class FileStorageObjectMapperWithFSOPrefix extends FileStorageObjectMapper implements RowMapper<FileStorageObject>, RowMapperHelper {
+    @Override
+    public String getPrefix() {
+        return "fso_";
+    }
```
connorlbark, line 10: the story/logical thread behind this class is that:
- I need to be able to map to a `FileStorageObject` class in the Dataset reducer, so I can get the file objects in the dataset.
- I needed to add the `fso_` prefix, because Dataset also has a column named `create_date`
- Since there is custom logic in the FSO mapper, e.g. getting the `BlobId` from the gcs uri, and finding the category enumeration, I couldn't just use `@RegisterBeanMapper(FileStorageObject.class, prefix="fso")`, since it is a "dumb" mapper
- It seems no other model has had this specific issue, and it's impossible to pass parameters to a mapper. To solve these issues, I made a subclass mapper which uses the `fso_` prefix and the original mapper's logic. In the normal mapper, there is no prefix.

src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetReducer.java
```diff
@@ -75,6 +79,25 @@ && hasColumn(rowView, "property_value", String.class)) {
         }
       }
     }
+
+    if (hasColumn(rowView, "fso_file_storage_object_id", Integer.class)
+      && Objects.nonNull(rowView.getColumn("fso_file_storage_object_id", Integer.class))
+    ) {
+      FileStorageObject fileStorageObject = rowView.getRow(FileStorageObject.class);
+
+      switch (fileStorageObject.getCategory()) {
```
connorlbark, line 85: all it takes to add future files to the dataset would be to add a bit of logic into this switch case based upon the category of the file

src/main/java/org/broadinstitute/consent/http/util/gson/GsonUtil.java
```diff
@@ -0,0 +1,19 @@
...
+    public static Gson buildGson() {
+        return gsonBuilderWithAdapters().create();
+    }
+
+    public static GsonBuilder gsonBuilderWithAdapters() {
+        return new GsonBuilder()
+                .registerTypeAdapter(
+                        Instant.class,
+                        new InstantTypeAdapter());
+    }
```
connorlbark: The FSO class is the first instance of us using `Instant` in an object which is converted into Json. We needed to be able to use a custom type converter because Gson doesn't know about. modern DateTime objects. Thus, I created these methods, and updated the default Gson used by consent to use this custom Gson object. There were a bunch of places where it broke because of using the default Gson object... so I updated all of those locations.

src/main/resources/assets/schemas/FileStorageObject.yaml
```diff
@@ -0,0 +1,45 @@
+type: object
```
connorlbark, line 1: right now, all fields are sent, but do we want all of these fields sent to the frontend?

src/main/java/org/broadinstitute/consent/http/util/gson/GsonUtil.java
rushtong: Oooh, this is more than I expected. I'm not happy that Gson doesn't understand this. I learned recently that Gson doesn't like record classes. Might need to reflect on this for a bit.

rushtong:
I recommend reverting the migration scripts - those are not part of the running codebase.

src/main/java/org/broadinstitute/consent/http/db/FileStorageObjectDAO.java
```diff
@@ -43,6 +43,17 @@ void deleteFileById(@Bind("fileStorageObjectId") Integer fileStorageObjectId,
                         @Bind("deleteUserId") Integer deleteUserId,
                         @Bind("deleteDate") Instant deleteDate);
 
+
+    @SqlUpdate(
+            "UPDATE file_storage_object SET gcs_file_uri=:gcsFileUri, media_type=:mediaType, update_user_id=:updateUserId, update_date=:updateDate WHERE file_storage_object_id = :fileStorageObjectId"
+    )
```
rushtong: Minor suggestion, I find this format a little easier to read:
```
    @SqlUpdate("""
        UPDATE file_storage_object
        SET deleted=true, delete_user_id=:deleteUserId, delete_date=:deleteDate
        WHERE file_storage_object_id = :fileStorageObjectId
        """)
```

src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
```diff
@@ -22,6 +22,7 @@
 import org.broadinstitute.consent.http.models.sam.UserStatusDiagnostics;
 import org.broadinstitute.consent.http.models.sam.UserStatusInfo;
 import org.broadinstitute.consent.http.util.HttpClientUtil;
+import org.broadinstitute.consent.http.util.gson.GsonUtil;
```
rushtong: I recommend reverting any changes to files that are not directly affected by FSOs, especially since this is already a large lift. Changes here introduce a need for reviewers to test well outside the scope of the PR.

src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
```diff
@@ -2,9 +2,13 @@
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
 import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.broadinstitute.consent.http.util.gson.GsonUtil;
+import org.broadinstitute.consent.http.util.gson.InstantTypeAdapter;
 
 import java.sql.Timestamp;
+import java.time.Instant;
```
rushtong: The only new import that is in use here is `GsonUtil`

src/main/java/org/broadinstitute/consent/http/ConsentModule.java
```diff
@@ -1,5 +1,8 @@
 package org.broadinstitute.consent.http;
 
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonPrimitive;
+import com.google.gson.JsonSerializer;
```
rushtong: nit: optimize imports

rushtong (commented):
Still reviewing, but wanted to get this feedback out sooner rather than later. I think this is really 2 PRs, one dealing with Gson work and the other with the new file type.

src/main/java/org/broadinstitute/consent/http/db/mapper/FileStorageObjectMapper.java
rushtong: > value is coerced to an int primitive, it cannot be null, so postrgres makes null -> 0

I think this is actually a feature of java.sql, not postgres. Postgres is fine with null numeric values, it's Java that doesn't like null primitives. Another approach is to get the column as an Object and do type coercion if `Objects.nonNull()`

src/main/resources/assets/schemas/FileStorageObject.yaml
rushtong: > right now, all fields are sent, but do we want all of these fields sent to the frontend?

My default answer is `yes` for all fields unless there is a compelling reason not to. There are cases where it doesn't make sense to expose certain model properties in certain cases (I think we do this for signing official names/ids on the user profile page), but those are usually the exception, not the rule.

connorlbark: My only thought here is that maybe the GCS URI is a slight overcommunication to the user? I envision having an endpoint, e.g., `GET /api/fileStorageObject/{id}/content`, that would fetch GCS for the user. Thus, they would never need to know the GCS URI. 
I could also see that it would be considered a security concern to return it, but I am not sure if that's true.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -277,7 +278,7 @@ public Response getDataset(@PathParam("datasetId") Integer datasetId){
             if (Objects.isNull(dataset)) {
                 throw new NotFoundException("Could not find the dataset with id: " + datasetId.toString());
             }
-            return Response.ok(dataset, MediaType.APPLICATION_JSON).build();
+            return Response.ok(unmarshal(dataset)).build();
```
connorlbark, line 281: forced using Gson by calling `unmarshal` on all endpoints which return datasets. I manually checked against develop and it appears to send the exact same object back.

src/main/java/org/broadinstitute/consent/http/models/FileStorageObject.java
```diff
@@ -137,6 +153,18 @@ public void setBlobId(BlobId blobId) {
...
+    /**
+     * Computes the last time this file was changed, be it created, updated, or deleted.
+     * @return The last time the file has been changed.
+     */
+    public Instant getLatestUpdateDate() {
+        List<Instant> dates = new ArrayList<>();
+        dates.add(this.getCreateDate());
+        dates.add(this.getUpdateDate());
+        dates.add(this.getDeleteDate());
+        return dates.stream().filter(Objects::nonNull).max(Instant::compareTo).orElse(this.getCreateDate());
```
otchet-broad: In theory, can't this still return null?  Would it be better to change the signature to make the returned value optional?

src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetReducer.java
```diff
@@ -75,6 +76,25 @@ && hasColumn(rowView, "property_value", String.class)) {
...
+        case NIH_INSTITUTIONAL_CERTIFICATION -> {
+          if (Objects.isNull(dataset.getNihInstitutionalCertificationFile())
+                  || fileStorageObject.getLatestUpdateDate().isAfter(dataset.getNihInstitutionalCertificationFile().getLatestUpdateDate())) {
```
otchet-broad, line 88: Can dataset.getNihInstitutionalCertificationFile().getLatestUpdateDate() return null in this block?

src/main/java/org/broadinstitute/consent/http/models/FileStorageObject.java
```diff
@@ -5,10 +5,26 @@
 
 import java.io.InputStream;
 import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
 
 public class FileStorageObject {
 
+    public static final String QUERY_FIELDS_WITH_FSO_PREFIX =
```
otchet-broad, line 16: Would this be better suited in the DatasetDAO?  It looks like that's the only place this is used.
Consider @rushtong 's comment about using text blocks.  He left that one on a PR of mine too and I think it really does help with readability.

src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetReducer.java
connorlbark: Since there is a non-null constraint in the database for a create date, there should never be a null case here. Happy to add an optional for extra safety though

src/main/java/org/broadinstitute/consent/http/models/FileStorageObject.java
connorlbark: Agreed on the multi-line string: seems to be a new pattern in Consent which I like. As far as keeping this here or in the dataset dao, I think it should be kept here, because it's intended use is across Consent. If we added FSO objects to any other objects, they would need this string. Plus, there are other examples of other models with a similar static constant on them.

src/main/java/org/broadinstitute/consent/http/models/FileStorageObject.java
rushtong: Nit, prefer construction over mutation. In general, I like constructors in this kind of scenario. `List.of()` does that nicely and avoids add calls. The second thing to consider here is that if you are using temporary storage variables, there is usually a more concise/functional way to do what you want. In this case, you can replace the list construction with `Stream.of()` instead, something like:
```
        return Stream.of(
                        this.getCreateDate(),
                        this.getUpdateDate(),
                        this.getDeleteDate())
                .filter(Objects::nonNull)
                .max(Instant::compareTo)
                .orElse(this.getCreateDate());
```

rushtong: To follow up on Elliot's feedback, yes, it's true that there is a non-null so you're guaranteed a create date. Create date is already in your list, so the `.orElse(this.getCreateDate())` condition is moot. You can probably fix both concerns by using a new instant (now) in the or else case.

src/main/resources/assets/schemas/FileStorageObject.yaml
rushtong: > My only thought here is that maybe the GCS URI is a slight overcommunication to the user?

Our GCS bucket is protected by service account access which minimizes the risk. Also, we currently return full path information for Consent files, for example:

Consents:
```
{
  "consentId": "eac1d4f9-78c9-4c88-9b10-9d692e171b5b",
  "lastElectionStatus": "Closed",
  "lastElectionArchived": false,
  "requiresManualReview": false,
  "dataUseLetter": "https://storage.googleapis.com/{bucket}/{filename}",
  ...
}
```

For DARs, we return the last part of that url which, if the user is well informed, could sort it out.
```
{
      ...
      "itDirector": "Test",
      "collaborationLetterLocation": "216f516b-c5f0-474f-8e67-c711979e2b26",
      "collaborationLetterName": "err.json",
      ...
}
```

However, I think you're correct, we should strive to minimize this kind of info. Let's pull that from the response.

src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
```diff
@@ -615,4 +624,30 @@ protected DataAccessRequest createDarForCollection(User user, Integer collection
...
+                fileName,
+                category.getValue(),
+                bucketName,
+                gcsFileUri,
+                entityId,
+                createUser.getUserId(),
+                createDate
+        );
+        return fileStorageObjectDAO.findFileById(newFileStorageObjectId);
+    }
```
rushtong: This is a pattern I want to avoid. We should move all of this code to the test class that needs it. The anti-pattern we've fallen into here is that this helper class has rapidly expanded to become an uber-service class. I've been removing as much creation code as possible from this class as I get to it.

src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
```diff
@@ -145,6 +147,131 @@ public void testFindNeedsApprovalDataSetByDataSetId() {
...
+    public void testGetNIHInstitutionalFile_AlwaysLatestUpdated() throws InterruptedException {
+        Dataset dataset = createDataset();
+
+
+        FileStorageObject nihFileCreatedFirstUpdatedSecond = createFileStorageObject(
+                dataset.getDataSetId().toString(),
+                FileCategory.NIH_INSTITUTIONAL_CERTIFICATION
+        );
+
+        Thread.sleep(10);
```
rushtong: I would prefer not to introduce sleeps to general tests. One way to get around this would be to create a local file storage insert method here in your test class that takes an Instant as a create date. That way, you could create instants of different times and use those instead of sleeping to force time passing.

src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
connorlbark: Understood - I moved it back because I started using the `createFileStorage` function in multiple DAO test classes. Where is the correct place for the code, given that it now is used across DAO classes? Is it recommended to copy the code to both DAO classes?

src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
```diff
@@ -145,6 +147,131 @@ public void testFindNeedsApprovalDataSetByDataSetId() {
...
+        );
+        FileStorageObject nihFile = createFileStorageObject(
+                dataset.getDataSetId().toString(),
+                FileCategory.NIH_INSTITUTIONAL_CERTIFICATION
+        );
+        createFileStorageObject();
+        createFileStorageObject(
+                dataset.getDataSetId().toString(),
+                FileCategory.DATA_USE_LETTER
+        );
```
rushtong, line 172: Here and again on line 154, it would be nice to have some commentary about why these creates are here. Additionally, if we are creating these for some reason, we should have some asserts that ensure what it is that we do not want to see, i.e., there should be no data use letter file object on the dataset in this case.

src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
rushtong: They should be local to the test class that uses them, so for example, `FileStorageObjectDAOTest` and in `DatasetDAOTest`. I'm not concerned about DRY in tests, and in fact, DRY in test code can often lead to flaky tests. For example, if I ever needed to make a change to this in `DAOTestHelper` for some other test class, then I would be side-effecting other unit tests which is problematic. The real point of `DAOTestHelper` is to set up the application for startup and shutdown, but that's where it should end, really.

src/main/java/org/broadinstitute/consent/http/db/mapper/FileStorageObjectMapperWithFSOPrefix.java
rushtong: Took me a while to get this, thank you - I think this is the best approach given the usage. Conceptually, I would have preferred a way to invoke the mapper with different contexts (i.e. prefixes) in different queries, but I don't think that's technically possible in jdbi.

src/main/java/org/broadinstitute/consent/http/enumeration/DatasetPropertyType.java
```diff
@@ -3,6 +3,7 @@
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonSyntaxException;
+import org.broadinstitute.consent.http.util.gson.GsonUtil;
```
rushtong: unused

rushtong (commented):
Finished with this round of reviews, see feedback inline. Thank you for the changes so far!

src/main/java/org/broadinstitute/consent/http/models/FileStorageObject.java
connorlbark: So the weird thing about `Stream.of` and `List.of` is that they both have `@NonNull` constraints... the reason I took the approach I did was because of the fact that `List.of` was throwing me null pointer errors when there was no delete or update date. But, strangely, `Stream.of` seems prefectly happy to take `Stream.of(null, null, null)`, even though it has `@NonNull`. Strange... I like this approach a lot more though.

src/main/resources/assets/schemas/FileStorageObject.yaml
connorlbark: Since we're pulling it from the response, do we want to keep the ability to serialize `BlobId` objects? There is a custom serializer/deserializer I wrote; is it worth keeping it in case we send it in the future, or go ahead and rip it?

rushtong: > Since we're pulling it from the response, do we want to keep the ability to serialize `BlobId` objects? There is a custom serializer/deserializer I wrote; is it worth keeping it in case we send it in the future, or go ahead and rip it?

Missed this yesterday - I would leave it in place for now.

src/test/java/org/broadinstitute/consent/http/resources/TDRResourceTest.java
```diff
@@ -113,7 +114,7 @@ public void testGetDatasetByIdentifier() {
         Response r = resource.getDatasetByIdentifier(new AuthUser(), "DUOS-00003");
 
         assertEquals(200, r.getStatus());
-        assertEquals(d, r.getEntity());
+        assertEquals(GsonUtil.buildGson().toJson(d), r.getEntity());
```
rushtong, line 117: Not required for this PR, just thinking out loud. Another pattern that is helpful in this kind of case is to create an interface with default methods. That is, in essence, a `trait`, something Java doesn't have natively. Let's say we had this:
```
interface GsonUtil {
    default Gson gson() { ... make our custom gson here ... }
}
```
Then this class implements the interface and gets the `gson()` method automatically. We use this pattern for logging and also for mock server construction which helps clean up constructors like this just a tad.

rushtong (approved):
Thank you for the changes! 👍🏽

src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetReducer.java
otchet-broad: Fair enough. I tend to be on the side of expecting those values (null) and defending against them.  YMMV.

src/main/resources/assets/schemas/FileStorageObject.yaml
otchet-broad: Good discussion.  I agree with the strategy of not exposing our internals (e.g. what cloud provider we may be using, etc) in responses.  Nice work!

otchet-broad (approved):
Great work Connor.  Thanks for the tests and implementing various elements from the feedback!

PR #1833, opened 2022-11-30
[DUOS-2208][risk=no] Wrap all nih related user updates in a single transaction

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2208

## Changes
* Wrap all updates related to era commons/nih account information in a single transaction
* Add logging and some additional validation for visibility
* Minor renaming/refactoring for clarity
* Add a constraint to ensure unique user id/property key for user properties

----

- Verify all tests go green


src/main/resources/changesets/changelog-consent-105.0.xml
```diff
@@ -0,0 +1,11 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <changeSet id="105.0" author="grushton">
+        <sql>
+            ALTER TABLE user_property
+            ADD CONSTRAINT uc_user_property_userid_keyval
+            UNIQUE (userid, propertykey);
+        </sql>
```
connorlbark, line 9: would there be an advantage/disadvantage to using [`<addUniqueConstraint... />`](https://docs.liquibase.com/change-types/add-unique-constraint.html) here over raw SQL? it looks like it has support for multiple column unique constraints.

src/test/java/org/broadinstitute/consent/http/service/NihServiceTest.java
```diff
@@ -47,43 +46,64 @@ public void setUp() throws Exception {
...
-        when(researcherService.updateProperties(any(), any(),any())).thenReturn(props);
+        when(researcherService.describeUserProperties(any())).thenReturn(props);
+        User user = new User();
+        user.setUserId(1);
+        when(userDAO.findUserById(any())).thenReturn(user);
         initService();
         try {
             List<UserProperty> properties = service.authenticateNih(nihUserAccount, authUser, 1);
             assertEquals(1, properties.size());
             assertEquals(Integer.valueOf(1), properties.get(0).getPropertyId());
```
connorlbark, line 78: should check: `serviceDAO.updateUserNihStatus` is called with user that has updated properties.

src/test/java/org/broadinstitute/consent/http/service/NihServiceTest.java
```diff
@@ -47,43 +46,64 @@ public void setUp() throws Exception {
...
     public void testAuthenticateNih_BadRequest() {
+        User user = new User();
+        user.setUserId(1);
+        when(userDAO.findUserById(any())).thenReturn(user);
         nihUserAccount.setNihUsername("");
         initService();
         service.authenticateNih(nihUserAccount, authUser, 1);
     }
 
+    @Test (expected = BadRequestException.class)
```
connorlbark: should there be another test when `userDAO.findUserById(any())` returns `null`?

connorlbark (commented):
looks great! much stronger code than before. couple thoughts on cleaning up

src/main/java/org/broadinstitute/consent/http/resources/NihAccountResource.java
```diff
@@ -32,6 +34,10 @@ public NihAccountResource(NihService nihService, UserService userService) {
     @Produces("application/json")
     @PermitAll
     public Response registerResearcher(NIHUserAccount nihAccount, @Auth AuthUser authUser) {
+        if (Objects.isNull(nihAccount) || Objects.isNull(nihAccount.getStatus()) || Objects.isNull(nihAccount.getEraExpiration())) {
```
otchet-broad: Nice early return!

src/main/java/org/broadinstitute/consent/http/service/NihService.java
```diff
@@ -2,40 +2,55 @@
...
+    public NihService(ResearcherService researcherService, UserDAO userDAO, NihServiceDAO serviceDAO) {
         this.researcherService = researcherService;
-        this.libraryCardDAO = libraryCardDAO;
         this.userDAO = userDAO;
+        this.serviceDAO = serviceDAO;
     }
 
     public List<UserProperty> authenticateNih(NIHUserAccount nihAccount, AuthUser authUser, Integer userId) throws BadRequestException {
+        // fail fast
+        if (Objects.isNull(nihAccount) || Objects.isNull(nihAccount.getStatus()) || Objects.isNull(nihAccount.getEraExpiration())) {
```
otchet-broad: The same set of conditional checks are also used in NihAccountResource.registerResearcher.  You might consider extracting this into a public method in the service so both can use the same logic.

src/main/java/org/broadinstitute/consent/http/service/dao/NihServiceDAO.java
```diff
@@ -0,0 +1,45 @@
...
+    private final Jdbi jdbi;
+
+    @Inject
+    public NihServiceDAO(Jdbi jdbi) {
+        this.jdbi = jdbi;
+    }
+
+    public void updateUserNihStatus(User user, NIHUserAccount nihAccount) throws IllegalArgumentException {
+        // fail fast
+        if (Objects.isNull(nihAccount) || Objects.isNull(nihAccount.getStatus()) || Objects.isNull(nihAccount.getEraExpiration())) {
```
otchet-broad, line 29: Same check as in the NihService.

src/test/java/org/broadinstitute/consent/http/service/dao/NihServiceDAOTest.java
```diff
@@ -0,0 +1,150 @@
...
+        assertTrue(cards.isEmpty());
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testUpdateUserNihStatus_nullAccount() {
+        User user = createUser();
+        serviceDAO.updateUserNihStatus(user, null);
+    }
+
+    @Test(expected = Exception.class)
```
otchet-broad, line 136: I was bitten by this pattern, so I tend to avoid declaring a test with an expected exception.
In my case, it was one of the setup objects that started throwing an exception and the test passed "as designed."  It turned out to be a critical failure in the system and I decided to move back to explicitly wrapping the method(s) I expected to throwing the exception with try/catch blocks.  I'm not suggesting you need to change what you're doing, just explaining why I've avoided this pattern.  Let me know if I need to make changes on my end.

otchet-broad (approved):
Nice work and thanks for the discussion after standup.  I left a few notes, for your consideration but none require action.

At standup we discussed the fact that the client is providing this information and it is for display only.  It's not being used in an access decision.  If it were, we should probably consider using our backend obtain this information independently so that the front end (or end user) can't manipulate it.

connorlbark (approved):
since none of my comments require action, i will also approve !

src/test/java/org/broadinstitute/consent/http/service/dao/NihServiceDAOTest.java
rushtong: This is a great point. I think that this is also not supported in junit5 so we should move away from it.

src/main/resources/changesets/changelog-consent-105.0.xml
rushtong: In theory, there should be no difference. Under the hood, Liquibase turns the xml into sql.

rushtong:
FYI @otchet-broad @connorlbark - I did make a few more changes and implemented some feedback. PTAL and let me know if you have any further feedback, thank you!

connorlbark (approved):
👍  looks a lot cleaner - I noticed you're calling `validateNihUserAccount` at the resource and the service level, but it makes sense to me given that either could be called separately.

PR #1834, opened 2022-11-30
[DUOS-2205][risk=low] Support a user who has multiple chairperson or membership roles.  Allow updating DAC on a dataset.

otchet-broad:
[DUOS-2205](https://broadworkbench.atlassian.net/browse/DUOS-2205) was a user filed bug report.  The user reported they were not able to see the dataset name on a DAR request they were voting on.  In analyzing the front end for rendering, it was observed that the backend was not returning the complete set of User Roles and dataset the user was responsible for in DAC duties. 

Root cause determination was that code considered a User Role equal with only the same user and role.  This was insufficient because it did not include the DAC the role might pertain to.  Tests were added to verify this fact and validate the code fix would support this use case.

While performing end-to-end testing for this feature, it was noted that dataset updates were not setting an updated DAC ID when changed in the UI.  This PR includes a fix for that as well.

This PR:
1) Ensures UserRole equivalence validation includes the user, the role, and the DAC ID.
2) Adds support for updating the DAC ID on a dataset.
3) Updates git-ignore so it excludes dependency-reduced-pom.xml

----

- Verify all tests go green

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -154,3 +154,6 @@ automation/.bsp/*
 /config/
 /config_postgres/
 /env.properties
+
+## Shade
+dependency-reduced-pom.xml
--- a/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
@@ -184,14 +184,15 @@ public interface DatasetDAO extends Transactional<DatasetDAO> {
             + "WHERE dataset_id = :datasetId")
     void updateDatasetNeedsApproval(@Bind("datasetId") Integer datasetId, @Bind("needsApproval") Boolean needsApproval);
 
-    @SqlUpdate(
-            "UPDATE dataset " +
-                " SET name = :datasetName," +
-                    " update_date = :updateDate, " +
-                    " update_user_id = :updateUserId, " +
-                    " needs_approval = :needsApproval, " +
-                    " dac_id = :dacId " +
-                " WHERE dataset_id = :datasetId")
+    @SqlUpdate("""
+                 UPDATE dataset
+                 SET name = :datasetName, 
+                     update_date = :updateDate, 
+                     update_user_id = :updateUserId, 
+                     needs_approval = :needsApproval, 
+                     dac_id = :dacId 
+                 WHERE dataset_id = :datasetId
+                 """)
     void updateDataset(@Bind("datasetId") Integer datasetId,
        @Bind("datasetName") String datasetName, @Bind("updateDate") Timestamp updateDate, @Bind("updateUserId") Integer updateUserId,
        @Bind("needsApproval") Boolean needsApproval, @Bind("dacId") Integer updatedDacId);
--- a/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
@@ -1,14 +1,9 @@
 package org.broadinstitute.consent.http.db;
 
-import java.sql.Timestamp;
-import java.time.Instant;
-import java.util.Collection;
-import java.util.List;
-import java.util.Set;
 import org.apache.commons.lang3.tuple.Pair;
 import org.broadinstitute.consent.http.db.mapper.AssociationMapper;
-import org.broadinstitute.consent.http.db.mapper.DatasetMapper;
 import org.broadinstitute.consent.http.db.mapper.DatasetDTOWithPropertiesMapper;
+import org.broadinstitute.consent.http.db.mapper.DatasetMapper;
 import org.broadinstitute.consent.http.db.mapper.DatasetPropertyMapper;
 import org.broadinstitute.consent.http.db.mapper.DatasetReducer;
 import org.broadinstitute.consent.http.db.mapper.DictionaryMapper;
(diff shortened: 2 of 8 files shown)


src/main/java/org/broadinstitute/consent/http/models/User.java
```diff
@@ -279,7 +279,7 @@ public void addRole(UserRole userRole) {
             this.setRoles(new ArrayList<>());
         }
 
-        if (!this.getRoles().stream().map(UserRole::getRoleId).collect(Collectors.toList()).contains(userRole.getRoleId())) {
+        if (!this.getRoles().contains(userRole)) {
```
connorlbark, line 282: this is so sneaky, wow. I think I wrote this condition a while back, impressive catch.

src/main/java/org/broadinstitute/consent/http/models/UserRole.java
```diff
@@ -79,7 +79,7 @@ public void setDacId(Integer dacId) {
 
     @Override
     public int hashCode() {
-        return Objects.hashCode(this.userRoleId, this.name, this.dacId);
+        return Objects.hashCode(this.userId, this.name, this.dacId);
```
connorlbark, line 82: why not include both userId and userRoleId?

.gitignore
```diff
@@ -154,3 +154,6 @@ automation/.bsp/*
 /config/
 /config_postgres/
 /env.properties
+
+## Shade
+dependency-reduced-pom.xml
```
connorlbark, line 159: what does this do?

connorlbark (approved):
was able to add myself to two DAC chair roles at the same time using the correct endpoints. minor comments aside - looks great

.gitignore
otchet-broad: Great question!  If you were going to give our shaded JAR/WAR to someone else, this pom conveys the minimal set of dependencies they would need to include to leverage our shaded archive..
There's a great discussion on SO about this: https://stackoverflow.com/questions/22904573/what-is-the-purpose-of-dependency-reduced-pom-xml-generated-by-the-shade-plugin

I see it as a nice to have kind of thing.  

There's been a request to the shade team to allow specifying the path of this file in the backlog for some time because the current implementation is broken. You can specify the root of the project, but anything else seems to cause breakage.

src/main/java/org/broadinstitute/consent/http/models/UserRole.java
rushtong: I think Elliot has it right. Alternatively, `(this.userId, this.userRoleId, this.dacId)` should also work. Comparing role id and name together is not helpful since if I have two chair roles, both of them will have the same role id and name, they'd only be distinguishable by the dac id.

.gitignore
rushtong: I like having it in the ignore since it is generated by certain maven commands and can show up in your IDE as a new file that looks like it should be added.

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
```diff
@@ -188,9 +189,12 @@ Integer insertDataset(
                 " SET name = :datasetName," +
                     " update_date = :updateDate, " +
                     " update_user_id = :updateUserId, " +
-                    " needs_approval = :needsApproval " +
+                    " needs_approval = :needsApproval, " +
+                    " dac_id = :dacId " +
                 " WHERE dataset_id = :datasetId")
```
rushtong: Not a request, but something to consider - I think sql in text blocks is much easier to read. We should move to that format as needed.

src/test/java/org/broadinstitute/consent/http/db/UserDAOTest.java
```diff
@@ -209,7 +207,6 @@ public void testFindUsersWithLCsAndInstitution() {
         libraryCardDAO.insertLibraryCard(user2.getUserId(), user.getInstitutionId(), "asdf", user.getDisplayName(), user.getEmail(), user.getUserId(), new Date());
 
         List<User> users = userDAO.findUsersWithLCsAndInstitution();
-        System.out.println(users.stream().map((u) -> u.getUserId()).collect(Collectors.toList()));
```
rushtong, line 212: Nice

src/main/java/org/broadinstitute/consent/http/models/User.java
rushtong: I've recently started moving away from list `contains` due to performance issues compared to hash `contains`. It should not be an issue here, just pointing it out.

src/main/java/org/broadinstitute/consent/http/models/UserRole.java
otchet-broad: Good question.  We could, but I think we can do a refactor to get rid of the userRoleId all together.  I'd like to consolidate and reduce all of the user_role table actions into the UserRoleDAO.

rushtong (approved):
Solid work, great sleuthing 👍🏽

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
otchet-broad: Have an exemplar in our codebase?

rushtong: [DatasetServiceDAO](https://github.com/DataBiosphere/consent/blob/744c6b2838875d82f4a9fd8982ef48ac887d1549/src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java)

otchet-broad: Updated this one a text block.  Let me know what you think.

PR #1837, opened 2022-12-01
[DUOS-2185][risk=no] Use custom Gson adapters

connorlbark:
### Partially Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2185

This is only one component of 2185 - for the ticket, custom Gson adapters needed to be added. This PR just has the Gson adapters and updates the `Gson` object of all of the places (I believe) that broke due to the addition of the FileStorageObject. This needs to be merged first before the second half of the ticket can be merged.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
@@ -2,7 +2,9 @@ package org.broadinstitute.consent.http.models;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
 import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.broadinstitute.consent.http.util.gson.GsonUtil;
 
 import java.sql.Timestamp;
 import java.util.HashSet;
@@ -75,7 +77,7 @@ public class DarCollection {
   }
 
   public DarCollection deepCopy() {
-    Gson gson = new Gson();
+    Gson gson = GsonUtil.buildGson();
     String json = gson.toJson(this);
     return gson.fromJson(json, DarCollection.class);
   }
@@ -190,6 +192,6 @@ public class DarCollection {
 
   @Override
   public String toString() {
-    return new Gson().toJson(this);
+    return GsonUtil.buildGson().toJson(this);
   }
 }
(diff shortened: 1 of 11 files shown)


quazi-h (approved):
The logic in the new Gson adapters looks pretty straightforward and the tests look good to me.

I was attempting to test the endpoints where they were implemented but was getting errors that I wasn't passing in the correct parameters. I don't think that was due to your code though, I just must not be putting in the DAC attributes correctly. For a quick test for example I was trying to use the updateDac endpoint to just update the name/description of a dac and only provided those two params along with dacId but got the 400 response: DAC is required. I was observing the same against develop though, confirming this is not your changes - do you know what I might be doing wrong here?

quazi-h:
Hey Connor, @otchet-broad took a quick look at this and noticed that the updateDac functionality was working fine from the UI but when trying from swagger it also failed but that's because it's one of the affected endpoints listed under invalid - I think we are not passing the dac object when submitting through swagger but there's a ticket in the backlog to address this and the fact that it was working from the UI confirms your code is working great.

otchet-broad:
Nice work Connor!  I would recommend test coverage on the GsonUtil as well.  A couple of tests ought to do it (one for Instant, one for BlobId, one for something with both of those elements and one for something without those elements).

src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
```diff
@@ -2,7 +2,9 @@
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
```
rushtong: unused

rushtong (commented):
This looks great, thank you Connor! Same feedback as Elliot noted for some test coverage of `GsonUtil`. Outside of that, 👍🏽

src/test/java/org/broadinstitute/consent/http/util/gson/GsonUtilTest.java
```diff
@@ -0,0 +1,83 @@
+package org.broadinstitute.consent.http.util.gson;
```
otchet-broad, line 1: Thanks for adding this coverage!  Nice work!

otchet-broad (approved):
Great work!

PR #1838, opened 2022-12-02
[DUOS-699][risk=no] Use Gson as Default Json Provider

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-699

I don't think it's possible to remove Jackson entirely, in the sense that it is no longer a dependency. However, you can "mask over it" and make Jersey always use Gson instead, which I have done here. I have extensively checked the UI and was able to fix all of the issues I ran into; seems like the UI was robust enough to be able to handle the minor differences in how Gson and Jackson serialize.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
@@ -53,7 +53,7 @@ public class MetricsService {
   }
 
   public class DarMetricsSummary {
-    @JsonProperty final Timestamp updateDate;
+    final Timestamp updateDate;
     @JsonProperty final String projectTitle;
     @JsonProperty final String darCode;
     @JsonProperty final String nonTechRus;
--- a/src/test/java/org/broadinstitute/consent/http/util/gson/GsonUtilTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/util/gson/GsonUtilTest.java
@@ -9,6 +9,8 @@ import org.junit.Test;
 import org.testcontainers.shaded.org.apache.commons.lang3.RandomStringUtils;
 
 import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.time.temporal.TemporalUnit;
 
 import static org.junit.Assert.assertEquals;
 
@@ -33,11 +35,11 @@ public class GsonUtilTest {
 
         String instantAsJsonString = gson.toJson(instant);
 
-        assertEquals("\"" + instant.toString() + "\"", instantAsJsonString);
+        assertEquals(Long.toString(instant.toEpochMilli()), instantAsJsonString);
 
         Instant parsed = gson.fromJson(instantAsJsonString, Instant.class);
 
-        assertEquals(instant, parsed);
+        assertEquals(instant.truncatedTo(ChronoUnit.MILLIS), parsed);
     }
 
     @Test
@@ -70,13 +72,13 @@ public class GsonUtilTest {
 
         JsonObject parsedJsonObj = gson.fromJson(fsoAsJsonString, JsonObject.class);
 
-        assertEquals(fso.getCreateDate().toString(), parsedJsonObj.get("createDate").getAsString());
+        assertEquals(fso.getCreateDate().toEpochMilli(), parsedJsonObj.get("createDate").getAsLong());
         assertEquals(fso.getBlobId().toGsUtilUri(), parsedJsonObj.get("blobId").getAsString());
         assertEquals(fso.getFileName(), parsedJsonObj.get("fileName").getAsString());
 
         FileStorageObject parsedFso = gson.fromJson(fsoAsJsonString, FileStorageObject.class);
 
-        assertEquals(fso.getCreateDate(), parsedFso.getCreateDate());
+        assertEquals(fso.getCreateDate().truncatedTo(ChronoUnit.MILLIS), parsedFso.getCreateDate());
         assertEquals(fso.getBlobId(), parsedFso.getBlobId());
         assertEquals(fso.getFileName(), parsedFso.getFileName());
     }
--- a/src/test/java/org/broadinstitute/consent/http/util/gson/GsonUtilTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/util/gson/GsonUtilTest.java
@@ -10,8 +10,10 @@ import org.testcontainers.shaded.org.apache.commons.lang3.RandomStringUtils;
 
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
+import java.util.Date;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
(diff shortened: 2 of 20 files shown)


src/main/java/org/broadinstitute/consent/http/util/gson/InstantTypeAdapter.java
```diff
@@ -16,14 +16,14 @@
         implements JsonSerializer<Instant>, JsonDeserializer<Instant> {
     @Override
     public JsonElement serialize(Instant src, Type srcType, JsonSerializationContext context) {
-        return new JsonPrimitive(src.toString());
+        return new JsonPrimitive(src.toEpochMilli());
```
connorlbark, line 19: I changed this to make instants serialize dates as milliseconds, just like our current dates do. Is there any preference to using the more accurate ISO string format that the Instant's `toString()` has or the epoch milliseconds which mirrors our current implementation?

rushtong: I think this is the better choice right now. The complication would be in the UI if there are any calls that are expecting string values and I think we need to look at those instances individually and gauge the work required for updates.

src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
```diff
@@ -53,7 +53,7 @@ public MetricsService(DacService dacService, DatasetDAO dataSetDAO, DataAccessRe
   }
 
   public class DarMetricsSummary {
-    @JsonProperty final Timestamp updateDate;
+    final Timestamp updateDate;
     @JsonProperty final String projectTitle;
```
rushtong, line 57: In a follow on story, we should see if we can completely rip out all `@JsonProperty` annotations. This is a jackson thing and if we aren't using it, we should be able to excise it from dependencies too.

rushtong (approved):
Nice work! 👍🏽

src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
connorlbark: Agreed - I started to do that in this ticket, but then quickly realized it was probably out of the scope.

src/test/java/org/broadinstitute/consent/http/util/gson/GsonUtilTest.java
```diff
@@ -53,4 +57,31 @@ public void testBlobIdJson() {
...
+        JsonObject parsedJsonObj = gson.fromJson(fsoAsJsonString, JsonObject.class);
+
+        assertEquals(fso.getCreateDate().toEpochMilli(), parsedJsonObj.get("createDate").getAsLong());
+        assertNull(parsedJsonObj.get("blobId"));
+        assertEquals(fso.getFileName(), parsedJsonObj.get("fileName").getAsString());
+
+        FileStorageObject parsedFso = gson.fromJson(fsoAsJsonString, FileStorageObject.class);
+
+        assertEquals(fso.getCreateDate().truncatedTo(ChronoUnit.MILLIS), parsedFso.getCreateDate());
+        assertNull(parsedFso.getBlobId()); // should not be parsed, since transient
```
otchet-broad: You should add a test to confirm that deserializing a blobId from JSON blows up.  Why?
This could become an AppSec issue.  A value for blobId should **never** be deserialized from a user.  We can fetch it from our database all day long, because we generated it.

otchet-broad: The test as written relies on the output having not serialized the value.  What if one were present?  Would it be parsed?  Worse yet, could it be stored?

otchet-broad (changes requested):
Nice work!  Requesting a test and possibly one change related to BlobIds.  We should blow up  and not process requests that contain a BlobId - it's more than just asserting null.  We should **never** attempt to deserialize blobIds from the gson/json flow.  IMHO, these are bad requests and should throw an exception.

src/test/java/org/broadinstitute/consent/http/util/gson/GsonUtilTest.java
connorlbark: It would not be parsed; however, I see what you're saying. It has the desired behavior, I'll just add a test to ensure no regressions.

connorlbark: I added an extra test to `models/FileStorageObjectTest` to address this.

connorlbark:
> Nice work! Requesting a test and possibly one change related to BlobIds. We should blow up and not process requests that contain a BlobId - it's more than just asserting null. We should **never** attempt to deserialize blobIds from the gson/json flow. IMHO, these are bad requests and should throw an exception.

I see what you're saying about blowing up if a blobid is received. But imo that's up to the resource and service classes to do, not to Gson. I believe that Gson should ensure it isn't deserialized, then the resource/service functions should do the error handling.

otchet-broad:
> > Nice work! Requesting a test and possibly one change related to BlobIds. We should blow up and not process requests that contain a BlobId - it's more than just asserting null. We should **never** attempt to deserialize blobIds from the gson/json flow. IMHO, these are bad requests and should throw an exception.
> 
> I see what you're saying about blowing up if a blobid is received. But imo that's up to the resource and service classes to do, not to Gson. I believe that Gson should ensure it isn't deserialized, then the resource/service functions should do the error handling.

I think we're saying something very similar to each other.  I'm suggesting that we should ensure that GSON won't deserialize it.  ever.  And that attempts to provide values in this field raise exceptions during deserialization so that our service layers can act appropriately.   If exceptions aren't raised we run the risk of controllers/services acting on these requests, when they shouldn't.  The AppSec risk on this should be updated.  I would suggest moderate because of the damage that could occur if a blobId to a bad document or a document this user wasn't intentionally authorized to use were processed.

connorlbark:
> > > Nice work! Requesting a test and possibly one change related to BlobIds. We should blow up and not process requests that contain a BlobId - it's more than just asserting null. We should **never** attempt to deserialize blobIds from the gson/json flow. IMHO, these are bad requests and should throw an exception.
> > 
> > 
> > I see what you're saying about blowing up if a blobid is received. But imo that's up to the resource and service classes to do, not to Gson. I believe that Gson should ensure it isn't deserialized, then the resource/service functions should do the error handling.
> 
> I think we're saying something very similar to each other. I'm suggesting that we should ensure that GSON won't deserialize it. ever. And that attempts to provide values in this field raise exceptions during deserialization so that our service layers can act appropriately. If exceptions aren't raised we run the risk of controllers/services acting on these requests, when they shouldn't. The AppSec risk on this should be updated. I would suggest moderate because of the damage that could occur if a blobId to a bad document or a document this user wasn't intentionally authorized to use were processed.

I can just remove the blob id handler and Gson will throw an exception since it doesn't recognize the object. I don't necessarily agree on increasing the risk here, because there are no endpoints which take an FSO as input. So this is all theoretical.

otchet-broad:
@connorlbark Thanks again for humoring me with pairing.  I think we landed in a great spot together!

otchet-broad (approved):

rushtong:
We have a form linked from the jira board to find out what the app sec risk level is. In general, file uploads are a concern for them and xml processing in particular.

quazi-h (approved):
Looks good to me, changes to use gson as the default provider looks good, and I think the excceptions thrown in BlobIdTypeAapter are a good safety. Tests look good too and all are passing when running locally.

PR #1839, opened 2022-12-02
[DUOS-2116][risk=medium] Allow SOs and Admins to add/remove user roles.

otchet-broad:
[DUOS-2116](https://broadworkbench.atlassian.net/browse/DUOS-2116) contains rules and conditions for when specific roles should be authorized to change specific user role attributes.  This PR implements those rules.  In doing so, these interfaces support delegating specific operations which were exclusive to administrators to signing officials.

Signing Officials should be able to:
Add the IT Director, Data Submitter, and Signing Official role to users with the same institutional Id.
Add the IT Director, Data Submitter, and Signing Official role to users with no institutional id.  When doing so for these cases, the user's institutional id will be set to the signing offical's institution id.
Remove the IT Director, Data Submitter roles from users with the same and non-null institutional Id.
Remove the Signing Official role from _other_ users with the same and non-null institutional Id. 

This PR also addresses:

- A new pattern for performing actions across DAOs in an SQL Transactional context.
- Improving performance of the UserRoles enumeration helper methods by reducing array creation and evaluation
- Removes a signing official's ability to remove themselves as a signing official via the api/user PUT (updateSelf) method

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -114,6 +117,15 @@ public User updateUserFieldsById(UserUpdateFields userUpdateFields, Integer user
...
     }
 
+    public void insertRoleAndInstitutionForUser(UserRole role, Integer institutionId, Integer userId) {
+        jdbi.useTransaction(transactionHandle -> {
+            UserDAO userDAOT = transactionHandle.attach(UserDAO.class);
+            UserRoleDAO userRoleDAOT = transactionHandle.attach(UserRoleDAO.class);
+            userDAOT.updateInstitutionId(userId, institutionId);
+            userRoleDAOT.insertSingleUserRole(role.getRoleId(), userId);
+        });
+    }
```
rushtong: One thing I'd like to try and maintain is a clean separation between the resource, service, and dao layers. If we could move this into a separate dao class, that would allow for other services to make use of it in the future. I'm open to new organizational patterns on this subject. The current pattern we've followed is to have a service-specific dao class, but I think that might be too limiting. Maybe something along the lines of a package for our traditional single statement transactional dao pattern and a separate package for more complex transactional statements like this one.

rushtong: I'm gonna flip flop on my opinion here. As long as we can keep the sql localized to the DAO classes and out of the service classes, I'm happy with this approach.

src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
```diff
@@ -78,7 +82,7 @@ public void setUp() {
     }
 
     private void initService() {
-        service = new UserService(userDAO, userPropertyDAO, userRoleDAO, voteDAO, institutionDAO, libraryCardDAO, samDAO);
+        service = new UserService(userDAO, userPropertyDAO, userRoleDAO, voteDAO, institutionDAO, libraryCardDAO, samDAO, jdbi);
```
rushtong: A disadvantage of this approach in the service class is that we can't (and shouldn't) have an integration test on the code block that does the multi-statement transaction.

src/test/java/org/broadinstitute/consent/http/service/dao/UserServiceDAOTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Test
+    public void testTransactionRollbackAfterMultipleInserts(){
+        boolean exceptionCaught = false;
+        User testUser = createUser();
+        Institution institution = createInstitution();
+        assertTrue(Optional.ofNullable(testUser.getInstitutionId()).isEmpty());
+        UserRole userRole = new UserRole(UserRoles.SIGNINGOFFICIAL.getRoleId(), UserRoles.SIGNINGOFFICIAL.getRoleName());
+        try{
+            //it's necessary to copy the code in from the service layer because we're testing that the transaction
```
otchet-broad:
```suggestion
            //it's necessary to copy the code in from the service dao layer because we're testing that the transaction
```

src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
```diff
@@ -22,6 +21,18 @@ public enum UserRoles {
 
     private final String roleName;
     private final Integer roleId;
+    private static final HashSet<Integer> LIST_OF_NON_DAC_ROLE_IDS = new HashSet<>(Arrays.asList(
```
rushtong: Optional. I generally lean towards `Lis.of()` since it makes an immutable copy of the elements. No real difference in this case with `Arrays.asList()`

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -258,30 +282,69 @@ public Response addRoleToUser(@Auth AuthUser authUser, @PathParam("userId") Inte
...
-                JsonObject userJson = userService.findUserWithPropertiesByIdAsJsonObject(authUser, userId);
-                return Response.ok().entity(gson.toJson(userJson)).build();
+            if (activeUser.hasUserRole(UserRoles.ADMIN)) {
+                if (!UserRoles.isValidNonDACRoleId(roleId)) {
+                    throw new BadRequestException("Invalid Role Id");
+                }
+                return doDelete(authUser, userId, roleId, activeUser, user);
+            } else if (activeUser.hasUserRole(UserRoles.SIGNINGOFFICIAL)) {
+                if (!UserRoles.isValidSoAdjustableRoleId(roleId)) {
+                    throw new ForbiddenException("A Signing Official may only remove the following role ids: [6, 7, 8] ");
```
rushtong, line 317: We should document this failure case in `api-docs.yaml` ... look near line 1916. One more thing we're doing is trying to slim down this file by moving the method docs into separate yamls in `src/main/resources/assets/paths`. The `User` schema is already defined in `src/main/resources/assets/schemas` so there shouldn't be any additional code for that.

src/test/java/org/broadinstitute/consent/http/service/dao/UserServiceDAOTest.java
```diff
@@ -0,0 +1,74 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.MockitoAnnotations.openMocks;
+
+// This is a utility test to verify a pattern for Database Transactions continues to be supported and works as expected.
+// It should be updated to include new patterns that are developed.
+public class UserServiceDAOTest extends DAOTestHelper {
+
+    UserServiceDAO serviceDAO;
```
rushtong: This can be private

rushtong (commented):
This is very solid, thank you for the changes! See minor comments inline. I think the only thing that is important is the docs one. Love the tests!

src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
connorlbark: I think you can also use `Set.of` for extra simplicity

otchet-broad: Interestingly enough, I've looked at the Java SE Documentation for Java 17 and TIL:
Both List.of and Set.of allow the underlying elements to be modified, even though elements in the collection cannot be added, removed, or replaced.
See: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/List.html and https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Set.html
They're very careful to say the underlying elements may be mutable. `However, if the contained elements are themselves mutable, this may cause the List's contents to appear to change.`

For now, I think I'm going to use Set.of.  Nice suggestion Connor.

rushtong (approved):
Thank you for the changes 👍🏽

connorlbark (approved):
👍 looks great!

PR #1840, opened 2022-12-02
[DUOS-2213][risk=no] Adds support for admins to delete one of their own acknowledgements.

otchet-broad:
Testing DUOS-1758 required modifying code to test more than once.  [DUOS-2213](https://broadworkbench.atlassian.net/browse/DUOS-2213) allows an ADMIN user the ability to remove an acknowledgement.

This PR:

- Adds resource, service, and DAO support for removing an acknowledgement for the currently authenticated & authorized user
- Limits this capability to ADMIN users.
- Updates the API docs to include the new endpoint.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/AcknowledgementDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/AcknowledgementDAO.java
@@ -31,4 +31,6 @@ public interface AcknowledgementDAO extends Transactional<AcknowledgementDAO> {
           + " FROM acknowledgement WHERE user_id = :userId and ack_key IN (<key_list>)")
     List<Acknowledgement> findAcknowledgementsForUser(@BindList("key_list") List<String> keys, @Bind("userId") Integer userId);
 
+    @SqlUpdate("DELETE FROM acknowledgement where user_id = :userId AND ack_key = :key")
+    void deleteAcknowledgement(@Bind("key") String key, @Bind("userId") Integer userId);
 }
--- a/src/main/resources/assets/api-docs.yaml
+++ b/src/main/resources/assets/api-docs.yaml
@@ -1965,6 +1965,31 @@ paths:
           description: No acknowledgement found for the given key.
         500:
           description: Error processing the request.
+    delete:
+      summary: ADMIN ONLY - Delete Acknowledgement by key
+      description: Deletes a specific acknowledgement for the current user by acknowledgement key.  Returns the deleted acknowledgement.
+      parameters:
+        - name: key
+          in: path
+          description: The name of the key.
+          required: true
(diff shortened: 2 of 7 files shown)


src/main/java/org/broadinstitute/consent/http/db/AcknowledgementDAO.java
```diff
@@ -31,4 +31,6 @@ public interface AcknowledgementDAO extends Transactional<AcknowledgementDAO> {
           + " FROM acknowledgement WHERE user_id = :userId and ack_key IN (<key_list>)")
     List<Acknowledgement> findAcknowledgementsForUser(@BindList("key_list") List<String> keys, @Bind("userId") Integer userId);
 
+    @SqlUpdate("DELETE FROM acknowledgement where user_id = :userId AND ack_key = :key")
```
rushtong, line 34: minor nit: 
```suggestion
    @SqlUpdate("DELETE FROM acknowledgement WHERE user_id = :userId AND ack_key = :key")
```

src/main/resources/assets/api-docs.yaml
```diff
@@ -1965,6 +1965,31 @@ paths:
...
+      description: Deletes a specific acknowledgement for the current user by acknowledgement key.  Returns the deleted acknowledgement.
+      parameters:
+        - name: key
+          in: path
+          description: The name of the key.
+          required: true
+          schema:
+            type: string
+      tags:
+        - User
```
rushtong, line 1979: We also have an Admin tag which makes sense for this API

rushtong (approved):
Nice, see minor comments inline 👍🏽

src/main/resources/assets/api-docs.yaml
```diff
@@ -1965,6 +1965,31 @@ paths:
           description: No acknowledgement found for the given key.
         500:
           description: Error processing the request.
+    delete:
+      summary: ADMIN ONLY - Delete Acknowledgement by key
+      description: Deletes a specific acknowledgement for the current user by acknowledgement key.  Returns the deleted acknowledgement.
```
connorlbark, line 1970: any reason only for the current user? it seems plausible that the Admin API would allow deleting anyone's acknowledgement.

connorlbark (approved):
as-is the code looks great and functioned as expected locally.

src/main/resources/assets/api-docs.yaml
otchet-broad: Yes.  There may be compliance issues with doing so. Adding that to the scope of the ticket wasn't needed to verify the correct behavior in your work.  I didn't want/need to open that can of worms at this time.

PR #1853, opened 2022-12-12
[DUOS-2184][risk=no] Add alternative data sharing plan to dataset

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2184

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
@@ -49,11 +49,8 @@ public class Dataset {
 
     private Boolean deletable;
 
-    private String sharingPlanDocument;
-
-    private String sharingPlanDocumentName;
-
     private FileStorageObject nihInstitutionalCertificationFile;
+    private FileStorageObject alternativeDataSharingPlanFile;
 
     private Set<DatasetProperty> properties;
 
@@ -290,21 +287,6 @@ public class Dataset {
         this.deletable = deletable;
     }
 
-    public String getSharingPlanDocument() {
-        return sharingPlanDocument;
-    }
-
-    public void setSharingPlanDocument(String sharingPlanDocument) {
-        this.sharingPlanDocument = sharingPlanDocument;
-    }
-
-    public String getSharingPlanDocumentName() {
-        return sharingPlanDocumentName;
-    }
-
-    public void setSharingPlanDocumentName(String sharingPlanDocumentName) {
-        this.sharingPlanDocumentName = sharingPlanDocumentName;
-    }
 
     /**
      * Checks if the Dataset matches a raw search query. Searches on all dataset properties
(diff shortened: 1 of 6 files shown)


src/main/java/org/broadinstitute/consent/http/models/Dataset.java
```diff
@@ -49,11 +49,8 @@ public class Dataset {
 
     private Boolean deletable;
 
-    private String sharingPlanDocument;
-
-    private String sharingPlanDocumentName;
-
```
connorlbark, line 55: Correct me if I am wrong here, but I didn't see any usages of this field in consent or in the frontend. No datasets on dev had any value for this; in fact, since the value was null, no datasets on dev even had this field. Therefore, it seems safe to remove.

quazi-h (approved):
Overall changes look good to me. We don't have to specify the new alternativeDataSharingPlanFile explicitly in the DatasetDAO queries because it's all being handled within the DatasetReducer, correct? I was looking for a way to test the changes locally - I see the json input for alternativeDataSharingPlanFile under the Update Dataset endpoint but I think our swagger is still a little messed up because I keep getting invalid URL errors when trying to use that endpoint to update the alternative file details.

connorlbark:
> Overall changes look good to me. We don't have to specify the new alternativeDataSharingPlanFile explicitly in the DatasetDAO queries because it's all being handled within the DatasetReducer, correct? I was looking for a way to test the changes locally - I see the json input for alternativeDataSharingPlanFile under the Update Dataset endpoint but I think our swagger is still a little messed up because I keep getting invalid URL errors when trying to use that endpoint to update the alternative file details.

So that endpoint will not update those files. Unfortunately, there is no endpoint to add them. The only way to add them (for now) is to attach to the local postgres instance and add the FSOs manually, which is a bit annoying.

src/main/java/org/broadinstitute/consent/http/models/Dataset.java
```diff
@@ -49,11 +49,8 @@ public class Dataset {
...
     private FileStorageObject nihInstitutionalCertificationFile;
+    private FileStorageObject alternativeDataSharingPlanFile;
```
otchet-broad, line 53: Had you considered modifying Dataset so it holds a set of FileStorageObjects?
The juice might not be worth the squeeze, but I could see where it might make this object DRYer.

otchet-broad (approved):
Nice work and I appreciate the tests! Thank you!
Added one comment for your consideration, but not required to change.

src/main/java/org/broadinstitute/consent/http/models/Dataset.java
connorlbark: I like the idea - I did consider it, but I felt that a) it would return irrelevant files, b) it puts more effort on the frontend to find the relevant file. This approach ensures that ONLY the relevant files are returned and it's easy to for the frontend to get the file that it needs - I am open to other approaches though.

connorlbark:
@otchet-broad @quazi-broad thanks for the 👍 s! FYI i think i will wait for the Gson changes to roll out before merging this, since these codes paths are heavily related and I think that PR handles FSO objects more elegantly.

src/main/java/org/broadinstitute/consent/http/models/Dataset.java
rushtong: Agreed Connor, nice compromise. I also like the idea of a bucket of files, we should think more on that.

PR #1858, opened 2022-12-16
[DUOS-2137][risk=no] Send New Researcher Email to SOs

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2137

Sends an email if and only if the SO went from being null to non-null (only the case the first time it was specified, since the UI does not allow unselecting an SO).

----

- Verify all tests go green


quazi-h (approved):
New logic to check for SO before and after an update of user properties looks good to me. Tests for initial SO designation vs SO change are good.

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
```diff
@@ -111,6 +111,7 @@ private void saveEmailAndResponse(
             Writer template) {
         Instant now = Instant.now();
         Instant dateSent = (Objects.nonNull(response) && response.getStatusCode() < 400) ? now : null;
+        System.out.println(template.toString());
```
otchet-broad: Did you mean to log something here?

src/main/resources/freemarker/new-researcher.html
```diff
@@ -0,0 +1,38 @@
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml" style="font-family: 'Roboto', sans-serif ;">
+<head>
+    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>Broad Data Use Oversight System - Admin - Dataset Approved Notification</title>
```
otchet-broad: That's probably not the correct title for this document.

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
```diff
@@ -96,4 +122,75 @@ public void testSendDataCustodianApprovalMessage() {
...
+                eq(1234),
+                eq(EmailType.NEW_RESEARCHER.getTypeInt()),
+                any(),
+                eq("""
+                <!DOCTYPE html>
+                <html xmlns="http://www.w3.org/1999/xhtml" style="font-family: 'Roboto', sans-serif ;">
+                <head>
+                    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+                    <meta name="viewport" content="width=device-width, initial-scale=1">
+                    <title>Broad Data Use Oversight System - Admin - Dataset Approved Notification</title>
```
otchet-broad: Incorrect title copy/pasta?

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
```diff
@@ -96,4 +122,75 @@ public void testSendDataCustodianApprovalMessage() {
...
+                            <td id="userName" align="left" style="border-collapse: collapse; font-family: 'Roboto', sans-serif ; font-size: 22px; color: #777777; text-align: left; line-height: 21px; display: block; font-weight: 500; padding: 25px 30px 20px 30px;">Hello,</td>
+                        </tr>
+                        <tr width="600" style="font-family: 'Roboto', sans-serif ;">
+                            <td align="left" style="border-collapse: collapse; font-family: 'Roboto', sans-serif ; font-size: 16px; color: #777777; text-align: left; line-height: 25px; padding: 0px 30px 20px 30px;">
+                                <p>A researcher from your institution, John Doe, has registered in DUOS and listed you as their Signing Official. In order to request access to data, they will need to be issued a Library Card. Please log in to DUOS here: Broad Data Use Oversight System and review the terms of the Library Card Agreements for this researcher in your Signing Official Console. If you have any questions, please contact <a href="mailto:duos-support@broadinstitute.zendesk.com">duos-support@broadinstitute.zendesk.com</a>.</p>
```
otchet-broad: Should we include a link to DUOS after "here:"?

otchet-broad (changes requested):
Nice work Connor!

Two things I think need to be addressed: 1) The title in the email template seems to be incorrect for the purpose of _this_ message.  2) The template tells the user to log into DUOS "here:" but doesn't seem to include a link to DUOS.  Should it?  See ~ line 25 in new-researcher.html template.

connorlbark:
> Two things I think need to be addressed: 1) The title in the email template seems to be incorrect for the purpose of _this_ message. 2) The template tells the user to log into DUOS "here:" but doesn't seem to include a link to DUOS. Should it? See ~ line 25 in new-researcher.html template.

Good catches. I was a bit forgetful about updating the template. That should all be fixed, and there should be better styling as well.

src/main/java/org/broadinstitute/consent/http/mail/message/NewResearcherMessage.java
```diff
@@ -0,0 +1,20 @@
+package org.broadinstitute.consent.http.mail.message;
+
+import com.sendgrid.helpers.mail.Mail;
+
+import javax.mail.MessagingException;
+import java.io.Writer;
+
+public class NewResearcherMessage extends MailMessage{
```
rushtong: nit, naming. This is actually a new library card request. Something like `NewResearcherLibraryCardMessage` would be more accurate. Same feedback applies to other similarly named files/methods.

src/main/java/org/broadinstitute/consent/http/mail/SendGridAPI.java
```diff
@@ -212,4 +214,9 @@ public Optional<Response> sendDatasetDeniedMessage(String toAddress, Writer temp
         return sendMessage(message);
     }
 
+    public Optional<Response> sendNewResearcherMessage(String toAddress, Writer template) throws MessagingException {
+        Mail message = newResearcherMessage.newResearcherMessage(toAddress, fromAccount, template);
+        return sendMessage(message);
+    }
```
rushtong: I can't find any usages of this method.

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
```diff
@@ -330,6 +330,20 @@ public void sendDatasetDeniedMessage(User user,
         );
     }
 
+    public void sendNewResearcherMessage(User researcher,
+                                         User signingOfficial) throws Exception {
+        Writer template = templateHelper.getNewResearcherTemplate(researcher.getDisplayName(), this.SERVER_URL);
+        Optional<Response> response = sendGridAPI.sendDatasetDeniedMessage(signingOfficial.getEmail(), template);
```
rushtong: Copy/paste error?

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -88,13 +91,32 @@ public User updateUserFieldsById(UserUpdateFields userUpdateFields, Integer user
             if (Objects.nonNull(userUpdateFields.getEraCommonsId())) {
                 userDAO.updateEraCommonsId(userId, userUpdateFields.getEraCommonsId());
             }
+
+            Integer soIdBeforeUpdate = this.getSigningOfficialForUser(userId);
```
rushtong: nit, unnecessary `this.`, here and on line 104

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -35,6 +35,7 @@
 import java.util.Date;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
```
rushtong, line 38: unused

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -117,6 +139,10 @@ public User updateUserFieldsById(UserUpdateFields userUpdateFields, Integer user
         return findUserById(userId);
     }
 
+    public void notifySigningOfficial(UserUpdateFields update, Integer userId) {
+
+    }
```
rushtong: Unused and empty

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -342,6 +368,29 @@ public void insertUserRoles(List<UserRole> roles, Integer userId) {
...
 
+    private Integer getSigningOfficialForUser(Integer userId) {
+        List<UserProperty> props =
+                userPropertyDAO.findUserPropertiesByUserIdAndPropertyKeys(
+                        userId,
+                        List.of(UserFields.SELECTED_SIGNING_OFFICIAL_ID.getValue())
+                );
+
+        if (props.size() == 0) {
+            return null;
```
rushtong: I generally try to avoid using null as a state (and am guilty of doing this myself). If we find ourselves needing it, try returning `Optional<T>` instead so that callers can know that the method may or may not have a valid value.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -342,6 +368,29 @@ public void insertUserRoles(List<UserRole> roles, Integer userId) {
...
+        }
+
+        UserProperty soIdProp = props.get(0);
+
+        int soId;
+        try {
+            soId = Integer.parseInt(soIdProp.getPropertyValue());
+        } catch (NumberFormatException e) {
+            return null;
+        }
```
rushtong: I think we have another error case here which is if we don't have an actual user matching the id.

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
```diff
@@ -106,10 +131,81 @@ public void testSendDataCustodianApprovalMessage() {
...
+        verify(emailDAO, times(1)).insert(
+                eq("1234"),
+                eq(null),
+                eq(1234),
+                eq(EmailType.NEW_RESEARCHER.getTypeInt()),
+                any(),
+                eq(
+                """
+                <!DOCTYPE html>
+                <html xmlns="http://www.w3.org/1999/xhtml" style="font-family: 'Roboto', sans-serif ;">
```
rushtong: I would lean towards not checking for the actual email content - this leads to brittle testing. The service call really only cares that it send **_some_** message to sendgrid. Looking at the contents of the message is going beyond the boundaries of the tested method call. If you want to test the actual content, we can do that in the test that creates the message itself, i.e. the template helper call.

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
```diff
@@ -106,10 +131,81 @@ public void testSendDataCustodianApprovalMessage() {
...
+
+        User so = new User();
+        user.setEmail("fake_email@asdf.com");
+        try {
+            service.sendNewResearcherMessage(user, so);
+        } catch (Exception e) {
+            fail("Should not fail sending message: " + e);
+        }
+
+        verify(sendGridAPI, times(1)).sendDatasetDeniedMessage(any(), any());
```
rushtong: This is the wrong call to be verifying, see earlier comments.

rushtong (changes requested):
See comments inline, thank you!

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
connorlbark: Hmm, sure. I can move this to a template helper call; I think here makes sense, too, but I don't feel strongly. I do, however, feel strongly that we do check it - before we never did check message content and I always felt weird about adding new emails / changing template content, since it was a manual test process.

rushtong: > I do, however, feel strongly that we do check it

I'm not advocating for NOT checking it, I'm advocating for respecting the boundaries of the current method under test. `service.sendNewResearcherMessage` does not care what the content is, it just delegates the message to another class. The place for testing the html content is the method that generates that content which is `templateHelper.getNewResearcherTemplate`

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
```diff
@@ -106,10 +135,41 @@ public void testSendDataCustodianApprovalMessage() {
...
+        User so = new User();
+        user.setEmail("fake_email@asdf.com");
+        try {
+            service.sendNewResearcherMessage(user, so);
+        } catch (Exception e) {
+            fail("Should not fail sending message: " + e);
+        }
+
+        verify(sendGridAPI, times(1)).sendNewResearcherLibraryRequestMessage(any(), any());
+        verify(templateHelper, times(1)).getNewResearcherLibraryRequestTemplate("John Doe", serverUrl);
```
connorlbark, line 136: I added this in place of checking that the template was right, in addition to the new template test. My main concern was that we don't know if the template was being generated "correctly" with the right arguments... this alleviates that.

src/main/java/org/broadinstitute/consent/http/mail/message/NewResearcherLibraryRequestMessage.java
```diff
@@ -0,0 +1,20 @@
...
+    private final String NEW_RESEARCHER = "New Library Card Request in DUOS";
+
+    public Mail newResearcherLibraryRequestMessage(String toAddress, String fromAddress, Writer template) throws MessagingException {
+        return generateEmailMessage(toAddress, fromAddress, template, null, null);
+    }
+
+    @Override
+    String assignSubject(String referenceId, String type) {
+        return NEW_RESEARCHER;
+    }
```
rushtong, line 19: Not your code, so this is not a request for a change. I'm not thrilled with this pattern of passing in nulls and overriding `assignSubject`. We really should refactor this to be less brittle.

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -18,6 +18,7 @@
 import org.broadinstitute.consent.http.models.Error;
 import org.broadinstitute.consent.http.service.AcknowledgementService;
 import org.broadinstitute.consent.http.service.DatasetService;
+import org.broadinstitute.consent.http.service.EmailService;
```
rushtong: Unused

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -342,6 +364,31 @@ public void insertUserRoles(List<UserRole> roles, Integer userId) {
...
+
+        int soId;
+        try {
+            soId = Integer.parseInt(soIdProp.getPropertyValue());
+        } catch (NumberFormatException e) {
+            return Optional.empty();
+        }
+
+
+
```
rushtong: minor nit: lots of white space

src/test/java/org/broadinstitute/consent/http/mail/freemarker/FreeMarkerTemplateHelperTest.java
```diff
@@ -121,6 +121,55 @@ public void testGetClosedDatasetElectionsTemplate() throws Exception {
         assertTrue(parsedTemplate.getElementById("userName").text().equals("Hello Admin!"));
     }
 
+    @Test
+    public void testGetNewResearcherLibraryRequestTemplate() throws Exception {
+        Writer template = helper.getNewResearcherLibraryRequestTemplate("John Doe", "http://localhost:8000/#/");
+        String templateString = template.toString();
+
+        assertEquals(
```
rushtong: I think this could be a much simpler test. What we really care about here is that we _**don't**_ have the replacement variables and instead we _**do**_ have the replacement text. If we tested that `templateString` did not contain two things and that `templateString` did contain the replacements, you wouldn't need to essentially copy the entire contents of the email here.

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
```diff
@@ -80,14 +96,27 @@ private void initService() {
...
+
+        openMocks(this);
+
+        FreeMarkerConfiguration fmConfig = new FreeMarkerConfiguration();
+        fmConfig.setDefaultEncoding("UTF-8");
+        fmConfig.setTemplateDirectory("/freemarker");
+        templateHelper = spy(new FreeMarkerTemplateHelper(fmConfig));
+        service = new EmailService(collectionDAO, consentDAO, voteDAO, electionDAO, userDAO,
+                emailDAO, sendGridAPI, templateHelper, serverUrl);
     }
```
rushtong: Given that we need two different set ups to test these cases, I think this might be enough to move these tests to a separate test class.

src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
```diff
@@ -135,6 +142,125 @@ public void testUpdateUserFieldsById() {
...
+            UserUpdateFields fields = new UserUpdateFields();
+            fields.setSelectedSigningOfficialId(1);
+
+            assertEquals(1, fields.buildUserProperties(user.getUserId()).size());
+            service.updateUserFieldsById(fields, user.getUserId());
+        } catch (Exception e) {
+            fail(e.getMessage());
+        }
+        // We added 3 user property values, we should have props for them:
+        verify(userDAO, times(0)).updateDisplayName(any(), any());
```
rushtong: Minor suggestion: `times(0)` can also be `never()`

rushtong (commented):
See comments inline, this looks good. I think the biggest concern I have is the complication of `EmailServiceTest` - I think we can clean that up with a separate suite of tests that require a more customized initialization.

src/test/java/org/broadinstitute/consent/http/mail/freemarker/FreeMarkerTemplateHelperTest.java
connorlbark: Ok so I personally like that you have the entire email here. Understandably, the code gets bloated. However, it's very easy to check how it actually renders. If I want to know what the email looks like, I just check this. Otherwise, to test styling, links, etc, I would have to manually put in the replacement variables, which allows for human error and slower testing, and if the email accidentally changes and style breaks, we don't know that. Having the code this way makes it very easy to see how the email renders, and ensure that styling never breaks (unless intentional). Perhaps, we could clean this up by putting this in its own file.

rushtong: > if the email accidentally changes and style breaks, we don't know that

We'll still not know if styles break - in practice, any changes to the template file will be copied here just to make the tests pass. Philosophically, we're going beyond the unit test by expecting things like styling to "work", so to speak. What I think you want is something that tests `new-researcher-library-request.html` and only the content of that file, which is not in the scope of this unit test. The boundary around `getNewResearcherLibraryRequestTemplate` is in two variables, the name and the url, so that's what a unit test around this method should be testing.

connorlbark: >  Philosophically, we're going beyond the unit test by expecting things like styling to "work", so to speak

Sure, but we would still be checking the content of the file right? We would still be rendering the HTML and checking that the content is replaced properly. If it is in the scope of the test to check the raw content of the HTML, I don't see why styling isn't also included with that. I would argue that styling is an aspect of rendering, and here are the tests to ensure proper rendering, styling and replacement both included.

Either way, we don't have any other place to check that... I am ok putting this elsewhere if you prefer. 

As far as copying and pasting, I think that forcing the programmer to copy over the template and change the variables forces them to actively make the decision to not check that the HTML renders properly, which I think would be strange. As opposed to, the template changes, and then the person testing has to manually make those changes to test the final rendered email, if they even notice the diff changes, which is subject to a multitude of human errors. Basically, I feel like there has to be some place that we can make sure the final email is all good, and I don't see where that would be other than here. And I don't feel comfortable _not_ having that for emails going forward, because it adds friction to testing something that actual users in production will see.

Idk though, maybe I'm off base here. If you feel strongly, I can change it to match other tests in this class.

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
connorlbark: Turns out i was being silly and I didn't need two different setups, so this should be all good now.

src/test/java/org/broadinstitute/consent/http/mail/freemarker/FreeMarkerTemplateHelperTest.java
rushtong: I'm not against testing the content - I'm advocating for proper testing boundaries around methods. What I think you're doing here is conflating the need for testing style with the right place to make that test. The unit test of `getNewResearcherLibraryRequestTemplate` doesn't have any concerns about the styling of the template. By adding additional styling logic to a test for **_that_** method makes the test do double duty which is what I'm trying to avoid. 

My suggestion is to isolate the testing for this method to what the method cares about. Secondly, if we really want to test the styling, then create a separate test that reads `new-researcher-library-request.html`, does variable interpolation on it, and ensures that it works the way we want it to.

src/test/java/org/broadinstitute/consent/http/mail/freemarker/FreeMarkerTemplateHelperTest.java
```diff
@@ -125,48 +126,25 @@ public void testGetClosedDatasetElectionsTemplate() throws Exception {
...
         String templateString = template.toString();
+        final Document parsedTemplate = getAsHtmlDoc(templateString);
+
+        assertTrue(parsedTemplate.title().equals("Broad Data Use Oversight System - New Researcher Library Request"));
+
+        assertTrue(
+                parsedTemplate
+                        .getElementById("content")
+                        .text()
+                        .contains("A researcher from your institution, John Doe, has registered in DUOS"));
```
rushtong, line 137: I think this is a great solution to previous concerns!

rushtong (approved):
Thank you for the changes! 👍🏽

PR #1862, opened 2022-12-30
[DUOS-2061][risk=low] Add Mail API.

otchet-broad:
[DUOS-2061](https://broadworkbench.atlassian.net/browse/DUOS-2061) specifies a new mail API be added so administrators can view metadata for and content of messages sent by the DUOS system.

Work in this pull request:
1) Adds the new MailResource
2) Updates the EmailService and MailMessageDAO to support fetching query results
3) Adds Swagger documentation and test coverage for the changes.

Item to note:
Since the API included adding a limit to the number of results returned, an offset option was added as well.  This offset option enables paging of results back to a client.

----
Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

[DUOS-2061]: https://broadworkbench.atlassian.net/browse/DUOS-2061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DUOS-2061]: https://broadworkbench.atlassian.net/browse/DUOS-2061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1864, opened 2022-12-30
[DUOS-2112][risk=low] Add support for searching mail sent by date range.

otchet-broad:
[DUOS-2112](https://broadworkbench.atlassian.net/browse/DUOS-2112) builds on work completed as part of [DUOS-2061](https://github.com/DataBiosphere/consent/pull/1862) and makes the following enhancements:

1. Endpoint and class updates for searching mail by date range added.
2. Documentation and tests added to describe and validate the feature.
3.  Limit and Offset are now validated to be greater than -1 for both searching by EmailType and Date Range.

----

- Verify all tests go green

[DUOS-2112]: https://broadworkbench.atlassian.net/browse/DUOS-2112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DUOS-2112]: https://broadworkbench.atlassian.net/browse/DUOS-2112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DUOS-2061]: https://broadworkbench.atlassian.net/browse/DUOS-2061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/MailMessageDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/MailMessageDAO.java
@@ -11,6 +11,7 @@ import org.jdbi.v3.sqlobject.transaction.Transactional;
 
 import javax.annotation.Nullable;
 import java.time.Instant;
+import java.util.Date;
 import java.util.List;
 
 @RegisterRowMapper(MailMessageMapper.class)
@@ -43,4 +44,13 @@ public interface MailMessageDAO extends Transactional<MailMessageDAO> {
              LIMIT :limit
              """)
     List<MailMessage> fetchMessagesByType(@Bind("emailType") Integer emailType, @Bind("limit") Integer limit, @Bind("offset") Integer offset);
+
+    @SqlQuery("""
+             SELECT entity_reference_id, email_entity_id, vote_id, user_id, email_type, date_sent, email_text, sendgrid_response, sendgrid_status, create_date FROM email_entity e
+             WHERE create_date BETWEEN SYMMETRIC :start AND :end 
+             ORDER BY create_date DESC
+             OFFSET :offset
+             LIMIT :limit
+             """)
+    List<MailMessage> fetchMessagesByCreateDate(@Bind("start") Date start, @Bind("end") Date end, @Bind("limit") Integer limit, @Bind("offset") Integer offset);
 }
--- a/src/main/java/org/broadinstitute/consent/http/db/MailMessageDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/MailMessageDAO.java
@@ -1,6 +1,7 @@
 package org.broadinstitute.consent.http.db;
 
 import org.broadinstitute.consent.http.db.mapper.MailMessageMapper;
+import org.broadinstitute.consent.http.models.mail.MailMessage;
 import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
 import org.jdbi.v3.sqlobject.customizer.Bind;
 import org.jdbi.v3.sqlobject.statement.GetGeneratedKeys;
@@ -10,6 +11,7 @@ import org.jdbi.v3.sqlobject.transaction.Transactional;
 
 import javax.annotation.Nullable;
 import java.time.Instant;
+import java.util.List;
 
 @RegisterRowMapper(MailMessageMapper.class)
 public interface MailMessageDAO extends Transactional<MailMessageDAO> {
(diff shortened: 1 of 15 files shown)


src/main/java/org/broadinstitute/consent/http/db/MailMessageDAO.java
```diff
@@ -32,4 +35,22 @@ Integer insert(@Nullable @Bind("entityReferenceId") String entityReferenceId,
...
+             WHERE email_type = :emailType
+             ORDER BY create_date DESC
+             OFFSET :offset
+             LIMIT :limit
+             """)
+    List<MailMessage> fetchMessagesByType(@Bind("emailType") Integer emailType, @Bind("limit") Integer limit, @Bind("offset") Integer offset);
+
+    @SqlQuery("""
+             SELECT entity_reference_id, email_entity_id, vote_id, user_id, email_type, date_sent, email_text, sendgrid_response, sendgrid_status, create_date FROM email_entity e
+             WHERE create_date BETWEEN SYMMETRIC :start AND :end
```
otchet-broad, line 50: BETWEEN SYMMETRIC allows for the end date to be before the start date.  The values are sorted before the expression is executed.  See: [https://www.postgresql.org/docs/current/functions-comparison.html#:~:text=datatype-,BETWEEN%20SYMMETRIC,-datatype%20AND%20datatype](https://www.postgresql.org/docs/current/functions-comparison.html#:~:text=datatype-,BETWEEN%20SYMMETRIC,-datatype%20AND%20datatype)

src/test/java/org/broadinstitute/consent/http/db/MailMessageDAOTest.java
```diff
@@ -207,4 +212,105 @@ public void testInsert_MissingCreateDate() {
...
+
+        List<MailMessage> mailMessageList4 = mailMessageDAO.fetchMessagesByType(EmailType.COLLECT.getTypeInt(), 20, 0);
+        assertEquals(2, mailMessageList4.size());
+    }
+
+    @Test
+    public void testFetchByCreateDate_with_limit_and_offset(){
+        Date start = new Date();
+        generateMessagesInTime(2, 1);
+        Date end = new Date();
```
connorlbark, line 282: could using the current time via `new Date()` lead to any sort of flakiness? maybe we should have more fixed dates in the test. also, I am having trouble understanding exactly what this test does / if it is complete, maybe splitting into a couple smaller tests would be good. also having the test verify the exact messages it gets instead of just the size could be good too

src/main/resources/assets/paths/mailByRange.yaml
```diff
@@ -0,0 +1,46 @@
+get:
+  summary: Get Mail Messages by Date Range.
+  description: |
+    Admin only feature returns MailMessage list filtered by Date Range.
+  parameters:
```
connorlbark, line 5: would an optional parameter to allow selecting on type of message be helpful as well? I could imagine a use case of an admin investigating emails of a certain type in the other endpoint, then wanting to move to this endpoint to find messages from a certain date.

src/main/resources/assets/paths/mailByRange.yaml
```diff
@@ -0,0 +1,46 @@
...
+      description: Earliest Date to search in MM/dd/yyyy format. (e.g. 05/19/2019)
+      required: true
+      schema:
+        type: string
+    - name: end
+      in: query
+      description: Latest Date to search in MM/dd/yyyy format. (e.g. 05/20/2022)
+      required: true
+      schema:
+        type: string
```
connorlbark, line 17: what if it's optional, and defaults to the time of being called if not specified? similarly for start, could be optional and default to a time far in the past.

connorlbark (commented):
Looking very good. Works great locally, couple things but overall great.

src/main/resources/assets/paths/mailByRange.yaml
rushtong: I do think this is a good idea, but outside the scope of the current ticket. We don't have any UI for this yet, so I think this idea could wait for that so we have some user-experience to rely on and inform new features.

src/test/java/org/broadinstitute/consent/http/db/MailMessageDAOTest.java
rushtong: I am similarly confused by this test. I get the general idea that we want to find messages between dates with various offsets and limits - each of the 4 cases need some commentary on what they should be finding. I think what this is also testing is `BETWEEN SYMMETRIC`, but I can't be sure. The confusing part to me is start and end dates that are effectively just milliseconds apart.

Here's my understanding
---
Given: 
* A `start` date of `now`
* A list of 2 mail messages. First message is created with an instant of `now` + a few milliseconds. Second message is created with the same instant, but one day earlier.
* An `end` date that's a few milliseconds after the first message creation date.
* A `startCalendar` which represents a 3rd search date that should be a few milliseconds after `end`

Case 1, `mailMessageList`: This makes sense - this should return the first mail message since it is the only one between start and end

Case 2, `mailMessageList2`: This doesn't make sense to me. It's expecting 2 results, but `startCalendar` and `end` are just a few milliseconds apart. This should not be finding message #2

Case 3, `mailMessageList3`: This also doesn't make sense to me for the same reason. The offset makes sense - if you're expecting 2 messages, and offsetting it by 1, the list should be 1 element long. Since the date range is just a few milliseconds apart, and both are after **_start_**, it shouldn't find any of the messages.

Case 4, `mailMessageList4`: Same here, searching between two dates that are both after the creation dates of both messages should return 0 messages.

rushtong:
Closing this PR in favor of https://github.com/DataBiosphere/consent/pull/1870

PR #1868, opened 2023-01-04
[DUOS-2229][risk=no] Add default timeout to status checks

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2229

This PR adds a default timeout to the external http calls we use to access Ontology, Sam, and Sendgrid. This PR does not address our ElasticSearch status check since that uses a different client (ES vs Apache) and has not been problematic in practice. When a service fails due to a request taking longer than the configured value, we will have a functional status call but with a degraded state which is a more accurate status representation than hanging on long downstream status checks.

## Testing
Add a line to the services block of your local config with zero seconds so all calls immediately fail:
```
services:
  localURL: foo
  ontologyURL: bar
  samUrl: baz
  timeoutSeconds: 0
```

The app should run with or without this configuration as there is a default value specified in code. Future PRs will address adding this line to Consent's configuration in its helm chart.

----

- Verify all tests go green

[DUOS-2229]: https://broadworkbench.atlassian.net/browse/DUOS-2229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -370,15 +370,9 @@
     </dependency>
 
     <dependency>
-      <groupId>org.apache.httpcomponents</groupId>
-      <artifactId>httpclient</artifactId>
-      <version>4.5.14</version>
-      <exclusions>
-        <exclusion>
-          <groupId>commons-codec</groupId>
-          <artifactId>commons-codec</artifactId>
-        </exclusion>
-      </exclusions>
+        <groupId>org.apache.httpcomponents.client5</groupId>
+        <artifactId>httpclient5</artifactId>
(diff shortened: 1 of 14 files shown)


connorlbark (approved):
👍 code looks good, status checks for me locally

quazi-h (approved):
Status endpoints are working well locally. Test logic looks good, no issues when running tests locally.

PR #1870, opened 2023-01-05
[DUOS-2112][risk=low] Add support for searching mail sent by date range.

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2112

[DUOS-2112](https://broadworkbench.atlassian.net/browse/DUOS-2112) builds on work completed as part of https://github.com/DataBiosphere/consent/pull/1862 and makes the following enhancements:

- Endpoint and class updates for searching mail by date range added.
- Documentation and tests added to describe and validate the feature.
- Limit and Offset are now validated to be greater than -1 for both searching by EmailType and Date Range.

### Notes
This PR is a branch of https://github.com/DataBiosphere/consent/pull/1864

----

- Verify all tests go green

[DUOS-2112]: https://broadworkbench.atlassian.net/browse/DUOS-2112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/resources/assets/paths/mailByRange.yaml
+++ b/src/main/resources/assets/paths/mailByRange.yaml
@@ -5,14 +5,16 @@ get:
   parameters:
     - name: start
       in: query
-      description: Earliest Date to search in MM/dd/yyyy format. (e.g. 05/19/2019)
+      description: | 
+        Earliest Date to search in MM/dd/yyyy format. (e.g. 05/19/2019)
       required: true
       schema:
         type: string
     - name: end
       in: query
-      description: Latest Date to search in MM/dd/yyyy format. (e.g. 05/20/2022)
-      required: true
+      description: |
+        Latest Date to search in MM/dd/yyyy format. (e.g. 05/20/2022). Defaults to the future if empty.
+      required: false
       schema:
         type: string
     - name: limit
@@ -28,6 +30,7 @@ get:
       schema:
         type: integer
   tags:
+    - Admin
     - Mail
   responses:
     200:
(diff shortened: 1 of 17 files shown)


src/main/resources/assets/paths/mailByRange.yaml
```diff
@@ -0,0 +1,49 @@
...
+      required: false
+      schema:
+        type: string
+    - name: limit
+      in: query
+      description: Number of results to return.  Defaults to 20 if unset. May not be negative
+      required: false
+      schema:
+        type: integer
+    - name: offset
```
connorlbark, line 26: feedback from before, but i am just curious on thoughts: any interest in an optional parameter to filter by type as well? i feel usually you would be debugging / trying to find specific emails of a known type when searching for date as well.

connorlbark (approved):
this is much cleaner, and i like the new tests a lot more. i will go ahead and approve because it works great and code is good. optional feedback is optional, up to what you think the use case of this is

src/main/resources/assets/paths/mailByRange.yaml
rushtong: Filtering on type is a different use case that we don't actually have yet. I can foresee the need for it, but I would prefer that we build that feature when we actually need it.

quazi-h (approved):
Changes make sense to me, tests passed when running locally. I was unable to find the /mail endpoints when running consent locally, I made sure that my user had an Admin role when checking /user/me. Is there a way to view/test those endpoints on swagger locally?

rushtong:
> I was unable to find the /mail endpoints when running consent locally, I made sure that my user had an Admin role when checking /user/me. Is there a way to view/test those endpoints on swagger locally?

The endpoint is defined in both `Admin` and `Mail` headings as seen here:
Screen Shot 2023-01-10 at 2 24 07 PM

PR #1873, opened 2023-01-08
[DUOS-2107][DUOS-2108][risk=no] Ensure that closed, canceled, and final elections are archived

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2107
https://broadworkbench.atlassian.net/browse/DUOS-2108

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/ElectionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/ElectionService.java
@@ -2,6 +2,7 @@ package org.broadinstitute.consent.http.service;
 
 import com.google.inject.Inject;
 import freemarker.template.TemplateException;
+import java.util.EnumSet;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.consent.http.db.ConsentDAO;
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/consent/http/service/ElectionService.java
```diff
@@ -101,13 +102,29 @@ public Election updateElectionById(Election rec, Integer electionId) {
         }
         Date lastUpdate = new Date();
         electionDAO.updateElectionById(electionId, rec.getStatus(), lastUpdate);
-        if(rec.getArchived() != null && rec.getArchived()) {
-            electionDAO.archiveElectionById(electionId, lastUpdate);
-        }
+        archiveElectionByIdAndStatus(electionId, rec.getStatus());
```
connorlbark: This pattern is a bit confusing - it's looks like all elections are being archived at this step, but it's really `archiveElectionIfNewStatusIsArchived` which is confusing... perhaps it should just be:

```
if (isArchiveStatus(rec.getStatus()) {
    electionDAO.archiveElectionById(electionId, update);
}
```

connorlbark (commented):
Looking good. Just some style nits. In addition, should some new tests be added?

src/main/java/org/broadinstitute/consent/http/service/ElectionService.java
rushtong: Yeah, the method name is confusing. I want to have this in a separate method since we need to call it multiple times ... will rework this a bit.

connorlbark (approved):
👍

quazi-h (approved):
New logic to archive elections after cancelling and closing them looks correct. Back population update looks good, too.

PR #1876, opened 2023-01-08
[DUOS-2231][risk=no] Add optional email to dac create/update

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2231

This PR adds an optional email field to Data Access Committees and updates the POST/PUT endpoints to accommodate it.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
@@ -71,7 +71,12 @@ public class DacResource extends Resource {
         if (dac.getDescription() == null) {
             throw new BadRequestException("DAC Description is required");
         }
-        Integer dacId = dacService.createDac(dac.getName(), dac.getDescription());
+        Integer dacId;
+        if (Objects.isNull(dac.getEmail())) {
+            dacId = dacService.createDac(dac.getName(), dac.getDescription());
+        } else {
+            dacId = dacService.createDac(dac.getName(), dac.getDescription(), dac.getEmail());
+        }
         if (dacId == null) {
             throw new Exception("Unable to create DAC with name: " + dac.getName() + " and description: " + dac.getDescription());
         }
@@ -96,7 +101,11 @@ public class DacResource extends Resource {
         if (dac.getDescription() == null) {
             throw new BadRequestException("DAC Description is required");
         }
-        dacService.updateDac(dac.getName(), dac.getDescription(), dac.getDacId());
+        if (Objects.isNull(dac.getEmail())) {
+            dacService.updateDac(dac.getName(), dac.getDescription(), dac.getDacId());
+        } else {
+            dacService.updateDac(dac.getName(), dac.getDescription(), dac.getEmail(), dac.getDacId());
+        }
         Dac savedDac = dacService.findById(dac.getDacId());
         return Response.ok().entity(unmarshal(savedDac)).build();
     }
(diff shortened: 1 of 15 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
```diff
@@ -71,7 +71,12 @@ public Response createDac(@Auth AuthUser authUser, String json) throws Exception
         if (dac.getDescription() == null) {
             throw new BadRequestException("DAC Description is required");
         }
-        Integer dacId = dacService.createDac(dac.getName(), dac.getDescription());
+        Integer dacId;
+        if (Objects.isNull(dac.getEmail())) {
```
connorlbark, line 75: unnecessary check? for clarity sake maybe it makes sense to just pass in the email here since it is null anyway. I see how on update you don't want to delete old emails if they are present, so the check makes sense there.

src/main/java/org/broadinstitute/consent/http/service/DacService.java
```diff
@@ -141,11 +141,21 @@ public Integer createDac(String name, String description) {
...
     public void updateDac(String name, String description, Integer dacId) {
         Date updateDate = new Date();
         dacDAO.updateDac(name, description, updateDate, dacId);
     }
 
+    public void updateDac(String name, String description, String email, Integer dacId) {
+        Date updateDate = new Date();
+        dacDAO.updateDac(name, description, email, updateDate, dacId);
+    }
+
```
connorlbark, line 158: do we need separate code paths for `with email` and `without email`? it seems like going forward we should always use the `with email` code path, so maybe getting ride of the non-email ones would make sense.

src/main/resources/assets/api-docs.yaml
```diff
@@ -254,125 +254,9 @@ paths:
...
-          description: All Data Access Committees with chairpersons and members
-          content:
-            application/json:
-              schema:
-                type: array
-                items:
-                  $ref: '#/components/schemas/Dac'
-        500:
-          description: Internal Server Error
+    $ref: 'paths/dac.yaml'
```
connorlbark, line 257: nice

src/test/java/org/broadinstitute/consent/http/db/DacDAOTest.java
```diff
@@ -309,4 +333,23 @@ public void testFindDatasetsAssociatedWithDac_SuggestedDacId() {
...
+
+    private Dac insertDacWithEmail() {
+        String testEmail = "test@email.com";
+        Integer id = dacDAO.createDac(
+                "Test_" + RandomStringUtils.random(20, true, true),
+                "Test_" + RandomStringUtils.random(20, true, true),
+                testEmail,
+                new Date());
+        return dacDAO.findById(id);
+    }
```
connorlbark, line 353: same feedback as above - seems redundant?

connorlbark (commented):
looking great! just some feedback on code pathing

src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
rushtong: I would like to keep the two separate code paths until we need to make email a required field.

src/main/java/org/broadinstitute/consent/http/service/DacService.java
rushtong: Yes, I do think we need both right now. If we determine that we want to make email required, we can consolidate them.

connorlbark: Ah ok, I was under the impression it was required.

connorlbark (approved):
if email is not required, then I am ok with these changes 👍

PR #1881, opened 2023-01-11
[DUOS-2243][risk=no] Add cache for external status checks

rushtong:
## Addresses
Along with https://github.com/broadinstitute/terra-helmfile/pull/3714, this PR partially addresses https://broadworkbench.atlassian.net/browse/DUOS-2243

This PR adds caching to most external status checks to reduce the latency of consent's status check.

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/configurations/ServicesConfiguration.java
```diff
@@ -19,6 +19,8 @@ public class ServicesConfiguration {
 
   private Integer poolSize = 10;
 
+  private Integer cacheExpireMinutes = 1;
```
nmalfroy, line 37: Is it weird to have this in minutes but you've got `timeoutSeconds` right above?

nmalfroy (approved):
This looks good.  Looking at the handler in `HttpClientUtil.getHttpResponse`, it seems like it's correctly consuming the entity response, so that's likely not where your memory leak is coming from

src/main/java/org/broadinstitute/consent/http/configurations/ServicesConfiguration.java
rushtong: I can certainly change it for consistency, but they are conceptually very different things.

rushtong: Maybe some additional commentary around each value would be helpful

nmalfroy: Yeah, that's totally fair.  And there's likely no need to have a more exact cache duration than minutes.

Re docs, that's maybe more general than this PR.  But yeah, maybe java docs around the service config might be useful. Totally not a blocker to merge BTW

connorlbark (approved):
👍 works great locally for me, much snappier after the first call. code looks good, thorough comments and testing.

PR #1889, opened 2023-01-16
[DUOS-1740][risk=no] Use standard OWASP check action

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1740

This updates our owasp check to use a standard action described [here](https://github.com/marketplace/actions/dependency-check) instead of running it via a maven plugin. The advantage of this action is that the CVE databases are built into the image daily and don't need to be downloaded for every run.

Example successful job run: https://github.com/DataBiosphere/consent/actions/runs/3935656264

----

- Verify all tests go green


connorlbark (approved):
👍 nice!

quazi-h (approved):

PR #1894, opened 2023-01-18
[DUOS-2120][risk=medium] Implement Save for Dataset V3 Endpoint

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2120

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/enumeration/DatasetPropertyType.java
```diff
@@ -64,9 +64,9 @@ public static Boolean coerceToBoolean(String value) throws IllegalArgumentExcept
         return java.lang.Boolean.valueOf(value);
     }
 
-    public static Instant coerceToDate(String value) throws IllegalArgumentException {
+    public static LocalDate coerceToDate(String value) throws IllegalArgumentException {
```
connorlbark: context: instants require information about timezone and the exact ms when things happen. LocalDate, instead, is the simplest form of date: no knowledge of timezones, no time of day, just "YYYY-MM-DD", which is all we want to capture in the dataset registration. Perhaps in the future we will need a dataset property with more granular time; in which case, we could perhaps make a `DateTime` or `Instant` dataset property type.

src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/DatasetRegistrationSchemaV1.java
```diff
@@ -6,12 +6,16 @@
...
+import org.broadinstitute.consent.http.enumeration.DatasetPropertyType;
+import org.broadinstitute.consent.http.models.DatasetProperty;
+import org.broadinstitute.consent.http.util.gson.GsonUtil;
 
 import javax.annotation.processing.Generated;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
```
rushtong: nit: The added imports are unused

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
```diff
@@ -0,0 +1,426 @@
...
+        dataUse.setCollaboratorRequired(group.getCol());
+        dataUse.setDiseaseRestrictions(group.getDiseaseSpecificUse());
+        dataUse.setEthicsApprovalRequired(group.getIrb());
+        dataUse.setGeneralUse(group.getGeneralResearchUse());
+        dataUse.setGeographicalRestrictions(group.getGs());
+        dataUse.setGeneticStudiesOnly(group.getGso());
+        dataUse.setHmbResearch(group.getHmb());
+        dataUse.setPublicationMoratorium(Objects.nonNull(group.getMor()) && group.getMor() ? group.getMorDate() : null);
+
+        dataUse.setMethodsResearch(Objects.nonNull(group.getNmds()) ? !group.getNmds() : null); // TODO: is this right?
```
rushtong: For a dataset, these are the cases as defined in two places, [here](https://github.com/DataBiosphere/consent-ontology/blob/develop/src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java#L212) and [here](https://github.com/DataBiosphere/consent-ontology/blob/develop/src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java#L362). Both look like they use the same logic:

data use methods research: 
- true: "Data use for methods development research irrespective of the specified data use limitations is not prohibited."
- false:
  -  non-null, or falsy: "Data use for methods development research ONLY within the bounds of other data use limitations. [NMDS]"
  - null: "Data use for methods development research irrespective of the specified data use limitations is not prohibited."

Pulling all this together, if our schema has `nmds==true` (which comes from the question `"No Methods Development or validation studies (NMDS)"`), then `dataUse.methodsResearch` should be `false`

However, this doesn't align exactly, so I think we need to bring this point back to Jonathan. We're trying to align a newish version of the NMDS question to the[ V2 algorithm logic](https://github.com/DataBiosphere/consent-ontology/blob/develop/docs/Algorithms.md) and it doesn't exactly match.

rushtong (commented):
Still reviewing - wanted to get the comment about NMDS out sooner rather than later

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
```diff
@@ -0,0 +1,426 @@
...
+        if (Objects.isNull(dacDAO.findById(registration.getDataAccessCommitteeId()))) {
+            throw new IllegalArgumentException("Invalid DAC id");
+        }
+
+        Map<String, BlobId> uploadedFileCache = new HashMap<>();
+
+        List<DatasetServiceDAO.DatasetInsert> datasetInserts = new ArrayList<>();
+        for (int consentGroupIdx = 0; consentGroupIdx < registration.getConsentGroups().size(); consentGroupIdx++) {
+            datasetInserts.add(createDatasetInsert(registration, user, files, uploadedFileCache, consentGroupIdx));
+        }
```
rushtong: I generally lean towards lambdas for this kind of thing, but `createDatasetInsert` throws `IOException` so it doesn't work inside the lambda unless it's re-thrown as a `RuntimeException`. So, if we hit an exception on index 3 out of 5, for example, then 0, 1, and 2 would have inserted files via `uploadFile` which is called inside the loop. Then we'd have `N` files left behind.

I think one solution to this problem would be to split up the file uploads into one block, and then, if they all succeed, proceed with the dataset inserts using the cached files stored in uploadedFileCache. That would allow for:

1. rolling back the file uploads if any fail
2. generating the inserts without throwing an `IOException`
3. rolling back file uploads if the dataset inserts fail

rushtong (commented):
Still reviewing ... will pick back up tomorrow.

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
connorlbark: I originally did that solution, but I didn't want to upload all files, because it was possible that the user uploaded a file we don't need, and I didn't want to upload anything that we wouldn't use. I also think a lambda here would muddy up the function, since `createDatasetInsert` is a non-trivial function. The function is already doing a lot, I think. I do agree on rolling back files though, I can add some try/catch blocks to delete the files if something goes wrong.

rushtong: My thought on the lambda was not to replace `createDatasetInsert`, but to call that function inside the lambda instead of using placeholder variables and mutating lists - this is where we often find bugs. The blocker for that was the `IOException`, hence my suggestion to get around that with pre-uploads. One (probably very) minor benefit to the separate upload/dataset insert is that you're not blocking the open db handle while file uploads to GCS are happening. Breaking those two things into discrete chunks will ease the pressure on the handle. 

That said, I can live with this approach :-)

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
```diff
@@ -0,0 +1,426 @@
...
+
+    /**
+     * Extracts an individual field as a dataset property.
+     *
+     * @param name The human-readable name of the field
+     * @param schemaProp The schema property name (camelCase)
+     * @param type The type of the field, e.g. Boolean, String
+     * @param getField Lambda which gets the field's value
+     */
+    private record DatasetPropertyExtractor(
```
rushtong: It would be nice to extract this to a public record so it could be independently testable. As it stands now, all usages are private so testing needs to go through the service class.

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
```diff
@@ -0,0 +1,426 @@
...
+
+            return Optional.of(datasetProperty);
+
+        }
+    };
+
+    private static final List<DatasetPropertyExtractor> datasetPropertyExtractors = List.of(
+            new DatasetPropertyExtractor(
+                    "PI Name", "piName", DatasetPropertyType.String,
+                    (registration, idx) -> registration.getPiName()),
```
rushtong: Thinking out loud here - I don't have an immediate solution.

`"piName"` exists in several places as a string value and as a field name. It would be nice if we could pull this from either the schema object or from something related to `DatasetProperty`, but either way would require some refactoring outside the scope of this PR. Same idea applies to all of the extractors.

The dataset property name comes from associated `Dictionary` key. Here, again, I wish there were a better way of pulling these values from some source of truth as opposed to defining them here as raw strings. Again, outside the scope of this PR.

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
```diff
@@ -0,0 +1,426 @@
...
+
+    private List<DatasetProperty> convertRegistrationToDatasetProperties(DatasetRegistrationSchemaV1 registration, int consentGroupIdx) {
+        List<DatasetProperty> datasetProperties = new ArrayList<>();
+        for (DatasetPropertyExtractor datasetPropertyExtractor : datasetPropertyExtractors) {
+
+            Optional<DatasetProperty> extractedProperty = datasetPropertyExtractor.extract(registration, consentGroupIdx);
+
+            if (extractedProperty.isPresent()) {
+                datasetProperties.add(extractedProperty.get());
+            }
```
rushtong: nit: could be simpler:
```suggestion
            extractedProperty.ifPresent(datasetProperties::add);
```

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
```diff
@@ -0,0 +1,426 @@
...
+                            return registration.getConsentGroups().get(idx).getUrl().toString();
+                        }
+                        return null;
+                    })
+    );
+
+
+    private List<DatasetProperty> convertRegistrationToDatasetProperties(DatasetRegistrationSchemaV1 registration, int consentGroupIdx) {
+        List<DatasetProperty> datasetProperties = new ArrayList<>();
+        for (DatasetPropertyExtractor datasetPropertyExtractor : datasetPropertyExtractors) {
```
rushtong: nit: avoid mutation - this can easily be a one-liner which, to me, reads a lot more clearly than a for loop with internal if and placeholder variables:
```
        return datasetPropertyExtractors
            .stream()
            .map(e -> e.extract(registration, consentGroupIdx))
            .filter(Optional::isPresent)
            .map(Optional::get)
            .toList();
```

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -881,9 +887,9 @@ public void testCreateDatasetRegistration_invalidSchema_case3() {
     }
 
     @Test
-    public void testCreateDatasetRegistration_validSchema() {
+    public void testCreateDatasetRegistration_validSchema() throws SQLException, IOException {
```
rushtong, line 890: For tests, its perfectly fine to throw a general exception:
```suggestion
    public void testCreateDatasetRegistration_validSchema() throws Exception {
```

src/main/java/org/broadinstitute/consent/http/enumeration/DatasetPropertyType.java
nmalfroy: Hmmm...but how do you know the global time?  Like, if two users call the endpoint from a different timezone (not sure where the original time comes from) at the same time, wouldn't they have different times then?  I thought instant defaulted to Greenwich Mean Time (e.g. no timezone offset) and then you lazily convert the time based on the calling user's locale

src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/DatasetRegistrationSchemaV1.java
nmalfroy: You can get IntelliJ to clean those up for you on save

rushtong (commented):
Still reviewing - some comments to consider inline

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
```diff
@@ -0,0 +1,426 @@
...
+     * @param user The User creating these datasets
+     * @param files Map of files, where the key is the name of the field
+     * @return List of created Datasets from the provided registration schema
+     */
+    public List<Dataset> createDatasetsFromRegistration(
+            DatasetRegistrationSchemaV1 registration,
+            User user,
+            Map<String, FormDataBodyPart> files) throws IOException, SQLException, IllegalArgumentException {
+
+        if (Objects.isNull(dacDAO.findById(registration.getDataAccessCommitteeId()))) {
```
nmalfroy, line 70: Is there value in explicitly throwing a 404-type error here along with a message to that effect?

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
nmalfroy: Can you make the DB bits transactional so that the only rollback would be deleting the orphaned files? (and also you could make `createDatasetInsert` no throw a checked exception so that you could call it in a lambda...Man I with there was a way to handle errors in lambdas more cleanly)

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
```diff
@@ -0,0 +1,426 @@
...
+        return dataUse;
+    }
+
+    private List<FileStorageObject> uploadFiles(Map<String, FormDataBodyPart> files,
+                                                Map<String, BlobId> uploadedFileCache,
+                                                Integer consentGroupIdx,
+                                                User user) throws IOException {
+        List<FileStorageObject> consentGroupFSOs = new ArrayList<>();
+
+        if (files.containsKey("alternativeDataSharingPlan")) {
```
nmalfroy: These strings should probably be constants

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
nmalfroy: yeah, at the very least package private

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
```diff
@@ -0,0 +1,426 @@
...
+            datasetProperty.setPropertyType(this.type);
+            datasetProperty.setSchemaProperty(this.schemaProp);
+            datasetProperty.setPropertyValue(this.type.coerce(value.toString()));
+
+            return Optional.of(datasetProperty);
+
+        }
+    };
+
+    private static final List<DatasetPropertyExtractor> datasetPropertyExtractors = List.of(
```
nmalfroy: There's a common pattern to have static finals be in snake case.  Not sure if this a pattern for you (e.g. 
```suggestion
    private static final List<DatasetPropertyExtractor> DATASET_PROPERTY_EXTRACTORS = List.of(
```

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
```diff
@@ -0,0 +1,426 @@
...
+                        if (Objects.nonNull(registration.getAlternativeDataSharingPlanControlledOpenAccess())) {
+                            return registration.getAlternativeDataSharingPlanControlledOpenAccess().value();
+                        }
+                        return null;
+                    }),
+            new DatasetPropertyExtractor(
+                    "Data Location", "consentGroup.dataLocation", DatasetPropertyType.Json,
+                    (registration, idx) -> {
+                        if (Objects.nonNull(registration.getConsentGroups().get(idx).getDataLocation())) {
+                            return GsonUtil.getInstance().toJson(registration.getConsentGroups().get(idx).getDataLocation().stream().map(DataLocation::value).toList());
```
nmalfroy: Any risk to the index being out of range?

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
```diff
@@ -0,0 +1,426 @@
...
+                        return null;
+                    })
+    );
+
+
+    private List<DatasetProperty> convertRegistrationToDatasetProperties(DatasetRegistrationSchemaV1 registration, int consentGroupIdx) {
+        List<DatasetProperty> datasetProperties = new ArrayList<>();
+        for (DatasetPropertyExtractor datasetPropertyExtractor : datasetPropertyExtractors) {
+
+            Optional<DatasetProperty> extractedProperty = datasetPropertyExtractor.extract(registration, consentGroupIdx);
```
nmalfroy: you can do this in one line with lambdas:
```suggestion
            Optional<DatasetProperty> extractedProperty = datasetPropertyExtractor.extract(registration, consentGroupIdx)
            .ifPresent(datasetProperties::add);
```

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -113,7 +112,7 @@ public List<Dataset> findNeedsApprovalDataSetByObjectId(List<Integer> datasetIdL
     }
 
     public Set<DatasetDTO> findDatasetsByDacIds(List<Integer> dacIds) {
-        if(Objects.isNull(dacIds) || dacIds.isEmpty()) {
+        if (Objects.isNull(dacIds) || dacIds.isEmpty()) {
```
nmalfroy: If you use CollectionUtils, I think you can do something like `CollectionUtils.isEmpty(dacIds)` which does both for you.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -186,23 +185,23 @@ public Consent createConsentForDataset(DatasetDTO dataset) {
...
-              (Objects.nonNull(dataUse.getPsychologicalTraits()) && dataUse
-                    .getPsychologicalTraits()) ||
-              (Objects.nonNull(dataUse.getSexualDiseases()) && dataUse.getSexualDiseases()) ||
-              (Objects.nonNull(dataUse.getStigmatizeDiseases()) && dataUse.getStigmatizeDiseases())
-              ||
-              (Objects.nonNull(dataUse.getVulnerablePopulations()) && dataUse
-                    .getVulnerablePopulations());
+                (Objects.nonNull(dataUse.getPopulationRestrictions()) && !dataUse
+                        .getPopulationRestrictions().isEmpty()) ||
+                (Objects.nonNull(dataUse.getAddiction()) && dataUse.getAddiction()) ||
```
nmalfroy, line 199: feel like there's a repeating pattern that should be put into a method.  Or maybe use a lamba to shorten, e.g
```suggestion
                Optional.ofNullable(dataUse.getAddiction()).orElse(false) ||
```

Probably makes sense to have the logic pulled into a method that checks for truthiness.  Justing thinking of making this giant boolean statement easier to read

src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
```diff
@@ -34,24 +104,41 @@ public List<DatasetProperty> synchronizeDatasetProperties(Integer datasetId, Lis
...
-                    updates.addAll(generatePropertyDeletes(handle, properties, existingProps));
-                    updates.forEach(Update::execute);
+                    executeSynchronizeDatasetProperties(handle, datasetId, properties);
                     handle.commit();
                 }
         );
         return datasetDAO.findDatasetPropertiesByDatasetId(datasetId).stream().toList();
     }
 
+    private void executeInsertFilesForDataset(List<FileStorageObject> files, Integer userId, Integer datasetId) {
```
nmalfroy: Seems like this could end up with the DB in a half committed state.  One pattern I've used with JDBI is to make sure that handle gets passed around in all methods an then the first method gets the handle.  There are ways to have the transaction management handled a little more automatically but that's a bit more involved

src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
```diff
@@ -524,29 +527,31 @@ public void testCreateNumberTypedDatasetProperty() {
     @Test
     public void testCreateDateTypedDatasetProperty() {
         Dataset d = insertDataset();
-        Instant now = Instant.now();
+        LocalDate date = LocalDate.now();
```
nmalfroy: Same thing: not really sure about switching to LocalDate...might be worth talking through. I may not be understanding the logic

nmalfroy (commented):
Looks good.  I just have a couple of comments.  My biggest concern isn't even really with what this PR is explicitly trying to do but more with how transactions are handled, especially cross dao.

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
rushtong: > Can you make the DB bits transactional so that the only rollback would be deleting the orphaned files? (and also you could make `createDatasetInsert` no throw a checked exception so that you could call it in a lambda...Man I with there was a way to handle errors in lambdas more cleanly)

The DB bits are transactional, but we have two systems to save to, so combining them is the thing that's problematic to me. We do the following in a single block:
1. Save a dataset with file metadata in postgres (transactional and rollback-able)
2. Save the actual file content in GCS

What I would prefer would be to break them out into two separate rollback-able things so we can fail faster if there are problems in either logical block.

src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
connorlbark: So, here the user is inputting a simple date. e.g., the alternative data sharing plan release date is `2022-10-11`. That's all the information we have. A LocalDate makes sense because it makes no assumptions other than keeping the year, day, and month, which is all we want to store. To make an Instant we would have to add information which we don't have, unless we changed what the frontend should send. I think an instant does make sense for most use cases. But, I think it makes sense here, because all we need to know is the day of the year it happened; we don't need to know timezone really. I tried keeping the Instant at first, but it became somewhat complicated having to insert timezone information, and I ended up just giving it the system's timezone, which isn't really any useful information. After spending more than like 20 minutes trying to get all the date conversions and format things right and it not working, I figured it was simpler. I am happy  convert it back to an Instant if you all prefer, but it takes some non-trivial code to add timezone information.

src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
connorlbark: Oh okay, this is the first time I've ever made my own service DAO function so I may have misunderstood how the whole handle thing worked. I'll pass the handle in and use that.

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
connorlbark: To clarify... creating the dataset insert is just creating a record object. So there is no database transaction happening. Basically, what's happening:

1. Upload all files & create dataset insert records at same time (no db changes)
2. Commit all dataset inserts in a rollback-able way at the same time

They are happening sequentially, the only thing I'm missing (if I understand the discourse here) is the ability to rollback files.

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
connorlbark: hmm, I agreed at first, but I have always been taught in classes that things shouldn't be exposed just for testing. I was also able to use argument captors to get into the nitty gritty of testing exactly what was sent to the dataset insert, and I was able to test dataset property creation then. It is a lot for that one method to test, but the method is basically "convert dataset registration into dataset insert files" and from that perspective, it makes sense to test it all. I'm ok if y'all both have a preference for breaking it up, though.

rushtong: > I have always been taught in classes that things shouldn't be exposed just for testing

TBH, I disagree with that, but it isn't `just for testing`, it's more about organization. Inner, private classes are good for small units of functionality but when there's any significant logic, it needs to stand on its own. This one is doing a lot of work that needs to be exercised properly. It's all really good stuff and there's a lot of potential for reuse. Imagine trying to do an insert with a V2 schema that subclassed V1 but with more fields ... there'd need to be some work done here to handle that and I'd want to make sure tests covered all cases. Probably not the best example, but it's the first to come to mind.

connorlbark: That makes sense, actually. I'll make it public and add test cases for it.

src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/DatasetRegistrationSchemaV1.java
connorlbark: no way, i'm gonna find that setting

nmalfroy: haha, wait sorry, I should have also given you the setting...
image


Note the "optimize imports on the fly"

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
connorlbark: Sounds good on talking to Jonathan - I am unsure about `NPU` as well. I'll bring it up in standup.

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
connorlbark: good point, sent down consent grouop object instead

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
connorlbark: I added rollback for files. I am gonna add a few tests, and I'll also add some tests for file rollback.

src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/DatasetRegistrationSchemaV1.java
rushtong: Still an issue

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
rushtong: I agree with this - and wish we had a better way of handling `"consentGroups["+consentGroupIdx+"].nihInstitutionalCertificationFile"` as well.

rushtong (commented):
Just a couple minor nits, then 👍🏽

rushtong (approved):
Thank you for the changes! 👍🏽

src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
fboulnois: One question is: does the service need to keep a globally consistent store of time and dates in the system? If so, then I think it's worth thinking about using something like an `Instant`. If you say your sharing plan release date is 2022-10-11, is this in relation to a particular geography (say, where the NIH is headquartered) or something else? I agree with Nick that it's worth having a larger conversation around this.

src/test/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAOTest.java
```diff
@@ -126,7 +132,138 @@ public void testSynchronizeDatasetProperties_case6() throws Exception {
...
+        assertEquals(insert2.dacId(), dataset2.getDacId());
+        assertEquals(true, dataset2.getDataUse().getIllegalBehavior());
+        assertEquals(2, dataset2.getProperties().size());
+        assertNull(dataset2.getNihInstitutionalCertificationFile());
+        assertNull(dataset2.getAlternativeDataSharingPlanFile());
+
+    }
+
+
+        // Helper methods
```
fboulnois:
```suggestion
```

nit: is this still needed?

src/test/java/org/broadinstitute/consent/http/service/DatasetRegistrationServiceTest.java
```diff
@@ -0,0 +1,531 @@
...
+        when(user.getUserId()).thenReturn(1);
+        schemaV1.setDataSubmitterUserId(user.getUserId());
+        schemaV1.setDataCustodianEmail(List.of(RandomStringUtils.randomAlphabetic(10)+"@domain.org"));
+        schemaV1.setPublicVisibility(true);
+        schemaV1.setDataAccessCommitteeId(1);
+        schemaV1.setSubmittingToAnvil(true);
+        schemaV1.setDbGaPPhsID(RandomStringUtils.randomAlphabetic(10));
+        schemaV1.setDbGaPStudyRegistrationName(RandomStringUtils.randomAlphabetic(10));
+        schemaV1.setEmbargoReleaseDate("2007-12-03");
+        schemaV1.setSequencingCenter(RandomStringUtils.randomAlphabetic(10));
```
fboulnois, line 489: I wonder if there's a way to improve creating random fields without having to manually set all the fields like this.

src/test/java/org/broadinstitute/consent/http/service/DatasetRegistrationServiceTest.java
```diff
@@ -0,0 +1,531 @@
...
+        assertEquals(files.get("alternativeDataSharingPlan").getContentDisposition().getFileName(), inserts.get(0).files().get(0).getFileName());
+        assertEquals(BlobId.of("asdf", "hjkl"), inserts.get(0).files().get(0).getBlobId());
+
+        assertEquals(FileCategory.NIH_INSTITUTIONAL_CERTIFICATION, inserts.get(0).files().get(1).getCategory());
+        assertEquals(files.get("consentGroups[0].nihInstitutionalCertificationFile").getContentDisposition().getFileName(), inserts.get(0).files().get(1).getFileName());
+        assertEquals(BlobId.of("qwer", "tyuio"), inserts.get(0).files().get(1).getBlobId());
+
+
+        List<DatasetProperty> props = inserts.get(0).props();
+        assertContainsDatasetProperty(props, "piName", schema.getPiName());
```
fboulnois, line 127: How can we ensure that we haven't missed a property in this list?

src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
```diff
@@ -27,31 +31,113 @@ public DatasetServiceDAO(Jdbi jdbi, DatasetDAO datasetDAO) {
...
+            );
+        }
+    }
+
     // Helper methods to generate Dictionary inserts
+    private void executeSynchronizeDatasetProperties(Handle handle, Integer datasetId, List<DatasetProperty> properties) {
+        // 1. Generate inserts for missing dictionary terms
+        // 2. Generate inserts for new dataset properties
+        // 3. Generate updates for existing dataset properties
+        // 4. Generate deletes for outdated dataset properties
```
fboulnois: While we are improving this code, perhaps it's worth moving these comments above the line of code that is executed?

src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
```diff
@@ -27,31 +31,113 @@ public DatasetServiceDAO(Jdbi jdbi, DatasetDAO datasetDAO) {
...
+                name,
+                new Timestamp(new Date().getTime()),
+                userId,
+                null,
+                false,
+                dataUse.toString(),
+                dacId
+        );
+
+
```
fboulnois: nit: the spacing looks a little strange here, is this expected?

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
```diff
@@ -0,0 +1,440 @@
...
+
+            return Optional.of(datasetProperty);
+
+        }
+    };
+
+    private static final List<DatasetPropertyExtractor> DATASET_REGISTRATION_V1_PROPERTY_EXTRACTORS = List.of(
+            new DatasetPropertyExtractor(
+                    "PI Name", "piName", DatasetPropertyType.String,
+                    (registration, consentGroup) -> registration.getPiName()),
```
fboulnois, line 252: Once again I wonder if there's a better way to do this in an automatic way so we don't miss anything.

fboulnois (commented):
A few comments, I think it's worth talking about the PR in a larger discussion:

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
rushtong: > Once again I wonder if there's a better way to do this in an automatic way so we don't miss anything.

I have the same concern. `"piName"` is used in several places, for example. We don't have anything that, say, correlates the two strings, `"PI Name"` and `"piName"` however. Since these come from `DatasetRegistrationSchemaV1`, I'm not seeing a clear solution to aligning all of them. The other complicating factor here is that `DatasetRegistrationSchemaV1` is made with a code generator so maintaining the original json schema with manual changes to that class is challenging.

src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
connorlbark: I talked to Jonathan, and I don't think it's relative to anything. He said it would be ok to default to UTC. If it's preferred, I could convert these to an Instant with a UTC timezone, but it introduced a bit of complexity in parsing since it has to handle the timezoned and timezoneless case.

src/test/java/org/broadinstitute/consent/http/service/DatasetRegistrationServiceTest.java
connorlbark: I agree - I'm not sure there is a great way, though. A similar pattern can be seen throughout the app, but this is just one of the bigger objects.

src/test/java/org/broadinstitute/consent/http/service/DatasetRegistrationServiceTest.java
connorlbark: Hmm, the complication is that not all fields of the dataset registration are turned into properties - only a subset - so we cannot use reflection or something like that to check they all exist. Unless we basically copied over the extractors - which would make the tests depend and use the potentially flawed implementation of the registration system, I am not sure there is a great way to ensure that. I am with you that it's not ideal though. I could break it up visually more, but I'm not sure that solves the problem.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
connorlbark: I agree it's not pretty, but this is a super old code path using objects we don't use anymore, and it doesn't look tested very well, so I'm wary to change it. I think it will probably be removed soon, as new registration stuff rolls out and deprecates old creation paths.

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
fboulnois: Perhaps we can add a TODO comment and investigate this in another ticket. This PR is already large and I don't want to block merging if the PR is otherwise in a good state.

src/test/java/org/broadinstitute/consent/http/service/DatasetRegistrationServiceTest.java
fboulnois: Perhaps worth adding a TODO comment here.

src/test/java/org/broadinstitute/consent/http/service/DatasetRegistrationServiceTest.java
fboulnois: Perhaps worth adding a TODO comment here.

fboulnois (approved):
lgtm, minor suggestions:

src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
nmalfroy: I think typically it's better to default to UTC and wait until the data gets to the client to render the date but it's your call.  My main concern here is if we are recording dates based on the user's locale but not recording the timezone, because we then don't actually know the time.  Is this getting stored in the DB?

nmalfroy (commented):
This looks good but I'd love to confirm that we're not storing dates without timezones (just want to make sure we're not storing partial dates since that's tough/impossible to reconstitute)

rushtong (approved):
👍🏽

connorlbark:
> This looks good but I'd love to confirm that we're not storing dates without timezones (just want to make sure we're not storing partial dates since that's tough/impossible to reconstitute)

👍, I made it default to UTC if not provided by the user

src/main/java/org/broadinstitute/consent/http/enumeration/DatasetPropertyType.java
```diff
@@ -66,9 +66,14 @@ public static Boolean coerceToBoolean(String value) throws IllegalArgumentExcept
 
     public static Instant coerceToDate(String value) throws IllegalArgumentException {
         try {
+
             return Instant.parse(value);
         } catch (DateTimeParseException e) {
-            throw new IllegalArgumentException("Could not parse as Date: " + e.getMessage());
+            try {
+                return Instant.parse(value + "T00:00:00Z"); // try adding timezones
```
rushtong, line 73: I was looking into the `java.util.TimeZone` object to see if there's a better way to use a string-ified default version, but came up short. I think this is OK for now.

rushtong (approved):
👍🏽

PR #1895, opened 2023-01-18
[DUOS-2197][risk=low] New Admin-only Update DataUse API

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2197

This PR adds a new admin-only endpoint to update the data use on existing datasets. There is no logic here for downstream side effects such as existing DAR<->Dataset Match entities no longer being valid. This is a conscious decision for the time being and any potential effects can be mitigated by regenerating elections and reprocessing match results. 

### Note
Recommended that you hide white space in the review as I did a lot of reformatting

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
@@ -7,6 +7,7 @@ import io.dropwizard.auth.Auth;
 import org.apache.commons.collections.CollectionUtils;
 import org.broadinstitute.consent.http.enumeration.UserRoles;
 import org.broadinstitute.consent.http.models.AuthUser;
+import org.broadinstitute.consent.http.models.DataUse;
 import org.broadinstitute.consent.http.models.Dataset;
 import org.broadinstitute.consent.http.models.Dictionary;
 import org.broadinstitute.consent.http.models.User;
@@ -503,6 +504,22 @@ public class DatasetResource extends Resource {
             Dataset dataset = datasetService.updateNeedsReviewDatasets(dataSetId, needsApproval);
             return Response.ok().entity(unmarshal(dataset)).build();
         }catch (Exception e){
+            return createExceptionResponse(e);
+        }
+    }
+
+    @PUT
+    @Produces("application/json")
+    @RolesAllowed(ADMIN)
+    @Path("/{id}/datause")
+    public Response updateDatasetDataUse(@Auth AuthUser authUser, @PathParam("id") Integer id, String dataUseJson) {
+        try {
+            User user = userService.findUserByEmail(authUser.getEmail());
+            Gson gson = new Gson();
+            DataUse dataUse = gson.fromJson(dataUseJson, DataUse.class);
+            Dataset dataset = datasetService.updateDatasetDataUse(user, id, dataUse);
+            return Response.ok().entity(unmarshal(dataset)).build();
+        } catch (Exception e) {
             return createExceptionResponse(e);
         }
     }
--- a/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
@@ -177,7 +177,7 @@ public class DatasetResource extends Resource {
                     files);
 
             URI uri = UriBuilder.fromPath("/api/dataset/v2").build();
-            return Response.created(uri).entity(unmarshal(datasets)).build();
+            return Response.created(uri).entity(datasets).build();
         } catch (Exception e) {
             return createExceptionResponse(e);
         }
@@ -274,7 +274,7 @@ public class DatasetResource extends Resource {
         try {
             User user = userService.findUserByEmail(authUser.getEmail());
             List<Dataset> datasets = datasetService.findAllDatasetsByUser(user);
-            return Response.ok(unmarshal(datasets)).build();
+            return Response.ok(datasets).build();
         } catch (Exception e) {
             return createExceptionResponse(e);
         }
@@ -304,7 +304,7 @@ public class DatasetResource extends Resource {
             if (Objects.isNull(dataset)) {
                 throw new NotFoundException("Could not find the dataset with id: " + datasetId.toString());
             }
-            return Response.ok(unmarshal(dataset)).build();
+            return Response.ok(dataset).build();
         } catch (Exception e){
             return createExceptionResponse(e);
         }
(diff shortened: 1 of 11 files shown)


src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
```diff
@@ -828,4 +844,15 @@ private FileStorageObject createFileStorageObject(String entityId, FileCategory
...
+    private Dataset insertDataset() {
+        User user = createUser();
+        String name = "Name_" + RandomStringUtils.random(20, true, true);
+        Timestamp now = new Timestamp(new Date().getTime());
+        String objectId = "Object ID_" + RandomStringUtils.random(20, true, true);
+        DataUse dataUse = new DataUseBuilder().setGeneralUse(true).build();
+        Integer id = datasetDAO.insertDataset(name, now, user.getUserId(), objectId, true, dataUse.toString(), null);
+        createDatasetProperties(id);
+        return datasetDAO.findDatasetById(id);
+    }
```
rushtong, line 857: This is an intentional copy from `DAOTestHelper` - we're trying to move away from common methods in that class in favor of replicating the required behavior in specific DAO testing classes.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -507,6 +505,31 @@ public Response updateNeedsReviewDataSets(@QueryParam("dataSetId") Integer dataS
...
+        try {
+            User user = userService.findUserByEmail(authUser.getEmail());
+            Gson gson = new Gson();
+            DataUse dataUse = gson.fromJson(dataUseJson, DataUse.class);
+            Dataset originalDataset = datasetService.findDatasetById(id);
+            if (Objects.isNull(originalDataset)) {
+                throw new NotFoundException("Dataset not found: " + id);
+            }
+            if (Objects.equals(dataUse, originalDataset.getDataUse())) {
+                return Response.notModified().entity(unmarshal(originalDataset)).build();
```
connorlbark: oh cool, i didn't know you could do more than `ok` - side note, you don't need to use `unmarshal` here, in fact we could completely remove it since we're using Gson by default (same goes for the `.ok` call below)

src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
```diff
@@ -421,19 +419,19 @@ public void testFindDataSetsByAuthUserEmail() {
...
-        List<Integer> datasetIds = datasets.stream().map(Dataset::getDataSetId).collect(Collectors.toList());
+        List<Integer> datasetIds = datasets.stream().map(Dataset::getDataSetId).toList();
         assertTrue(datasetIds.contains(dataset.getDataSetId()));
     }
 
     @Test
     public void testFindNonDACDataSets() {
-        Dataset dataset = createDataset();
+        Dataset dataset = insertDataset();
         Consent consent = createConsent();
```
connorlbark, line 429: curious: why preference for `insertDataset`? should we start migrating verbage of others, e.g., `createConsent`, into insert as well?

src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
```diff
@@ -828,4 +844,15 @@ private FileStorageObject createFileStorageObject(String entityId, FileCategory
         );
         return fileStorageObjectDAO.findFileById(newFileStorageObjectId);
     }
+
+    private Dataset insertDataset() {
+        User user = createUser();
+        String name = "Name_" + RandomStringUtils.random(20, true, true);
+        Timestamp now = new Timestamp(new Date().getTime());
+        String objectId = "Object ID_" + RandomStringUtils.random(20, true, true);
```
connorlbark, line 852: unrelated to this ticket, but i have never understood what objectId is - is it just a UUID? what is it used for, aside from datasetId?

src/test/java/org/broadinstitute/consent/http/service/DatasetServiceTest.java
```diff
@@ -377,6 +378,49 @@ public void testUpdateDatasetNotModified() {
...
+            UserRoles.MEMBER,
+            UserRoles.RESEARCHER,
+            UserRoles.SIGNINGOFFICIAL,
+            UserRoles.DATAOWNER,
+            UserRoles.DATASUBMITTER,
+            UserRoles.ITDIRECTOR,
+            UserRoles.ALUMNI
+        ).forEach(r -> {
+            u.addRole(new UserRole(r.getRoleId(), r.getRoleName()));
+        });
```
connorlbark, line 414: i like this testing pattern, never seen it before

connorlbark (approved):
👍 i like the tests, looking great.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
rushtong: Thanks, that's great, this is what copy paste gets me.

src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
rushtong: I couldn't override `createDataset` in `DAOTestHelper` with a private method. The goal, over time, is to reduce dependency on the the common `DAOTestHelper` methods that create things that have other things in them, i.e., datasets with dacs and users, etc. I want to slim that class down since it has grown into an uber-service class. The only thing it should be doing is providing the base DAO classes, setup, and tear-down.

src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
rushtong: It's an old value that we don't use any more. It used to refer to a Broad sample set (i.e. `SM-1234`) that we got when it was exported from ORSP. That feature was removed from ORSP ages ago and we don't really support it. We should file a ticket to get rid of it.

src/test/java/org/broadinstitute/consent/http/service/DatasetServiceTest.java
rushtong: I think it would be cleaner to have a `User.addRoles(UserRoles... roles)` method purely for testing, but this will do.

quazi-h (approved):
Changes look good and tests are running successfully for me locally.

PR #1896, opened 2023-01-18
[DUOS-2246][risk=no] Fix update self institution logic

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2246

Minor update to the logic around a user updating their own institution. Previous logic was preventing non-admin, non-SO, non-IT users from updating their institution. 

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
@@ -2,6 +2,7 @@ package org.broadinstitute.consent.http.resources;
 
 
 import com.google.api.client.http.HttpStatusCodes;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
 import com.google.gson.reflect.TypeToken;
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -227,17 +227,15 @@ private boolean canUpdateInstitution(User user, Integer newInstitutionId) {
...
             }
-
             if (user.hasUserRole(UserRoles.SIGNINGOFFICIAL) || user.hasUserRole(UserRoles.ITDIRECTOR)) {
                 // can only update institution if not set.
                 return Objects.isNull(user.getInstitutionId()) && Objects.nonNull(newInstitutionId);
             }
-
-            return false;
+            // User is not restricted based on role
+            return true;
```
connorlbark, line 237: should we add a quick test for this?

rushtong: Yeah, I should have done that. The method is private and only used once - will open it up for tests

quazi-h (approved):
Changes in logic look good and tests are passing. I tested this locally and was able to update my institution using the user update endpoint.

PR #1900, opened 2023-01-22
[DUOS-2171][risk=no] Replace set-output in action

rushtong:
## Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2171

See also: 
* https://github.com/DataBiosphere/consent-ontology/pull/761
* https://github.com/DataBiosphere/duos-ui/pull/1996

----

- Verify all tests go green

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -21,26 +21,26 @@ jobs:
       uses: actions/checkout@v3
     - name: Get Short Sha
       id: short-sha
-      run: echo "sha=$(git rev-parse --short=12 HEAD)" >> $GITHUB_ENV
+      run: echo "sha=$(git rev-parse --short=12 HEAD)" >> $GITHUB_OUTPUT
     - name: 'Set up Cloud SDK'
       uses: 'google-github-actions/setup-gcloud@v1'
     - name: Construct tags
       id: construct-tags
       run: |
-        SHA_TAG="${REGISTRY_HOST}/${GOOGLE_PROJECT}/${SERVICE_NAME}:${{ env.sha }}"
+        SHA_TAG="${REGISTRY_HOST}/${GOOGLE_PROJECT}/${SERVICE_NAME}:${{ steps.short-sha.outputs.sha }}"
         ENVIRONMENT_TAG=""
         if ${{ github.event_name == 'pull_request'}}; then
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -21,26 +21,26 @@ jobs:
...
         ENVIRONMENT_TAG=""
         if ${{ github.event_name == 'pull_request'}}; then
-          ENVIRONMENT_TAG="${REGISTRY_HOST}/${GOOGLE_PROJECT}/${SERVICE_NAME}:pr-${{ steps.short-sha.outputs.sha }}"
+          ENVIRONMENT_TAG="${REGISTRY_HOST}/${GOOGLE_PROJECT}/${SERVICE_NAME}:pr-${{ env.sha }}"
         elif ${{github.event_name == 'push' }}; then
           ENVIRONMENT_TAG="${REGISTRY_HOST}/${GOOGLE_PROJECT}/${SERVICE_NAME}:dev"
         fi
-        echo ::set-output name=sha-tag::$SHA_TAG
-        echo ::set-output name=environment-tag::$ENVIRONMENT_TAG
+        echo "sha-tag=$SHA_TAG" >> $GITHUB_ENV
```
nmalfroy: Not asking as a change request but just curiosity: any reason you changed to `GITHUB_ENV` as opposed to `GITHUB_OUTPUT` (I think the later would avoid the changes like on line 33)?  I actually think it makes more sense to do what you did and change to GITHUB_ENV (seems a lot easier to deal with than worry about which step an output belongs to) but more for my own education

nmalfroy (approved):
lgtm!

.github/workflows/build.yaml
rushtong: I think I was using [this](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions) as a guide but yeah, I could go either way here.

connorlbark (approved):
👍

.github/workflows/build.yaml
rushtong: Turns out I was very confused about the difference between `GITHUB_OUTPUT` and `GITHUB_ENV`. Using `GITHUB_OUTPUT` simplifies the PR quite a bit.

fboulnois (approved):

PR #1909, opened 2023-01-30
[DUOS-2277][risk=no] New API to replace existing one for user relevant datasets

rushtong:
## Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DUOS-2277

Related to [DUOS-1929](https://broadworkbench.atlassian.net/browse/DUOS-1929)
This covers API work that was discovered to be missing for chairs and members while working on a similar page for DAC-dataset-approval. This PR creates a new version of the API to get relevant datasets for a user but returning full `Dataset` objects instead of the deprecated `DatasetDTO` objects. The one main difference between the two are how they treat the `alias` value. `Dataset` treats it as a numeric value and adds a `datasetIdentifier` field for the parsed version of the numeric `alias` value while the DTO puts the identifier into the alias.

See also https://github.com/DataBiosphere/duos-ui/pull/1999 where there is some UI work related to this.

----

- Verify all tests go green

[DUOS-1929]: https://broadworkbench.atlassian.net/browse/DUOS-1929?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
@@ -1,5 +1,10 @@
 package org.broadinstitute.consent.http.db;
 
+import java.sql.Timestamp;
+import java.time.Instant;
+import java.util.Collection;
+import java.util.List;
+import java.util.Set;
 import org.apache.commons.lang3.tuple.Pair;
 import org.broadinstitute.consent.http.db.mapper.AssociationMapper;
 import org.broadinstitute.consent.http.db.mapper.DatasetDTOWithPropertiesMapper;
@@ -29,12 +34,6 @@ import org.jdbi.v3.sqlobject.statement.SqlUpdate;
 import org.jdbi.v3.sqlobject.statement.UseRowMapper;
 import org.jdbi.v3.sqlobject.statement.UseRowReducer;
 import org.jdbi.v3.sqlobject.transaction.Transactional;
-
-import java.sql.Timestamp;
-import java.time.Instant;
-import java.util.Collection;
-import java.util.List;
-import java.util.Set;
 
 @RegisterRowMapper(DatasetMapper.class)
 @RegisterRowMapper(FileStorageObjectMapperWithFSOPrefix.class)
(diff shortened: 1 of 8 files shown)


src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
```diff
@@ -456,6 +455,18 @@ void updateDataset(@Bind("datasetId") Integer datasetId,
...
+    @SqlQuery("""
+            SELECT distinct d.*, k.key, p.property_value, c.consent_id, d.dac_id, c.translated_use_restriction
+            FROM dataset d
+            LEFT JOIN dataset_property p ON p.dataset_id = d.dataset_id
+            LEFT JOIN dictionary k ON k.key_id = p.property_key
+            LEFT JOIN consent_associations a ON a.dataset_id = d.dataset_id
+            LEFT JOIN consents c ON c.consent_id = a.consent_id
+            WHERE d.dac_id IN (<dacIds>)
+        """)
+    List<Dataset> findDatasetListByDacIds(@BindList("dacIds") List<Integer> dacIds);
```
nmalfroy, line 468: I just want to say I really like this dao pattern

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -136,6 +137,24 @@ public Response getDatasetsFromUserDacs(@Auth AuthUser authUser) {
...
+    @Produces("application/json")
+    @RolesAllowed({CHAIRPERSON, MEMBER})
+    public Response getDatasetsFromUserDacsV2(@Auth AuthUser authUser) {
+        try {
+            User user = userService.findUserByEmail(authUser.getEmail());
+            List<Integer> dacIds = user.getRoles().stream()
+                .map(UserRole::getDacId)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList());
+            List<Dataset> datasets = dacIds.isEmpty() ? List.of() : datasetService.findDatasetListByDacIds(dacIds);
```
nmalfroy, line 151: Would it make sense to throw a 404 if the list is empty?

nmalfroy (approved):
Looks good!  Just that one question about returning an specific error

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
rushtong: I'm enjoying this too - it's so much easier to read as well as test if you have a db app up and running.

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
rushtong: Great point, I think that does make sense.

fboulnois (approved):
looks good!

PR #1916, opened 2023-02-07
[DUOS-2303][risk=no] More information Dataset Registration errors

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2303

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
@@ -11,7 +11,6 @@ import org.broadinstitute.consent.http.models.DataUse;
 import org.broadinstitute.consent.http.models.Dataset;
 import org.broadinstitute.consent.http.models.Dictionary;
 import org.broadinstitute.consent.http.models.User;
-import org.broadinstitute.consent.http.models.Error;
 import org.broadinstitute.consent.http.models.UserRole;
 import org.broadinstitute.consent.http.models.dataset_registration_v1.DatasetRegistrationSchemaV1;
 import org.broadinstitute.consent.http.models.dto.DatasetDTO;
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -156,14 +157,12 @@ public Response createDatasetRegistration(
...
-            try {
-                if (!jsonSchemaUtil.isValidSchema_v1(json)) {
-                    throw new BadRequestException("Invalid schema");
-                }
-            } catch (Exception e) {
-                throw new BadRequestException("Invalid schema");
+            List<String> errors = jsonSchemaUtil.validateSchema_v1(json);
+            if (!errors.isEmpty()) {
+                throw new BadRequestException("Invalid schema:\n" + String.join("\n", errors));
             }
```
rushtong, line 163: Optional - we have an `Error` record class meant for this kind of thing and works well when converted to json: `org.broadinstitute.consent.http.models.Error`

rushtong (approved):
See optional suggestion inline 👍🏽

src/main/java/org/broadinstitute/consent/http/util/JsonSchemaUtil.java
```diff
@@ -83,7 +85,9 @@ public boolean isValidSchema_v1(String datasetRegistrationInstance) {
           logger.trace("Validation error: " + m);
         }
       }
-      return false;
+      return ve.getAllMessages();
+    } catch (Exception e) {
```
fboulnois, line 89: Are we able to catch anything more specific here?

fboulnois (approved):
Overall looks good, one minor comment:

src/main/java/org/broadinstitute/consent/http/util/JsonSchemaUtil.java
connorlbark: I considered sending back the error message, but I didn't want to send the user too many details if we're not sure what the exception is. I thought a generic error here was appropriate as such

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
connorlbark: Good idea. I'll change that.

PR #1921, opened 2023-02-09
[DUOS-2291][risk=no] Add create user info to select Dataset APIs

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2291

This PR adds the create user, when available, to the following dataset APIs:
* GET /api/dac/{dacId}/datasets
* GET /api/dataset/batch
* GET /api/dataset/search
* GET /api/dataset/v2
* GET /api/dataset/v2/{id}
* GET /api/tdr/{identifier}

Additionally, this PR cleans up a lot of duplicate and unused code.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
@@ -248,6 +248,41 @@ public interface DatasetDAO extends Transactional<DatasetDAO> {
     """)
     List<Dataset> findDatasetsAssociatedWithDac(@Bind("dacId") Integer dacId);
 
+    @RegisterBeanMapper(value = User.class, prefix = "u")
+    @UseRowReducer(DatasetReducer.class)
+    @SqlQuery("""
+        SELECT d.dataset_id, d.name, d.create_date, d.create_user_id, d.update_date,
+            d.update_user_id, d.object_id, d.active, d.dac_id, d.alias, d.data_use, d.dac_approval,
+            dar_ds_ids.id AS in_use,
+            u.user_id AS u_user_id, u.email AS u_email, u.display_name AS u_display_name,
+            u.create_date AS u_create_date, u.email_preference AS u_email_preference,
+            u.institution_id AS u_institution_id, u.era_commons_id AS u_era_commons_id,
+            k.key, dp.property_value, dp.property_key, dp.property_type, dp.schema_property, dp.property_id,
+            ca.consent_id, c.translated_use_restriction,
+            fso.file_storage_object_id AS fso_file_storage_object_id,
+            fso.entity_id AS fso_entity_id,
+            fso.file_name AS fso_file_name,
+            fso.category AS fso_category,
+            fso.gcs_file_uri AS fso_gcs_file_uri,
+            fso.media_type AS fso_media_type,
+            fso.create_date AS fso_create_date,
+            fso.create_user_id AS fso_create_user_id,
+            fso.update_date AS fso_update_date,
+            fso.update_user_id AS fso_update_user_id,
+            fso.deleted AS fso_deleted,
+            fso.delete_user_id AS fso_delete_user_id
+        FROM dataset d
+        LEFT JOIN users u on d.create_user_id = u.user_id
+        LEFT JOIN (SELECT DISTINCT dataset_id AS id FROM dar_dataset) dar_ds_ids ON dar_ds_ids.id = d.dataset_id
+        LEFT JOIN dataset_property dp ON dp.dataset_id = d.dataset_id
+        LEFT JOIN dictionary k ON k.key_id = dp.property_key
+        LEFT JOIN consent_associations ca ON ca.dataset_id = d.dataset_id
+        LEFT JOIN file_storage_object fso ON fso.entity_id = d.dataset_id::text AND fso.deleted = false
+        LEFT JOIN consents c ON c.consent_id = ca.consent_id
+        WHERE d.name IS NOT NULL AND d.active = true
+    """)
+    List<Dataset> getActiveDatasets();
+
     @UseRowReducer(DatasetReducer.class)
     @SqlQuery(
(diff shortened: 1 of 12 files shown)


src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
```diff
@@ -372,28 +403,12 @@ public void testGetDictionaryTerms() {
             assertNotNull(t.getKey());
         });
     }
-    @Test
-    public void testGetDataSetsForObjectIdList() {
```
fboulnois, line 376: Can you explain why some of these tests are no longer relevant? Is it because it was testing the duplicated codepaths / queries you mentioned? Also, what makes you confident that these should be deleted?

fboulnois (approved):
Overall, the inclusion of the user info makes sense to me. One comment around testing strategy as I familiarize myself with the codebase:

src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
rushtong: It's hard to tell from such a convoluted change list, but the underlying method under test here was removed so I removed the test method. This is something we unfortunately run into too often. We have extensive unit tests and sometimes the only usage of a method is in a test and not application code. I routinely prune these and look for this case with most PRs.

src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
```diff
@@ -872,4 +872,33 @@ private Dataset insertDataset() {
...
+    }
+
+    protected Consent insertConsent() {
+        String consentId = UUID.randomUUID().toString();
+        consentDAO.insertConsent(consentId,
+                false,
+            """
+                {"type":"everything"}
+                """,
+            """
```
connorlbark: nit: formatting

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -558,12 +542,8 @@ public List<Dataset> findAllDatasetsByUser(User user) {
         }
     }
 
-    public Dataset getDataset(Integer datasetId) {
-        return this.datasetDAO.findDatasetById(datasetId);
-    }
-
     public List<Dataset> getDatasets(List<Integer> datasetIds) {
```
connorlbark: should this be renamed to `findDatasets`? it seems like the naming is out of convention, and it might make sense to update now since we're changing names in other places

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
```diff
@@ -57,33 +57,265 @@ Integer insertDataset(
         @Bind("dataUse") String dataUse,
         @Bind("dacId") Integer dacId);
 
+    @RegisterBeanMapper(value = User.class, prefix = "u")
     @UseRowReducer(DatasetReducer.class)
-    @SqlQuery(
-        "SELECT d.*, k.key, dp.property_value, dp.property_key, dp.property_type, dp.schema_property, dp.property_id, " +
-            " ca.consent_id, c.translated_use_restriction, dar_ds_ids.id as in_use, " +
-            FileStorageObject.QUERY_FIELDS_WITH_FSO_PREFIX +
```
connorlbark, line 64: should we keep using the constant strings? now with use and fso both spelled out the query is very long / hard to parse, and maybe more brittle in case those change

connorlbark: (same thought goes for all the below DAO changes)

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
```diff
@@ -57,33 +57,265 @@ Integer insertDataset(
...
+        LEFT JOIN dataset_property dp ON dp.dataset_id = d.dataset_id
+        LEFT JOIN dictionary k ON k.key_id = dp.property_key
+        LEFT JOIN consent_associations ca ON ca.dataset_id = d.dataset_id
+        LEFT JOIN file_storage_object fso ON fso.entity_id = d.dataset_id::text AND fso.deleted = false
+        LEFT JOIN consents c ON c.consent_id = ca.consent_id
+        WHERE d.dataset_id = :datasetId
+    """)
     Dataset findDatasetById(@Bind("datasetId") Integer datasetId);
 
+    @RegisterBeanMapper(value = User.class, prefix = "u")
```
connorlbark: nit: maybe move this to the top of the class? so we don't need it on every call.

connorlbark (commented):
looks great! couple nits, then 👍

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
rushtong: So ... I had a similar concern. I tried a couple different ways of incorporating the query string constants into text blocks and they all either didn't work syntactically or looked horrible. I went this route because it was much easier to run the dev/test cycle with an open sql editor than trying to finagle query strings between each run.

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
rushtong: I didn't think that would work since it is not necessary for most of the other queries, but it turns out that it does, nice suggestion!

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
rushtong: I'll make that change - yes, it should. We have a fairly mixed bag of naming approaches here.

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
connorlbark: that's fair - i think it would be worth figuring out how to do both, but we don't have to do that in this ticket. something to think about

connorlbark (approved):
👍

PR #1951, opened 2023-03-22
[DUOS-2377][risk=no] Update backend DS schema

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2377

- Everit library is replaced with networknt JsonSchema library
- The new library has a fundamental issue with the `items` keyword in V2020-12 - however, we don't use any of the new 2020-12 functionality, so downgraded to 2019-09 without any changes needed until the library is updated.
- New dataset field: `openAccess`

Schema:
- Makes a lot of fields conditionally required in accordance to how the frontend works
- Reviewed all required fields with jonathan
- Removed some old fields
- Reworked consent group to: only allow 1 primary consent group, and only require DAC id if not open access.
- Added unit tests to ensure the schema's complex functionality works as expected

Testing
- Make calls to the create dataset endpoint, and it should give you human-readable validation messages back.

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/models/DataUse.java
```diff
@@ -88,6 +89,13 @@ public Boolean getPopulationOriginsAncestry() {
     public void setPopulationOriginsAncestry(Boolean populationOriginsAncestry) {
         this.populationOriginsAncestry = populationOriginsAncestry;
     }
+    public Boolean getOpenAccess() {
```
fboulnois:
```suggestion

    public Boolean getOpenAccess() {
```

nit: spacing

src/main/java/org/broadinstitute/consent/http/models/DataUse.java
```diff
@@ -22,6 +22,7 @@ public class DataUse {
     private Boolean hmbResearch;
     private List<String> diseaseRestrictions;
     private Boolean populationOriginsAncestry;
+    private Boolean openAccess;
```
rushtong: This should not be part of DataUse - in essence, it is a dataset property, not a use restriction condition that we need for doing a logical match between a DAR and a Dataset.

connorlbark: I made this a field on the dataset. This PR is now getting pretty big, I am trying to think of a good way to split it up but I am not sure there is.

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
```diff
@@ -43,25 +44,28 @@ public interface DatasetDAO extends Transactional<DatasetDAO> {
...
+            (SELECT :name, :createDate, :createUserId, :createDate,
+                :createUserId, :objectId, :openAccess, :active, :dacId, COALESCE(MAX(alias),0)+1, :dataUse FROM dataset)
+            """)
     @GetGeneratedKeys
     Integer insertDataset(
         @Bind("name") String name,
         @Bind("createDate") Timestamp createDate,
         @Bind("createUserId") Integer createUserId,
         @Bind("objectId") String objectId,
+        @Bind("openAccess") Boolean openAccess,
```
fboulnois: I think if you really wanted to, you could split out the openAccess piece into a separate PR, but since it's coherent with this one I would just keep it here.

fboulnois (approved):
Looks great! Tests are also thorough.

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
rushtong: Not sure if it would reduce the PR size, but it could be a property instead of a top level field on dataset. The only benefit there would be to trim the sql/dao changes down a bit

connorlbark: Hmm, it's up to you, I could see it either way. It feels like being a top-level property makes sense, because it's fundamental to the operation of the dataset. But I could see it being a property as well, because it's a lot simpler.

rushtong: I'm nervous about adding something this important as a last-minute addition without going through all the uses of dataset to ensure it doesn't have any side effects. If it were a property, then the risk is much lower. I don't see it as a primary aspect of a dataset - it seems to have the same level of importance as many of the other properties that we're capturing separately as props.

connorlbark: Sure, can do. I'll make it a property.

connorlbark: This has been finished - ready for a re-review!

rushtong (approved):
Thank you for the changes!

PR #1960, opened 2023-03-29
[DUOS-2391][risk=low] Update approved users query

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2391

This PR updates the query for approved dataset users to look for the **_last_** final `true` vote for a specific dar+dataset election to check if the user is approved or not. Previously, we were looking for **_any_** approved election and would return a user id for the case where the user was first approved, then denied access in a subsequent election.

One edge case this PR does not cover, and is not intended to cover, is the case where a user has requested access to a dataset in more than one data access request. If the user is approved for one DAR, but denied for the other DAR, we still list that user as an approved user since they do, in fact, have an approved DAR against that dataset. To revoke that user's access, ALL DARs would need to be revoked.

### Testing
To test, create and approve (as a DAC Chair) a DAR against a dataset. Then re-open that election and vote "No" so as to deny that user access. Look at the approved user list (`GET` `/api/tdr/{identifier}/approved/users`) for the dataset in question. Before voting "No", your researcher user should show up as an approved user for the dataset. After voting "No", that user should NOT be visible in the approved user list. You can again re-open that DAR, approve it as a Chair, and then the user should now be visible again in the approved users list.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
@@ -63,7 +63,20 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
   )
   List<DataAccessRequest> findAllApprovedDataAccessRequestsByDatasetId(@Bind("datasetId") Integer datasetId);
 
-
+  /**
+   * This query finds unique user ids on dar-dataset combinations where the most recent
+   * vote is true. The query accomplishes this by creating a view that is a grouping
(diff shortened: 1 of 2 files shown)


fboulnois (approved):
The new query is pretty complex! The test makes sense, though.

connorlbark (approved):
I don't fully understand the query but the test makes sense

rushtong:
> I don't fully understand the query but the test makes sense

I will add some commentary since it is so complex and then merge after that.

PR #1972, opened 2023-04-05
[DUOS-2383][risk=low] Add Validate DS Study Name API

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2383

----

- Verify all tests go green


connorlbark:
I do not fully understand the semgrep error here. It seems to just say "you added a new rest endpoint, check that it is secure" and doesn't have a clear way to fix.

Edit: The error didn't suggest any fixes but I looked into the raw semgrep rule and added `StringEscapeUtils.unescapeJava` to the input from the user and it was happy with that.

rushtong:
> I do not fully understand the semgrep error here. It seems to just say "you added a new rest endpoint, check that it is secure" and doesn't have a clear way to fix.
> 
> Edit: The error didn't suggest any fixes but I looked into the raw semgrep rule and added `StringEscapeUtils.unescapeJava` to the input from the user and it was happy with that.

The latest run still shows an error:
>   BLOCKING CODE RULES FIRED:
>    gitlab.find_sec_bugs.JAXRS_ENDPOINT-1

However, I really think this should not be a blocking error from the description. I think our use of this action might be a little outdated so we should look into updating it separately from this PR.

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
```diff
@@ -281,6 +281,40 @@ LEFT JOIN (SELECT DISTINCT dataset_id AS id FROM dar_dataset) dar_ds_ids ON dar_
...
+        LEFT JOIN users u on d.create_user_id = u.user_id
+        LEFT JOIN (SELECT DISTINCT dataset_id AS id FROM dar_dataset) dar_ds_ids ON dar_ds_ids.id = d.dataset_id
+        LEFT JOIN dataset_property dp ON dp.dataset_id = d.dataset_id
+        LEFT JOIN dictionary k ON k.key_id = dp.property_key
+        LEFT JOIN consent_associations ca ON ca.dataset_id = d.dataset_id
+        LEFT JOIN file_storage_object fso ON fso.entity_id = d.dataset_id::text AND fso.deleted = false
+        LEFT JOIN consents c ON c.consent_id = ca.consent_id
+        WHERE (dp.schema_property = 'studyName' AND UPPER(dp.property_value) = UPPER(:name))
+    """)
+    Dataset findDatasetByStudyName(@Bind("name") String name);
```
rushtong: There can be many datasets with the same study name. When we parse an incoming dataset submission form, if there are N consent groups, there will be N datasets and each one will have the same study name.

connorlbark: Oh interesting. I can make this return a list of ids instead. It seems somewhat unintuitive to me that the studyName property is how multiple datasets are linked as being under the same submission, and that the linking is unique. But trying to envision another approach makes things super complicated so it makes sense.

rushtong: Yeah, I'm thinking that we might want to trim this feature down a bit. The only use case here is to ensure unique study names. If we instead returned just that ... a list of unique study names, we'd speed up the query quite a bit. Separately, the the UI could client-side filter to ensure uniqueness. Just a suggestion - making this return a list is just as functional.

connorlbark: Hmm... I feel like returning a list of IDs might be simpler. It could also just return a `true`/`false` -- all it would have to do is check if there is a dataset property with `studyName` of the provided input, which would make it fast. Let me know what you think, and I can implement it either way

rushtong: The UI will need to check string values, so if it has them in memory (i.e., make a single API call), it would be a trivial check. With IDs, you might have to submit the query multiple times to see if the string matches. Depending on how that UI widget is implemented, it could be many API calls, like our ontology autocomplete. I think a list of existing study names is easier for the UI to process.

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
```diff
@@ -281,6 +281,14 @@ LEFT JOIN (SELECT DISTINCT dataset_id AS id FROM dar_dataset) dar_ds_ids ON dar_
     """)
     List<Dataset> getActiveDatasets();
 
+    @SqlQuery("""
+        SELECT dp.property_value
```
rushtong: nit: adding DISTINCT will offload the work of set coercion to the db
```suggestion
        SELECT DISTINCT dp.property_value
```

src/main/resources/assets/paths/findAllStudyNames.yaml
```diff
@@ -0,0 +1,16 @@
+get:
+  summary: Find All Study Names
+  description: Returns a list of all the actively used study names in the backend.
```
rushtong: nit, language. 
```suggestion
  description: Returns a list of all the actively used dataset study names.
```

src/main/resources/assets/paths/findAllStudyNames.yaml
```diff
@@ -0,0 +1,16 @@
...
+    200:
+      description: dataset id
+      content:
+        application/json:
+          schema:
+            type: array
+            items:
+              type: string
+    404:
+      description: Dataset not found
```
rushtong: I'm not clear on why we'd return a 404 from this API. I can see the need for a 500 in the case of a database error.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -353,6 +354,20 @@ public Response validateDatasetName(@QueryParam("name") String name) {
...
+    @Consumes("application/json")
+    @Produces("application/json")
+    @Path("/studyNames")
+    @PermitAll
+    public Response findAllActiveStudyNames() {
+        try {
+            Set<String> studyNames = datasetService.findAllActiveStudyNames();
+            return Response.ok(studyNames).build();
+        } catch (Exception e) {
+            throw new NotFoundException("Could not get all study names");
```
rushtong: If you hit an exception here, there's something else wrong, this doesn't seem like a 404 to me, perhaps a 500 error since we should never have an error case here. An empty list might still be a valid list of study names in use. Also, see how we use the super method of `createExceptionResponse(e)` which will try to examine the exception type and return an appropriate error code.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -364,6 +365,14 @@ public void testValidateDatasetNameNotFound() {
         resource.validateDatasetName("test");
     }
 
+    @Test
+    public void testFindAllActiveStudyNamesSuccess() {
+        when(datasetService.findAllActiveStudyNames()).thenReturn(Set.of("Hi", "Hello"));
+        initResource();
+        Response response = resource.findAllActiveStudyNames();
+        assertEquals(200, response.getStatus());
+    }
```
rushtong, line 374: Might be nice to test an exception case here too. For example, we don't want to percolate Sql exceptions back up to the client since they often contain somewhat sensitive info useful in security-fishing. The base `Resource` class should help handle that for you.

rushtong (changes requested):
Looking good, see minor comments inline.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
connorlbark: Oops yeah, this was all copied over code and schema. My bad, I should have looked more carefully at it.

rushtong (approved):
Thank you for the changes! 👍🏽

fboulnois (approved):
logic and tests lgtm 👍

PR #1988, opened 2023-04-27
[DUOS-2446][risk=no] Add Study/StudyProperty Tables

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2446

This ended up being honestly a fair bit more complicated than I imagined.

Originally, I was hoping to have the study have a list of dataset objects:

```
{
   study_id: 2,
   datasets: [
      { dataset_id: 3, ... } 
   ]
}
```

But this proved to be pretty complicated with how datasets are currently represented, mapped, and reduced.

If we wanted to have datasets underneath studies, then we would have to: rewrite the dataset reducer to be more flexible, and probably change every single DAO call for datasets pretty significantly. At least, it seemed this way when I tried to do this.

Instead, I opted to have studies just have a list of dataset ids (not the full object), and datasets have the full study object (if present), like so:

```
{
   dataset_id: 3,
   study: {
      study_id: 1,
      ....
   }
}
```

Given that not all datasets are underneath a study right now, I think this approach makes more sense for now. E.g., in the new dataset catalog we're working on, I assume we want to be backwards compatible with datasets which don't have a study. This approach makes it easy - get all datasets, like we're doing now, and simply group them by study wherever possible.

However, long term, once all datasets have a study, I do think it makes sense to have datasets be returned underneath studies. In this scheme, the dataset catalog would do a call to get all studies (instead of all datasets). The studies themselves would contain the full dataset object. These datasets would already be properly grouped, and rendering them as such would be really straightforward.

Until then, though, this more complicated / less intuitive (imo) approach is necessary.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
@@ -74,11 +74,22 @@ public interface DatasetDAO extends Transactional<DatasetDAO> {
             k.key, dp.property_value, dp.property_key, dp.property_type, dp.schema_property, dp.property_id,
             ca.consent_id, c.translated_use_restriction,
             fso.file_storage_object_id AS fso_file_storage_object_id,
-            s.study_id AS s_study_id, s.name AS s_name,
-            s.description AS s_description, s.data_types AS s_data_types,
-            s.pi_name AS s_pi_name, s.public_visibility AS s_public_visibility,
-            sp.study_property_id AS sp_study_property_id, sp.key AS sp_key,
-            sp.type AS sp_type, sp.value AS sp_value,
+            s.study_id AS s_study_id,
+            s.name AS s_name,
+            s.description AS s_description,
+            s.data_types AS s_data_types,
+            s.pi_name AS s_pi_name,
+            s.create_user_id AS s_create_user_id,
+            s.create_date AS s_create_date,
+            s.update_user_id AS s_user_id,
+            s.update_date AS s_update_date,
+            s.public_visibility AS s_public_visibility,
+            s_dataset.dataset_id AS s_dataset_id,
+            sp.study_property_id AS sp_study_property_id,
+            sp.study_id AS sp_study_id,
+            sp.key AS sp_key,
+            sp.value AS sp_value,
+            sp.type AS sp_type,
             fso.entity_id AS fso_entity_id,
             fso.file_name AS fso_file_name,
             fso.category AS fso_category,
@@ -96,10 +107,11 @@ public interface DatasetDAO extends Transactional<DatasetDAO> {
         LEFT JOIN dataset_property dp ON dp.dataset_id = d.dataset_id
         LEFT JOIN dictionary k ON k.key_id = dp.property_key
         LEFT JOIN consent_associations ca ON ca.dataset_id = d.dataset_id
-        LEFT JOIN file_storage_object fso ON fso.entity_id = d.dataset_id::text AND fso.deleted = false
-        LEFT JOIN consents c ON c.consent_id = ca.consent_id
         LEFT JOIN study s ON s.study_id = d.study_id
         LEFT JOIN study_property sp ON sp.study_id = s.study_id
+        LEFT JOIN dataset s_dataset ON s_dataset.study_id = s.study_id
+        LEFT JOIN file_storage_object fso ON (fso.entity_id = d.dataset_id::text OR fso.entity_id = s.uuid::text) AND fso.deleted = false
+        LEFT JOIN consents c ON c.consent_id = ca.consent_id
         WHERE d.dataset_id = :datasetId
     """)
     Dataset findDatasetById(@Bind("datasetId") Integer datasetId);
@@ -114,6 +126,22 @@ public interface DatasetDAO extends Transactional<DatasetDAO> {
             u.institution_id AS u_institution_id, u.era_commons_id AS u_era_commons_id,
             k.key, dp.property_value, dp.property_key, dp.property_type, dp.schema_property, dp.property_id,
             ca.consent_id, c.translated_use_restriction,
+            s.study_id AS s_study_id,
+            s.name AS s_name,
+            s.description AS s_description,
+            s.data_types AS s_data_types,
+            s.pi_name AS s_pi_name,
+            s.create_user_id AS s_create_user_id,
+            s.create_date AS s_create_date,
+            s.update_user_id AS s_user_id,
+            s.update_date AS s_update_date,
+            s.public_visibility AS s_public_visibility,
+            s_dataset.dataset_id AS s_dataset_id,
+            sp.study_property_id AS sp_study_property_id,
+            sp.study_id AS sp_study_id,
+            sp.key AS sp_key,
+            sp.value AS sp_value,
+            sp.type AS sp_type,
             fso.file_storage_object_id AS fso_file_storage_object_id,
             fso.entity_id AS fso_entity_id,
             fso.file_name AS fso_file_name,
@@ -132,6 +160,9 @@ public interface DatasetDAO extends Transactional<DatasetDAO> {
         LEFT JOIN dataset_property dp ON dp.dataset_id = d.dataset_id
         LEFT JOIN dictionary k ON k.key_id = dp.property_key
         LEFT JOIN consent_associations ca ON ca.dataset_id = d.dataset_id
+        LEFT JOIN study s ON s.study_id = d.study_id
+        LEFT JOIN study_property sp ON sp.study_id = s.study_id
+        LEFT JOIN dataset s_dataset ON s_dataset.study_id = s.study_id
         LEFT JOIN file_storage_object fso ON fso.entity_id = d.dataset_id::text AND fso.deleted = false
         LEFT JOIN consents c ON c.consent_id = ca.consent_id
         WHERE d.dataset_id in (<datasetIds>)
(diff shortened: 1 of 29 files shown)


src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
```diff
@@ -141,6 +141,7 @@ public static void startUp() throws Exception {
         electionDAO = jdbi.onDemand(ElectionDAO.class);
         userRoleDAO = jdbi.onDemand(UserRoleDAO.class);
         voteDAO = jdbi.onDemand(VoteDAO.class);
+        studyDAO = jdbi.onDemand(StudyDAO.class);
```
rushtong, line 144: In the teardown method, you'll need to delete all studies.

rushtong (commented):
One comment while I'm looking at this.

src/main/java/org/broadinstitute/consent/http/db/mapper/StudyReducer.java
```diff
@@ -0,0 +1,78 @@
...
+  public void accumulate(Map<Integer, Study> map, RowView rowView) {
+    Study study =
+            map.computeIfAbsent(
+                    rowView.getColumn("study_id", Integer.class), id -> rowView.getRow(Study.class));
+
+    reduceStudy(study, rowView);
+  }
+
+
+  public void reduceStudy(Study study, RowView rowView) {
```
connorlbark, line 25: Note: doing this allowed me to use the reducer in a more generic way - the reducer could be used both when reducing the study field for a dataset in the DatasetDAO and for reducing normal study objects.

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
```diff
@@ -89,7 +107,10 @@ LEFT JOIN (SELECT DISTINCT dataset_id AS id FROM dar_dataset) dar_ds_ids ON dar_
         LEFT JOIN dataset_property dp ON dp.dataset_id = d.dataset_id
         LEFT JOIN dictionary k ON k.key_id = dp.property_key
         LEFT JOIN consent_associations ca ON ca.dataset_id = d.dataset_id
-        LEFT JOIN file_storage_object fso ON fso.entity_id = d.dataset_id::text AND fso.deleted = false
+        LEFT JOIN study s ON s.study_id = d.study_id
+        LEFT JOIN study_property sp ON sp.study_id = s.study_id
+        LEFT JOIN dataset s_dataset ON s_dataset.study_id = s.study_id
+        LEFT JOIN file_storage_object fso ON (fso.entity_id = d.dataset_id::text OR fso.entity_id = s.uuid::text) AND fso.deleted = false
```
connorlbark, line 113: I noticed in this PR that FSO entity ids for datasets are linked to `dataset_id` and not some UUID. This means if we linked to another table's `id` we could MAYBE run into collisions - however, I am assuming that all FSO categories will be associated which just one kind of entity - for example, `ALTERNATIVE_DATA_SHARING_PLAN`s will only be associated with Studies, etc. So, I don't think it's a big deal - just something that I noticed as a yellow-flag.

connorlbark (commented):
Some notes.

src/main/java/org/broadinstitute/consent/http/db/mapper/StudyReducer.java
```diff
@@ -0,0 +1,78 @@
...
+            && Objects.nonNull(rowView.getColumn("fso_file_storage_object_id", Integer.class))
+    ) {
+      FileStorageObject fileStorageObject = rowView.getRow(FileStorageObject.class);
+
+      switch (fileStorageObject.getCategory()) {
+        case ALTERNATIVE_DATA_SHARING_PLAN -> {
+          if (isFileNewer(fileStorageObject, study.getAlternativeDataSharingPlan())) {
+            study.setAlternativeDataSharingPlan(fileStorageObject);
+          }
+        }
```
connorlbark, line 68: Right now, datasets also have an alternative data sharing plan. However, that is actively used in the create endpoint - in the following ticket, then it should be removed from datasets alongside changing the FSO creation mechanism.

src/test/java/org/broadinstitute/consent/http/db/StudyDAOTest.java
```diff
@@ -0,0 +1,279 @@
...
+        Study study = studyDAO.findStudyById(id);
+
+        assertEquals(fso.getFileStorageObjectId(), study.getAlternativeDataSharingPlan().getFileStorageObjectId());
+        assertEquals(fso.getBlobId(), study.getAlternativeDataSharingPlan().getBlobId());
+
+    }
+
+    @Test
+    public void testIncludesDatasetIds() {
+        Study s = this.insertStudyWithProperties();
```
rushtong: Here and elsewhere, `this.` is extraneous in the current test context

src/test/java/org/broadinstitute/consent/http/db/StudyDAOTest.java
```diff
@@ -0,0 +1,279 @@
...
+        insertDataset();
+
+        s = studyDAO.findStudyById(s.getStudyId());
+
+        assertEquals(2, s.getDatasetIds().size());
+        assertTrue(s.getDatasetIds().contains(ds1.getDataSetId()));
+        assertTrue(s.getDatasetIds().contains(ds2.getDataSetId()));
+    }
+
+    private FileStorageObject createFileStorageObject(String entityId, FileCategory category) {
```
rushtong, line 197: Nice work pulling in these helpers private to the test class.

src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
```diff
@@ -177,8 +178,10 @@ public void tearDown() {
         testingDAO.deleteAllDatasetProperties();
         testingDAO.deleteAllDictionaryTerms();
         testingDAO.deleteAllDatasetAssociations();
-        testingDAO.deleteAllDatasetAudits();;
+        testingDAO.deleteAllDatasetAudits();
```
rushtong, line 181: nice catch

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
rushtong: One thing that we're not quite addressing yet is how to format a Dataset + Study where there are overlapping properties. For example, if a Study has a PI name and a Dataset has a PI name, which one do we show? Right now, there are none in the study table but as soon as we get one, we'll need to sort out what to do since the UI is expecting a suite of dataset properties with specific types to exist. I see two areas where we'll need work - the standard dataset catalog page and the study display page.

src/main/java/org/broadinstitute/consent/http/db/StudyDAO.java
```diff
@@ -0,0 +1,89 @@
...
+            s.*,
+            sp.study_property_id AS sp_study_property_id,
+            sp.study_id AS sp_study_id,
+            sp.key AS sp_key,
+            sp.value AS sp_value,
+            sp.type AS sp_type,
+            d.dataset_id AS s_dataset_id,
+        """
+        + FileStorageObject.QUERY_FIELDS_WITH_FSO_PREFIX + " " +
+        """
```
rushtong, line 37: I'm surprised this hasn't triggered a semgrep error since it looks for this kind of pattern for sql injection problems.

rushtong (approved):
Overall, great work! One minor nit inline, but otherwise, 👍🏽

connorlbark:
Thought: Since jonathan mentioned that, when we transition to the new catalog, all datasets will have a study, should we make a ticket to transition to the Study model containing full dataset objects, and the dataset model containing just simple Study ids? Essentially the opposite of the ^ approach.

src/test/java/org/broadinstitute/consent/http/db/StudyDAOTest.java
```diff
@@ -0,0 +1,279 @@
...
+        String fileName = org.apache.commons.lang3.RandomStringUtils.randomAlphabetic(10);
```
fboulnois:
```suggestion
        String fileName = randomAlphabetic(10);
```

I think these can be pulled out since you import the `RandomStringUtils`.

src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
```diff
@@ -946,4 +981,44 @@ protected Consent insertConsent() {
...
+    private Study insertStudyWithProperties() {
+        User u = createUser();
+
+        String name = RandomStringUtils.randomAlphabetic(20);
+        String description = RandomStringUtils.randomAlphabetic(20);
+        List<String> dataTypes = List.of(
+                RandomStringUtils.randomAlphabetic(20),
+                RandomStringUtils.randomAlphabetic(20)
+        );
+        String piName = org.testcontainers.shaded.org.apache.commons.lang3.RandomStringUtils.randomAlphabetic(20);
```
fboulnois:
```suggestion
        String piName = randomAlphabetic(20);
```

Ditto here.

fboulnois (changes requested):
The SQL, logic, and tests lgtm, two small suggestions:

rushtong:
@connorlbark 
> Thought: Since jonathan mentioned that, when we transition to the new catalog, all datasets will have a study, should we make a ticket to transition to the Study model containing full dataset objects, and the dataset model containing just simple Study ids? Essentially the opposite of the ^ approach.

I would prefer to keep the current approach as implemented in this PR. I think this gives us more flexibility especially with other systems which are more dataset-centric rather than study-centric. Refactoring the model at this point seems premature given that we have yet to see real-world usage.

fboulnois (approved):

PR #1991, opened 2023-04-28
[DUOS-2460][risk=low] Refactor google token validation to be non-fatal

rushtong:
## Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DUOS-2460

This PR removes the failure case when we cannot validate a bearer token against google. This code path no longer necessary as our proxy does this for us. Additionally, it only supports google tokens and we need to be able to accept authenticated calls from B2C tokens which will fail when queried against Google's token endpoint.

Future tasks/PRs will address support for B2C tokens in our proxy and B2C authentication via the UI.

----

- Verify all tests go green


okotsopoulos (approved):
Thanks for working on this so quickly and the discussion this morning!

fboulnois (approved):
lgtm, the justifications make sense and the tests look good. 👍

PR #1994, opened 2023-05-01
[DUOS-2460][risk=no] Ensure Sam user existence for all API calls

rushtong:
## Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DUOS-2460

This fixes a bug where if an API call comes through and has never registered in Sam, Sam would, of course, give us a 404. This PR attempts to auto-register any API user in the 404 case. This error was clear in the case of an existing Google identity trying to access Consent but using a token for a corresponding Microsoft identity with the same email address. Sam would return a 404 and therefore Consent would have no way to know what the authenticated user's email address was. Previously, we'd look up the user in Google first, but in the Microsoft/Azure case, we didn't have that info this point in the auth code path.

### Side Note
I'm seeing a test failure from a test that we should not have. Disabling and filing a ticket to remove.
```
Error:  Failures: 
Error:    IndexOntologyServiceTest.testTerms:157 Could not load imported ontology: <http://purl.obolibrary.org/obo/duo/releases/2017-02-17/duo.owl> Cause: OWLOntologyCreationIOException: java.io.IOException: Server returned HTTP response code: 502 for URL: https://raw.githubusercontent.com/EBISPOT/DUO/v2017-02-17/duo.owl
```

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
+++ b/src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
@@ -70,7 +70,6 @@ public class OAuthAuthenticator implements Authenticator<String, AuthUser>, Cons
                     throw new ServerErrorException("User not able to be registered", 500);
                 }
             } catch (Exception exc) {
-                // Same
                 logException("User not able to be registered: '" + authUser.getEmail(), exc);
             }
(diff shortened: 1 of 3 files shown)


fboulnois (approved):
👍

kaitlynoflaherty (approved):
👍

PR #1996, opened 2023-05-02
[DUOS-1520][risk=no] Add PR/Closeout fields

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1520

This PR adds new fields to the DAR Data object to support Progress Reports and Closeouts.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/Presentation.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Presentation.java
@@ -1,0 +1,4 @@
+package org.broadinstitute.consent.http.models;
+
+public record Presentation(String title, String link, String date, String authors, String citation) {
+}
--- a/src/main/java/org/broadinstitute/consent/http/models/Presentation.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Presentation.java
@@ -1,4 +1,4 @@
 package org.broadinstitute.consent.http.models;
 
-public record Presentation(String title, String link, String date, String authors, String citation) {
+public record Presentation(String title, String link, String date, String authors, String datasetCitation, Boolean citation) {
 }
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/consent/http/models/Presentation.java
```diff
@@ -0,0 +1,4 @@
+package org.broadinstitute.consent.http.models;
+
+public record Presentation(String title, String link, String date, String authors, String citation) {
```
connorlbark: Should `authors` be an array? Should date be a `Date` object, or link be `URI`? Not sure if whatever parsing library we're using supports those objects

src/main/java/org/broadinstitute/consent/http/models/Publication.java
```diff
@@ -0,0 +1,4 @@
+package org.broadinstitute.consent.http.models;
+
+public record Publication(String title, String pubmedId, String date, String authors, String citation) {
```
connorlbark: Same thoughts here

connorlbark (approved):
Looks good, matches what is expected according to the ticket!

kaitlynoflaherty (approved):
Besides Connor's questions, looks good!

src/main/java/org/broadinstitute/consent/http/models/Presentation.java
rushtong: I'm hesitant to model these any more granularly than generic strings. The current mocks are super simple. For example, this is a presentation:
Screenshot 2023-05-02 at 12 15 48 PM

And this is a publication:
Screenshot 2023-05-02 at 12 16 33 PM

However, looking at this again, I think citation needs to be a boolean, not a string. I think the earlier versions were not booleans but looking at the mocks again, they are.

PR #1997, opened 2023-05-02
[DUOS-2414][risk=no] Use match v3 for all match results

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2414

### Summary

Use V3 match for all match results.

### Notes

I originally tried to propagate the `APPROVE`, `DENY`, and `ABSTAIN` case across the codebase, but this added much more code to the codebase and importantly required me to change the `matchentity` column in the database from a boolean to an enum. This had additional consequences including requiring us to migrate all the current matches which made the problem more challenging.

As a result I picked a simpler approach which instead adds a boolean `abstain` column in the database and use the `APPROVE`, `DENY`, `ABSTAIN` where it makes sense, without propagating it to the whole codebase.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/Match.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Match.java
@@ -2,11 +2,15 @@ package org.broadinstitute.consent.http.models;
 
 
 import org.broadinstitute.consent.http.enumeration.MatchAlgorithm;
+import org.broadinstitute.consent.http.models.matching.DataUseMatchResultType;
 
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 import java.util.Objects;
+
+import static org.broadinstitute.consent.http.models.matching.DataUseMatchResultType.Abstain;
+import static org.broadinstitute.consent.http.models.matching.DataUseMatchResultType.Approve;
 
 public class Match {
 
@@ -18,6 +22,8 @@ public class Match {
 
     private Boolean match;
 
+    private Boolean abstain;
+
     private Boolean failed;
 
     private Date createDate;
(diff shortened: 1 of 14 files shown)


kaitlynoflaherty (approved):
Looks good!

src/main/java/org/broadinstitute/consent/http/service/MatchService.java
```diff
@@ -142,17 +144,17 @@ public List<Match> createMatchesForConsent(String consentId) {
...
+                    logWarn("Error finding matches for consent: " + consentId);
+                    matches.add(matchFailure(consentId, dar.getReferenceId(), List.of()));
                 }
             });
         });
         return matches;
     }
 
-    private Match singleEntitiesMatchV2(Dataset dataset, DataAccessRequest dar) {
+    private Match singleEntitiesMatchV3(Dataset dataset, DataAccessRequest dar) {
```
connorlbark: nit: should we add a test for this function or to `matchSuccess` to ensure that the logic for abstain/approve/deny is mapping to the right booleans?

connorlbark (approved):
looks great! i could see it with a few more simple tests to round it out / become more confident it all works, but it looks there and the abstain column is a great approach for keeping the update simple to integrate and to understand.

src/main/java/org/broadinstitute/consent/http/service/MatchService.java
fboulnois: I added tests to have complete coverage of `singleEntitiesMatchV3`.

connorlbark (approved):

src/main/java/org/broadinstitute/consent/http/models/Match.java
```diff
@@ -135,4 +140,12 @@ public void addFailureReason(String reason) {
...
         }
     }
+
+    public static Match matchFailure(String consentId, String purposeId, List<String> failureReasons) {
+        return new Match(consentId, purposeId, false, false, true, MatchAlgorithm.V3, failureReasons);
+    }
+
+    public static Match matchSuccess(String consentId, String purposeId, DataUseMatchResultType match, List<String> failureReasons) {
+        return new Match(consentId, purposeId, Approve(match), Abstain(match), false, MatchAlgorithm.V3, failureReasons);
+    }
```
rushtong, line 150: In theory, failure reasons should always be an empty list in the success case.

rushtong (approved):
Looks good 👍🏽 - see optional comment inline.

src/main/java/org/broadinstitute/consent/http/models/Match.java
fboulnois: The failure reasons are extracted from the [response entity](https://github.com/DataBiosphere/consent/pull/1997/files#diff-b2fbd6a813de88461f1c1800da590a32cf7ac7312857be8a6b5566509240f799R174), so from what I can tell it is possible to have a MatchResult with [various associated reasons](https://github.com/DataBiosphere/consent-ontology/blob/develop/src/main/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatcherV3.java#L75), and these reasons will be sent as [failure reasons](https://github.com/DataBiosphere/consent-ontology/blob/develop/src/main/java/org/broadinstitute/dsde/consent/ontology/resources/MatchResource.java#L97) in the json.

So although the POST is successful and the `consent` Match does not have the boolean `failed` set to true, it still may have one or more "failure reasons".

PR #1998, opened 2023-05-03
[DUOS-2447][risk=no] Create Study on V3 Dataset Create Endpoint

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2447

----

- Verify all tests go green


kaitlynoflaherty (approved):
👍

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
```diff
@@ -177,6 +196,21 @@ private List<FileStorageObject> uploadFiles(Map<String, FormDataBodyPart> files,
 
     }
 
+    private List<FileStorageObject> uploadFilesForStudy(Map<String, FormDataBodyPart> files,
```
rushtong, line 218: ~Minor naming nit. This method is extracting a list of Study FSOs from the mass of supplied form data. Something like `extractStudyFiles` might be more descriptive.~

Edit: I misread this at first, it's doing two things, extraction and upload.

rushtong: I think these private methods could use some javadoc.

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
```diff
@@ -246,197 +279,222 @@ Optional<DatasetProperty> extract(DatasetRegistrationSchemaV1 registration, Cons
...
-            new DatasetPropertyExtractor(
-                    "PI Name", "piName", PropertyType.String,
-                    (registration, consentGroup) -> registration.getPiName()),
-            new DatasetPropertyExtractor(
-                    "Study Name", "studyName", PropertyType.String,
-                    (registration, consentGroup) -> registration.getStudyName()),
-            new DatasetPropertyExtractor(
-                    "Study Type", "studyType", PropertyType.String,
-                    (registration, consentGroup) -> {
+    public record StudyPropertyExtractor(
```
rushtong, line 304: Can we pull this out into its own class?

rushtong (approved):
Looks really good 👍🏽 Comments inline are optional.

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
connorlbark: I am honestly not super happy with this approach. I want to rethink it - it's overly complex just to save some lines of code. I feel like we should keep it as-is and reconsider later refactoring it. I'll put in a comment to do that as a follow up, but I feel like it's low-priority (personally).

fboulnois (approved):
lgtm

PR #2001, opened 2023-05-04
[DUOS-2468][risk=no] DAR creation fails if user has no library card

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2468

----

- Verify all tests go green


kaitlynoflaherty (approved):
Looks good!

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -252,6 +253,11 @@ public DataAccessRequest createDataAccessRequest(User user, DataAccessRequest da
         if (Objects.isNull(user) || Objects.isNull(dataAccessRequest) || Objects.isNull(dataAccessRequest.getReferenceId()) || Objects.isNull(dataAccessRequest.getData())) {
             throw new IllegalArgumentException("User and DataAccessRequest are required");
         }
+
+        if (Objects.isNull(user.getLibraryCards()) || user.getLibraryCards().isEmpty()) {
+            throw new IllegalArgumentException("User must have a library card.");
+        }
```
rushtong, line 258: While important to have here, I like to do validation at the resource level and fail fast when we know we can't complete the process. The fact that the request gets here at all indicates that we should have returned an error previously.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
```diff
@@ -89,6 +90,10 @@ public Response createDataAccessRequest(
       @Auth AuthUser authUser, @Context UriInfo info, String dar) {
     try {
       User user = findUserByEmail(authUser.getEmail());
+      if (Objects.isNull(user.getLibraryCards()) || user.getLibraryCards().isEmpty()) {
+        throw new IllegalArgumentException("User must have a library card to create a DAR.");
```
rushtong: I think a `NotAllowedException` might be more appropriate here.
Edit: I can be persuaded otherwise too.

rushtong (approved):
Looks good, one minor and optional comment inline. Illegal argument exceptions are translated to Bad Request, which also works in this case.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
connorlbark: I think NotAllowedException is when the user does the wrong method - e.g. a GET on an endpoint that just has PUT. E.g., this example I found online:
``` java
public Response handle() {
  if (HttpMethod.GET.equals(request.getMethod())) {
       // logic
   }
  throw new NotAllowedException("GET"); // only GET is allowed.
}
```

It should translate to a 405 (according to the docs)

PR #2008, opened 2023-05-09
[DUOS-1527][risk=no] POST endpoint for progress report

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-1527

### Summary

POST endpoint for progress report

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
@@ -431,7 +431,12 @@ public class DataAccessRequestResourceVersion2 extends Resource {
 
   private DataAccessRequest populateDarFromJsonString(User user, String json) {
     DataAccessRequest newDar = new DataAccessRequest();
-    DataAccessRequestData data = DataAccessRequestData.fromString(json);
+    DataAccessRequestData data;
+    try {
+      data = DataAccessRequestData.fromString(json);
+    } catch (Exception e) {
+      throw new BadRequestException("Unable to parse DAR from JSON string");
+    }
     if (Objects.isNull(data)) {
       data = new DataAccessRequestData();
     }
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
@@ -316,13 +316,18 @@ public class DataAccessRequestResourceVersion2 extends Resource {
 
     for (Integer datasetId : childDar.getDatasetIds()) {
       Dataset dataset = datasetService.findDatasetById(datasetId);
+      if (dataset == null) {
+        throw new NotFoundException("Dataset " + datasetId + " not found");
+      }
       DataUse dataUse = dataset.getDataUse();
-      if (dataUse == null) {
-        throw new BadRequestException("Dataset " + datasetId + " has no data use");
+      if (dataUse == null || dataUse.getCollaboratorRequired() == null
+          || dataUse.getEthicsApprovalRequired() == null) {
+        throw new BadRequestException("Dataset " + datasetId + " is missing data use(s)");
       }
       if (dataUse.getCollaboratorRequired()) {
         String parentCollabLocation = parentDar.getData().getCollaborationLetterLocation();
-        if (collabFileDetails.getSize() <= 0 && Strings.isNullOrEmpty(parentCollabLocation)) {
+        if ((collabFileDetails == null || collabFileDetails.getSize() <= 0)
+            && Strings.isNullOrEmpty(parentCollabLocation)) {
           throw new BadRequestException("Collaboration document is required");
         }
         try {
@@ -334,7 +339,8 @@ public class DataAccessRequestResourceVersion2 extends Resource {
       }
       if (dataUse.getEthicsApprovalRequired()) {
         String parentEthicsLocation = parentDar.getData().getIrbDocumentLocation();
-        if (ethicsFileDetails.getSize() <= 0 && Strings.isNullOrEmpty(parentEthicsLocation)) {
+        if ((ethicsFileDetails == null || ethicsFileDetails.getSize() <= 0)
+            && Strings.isNullOrEmpty(parentEthicsLocation)) {
           throw new BadRequestException("Ethics approval document is required");
         }
         try {
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
@@ -317,6 +317,9 @@ public class DataAccessRequestResourceVersion2 extends Resource {
     for (Integer datasetId : childDar.getDatasetIds()) {
       Dataset dataset = datasetService.findDatasetById(datasetId);
       DataUse dataUse = dataset.getDataUse();
+      if (dataUse == null) {
+        throw new BadRequestException("Dataset " + datasetId + " has no data use");
+      }
       if (dataUse.getCollaboratorRequired()) {
         String parentCollabLocation = parentDar.getData().getCollaborationLetterLocation();
         if (collabFileDetails.getSize() <= 0 && Strings.isNullOrEmpty(parentCollabLocation)) {
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
@@ -1,8 +1,10 @@
 package org.broadinstitute.consent.http.resources;
 
 import com.google.cloud.storage.BlobId;
+import com.google.common.base.Strings;
 import com.google.inject.Inject;
 import io.dropwizard.auth.Auth;
+import jakarta.ws.rs.BadRequestException;
 import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.DELETE;
 import jakarta.ws.rs.ForbiddenException;
@@ -35,9 +37,12 @@ import org.broadinstitute.consent.http.enumeration.UserRoles;
 import org.broadinstitute.consent.http.models.AuthUser;
 import org.broadinstitute.consent.http.models.DataAccessRequest;
 import org.broadinstitute.consent.http.models.DataAccessRequestData;
+import org.broadinstitute.consent.http.models.DataUse;
+import org.broadinstitute.consent.http.models.Dataset;
 import org.broadinstitute.consent.http.models.Error;
 import org.broadinstitute.consent.http.models.User;
 import org.broadinstitute.consent.http.service.DataAccessRequestService;
+import org.broadinstitute.consent.http.service.DatasetService;
 import org.broadinstitute.consent.http.service.EmailService;
 import org.broadinstitute.consent.http.service.MatchService;
 import org.broadinstitute.consent.http.service.UserService;
@@ -52,6 +57,7 @@ public class DataAccessRequestResourceVersion2 extends Resource {
   private final GCSService gcsService;
   private final MatchService matchService;
   private final UserService userService;
+  private final DatasetService datasetService;
 
   @Inject
   public DataAccessRequestResourceVersion2(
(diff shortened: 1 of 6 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -5,13 +5,46 @@
 import com.google.inject.Inject;
```
fboulnois: This file somehow escaped the reformat and I'm not sure how. I removed the `extractFilesFromMultiPart` method and moved it to a static method in `Resource.java`.

src/main/resources/assets/api-docs.yaml
```diff
@@ -488,6 +488,36 @@ paths:
...
+          multipart/form-data:
+            schema:
+              type: object
+              properties:
+                collaboratorRequiredFile:
+                  type: string
+                  format: binary
+                ethicsApprovalRequiredFile:
+                  type: string
+                  format: binary
```
rushtong: Minor request - we're trying to extract these blocks to `src/main/resources/assets/paths` whenever possible

src/main/resources/assets/api-docs.yaml
```diff
@@ -488,6 +488,36 @@ paths:
             application/json:
               schema:
                 $ref: './schemas/DataAccessRequest.yaml'
+  /api/dar/progress_report/{parentReferenceId}:
+    post:
```
rushtong: We need a tags element somewhere so this shows up in the swagger UI
```
  tags:
    - Data Access Request
```

src/main/resources/assets/api-docs.yaml
```diff
@@ -488,6 +488,36 @@ paths:
...
+          description: The referenceId of the Data Access Request
+          required: true
+          schema:
+            type: string
+        - name: dar
+          in: body
+          description: The fields that represent a Data Access Request, json format.
+          required: true
+          schema:
+            $ref: './schemas/DataAccessRequest.yaml'
```
rushtong: This is in an older version of swagger syntax, the current version needs this to be in the `requestBody` section.

src/main/resources/assets/api-docs.yaml
rushtong: At the end of this, we also need a `responses` section.

src/main/resources/assets/api-docs.yaml
```diff
@@ -488,6 +488,36 @@ paths:
             application/json:
               schema:
                 $ref: './schemas/DataAccessRequest.yaml'
+  /api/dar/progress_report/{parentReferenceId}:
```
rushtong: The API location as defined in `DataAccessRequestResourceVersion2` is actually `/api/dar/v2/progress_report/{parentReferenceId}`

rushtong (changes requested):
Have not quite gotten to testing this - there are a number of errors in the api docs to sort through. Will tackle this again tomorrow.

src/main/resources/assets/paths/progressReport.yaml
```diff
@@ -0,0 +1,35 @@
...
+  parameters:
+    - name: parentReferenceId
+      in: path
+      description: The referenceId of the Data Access Request
+      required: true
+      schema:
+        type: string
+  requestBody:
+    content:
+      required: true
```
rushtong: This line needs to move up to the same indentation as `content:` As it is now, this renders as:
Screenshot 2023-06-01 at 2 48 30 PM

src/main/resources/assets/paths/progressReport.yaml
```diff
@@ -0,0 +1,35 @@
...
+      application/json:
+        schema:
+          $ref: '../schemas/DataAccessRequest.yaml'
```
rushtong: Posting this via swagger, I get
```
{
  "code": 415,
  "message": "HTTP 415 Unsupported Media Type"
}
```
I believe `content` needs to be under the multipart/form-data section (https://swagger.io/docs/specification/describing-request-body/multipart-requests/)

rushtong (commented):
Still not able to test this yet - the swagger doc needs a couple of tweaks to get it working.

src/main/resources/assets/paths/progressReport.yaml
fboulnois: Ok, I was able to post content by moving everything under `multipart/form-data`. I still can't get it to recognize the schema of the dar though.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
```diff
@@ -287,6 +295,57 @@ public Response uploadIrbDocument(
...
+      @FormDataParam("ethicsApprovalRequiredFile") FormDataContentDisposition ethicsFileDetails) {
+    User user = userService.findUserByEmail(authUser.getEmail());
+    DataAccessRequest parentDar = dataAccessRequestService.findByReferenceId(parentReferenceId);
+    checkAuthorizedUpdateUser(user, parentDar);
+    DataAccessRequest payload = populateDarFromJsonString(user, dar);
+    DataAccessRequest childDar = dataAccessRequestService.createDataAccessRequest(user, payload);
+
+    for (Integer datasetId : childDar.getDatasetIds()) {
+      Dataset dataset = datasetService.findDatasetById(datasetId);
+      DataUse dataUse = dataset.getDataUse();
```
rushtong, line 322: Connor and I found that `getDataUse()` can return null for about 25 or so datasets on dev

src/main/resources/assets/paths/progressReport.yaml
```diff
@@ -0,0 +1,35 @@
...
+  requestBody:
+    required: true
+    content:
+      multipart/form-data:
+        schema:
+          type: object
+          properties:
+            dar:
+              schema:
+                $ref: '../schemas/DataAccessRequest.yaml'
```
rushtong: The more I read, the more I think this isn't allowed in the spec. I am fine with adding documentation to describe this case.

src/test/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2Test.java
```diff
@@ -342,6 +352,46 @@ public void testUploadIrbDocumentWithPreviousIrbDocument() throws Exception {
...
+    when(dataset.getDataUse()).thenReturn(dataUseCollabAndEthics);
+    when(gcsService.storeDocument(any(), any(), any())).thenReturn(BlobId.of("bucket", "name"));
+    when(gcsService.deleteDocument(any())).thenReturn(true);
+    when(dataAccessRequestService.updateByReferenceId(any(), any())).thenReturn(childDar);
+    initResource();
+
+    Response response = resource.postProgressReport(authUser, "", "", collabFile.getLeft(),
+        collabFile.getRight(), ethicsFile.getLeft(), ethicsFile.getRight());
+    assertEquals(200, response.getStatus());
+  }
```
rushtong, line 393: Some additional test case ideas
* Null dataset.getDataUse()
* Null dataUse.getCollaboratorRequired()
* Null dataUse.getEthicsApprovalRequired()
* Non-create user trying to create a PR
* Parent DAR not found
* Requested datasets not found

rushtong (commented):
I think this is generally the right approach. There are bound to be some edge cases we've missed, but I'm sure we can manage that when we dive a little deeper.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
fboulnois: What should the behavior be if `getDataUse` is null?

rushtong: I think we need to throw a server error so we can follow up. This isn't a user problem, it's an us problem in that we have a bad dataset that needs work.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
```diff
@@ -316,13 +316,18 @@ public Response postProgressReport(
 
     for (Integer datasetId : childDar.getDatasetIds()) {
       Dataset dataset = datasetService.findDatasetById(datasetId);
+      if (dataset == null) {
+        throw new NotFoundException("Dataset " + datasetId + " not found");
```
fboulnois, line 320: I am not totally sure under which scenarios it is actually possible to hit this exception. Nevertheless, it should avoid at least one class of issues.

rushtong: Another pattern we often use is to throw a NFE in the service class. It doesn't look like we do that here though.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
```diff
@@ -287,6 +295,66 @@ public Response uploadIrbDocument(
...
+      @FormDataParam("collaboratorRequiredFile") FormDataContentDisposition collabFileDetails,
+      @FormDataParam("ethicsApprovalRequiredFile") InputStream ethicsInputStream,
+      @FormDataParam("ethicsApprovalRequiredFile") FormDataContentDisposition ethicsFileDetails) {
+    User user = userService.findUserByEmail(authUser.getEmail());
+    DataAccessRequest parentDar = dataAccessRequestService.findByReferenceId(parentReferenceId);
+    checkAuthorizedUpdateUser(user, parentDar);
+    DataAccessRequest payload = populateDarFromJsonString(user, dar);
+    DataAccessRequest childDar = dataAccessRequestService.createDataAccessRequest(user, payload);
+
+    for (Integer datasetId : childDar.getDatasetIds()) {
```
rushtong, line 317: There's an issue here if the user passes in a json body with no dataset ids. I think the bug is in `populateDarFromJsonString`. Passing in `{"projectTitle": "PR Testing"}` yields this error:
```
{
  "code": 404,
  "message": "Dataset 0 not found"
}
```

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
```diff
@@ -287,6 +295,66 @@ public Response uploadIrbDocument(
...
+      @PathParam("parentReferenceId") String parentReferenceId,
+      @FormDataParam("dar") String dar,
+      @FormDataParam("collaboratorRequiredFile") InputStream collabInputStream,
+      @FormDataParam("collaboratorRequiredFile") FormDataContentDisposition collabFileDetails,
+      @FormDataParam("ethicsApprovalRequiredFile") InputStream ethicsInputStream,
+      @FormDataParam("ethicsApprovalRequiredFile") FormDataContentDisposition ethicsFileDetails) {
+    User user = userService.findUserByEmail(authUser.getEmail());
+    DataAccessRequest parentDar = dataAccessRequestService.findByReferenceId(parentReferenceId);
+    checkAuthorizedUpdateUser(user, parentDar);
+    DataAccessRequest payload = populateDarFromJsonString(user, dar);
```
rushtong, line 314: Passing in malformed json throws an unhelpful error, maybe we can wrap some of this to return a better (400) error:
```
{
  "code": 500,
  "message": "There was an error processing your request. It has been logged (ID 2c75f2b5d758abe2)."
}
```

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
rushtong: There is something confusing happening with dataset ids. Even if I pass in valid dataset ids, the populated value seems to contain incorrect values. Passing in
```
{
  "projectTitle": "PR Testing",
  "datasetIds": [2]
}
```
shows this in the debugger:
Screenshot 2023-06-05 at 10 24 23 AM

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
fboulnois: Yes, I noticed that in the other services classes. I'll update the code to do that.

fboulnois: After taking a deeper look at this it would be non-trivial work that would have side effects on other code we've already written. I think I'll just create a tech debt ticket instead.

fboulnois: I created https://broadworkbench.atlassian.net/browse/DUOS-2540 as a tech debt ticket.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceVersion2.java
fboulnois: Bug fixed in latest commit.

rushtong (approved):
Thank you for the changes! 👍🏽

rushtong (approved):
👍🏽

kaitlynoflaherty (approved):
Everything looks great!

PR #2015, opened 2023-05-11
[DUOS-750][risk=no] Fix POST /api/dacuser Swagger Doc

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-750

### Summary
This PR fixes the swagger endpoint for `POST /api/dacuser` which is an admin endpoint used to manually create users. The swagger definition was syntactically incorrect.

----

- Verify all tests go green

### DIFF
--- a/src/main/resources/assets/api-docs.yaml
+++ b/src/main/resources/assets/api-docs.yaml
@@ -302,26 +302,7 @@ paths:
         500:
           description: Internal Server Error
   /api/dacuser:
-    post:
-      summary: Create User
-      description: Creates a User
-      parameters:
-        - name: dac
-          in: body
-          required: true
-          schema:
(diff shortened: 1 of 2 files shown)


fboulnois (approved):
Good catch. I wonder if we should have better tests for the openapi changes we're making.

rushtong:
> Good catch. I wonder if we should have better tests for the openapi changes we're making.

There are quite a few of these errors, this is just one. I'm sure we could add some kind of swagger validation hooks ... will look into that.

kaitlynoflaherty (approved):
👍

kaitlynoflaherty (approved):
👍

kaitlynoflaherty (approved):
👍

PR #2018, opened 2023-05-11
[DUOS-2186]V3 Dataset PUT API

kaitlynoflaherty:
### Addresses
https://broadworkbench.atlassian.net/jira/software/c/projects/DUOS/boards/123?modal=detail&selectedIssue=DUOS-2186

### Summary

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -221,6 +221,57 @@ private Map<String, FormDataBodyPart> extractFilesFromMultiPart(FormDataMultiPar
...
+    try {
+      Set<ValidationMessage> errors = jsonSchemaUtil.validateSchema_v1(json);
+      if (!errors.isEmpty()) {
+        throw new BadRequestException(
+            "Invalid schema:\n"
+                + String.join("\n", errors.stream().map(ValidationMessage::getMessage).toList()));
+      }
+
+      // get registration
+      DatasetRegistrationSchemaV1 registration = jsonSchemaUtil.deserializeDatasetRegistration(json);
```
connorlbark: Hmm, so I think we shouldn't use the whole validation schema, because there are a lot of unused properties and  it allows any number of datasets even though we're just updating one. However, it might make sense to make a new schema with just the consent group, instead of the whole data submission scheme. That shouldn't be too much work, I think, and we could import that new schema in the data submission form, so consent groups are only defined in one location. Thoughts?

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
```diff
@@ -62,6 +62,18 @@ Integer insertDataset(
...
+  @SqlUpdate("""
+            UPDATE dataset
+            SET name = :name,
+                update_date = :updateDate,
+                update_user_id = :updateUserId,
+                needs_approval = :needsApproval,
+                dac_id = :dacId
+            WHERE dataset_id = :datasetId
+            """)
+  Integer updateDatasetNew(@Bind("datasetName") String name, @Bind("updateDate") Timestamp updateDate, @Bind("updateUserId") Integer updateUserId,
```
rushtong: Nit, naming/duplication. Is there a reason we can't refactor the existing update?

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
rushtong: I think using any schema here might be overkill. Maintaining two schemas for essentially same kind of thing is a bigger commitment to this approach than I'm comfortable with at this time. I wonder if we could instead manually parse the json into a new dataset intake record class that we can adjust specifically for this endpoint.

src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
```diff
@@ -179,6 +188,56 @@ private void executeInsertFiles(Handle handle, List<FileStorageObject> files, In
     }
   }
 
+  public Integer updateDataset(DatasetUpdate updates) throws SQLException {
+    AtomicReference<Integer> updatedDataset = new AtomicReference<>();
```
rushtong: Nit: do we need this at all? We already know what the `datasetId` is from the `updates`. The method could be a void but wrapped with some error handling in case of a jdbi error.

src/main/resources/assets/paths/datasetV3.yaml
```diff
@@ -41,3 +41,41 @@ post:
       description: Dataset Name given is already in use by another dataset
     500:
       description: Internal Error (something went wrong processing a valid input)
+
+put:
+  summary: Updates a dataset from a valid dataset registration JsonSchema object
+  description: Updates a dataset from a valid dataset registration JsonSchema object
+  requestBody:
+    description: to do
```
rushtong: nit: flesh this out

src/main/resources/assets/paths/datasetV3.yaml
```diff
@@ -41,3 +41,41 @@ post:
...
+              type: object
+              $ref: '../schemas/DatasetRegistrationSchemaV1.yaml'
+            alternativeDataSharingPlan:
+              description: Optional alternative data sharing plan
+              type: string
+              format: binary
+            consentGroups[0].nihInstitutionalCertificationFile:
+              description: Certification file for first consent group (others omitted because swagger does not allow variable number of file uploads)
+              type: string
+              format: binary
```
rushtong: This is why I find the schema approach over-complicated. We're forced to replicate the multiple consent group approach when we really don't have to.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
connorlbark: I don't know if I agree with this. I don't think it would be more than what we're maintaining right now, because I think it could work like so:


DataRegistrationV1 -
```
{
   ...
   consentGroups: { type: array, items: { ref: /* consent group file */ }}
}
```
ConsentGroup - exactly like it is now.
```
{
   type: object
   properties: { ... }
}
```

This also ensures that the dataset updates don't put the dataset into an invalid state - what's to stop a user from simply updating their dataset from the backend and changing, e.g., the data submitter id, deleting key properties, etc.. 

To me, I feel like having a schema is important to making sure that the endpoint is secured and the user doesn't mess anything up. I find that a lot of our endpoints don't put a lot of bounds around user input, and schemas are super easy for doing that. They also sync really well with form validation in the frontend, which would be useful for the frontend piece of this ticket. I personally see a lot of value in them, and would argue we should move towards using them more, especially on important endpoints that are linked to a form in the app like this one.

rushtong: Designing a dataset schema is well beyond the scope of the assigned task/ticket. I'm not specifically opposed to the idea in general, but for this PR, we haven't done the homework to design and publish such a schema.

connorlbark: That's totally fair, I just get some code smell from sort of just accepting whatever the user puts in as a dataset, and it would make me a lot more comfortable if there were well understood/formalized bounds around user input. And since she's already using the consentGroup from the data submission, just making that its own schema seems like a small code lift, since it has a 1:1 mapping to dataset properties and fields. But i totally understand if you prefer the old approach, I'm just a big advocate for schemas

rushtong: If we used a record class as the intake model, we would be controlling the format of the data which has the same effect as a schema.

connorlbark: That's fair enough, I thought you were suggesting the same approach as the old endpoint. I am not opposed to the dataset intake class, so we can go for that, but I do feel like having the same schema as the create endpoint would be nice for consistency and frontend validation, but I won't hold the ticket up.

rushtong:
It looks like `DatasetService` isn't up to date with develop, can you merge the latest in? There's a failing import for me locally that should be using `com.google.inject.Inject` as opposed to `javax.inject.Inject`

src/main/resources/assets/paths/datasetV3.yaml
```diff
@@ -41,3 +41,45 @@ post:
...
+              schema:
+                type: array
+                items:
+                  $ref: '../schemas/Dataset.yaml'
+        400:
+          description: Bad Request (invalid input)
+        404:
+          description: Dataset Id not found
+        500:
+          description: Internal Error (something went wrong processing a valid input)
```
rushtong: These are indented too many places. Pull them both back so they are top level entries under `put:`

rushtong: I'm a little confused that I can't get this to show up in swagger, but there is a second issue here. The endpoint should be a PUT in `/api/dataset/v3/{datasetId}` ... however, this swagger fragment is defined under `/api/dataset/v3`. There needs to be a new path defined in api-docs.yaml for this new path. Separately, the `put` fragment defined here needs to be in its own file.

src/main/resources/assets/paths/datasetV3.yaml
```diff
@@ -41,3 +41,45 @@ post:
...
+        - Dataset
+      responses:
+        201:
+          description: Successfully updated Datasets
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '../schemas/Dataset.yaml'
```
rushtong: Technically, this is not the thing you are updating. We're parsing the input into a single `DatasetUpdate` object, so that's what we have to define as the input. Remove the array piece, and then create a new schema object to represent what you're actually expecting.

src/main/resources/assets/paths/datasetV3ById.yaml
```diff
@@ -0,0 +1,46 @@
...
+              format: binary
+  tags:
+    - Dataset
+  responses:
+    201:
+      description: Successfully updated Datasets
+      content:
+        application/json:
+          schema:
+            $ref: '../schemas/Dataset.yaml'
```
rushtong, line 72: This still needs to be more clearly defined not as a dataset, but as a new object that represents the updatable fields on the dataset.

src/main/resources/assets/paths/datasetV3ById.yaml
```diff
@@ -0,0 +1,46 @@
...
+              format: binary
+  tags:
+    - Dataset
+  responses:
+    201:
+      description: Successfully updated Datasets
+      content:
+        application/json:
+          schema:
+            $ref: '../schemas/DatasetUpdate.yaml'
```
rushtong: We're not returning a dataset update object, we're returning a full dataset object. However, we do need to somehow express that the values in `DatasetUpdate.yaml` are what to use for input. Since we can't use `$ref` where we want to, we should move all of the content in `DatasetUpdate.yaml` into the description and then reference `Dataset.yaml` here on line 40.

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
```diff
@@ -62,6 +62,25 @@ Integer insertDataset(
       @Bind("dataUse") String dataUse,
       @Bind("dacId") Integer dacId);
 
+  @SqlUpdate("""
+            UPDATE dataset
```
rushtong, line 66: nit, indentation

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
```diff
@@ -122,6 +122,45 @@ private DatasetServiceDAO.StudyInsert createStudyInsert(DatasetRegistrationSchem
...
+
+    } catch (IOException e) {
+      // uploading files to GCS failed. rollback files...
+      uploadedFileCache.values().forEach((id) -> gcsService.deleteDocument(id.getName()));
+      throw e;
+    }
+
+    // Update or create the objects in the database
+    datasetServiceDAO.updateDataset(datasetUpdates);
+    return datasetDAO.findDatasetById(datasetId);
```
rushtong: This could all be inside the try so that you don't have to declare, then populate, `datasetUpdates`

rushtong (commented):
Outside of the documentation, I think this works as expected!

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -222,6 +223,45 @@ private Map<String, FormDataBodyPart> extractFilesFromMultiPart(FormDataMultiPar
...
+  @RolesAllowed({ADMIN, CHAIRPERSON})
+  public Response updateByDatasetUpdate(
+      @Auth AuthUser authUser,
+      @PathParam("datasetId") Integer datasetId,
+      FormDataMultiPart multipart,
+      @FormDataParam("dataset") String json) {
+
+    try {
+
+      DatasetUpdate update = new Gson().fromJson(json, DatasetUpdate.class);
```
fboulnois, line 242: I think you may need to try-catch in case the JSON in invalid, too.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -1058,6 +1061,125 @@ public void testCreateDatasetRegistration_invalidFileName() {
...
+    when(formDataMultiPart.getFields()).thenReturn(Map.of("file", List.of(formDataBodyPart)));
+    initResource();
+
+    Response response = resource.updateByDatasetUpdate(authUser, 1, formDataMultiPart, json);
+    assertEquals(200, response.getStatus());
+    assertEquals(Optional.of(preexistingDataset).get(), response.getEntity());
+  }
+
+  @Test
+  public void testUpdateDatasetWithNoJson() {
```
fboulnois, line 1095: I would add a test for invalid JSON, too.

fboulnois (commented):
One suggestion around invalid JSON:

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
kaitlynoflaherty: I created a test for this case using the current try-catch, and the response was a 400 which I would expect. Do you think it's redundant to add another try-catch? or am I looking for a different response?

fboulnois: Oh yeah I missed that, that works.

fboulnois (approved):

src/main/resources/assets/paths/datasetV3ById.yaml
```diff
@@ -0,0 +1,86 @@
...
+            alternativeDataSharingPlan:
+              description: Optional alternative data sharing plan
+              type: string
+              format: binary
+              $ref: './FileStorageObject.yaml'
+            consentGroups[0].nihInstitutionalCertificationFile:
+              description: Certification file
+              type: string
+              format: binary
+              $ref: './FileStorageObject.yaml'
```
rushtong: This is what I see when running swagger locally:
```
Resolver error at paths./api/dataset/v3/{datasetId}.put.requestBody.content.multipart/form-data.schema.properties.alternativeDataSharingPlan.$ref
Could not resolve reference: Not Found
Resolver error at paths./api/dataset/v3/{datasetId}.put.requestBody.content.multipart/form-data.schema.properties.consentGroups[0].nihInstitutionalCertificationFile.$ref
Could not resolve reference: Not Found
Resolver error at paths./api/dataset/v3/{datasetId}.put.requestBody.content.multipart/form-data.schema.properties.dataset.description.properties.datasetProperties.items.$ref
Could not resolve reference: Not Found
```
I think the syntax might be `$ref: '../schemas/FileStorageObject.yaml'` but haven't tested.

rushtong: There are a number of validation errors here.
* I don't think you can have a $ref for the two files. 
* `requestBody` can't have a description
* `properties.dataset.description` is empty
* Right below that, `type` is indented one place too far which impacts everything below it.

rushtong: This `requestBody` seems to work for me:
```
  requestBody:
    content:
      multipart/form-data:
        schema:
          type: object
          properties:
            dataset:
              description: A list of updatable dataset properties
              type: object
              properties:
                name:
                  type: string
                  description: The dataset name
                datasetName:
                  type: string
                  deprecated: true
                  description: Legacy value for backwards compatibility, same as the dataset name
                dataSetId:
                  type: integer
                  format: int32
                  description: The dataset id, integer format
                dacId:
                  type: string
                  description: The id of the DAC this dataset belongs to, if defined.
                datasetProperties:
                  type: array
                  items:
                    $ref: '../schemas/DatasetProperty.yaml'
                  example:
                    - propertyName: Dataset Name
                      propertyValue: Dataset Name
                    - propertyName: Data Type
                      propertyValue: Whole Genome
                    - propertyName: Species
                      propertyValue: Human
                    - propertyName: Phenotype/Indication
                      propertyValue: Cardiovascular System Disease
                    - propertyName: "# of participants"
                      propertyValue: 20
                    - propertyName: Description
                      propertyValue: Description of Dataset
                    - propertyName: dbGAP
                      propertyValue: http://...
                    - propertyName: Data Depositor
                      propertyValue: NHLBI
                    - propertyName: Principal Investigator(PI)
                      propertyValue: NHLBI
                    - propertyName: Sample Collection ID
                      propertyValue: SC-000
            alternativeDataSharingPlan:
              description: Optional alternative data sharing plan
              type: string
              format: binary
            consentGroups[0].nihInstitutionalCertificationFile:
              description: Certification file
              type: string
              format: binary
```

kaitlynoflaherty: Thank you!

src/main/resources/assets/paths/datasetV3ById.yaml
```diff
@@ -0,0 +1,82 @@
...
+                  type: string
+                  description: The dataset name
+                datasetName:
+                  type: string
+                  deprecated: true
+                  description: Legacy value for backwards compatibility, same as the dataset name
+                dataSetId:
+                  type: integer
+                  format: int32
+                  description: The dataset id, integer format
```
rushtong: Sorry I missed this before, but this isn't an updatable field in `DatasetUpdate`. We get the ID from the path parameter.

rushtong (commented):
One more fix and then I think this is complete!

rushtong (approved):
👍🏽

PR #2023, opened 2023-05-16
[DUOS-2501][risk=no] Remove unused paginated query endpoints

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2501

### Summary
This PR removes unused code that we had intended to use for server-side paginated DAR Collection queries. We have since moved away from that approach towards a summary-based approach as opposed to a full DAR Collection object based approach. These endpoints were never used in production and can be removed.

Removes:

- `GET /api/collections/role/{roleName}/paginated`
- `GET /api/collections/role/{roleName}/query`
- `GET /api/collections`
- `GET /api/collections/role/{roleName}`

----

- Verify all tests go green


kaitlynoflaherty (approved):
👍

connorlbark (approved):
Looks good - I have been curious why we didn't go this approach, and instead went for UI filtering? It feels like DarCollectionSummaries are a less scalable solution.

rushtong:
> Looks good - I have been curious why we didn't go this approach, and instead went for UI filtering? It feels like DarCollectionSummaries are a less scalable solution.

We discarded this approach for a couple reasons. One was the huge size of the dar collection payload. By sending a DAR Collection that may have had 600 datasets (we had examples like that) in it brought the UI to a halt. The second reason was the sheer volume of calculations necessary to perform in the UI on collection state. Even with a small number of datasets/collections, the UI would still struggle to calculate the variety of states we show in the summary. Summaries are much faster to calculate on the back end and we can adjust the logic more easily there to keep it centralized. For example, submission state (canceled, approved, denied, etc.) would have needed the same logic in both Consent and DUOS-UI to be consistent. Moving it all to the back end made it much easier to work with.

That being said, we may still need to move to a server-side paginated approach, but it won't be on the full DAR Collection object, it will instead be on the summary. We can implement that when it we need to scale this better.

PR #2024, opened 2023-05-16
[DUOS-2465][risk=no] POST endpoint for search index

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2465

### Summary

POST endpoint for search index

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
@@ -43,6 +43,7 @@ import org.broadinstitute.consent.http.enumeration.UserRoles;
 import org.broadinstitute.consent.http.models.AuthUser;
 import org.broadinstitute.consent.http.models.DataUse;
 import org.broadinstitute.consent.http.models.Dataset;
+import org.broadinstitute.consent.http.models.DatasetSearchTerm;
 import org.broadinstitute.consent.http.models.Dictionary;
 import org.broadinstitute.consent.http.models.User;
 import org.broadinstitute.consent.http.models.UserRole;
@@ -530,6 +531,20 @@ public class DatasetResource extends Resource {
       User user = userService.findUserByEmail(authUser.getEmail());
       List<Dataset> datasets = datasetService.searchDatasets(query, user);
       return Response.ok().entity(unmarshal(datasets)).build();
+    } catch (Exception e) {
+      return createExceptionResponse(e);
+    }
+  }
+
+  @POST
+  @Path("/search/index")
+  @Consumes("application/json")
+  @Produces("application/json")
+  public Response searchDatasetIndex(@Auth AuthUser authUser, String query) {
+    try {
+      User user = userService.findUserByEmail(authUser.getEmail());
+      DatasetSearchTerm datasetSearchTerm = new Gson().fromJson(query, DatasetSearchTerm.class);
+      return Response.ok().entity(unmarshal(List.of(datasetSearchTerm))).build();
     } catch (Exception e) {
       return createExceptionResponse(e);
     }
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -25,538 +59,574 @@
...
+  @Produces("application/json")
+  public Response searchDatasetIndex(@Auth AuthUser authUser, String query) {
+    try {
+      User user = userService.findUserByEmail(authUser.getEmail());
+      // Use the searchDataset method as a dummy method to validate the query
+      List<Dataset> datasets = datasetService.searchDatasets(query, user);
+      return Response.ok().entity(unmarshal(datasets)).build();
+    } catch (Exception e) {
+      return createExceptionResponse(e);
     }
```
fboulnois: Here's what was added.

rushtong: For this ticket, I think we need a little more to facilitate UI development. The return object should be a list of terms as defined in https://broadworkbench.atlassian.net/browse/DUOS-2469 ... it doesn't have to be exact, but should be something that has that structure so the UI devs can manipulate those terms in components.

Also, the input object should also be a json structure that looks like a term as defined in DUOS-2469. That way the UI can make real calls with a term query object and theoretically get back a filtered result. 

Lastly, this method should provide a minimal amount of filtering on the mock terms. If we had, say, 5 terms with some pre-defined filter fields, that should be enough for UI work.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -535,6 +535,21 @@ public Response searchDatasets(@Auth AuthUser authUser, @QueryParam("query") Str
...
+  @POST
+  @Path("/search/index")
+  @Consumes("application/json")
+  @Produces("application/json")
+  public Response searchDatasetIndex(@Auth AuthUser authUser, String query) {
+    try {
+      User user = userService.findUserByEmail(authUser.getEmail());
+      // Use the searchDataset method as a dummy method to validate the query
+      List<Dataset> datasets = datasetService.searchDatasets(query, user);
+      return Response.ok().entity(unmarshal(datasets)).build();
```
rushtong: I think maybe there's a misunderstanding of the required model. We're not returning real datasets here. We're returning a mock of what the elastic search query will return which is not a dataset, but instead is a term object that looks a lot like a dataset, but is different (see [DUOS-2469](https://broadworkbench.atlassian.net/browse/DUOS-2469) for more info). For this PR, it could look like this:
```
{
    datasetId: integer
    datasetIdentifier: string
    description: string
    studyName: string
    studyId: integer
    participantCount: integer
    phenotype: string
    species: string
    piName: string
    dataSubmitter: string
    dataCustodian: string
    dataUse: [list of strings (data use abbreviations)]
    dataTypes: [list of strings]
    dataLocation: string
    dacName: string
    openAccess: boolean
    approvedUsers: [list of strings (usernames? userids?)]
}
```

kaitlynoflaherty (approved):
Besides the semgrep failure, looks good

rushtong:
The semgrep is a false positive. We should disable that rule - `String description = ""` is not the cryptographic key [the rule describes](https://semgrep.dev/r?q=gitlab.find_sec_bugs.HARD_CODE_KEY-2).

rushtong (commented):
I think we still need one more piece of functionality here to complete the ticket which is to return a list of predefined mock objects filtered by the supplied query. Right now, it returns a single object regardless of query terms. One possibility is to push that to a separate ticket which I'm fine with. Outside of that, the content looks good.

rushtong (approved):
Thank you for the changes!

PR #2031, opened 2023-05-19
[DUOS-2469][risk=no] Add Dataset Index for Elastic Search

connorlbark:
### Addresses

Adds a new elastic search service and plumbing for indexing dataset terms.

### Summary

https://broadworkbench.atlassian.net/browse/DUOS-2469

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/models/elastic_search/DatasetTerm.java
```diff
@@ -0,0 +1,170 @@
...
+  private String piName;
+  private UserTerm dataSubmitter;
+  private String dataCustodian;
+  private DataUseSummary dataUse;
+  private List<String> dataTypes;
+  private String dataLocation;
+  private String dacName;
+  private Boolean openAccess;
+  private Boolean publicVisibility;
+  private List<UserTerm> approvedUsers;
```
connorlbark: In the ticket, it specified just using emails, or names. I figured it would be more robust if we just created a user term, so we could go email, name or id.

src/main/java/org/broadinstitute/consent/http/models/elastic_search/DatasetTerm.java
```diff
@@ -0,0 +1,170 @@
...
+import org.broadinstitute.consent.http.models.ontology.DataUseSummary;
+
+public class DatasetTerm {
+
+  private Integer datasetId;
+  private String datasetIdentifier;
+  private String description;
+  private String studyName;
+  private Integer studyId;
+  private Integer participantCount;
```
connorlbark, line 10: This one is problematic - there could be more than one `FileType` so there isn't just one "participant count". I wasn't sure what to do with it, so I am not filling it out. not sure how to change - could become an array or select the maximum or just get removed.

src/main/java/org/broadinstitute/consent/http/models/elastic_search/DatasetTerm.java
```diff
@@ -0,0 +1,170 @@
...
+  private String species;
+  private String piName;
+  private UserTerm dataSubmitter;
+  private String dataCustodian;
+  private DataUseSummary dataUse;
+  private List<String> dataTypes;
+  private String dataLocation;
+  private String dacName;
+  private Boolean openAccess;
+  private Boolean publicVisibility;
```
connorlbark: I added public visibility because I felt it would be essential to know when displaying datasets to the user.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -0,0 +1,180 @@
...
+            dataset.getDataSetId());
+
+    if (!approvedUserIds.isEmpty()) {
+      Collection<User> approvedUsers = userDAO.findUsers(approvedUserIds);
+
+      List<UserTerm> approvedUserTerms = approvedUsers.stream().map(this::toUserTerm).toList();
+      term.setApprovedUsers(approvedUserTerms);
+    }
+
+    term.setDataUse(useRestrictionConverter.translateDataUseSummary(dataset.getDataUse()));
```
connorlbark: per discussion w Greg, data use is saved as the output of `/translate/summary` on ontology, which might need more return fields. for example, with a DS restriction, there is no field which specifies the diseases it restricts to. Perhaps we want to add a ticket to do that?

connorlbark (commented):
Some thoughts - I also haven't actually ran this against a real elastic search index, but I almost feel like that's out of scope of this ticket. The indexing code is very simple/copied straight from the docs and implementing a way to trigger it (with, e.g., an endpoint) seems out of scope.

src/main/java/org/broadinstitute/consent/http/models/elastic_search/DatasetTerm.java
rushtong: To meet MVP, we can use just user ids.The UI will (eventually) have a need to filter on datasets that the user already has access to. Since the user id is known at that time, that's the minimum info we need to filter on. More info does provide additional filterable abilities, but we need to explore those when that need arises.

src/main/java/org/broadinstitute/consent/http/models/elastic_search/DatasetTerm.java
```diff
@@ -0,0 +1,170 @@
+package org.broadinstitute.consent.http.models.elastic_search;
+
+import java.util.List;
+import org.broadinstitute.consent.http.models.ontology.DataUseSummary;
+
+public class DatasetTerm {
```
rushtong, line 6: I think this can be a record class. Florian is doing something similar in his mock search PR. Depending on which one lands first, we'll need to coalesce these objects.

src/main/java/org/broadinstitute/consent/http/models/elastic_search/DatasetTerm.java
rushtong: This is a clear mismatch between the new form and the mocks ... leave it as is for now since we'll definitely need revised mocks to show multiple file types per dataset if, indeed, that is what we need.

src/main/java/org/broadinstitute/consent/http/models/elastic_search/DatasetTerm.java
```diff
@@ -0,0 +1,170 @@
...
+  private Integer participantCount;
+  private String phenotype;
+  private String species;
+  private String piName;
+  private UserTerm dataSubmitter;
+  private String dataCustodian;
+  private DataUseSummary dataUse;
+  private List<String> dataTypes;
+  private String dataLocation;
+  private String dacName;
```
rushtong: Revisiting the mocks ... it looks like we have a link to the DAC and data location so we might need a couple more fields here ... perhaps dacId a more complex version of data location.

src/main/java/org/broadinstitute/consent/http/models/elastic_search/UserTerm.java
```diff
@@ -0,0 +1,43 @@
+package org.broadinstitute.consent.http.models.elastic_search;
+
+import java.util.List;
+
+public class UserTerm {
+
+  private Integer userId;
+  private String email;
+  private String displayName;
+  private List<String> roles;
```
rushtong: This is something I would prefer that we do not push to ES.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -0,0 +1,180 @@
...
+import org.broadinstitute.consent.http.models.User;
+import org.broadinstitute.consent.http.models.UserRole;
+import org.broadinstitute.consent.http.models.elastic_search.DatasetTerm;
+import org.broadinstitute.consent.http.models.elastic_search.UserTerm;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ElasticSearchService {
+
+  private final Logger logger = LoggerFactory.getLogger(this.getClass());
```
rushtong: Use `ConsentLogger` instead.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
rushtong: We do have a separate API to get the full disease term (see `/search` in Ontology) so we can still proceed with this PR and improve the efficiency with an update to the `/translate/summary` call.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -0,0 +1,180 @@
...
+    this.userDAO = userDAO;
+  }
+
+  public BulkResponse indexDatasets(List<DatasetTerm> datasets) throws IOException {
+    BulkRequest.Builder br = new BulkRequest.Builder();
+
+    for (DatasetTerm dataset : datasets) {
+      br.operations(op -> op
+          .index(idx -> idx
+              .index("datasets")
```
rushtong: Something I think we need here is to check that the index exists before trying to push to it.
Also, the index name should be a config value instead of hard coded. We'll need a corresponding config update in https://github.com/broadinstitute/terra-helmfile as well.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -0,0 +1,180 @@
...
+    this.dataAccessRequestDAO = dataAccessRequestDAO;
+    this.dacDAO = dacDAO;
+    this.userDAO = userDAO;
+  }
+
+  public BulkResponse indexDatasets(List<DatasetTerm> datasets) throws IOException {
+    BulkRequest.Builder br = new BulkRequest.Builder();
+
+    for (DatasetTerm dataset : datasets) {
+      br.operations(op -> op
```
rushtong: I _think_ we can reverse this logic a bit. We're executing a bulk update for each dataset. Instead, I think we can start the bulk operation on all dataset term documents.

src/main/java/org/broadinstitute/consent/http/models/elastic_search/DatasetTerm.java
connorlbark: So, would you prefer to remove the UserTerm class and just move to IDs?

src/main/java/org/broadinstitute/consent/http/models/elastic_search/DatasetTerm.java
connorlbark: Hmm, I do agree, but in trying it out, the sheer number of properties makes it hard to work with, since records are immutable and have to be provided all the properties at once. Building it up one by one I think is a lot cleaner. If you don't mind a bit of clunkiness, I am happy to change it, though

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
connorlbark: https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/current/indexing-bulk.html

I don't know enough about elastic search to argue this, but I copied it straight from the official docs. I think this builds up a bulk request, so in the end they're sent in parallel instead of sequentially. But I could be wrong!

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
connorlbark: By check that the index exists, do you mean ensure that there is not a dataset with the ID provided already in elastic search? If so, [I think `index` is by default an `upsert` operation](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html): 

> `create` fails if a document with the same ID already exists in the target, `index` adds or replaces a document as necessary

Or, do you mean check that `datasets` is a valid elastic search index? I am not sure how to... The modern client's documentation never shows how to query the existence of an index or to create one. I think it handles that automatically. From what I can tell, I think it doesn't matter, because[ it will create the index if it doesn't exist already automatically](https://www.elastic.co/guide/en/elasticsearch/reference/8.7/docs-index_.html): 

> If the target doesn’t exist and matches the name or wildcard (*) pattern of an [index template with a data_stream definition](https://www.elastic.co/guide/en/elasticsearch/reference/8.7/set-up-a-data-stream.html#create-index-template), this request creates the data stream. See [Set up a data stream](https://www.elastic.co/guide/en/elasticsearch/reference/8.7/set-up-a-data-stream.html).

I agree on hard-coded though, I will fix that.

src/main/java/org/broadinstitute/consent/http/models/elastic_search/DatasetTerm.java
rushtong: I'm fine with this approach if that's what we're doing.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
rushtong: > Or, do you mean check that datasets is a valid elastic search index?

This is what I meant. We'll have to test this out - prod ES is version 5.4.0.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
rushtong: You are correct here, I mis-read what was happening. The operations are built up and only executed once on line 66.

rushtong: I think the only optimization I would make would be to stream the `DatasetTerm` objects but that's just a stylistic change, so no need to change this.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -0,0 +1,179 @@
...
+        study.getProperties(), "species"
+    ).ifPresent(
+        prop -> term.setSpecies((String) prop.getValue())
+    );
+
+    findStudyProperty(
+        study.getProperties(), "dataCustodianEmail"
+    ).ifPresent(
+        prop -> term.setDataCustodian((String) prop.getValue())
+    );
```
rushtong: In testing on dev, `study.getProperties()` can be null

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -0,0 +1,179 @@
...
+
+    User dataSubmitter = userDAO.findUserById(study.getCreateUserId());
+    term.setDataSubmitter(toUserTerm(dataSubmitter));
```
rushtong: In testing on dev, this can also be null.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -0,0 +1,179 @@
...
+
+    return term;
+  }
+
+  Optional<DatasetProperty> findDatasetProperty(Collection<DatasetProperty> props,
+      String schemaProp) {
+    return
+        props
+            .stream()
+            .filter((p) -> p.getSchemaProperty().equals(schemaProp))
```
rushtong, line 169: `p.getSchemaProperty()` can be null ...
```suggestion
            .filter(p -> Objects.nonNull(p.getSchemaProperty()))
            .filter(p -> p.getSchemaProperty().equals(schemaProp))
```

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
rushtong: Apparently, there are 25 datasets in dev that have null data use objects so we need to handle that here too.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -0,0 +1,179 @@
...
+      br.operations(op -> op
+          .index(idx -> idx
+              .index(esConfig.getDatasetIndexName())
+              .id(dataset.getDatasetId().toString())
+              .document(dataset)
+          )
+      );
+    }
+
+    return esClient.bulk(br.build());
```
rushtong: With all of the null checks handled, I was able to get to this statement. It ended up failing with about 600+ terms
```
{
  "message": "[es/bulk] failed: [action_request_validation_exception] Validation Failed: 1: type is missing;
...
```
It also took about 4 minutes to complete the execution (run against all datasets) so we should probably look into batching up groups of datasets as opposed to trying to do them all in one go.

rushtong: The failure message leads me to think there's an incompatibility between the most recent java library's bulk call and our very old ES server instance (see this [SO](https://stackoverflow.com/questions/73446339/elasticsearch-exception-type-action-request-validation-exception-reason-valida) post). 

In the past, due to these kinds of inherent issues, I've relied on the ~Http~ Rest client and not the java client. It might make sense to try that approach here.

connorlbark: I agree - although, what's the reason we're sticking to an old version? Is there a reason we can't update? It looks like we're 3 major versions behind, which seems pretty drastic.

rushtong: > what's the reason we're sticking to an old version?

We don't manage the ES instance, devops does, so it's not under our control. In theory, we could do a migration to a newer instance, but that's a separate effort that we'd need to request prioritization on. I'll start that process on our end, but it shouldn't hold up what we need to do.

connorlbark: Gotcha, that makes sense. I'll migrate to the old client then.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
connorlbark: https://www.elastic.co/guide/en/elasticsearch/reference/5.4/docs-index_.html#index-creation
I haven't tested yet but it looks like it was true back then, too.

connorlbark: Worked for me - didn't need to create `datasets` index manually before adding datasets.

rushtong: Nice!

connorlbark:
@rushtong This is FINALLY ready for a re-review. I can't get "index all datasets" to work because of the ontology call - if you remove that call, it works fine, but it takes so long with it that it just times out, even if it would eventually be successful. However, indexing individual datasets at a time (with ontology call) works. I think we can figure out some way to index in chunks or something to make sure it isn't too much.

src/test/java/org/broadinstitute/consent/http/service/UseRestrictionConverterTest.java
```diff
@@ -234,4 +238,85 @@ public void testTranslateTrueValues() {
     assertTrue(dataUse.getHmbResearch());
   }
 
+  @Test
+  public void testTranslateDataUseSummary() {
```
rushtong: Curious as to why this test is in the PR.

src/main/java/org/broadinstitute/consent/http/models/elastic_search/UserTerm.java
```diff
@@ -0,0 +1,14 @@
...
+  private Integer userId;
+
+  public Integer getUserId() {
+    return userId;
+  }
+
+  public void setUserId(Integer userId) {
+    this.userId = userId;
+  }
+}
```
rushtong, line 14: If we've reduced this to just an ID, it might be easier to just store the id instead of a wrapper around the ID.

src/test/java/org/broadinstitute/consent/http/service/UseRestrictionConverterTest.java
rushtong: OK, I can see this now. Not sure this is the right location at this time - `UseRestriction` work should be completely deprecated at this point since we're removed all of that from Ontology

connorlbark: I had to add it because we weren't ever calling the data use summary endpoint in consent, so I had to add the plumbing in `UseRestrictionCoverter` - which felt like an appropriate place, but I could be wrong. The correspond code was added in UseRestrictionConverter

src/main/java/org/broadinstitute/consent/http/models/elastic_search/DatasetTerm.java
connorlbark: Since there is now an actual participant count, I added code to grab it.

src/main/java/org/broadinstitute/consent/http/models/elastic_search/UserTerm.java
rushtong: After some testing, I think this would be an important change. For example, this is what we see now ...
```
          "approvedUsers": [
            {
              "userId": 5078
            },
            {
              "userId": 4949
            },
```
But, I think this would be easier for the UI to determine if the current user is in the list of approved users:
```
          "approvedUserIds": [5078, 4949],
```

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -0,0 +1,214 @@
...
+  public Response indexDatasets(List<DatasetTerm> datasets) throws IOException {
+    List<String> bulkApiCall = new ArrayList<>();
+
+    datasets.forEach((dsTerm) -> {
+      bulkApiCall.add(bulkHeader.formatted(dsTerm.getDatasetId()));
+      bulkApiCall.add(GsonUtil.getInstance().toJson(dsTerm) + "\n");
+    });
+
+    Request bulkRequest = new Request(
+        "PUT",
```
rushtong: Suggestion: use `jakarta.ws.rs.HttpMethod.PUT` here

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -0,0 +1,214 @@
...
+import org.broadinstitute.consent.http.util.gson.GsonUtil;
+import org.elasticsearch.client.Request;
+import org.elasticsearch.client.Response;
+import org.elasticsearch.client.RestClient;
+
+public class ElasticSearchService implements ConsentLogger {
+
+  private final RestClient esClient;
+  private final ElasticSearchConfiguration esConfig;
+  private final DatasetDAO datasetDAO;
```
rushtong: nit: unused

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -0,0 +1,214 @@
...
+  private final DataAccessRequestDAO dataAccessRequestDAO;
+  private final DacDAO dacDAO;
```
rushtong: nit: unused

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -0,0 +1,214 @@
...
+  private static final String bulkHeader = """
+      { "index": {"_type": "dataset", "_id": "%d"} }
+      """;
+
+  public Response indexDatasets(List<DatasetTerm> datasets) throws IOException {
+    List<String> bulkApiCall = new ArrayList<>();
+
+    datasets.forEach((dsTerm) -> {
+      bulkApiCall.add(bulkHeader.formatted(dsTerm.getDatasetId()));
+      bulkApiCall.add(GsonUtil.getInstance().toJson(dsTerm) + "\n");
```
rushtong, line 55: In testing, I hit this error when indexing all datasets. I suspect it is originating from this line.
```
{
  "message": "class com.google.gson.JsonArray cannot be cast to class java.lang.String (com.google.gson.JsonArray is in unnamed module of loader 'app'; java.lang.String is in module java.base of loader 'bootstrap')",
  "code": 500
}
```

rushtong (commented):
Still reviewing, but I think there's enough here for some minor revisions.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -0,0 +1,214 @@
...
+    findDatasetProperty(
+        dataset.getProperties(), "numberOfParticipants"
+    ).ifPresent(
+        datasetProperty -> term.setParticipantCount((Integer) datasetProperty.getPropertyValue())
+    );
+
+    findDatasetProperty(
+        dataset.getProperties(), "url"
+    ).ifPresent(
+        datasetProperty -> term.setUrl((String) datasetProperty.getPropertyValue())
```
rushtong: Here and in other places, use `getPropertyValueAsString()` as opposed to casting. We should try to avoid the casts here and in other places as well.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -0,0 +1,214 @@
...
+    findStudyProperty(
+        study.getProperties(), "species"
+    ).ifPresent(
+        prop -> term.setSpecies((String) prop.getValue())
+    );
+
+    findStudyProperty(
+        study.getProperties(), "dataCustodianEmail"
+    ).ifPresent(
+        prop -> term.setDataCustodian((String) prop.getValue())
```
rushtong: This can be an array in some cases and throws an error. On dev, you'll see this with dataset ids 1422 -> 1436
```
Unable to map dataset id 1422 to a term: class com.google.gson.JsonArray cannot be cast to class java.lang.String (com.google.gson.JsonArray is in unnamed module of loader 'app'; java.lang.String is in module java.base of loader 'bootstrap')
```

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -0,0 +1,214 @@
...
+      bulkApiCall.add(GsonUtil.getInstance().toJson(dsTerm) + "\n");
+    });
+
+    Request bulkRequest = new Request(
+        "PUT",
+        "/" + esConfig.getDatasetIndexName() + "/_bulk");
+
+    bulkRequest.setEntity(new NStringEntity(
+        String.join("", bulkApiCall) + "\n",
+        ContentType.DEFAULT_BINARY));
```
rushtong, line 64: I think either in the client builder code, we need to add a `Content-Type` header or maybe to specify the content type as json here instead of binary. Caveat, have not tested this out either way.

```
elastic  | [2023-06-01T16:48:35,881][WARN ][o.e.d.r.RestController   ] Content type detection for rest requests is deprecated. Specify the content type using the [Content-Type] header.
```

src/main/java/org/broadinstitute/consent/http/models/elastic_search/UserTerm.java
connorlbark: That's fair - I just didn't want to mix approaches, e.g., with `study` object and other object-y things, and I could see needing more info on users in the future. but i think this change makes sense.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
connorlbark: You have to use binary for this endpoint. Don't know why. But I found documentation online claiming this (only for v 5.4 - future versions of ES don't use binary content). But when I was testing it, anything else failed.

src/test/java/org/broadinstitute/consent/http/service/UseRestrictionConverterTest.java
connorlbark: Is there a place we put ontology calls? It feels weird to lump it into the rest of the service class I added, but also I don't see anywhere else appropriate.

rushtong (commented):
Thank you for the changes, these look great! We need to have a PR in terra-helmfile for the new config before we move this one forward.

kaitlynoflaherty (approved):
Everything looks good 👍

rushtong (approved):
👍🏽

PR #2039, opened 2023-05-26
[DUOS-2531][risk=no] Test helper cleanup

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2531

### Summary
Minor PR to start the cleanup of DAOTestHelper. Essentially, we need to move all protected object creation out to the test classes that need the populated objects. This PR doesn't partially addresses that. Future PRs will complete the task.

----

- Verify all tests go green

### DIFF
--- a/src/test/java/org/broadinstitute/consent/http/db/UserPropertyDAOTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/db/UserPropertyDAOTest.java
@@ -5,13 +5,13 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.Date;
 import java.util.List;
+import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.commons.lang3.RandomUtils;
 import org.broadinstitute.consent.http.enumeration.UserFields;
 import org.broadinstitute.consent.http.enumeration.UserRoles;
 import org.broadinstitute.consent.http.models.User;
 import org.broadinstitute.consent.http.models.UserProperty;
 import org.junit.jupiter.api.Test;
-import org.testcontainers.shaded.org.apache.commons.lang3.RandomStringUtils;
 
 public class UserPropertyDAOTest extends DAOTestHelper {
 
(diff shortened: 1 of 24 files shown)


connorlbark:
Still looking through this, but one thing I am concerned about with this approach is refactorability. If, e.g., datasets get a new required field, you have to change it in 30 different places, and I could see that leading to hard to catch bugs and more headache as a developer. I feel that if there are different needs between DAOTest classes, then there should be different/appropriately named functions within the DAOTestHelper. That way it's the most flexible to future changes. However, I won't hold this ticket up, since I know you prefer this approach, I just feel hesitant about it.

rushtong:
> but one thing I am concerned about with this approach is refactorability. If, e.g., datasets get a new required field, you have to change it in 30 different places, and I could see that leading to hard to catch bugs and more headache as a developer.

This is the problem that this PR is trying to solve, but what is true now is the reverse. Every test class needs some flavor of object creation and by consolidating all of that into one class, we end up creating an uber-service class that is in fact unmanageable, hard to debug, and forces the developer to go into every usage of that class across all sorts of unrelated test classes. Every change to `Institution`, for example, if it were consolidated in one place, will side affect every other class that needs it outside of the single class that needed the change. We've seen this happen with User, DAR, DAC, etc. Each time a test class needs a slightly new flavor of the class, we create a variation of a create method in the main helper class. The idea here is to make test classes independent.

connorlbark:
I totally get where you're coming from as far as creation code that might need variations across DAOTest, I just think this makes it even more complicated / adds more places code needs to change for the more common case that instantiation code does not change. But, I won't fight it - I will look through it and give a 👍  soon

src/test/java/org/broadinstitute/consent/http/db/UserPropertyDAOTest.java
```diff
@@ -62,4 +64,19 @@ public void testFindResearcherProperties() {
         (p.getPropertyKey().equals(UserFields.SUGGESTED_SIGNING_OFFICIAL.getValue())
             && p.getPropertyValue().equals(suggestedSigningOfficial.getPropertyValue()))));
   }
+
+  private User createUserWithRole(Integer roleId) {
+    int i1 = RandomUtils.nextInt(5, 10);
+    int i2 = RandomUtils.nextInt(5, 10);
+    int i3 = RandomUtils.nextInt(3, 5);
+    String email = org.apache.commons.lang3.RandomStringUtils.randomAlphabetic(i1) +
```
connorlbark: nit: should be imported

connorlbark (approved):
Looks good!

connorlbark (approved):
Looks good!

src/test/java/org/broadinstitute/consent/http/db/UserPropertyDAOTest.java
rushtong: Nice catch

kaitlynoflaherty (approved):
Looks good!

PR #2050, opened 2023-06-01
[DUOS-2515]Update Dataset Search Feature

kaitlynoflaherty:
### Addresses
https://broadworkbench.atlassian.net/jira/software/c/projects/DUOS/boards/123?modal=detail&selectedIssue=DUOS-2515

### Summary
After this update: `GET /api/dataset/search?query={term}&open=true|false`
The open parameter will have a default value of false in the case that it is not passed in as a value.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
@@ -313,7 +313,7 @@ public class Dataset {
    * @return if the Dataset matched query
    *
    */
-  public boolean isStringMatch(@NonNull String query, @NonNull Boolean openAccess) {
+  public boolean isDatasetMatch(@NonNull String query, @NonNull Boolean openAccess) {
     String lowerCaseQuery = query.toLowerCase();
     List<String> queryTerms = List.of(lowerCaseQuery.split("\\s+"));
 
--- a/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
@@ -1,11 +1,15 @@
 package org.broadinstitute.consent.http.models;
 
+import static org.broadinstitute.consent.http.models.dataset_registration_v1.DatasetRegistrationSchemaV1.AlternativeDataSharingPlanControlledOpenAccess.OPEN_ACCESS;
+
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringUtils;
 import org.checkerframework.checker.nullness.qual.NonNull;
 import org.broadinstitute.consent.http.models.dataset_registration_v1.ConsentGroup;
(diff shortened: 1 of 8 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -525,10 +526,13 @@ public Response datasetAutocomplete(@Auth AuthUser authUser,
   @Path("/search")
   @Produces("application/json")
   @PermitAll
-  public Response searchDatasets(@Auth AuthUser authUser, @QueryParam("query") String query) {
+  public Response searchDatasets(
+      @Auth AuthUser authUser,
+      @QueryParam("query") String query,
+      @QueryParam("open") Boolean openAccess) {
```
rushtong: This needs a default value
```suggestion
      @QueryParam("open") @DefaultValue("false") Boolean openAccess) {
```

src/main/java/org/broadinstitute/consent/http/models/Dataset.java
```diff
@@ -290,22 +303,51 @@ public void setDeletable(Boolean deletable) {
     this.deletable = deletable;
   }
 
+  // helper methods for open access string matching
+  public Optional<DatasetProperty> getPropertyByName(String name) {
```
rushtong: The only use of this is in the string matching method - we should move the logic there.

src/main/java/org/broadinstitute/consent/http/models/Dataset.java
```diff
@@ -290,22 +303,51 @@ public void setDeletable(Boolean deletable) {
...
+   * the raw search query and open access.
    *
-   * @param query Raw string query
+   * @param query       Raw string query
+   * @param openAccess  Nullable Boolean for open access
    * @return if the Dataset matched query
+   *
    */
-  public boolean isStringMatch(@NonNull String query) {
+  public boolean isStringMatch(@NonNull String query, Boolean openAccess) {
```
rushtong: `openAccess` can be a primitive which would force this to be non-null

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -428,9 +428,9 @@ public Set<DatasetDTO> describeDatasets(Integer userId) {
   }
 
 
-  public List<Dataset> searchDatasets(String query, User user) {
+  public List<Dataset> searchDatasets(String query, Boolean openAccess, User user) {
```
rushtong: Similar comment, this can be a primitive which forces it to be non-null

src/main/java/org/broadinstitute/consent/http/models/Dataset.java
fboulnois: In general I think it's a little unusual to overload an `isStringMatch` method with an `openAccess` parameter. Instead I would recommend a creating method which checks the `openAccess` parameter with the logic you added below and calls `isStringMatch`. This way you also wouldn't need to overload all the `stringMatch` tests.

fboulnois (commented):
One suggestion:

src/main/java/org/broadinstitute/consent/http/models/Dataset.java
kaitlynoflaherty: Since each dataset and it's properties are checked individually and return whether it is a match or not, I think it needs to be inside this method. But let me know your thoughts, maybe there's another approach I can take

fboulnois: Perhaps we should rename the method to something like `isDatasetMatch` instead then?

kaitlynoflaherty: Yes, I agree the name is confusing

rushtong (approved):
👍🏽

src/main/java/org/broadinstitute/consent/http/models/Dataset.java
```diff
@@ -307,6 +322,22 @@ public boolean isStringMatch(@NonNull String query) {
...
+          .filter((dp) -> Objects.nonNull(dp.getPropertyValue()))
+          .filter((dp) -> Objects.equals(dp.getPropertyName(), OPEN_ACCESS.toString()))
+          .findFirst();
+
+      if (openAccessProp.isEmpty()){
+        if (openAccess) {
+          return false;
+        }
+      }
+      else if (!Objects.equals(openAccessProp.get().getPropertyValue().toString(), Boolean.toString(openAccess)))
```
fboulnois, line 338: Perhaps we can add a minor `TODO` to investigate whether we can switch to try to coerce `getPropertyValue` to a boolean instead.

fboulnois (approved):
👍 , one tiny suggestion:

src/main/java/org/broadinstitute/consent/http/models/Dataset.java
kaitlynoflaherty: will do!

PR #2052, opened 2023-06-06
[DUOS-2128][risk=no] Add Get Dataset By Role API

connorlbark:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2128

### Summary

Adds an endpoint that returns the "visible" datasets according to their role, now with the concept of public visibility.

----

- Verify all tests go green


kaitlynoflaherty (approved):
Looks good!

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -613,6 +639,24 @@ public Response downloadDatasetApprovedUsers(@Auth AuthUser authUser,
...
+   * @param user     The User
+   * @param roleName The UserRole name
+   */
+  void validateUserHasRoleName(User user, String roleName) {
+    UserRoles thisRole = UserRoles.getUserRoleFromName(roleName);
+    if (Objects.isNull(thisRole) || !user.hasUserRole(thisRole)) {
+      throw new BadRequestException("Invalid role selection: " + roleName);
+    }
+  }
+
```
rushtong: This is a copy/paste from `Resource`

kaitlynoflaherty (approved):
These updates look good!

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
```diff
@@ -239,6 +239,185 @@ LEFT JOIN file_storage_object fso ON (fso.entity_id = d.dataset_id::text OR fso.
...
+            (
+              (s.public_visibility IS NULL OR s.public_visibility = TRUE)
+              AND d.active = TRUE
+              AND d.dac_approval = TRUE
+            )
+            OR
+            (
+              s.create_user_id = :userId
+              OR (sp.key = 'dataCustodianEmail' AND sp.value LIKE concat('%"', :email, '"%'))
+            )
```
fboulnois, line 277: Since the main difference between these queries is the where clause, I wonder if we should abstract that part away so it can be reused by these queries.

fboulnois (commented):
One suggestion, otherwise looks good.

fboulnois (approved):

PR #2054, opened 2023-06-06
[DUOS-2528]Get Study By Id API

kaitlynoflaherty:
### Addresses
https://broadworkbench.atlassian.net/jira/software/c/projects/DUOS/boards/123?modal=detail&selectedIssue=DUOS-2528

### Summary

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/ConsentModule.java
```diff
@@ -294,7 +294,7 @@ DatasetService providesDatasetService() {
         providesUserRoleDAO(),
         providesDacDAO(),
         providesUseRestrictionConverter(),
-        providesEmailService());
+        providesEmailService(), studyDAO);
```
connorlbark: should be `providesStudyDAO()`

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -370,6 +371,23 @@ public Response getDataset(@PathParam("datasetId") Integer datasetId) {
...
+  @GET
+  @Path("/study/{studyId}")
+  @Consumes("application/json")
+  @Produces("application/json")
+  @RolesAllowed({ADMIN, CHAIRPERSON, DATASUBMITTER})
+  public Response getStudyById(@PathParam("studyId") Integer studyId) {
+    try {
+      Study study = datasetService.findStudyById(studyId);
+      if (Objects.isNull(study)) {
+        throw new NotFoundException("Could not find the dataset with id: " + studyId.toString());
```
connorlbark: Should say `Could not find the study with id: `

src/main/resources/assets/paths/studyById.yaml
```diff
@@ -0,0 +1,23 @@
...
+        type: integer
+  tags:
+    - Study
+  responses:
+    200:
+      description: The Study with properties
+      content:
+        application/json:
+          schema:
+            $ref: '../schemas/DatasetRegistrationSchemaV1.yaml'
```
connorlbark: This should be`../schemas/Study.yaml` - it should exist, but i'm not sure.

connorlbark (commented):
Looking good! Just a couple minor things

src/main/resources/assets/paths/studyById.yaml
kaitlynoflaherty: It does exist and I added it, I wasn't sure which one was more appropriate based on the ticket description. We might need to make some edits in the future but for now this should work fine

src/main/resources/assets/paths/studyById.yaml
```diff
@@ -0,0 +1,23 @@
+get:
+  summary: Get Study by Id
+  description: Finds Study and Properties of the Study specified by id.
+  parameters:
+    - name: id
```
rushtong: The variable here should match what's in api-docs
```suggestion
    - name: studyId
```

src/main/resources/assets/paths/studyById.yaml
```diff
@@ -0,0 +1,23 @@
...
+      in: path
+      description: ID of the Study
+      required: true
+      schema:
+        type: integer
+  tags:
+    - Study
```
rushtong: This is a new category, which makes sense. We should probably put it in both, `Dataset` and `Study`. Also, since we're adding `Study` here, it would make sense to tag `/api/dataset/studyNames` as well.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -370,6 +371,23 @@ public Response getDataset(@PathParam("datasetId") Integer datasetId) {
     }
   }
 
+  @GET
+  @Path("/study/{studyId}")
+  @Consumes("application/json")
```
rushtong: Remove this line since we're not consuming any json. This bug took quite a while to debug. Hitting the endpoint from swagger, this always returned a 404 that was served from the root dropwizard application due to the consumes header mismatch.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -370,6 +371,23 @@ public Response getDataset(@PathParam("datasetId") Integer datasetId) {
...
+  @Produces("application/json")
```
rushtong: Nit: We don't do this consistently and so it's a general problem in the repo, but we should strive to use the constants instead of raw strings as that can easily be typo-ed
```suggestion
  @Produces(MediaType.APPLICATION_JSON)
```

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -370,6 +371,23 @@ public Response getDataset(@PathParam("datasetId") Integer datasetId) {
...
+  @RolesAllowed({ADMIN, CHAIRPERSON, DATASUBMITTER})
+  public Response getStudyById(@PathParam("studyId") Integer studyId) {
+    try {
+      Study study = datasetService.findStudyById(studyId);
```
rushtong: This really needs to return more than just the study info. The caller will need to have all of the associated datasets in addition to the study. Currently, this just returns a study with a list of dataset ids, but we should additionally pull the datasets as well.

.github/workflows/owasp.yaml
```diff
@@ -16,33 +16,15 @@ jobs:
         with:
```
rushtong: You'll need to merge the latest version of dev into your branch - this file shouldn't be in your changeset.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
rushtong: This still returns just a basic study object - it doesn't contain the datasets the UI will need. There are a couple ways to do this, the preferred approach would be to generate a new sql query that pulls all data together. Another approach is to first grab the study in the service class, then query on the dataset ids in a second call. Either one will accomplish the basic task but option 1 is more complex since we don't have that code now while we have all the code in place for option 2.

kaitlynoflaherty (commented):
I added the list of datasets which are returned with the study, but it seems like we don't need the entire dataset object because of it's redundant data. Let me know if you think I should edit this to only return certain dataset data!

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -1079,6 +1080,27 @@ public void testCreateDatasetRegistration_invalidFileName() {
     assertEquals(HttpStatusCodes.STATUS_CODE_BAD_REQUEST, response.getStatus());
   }
 
+  @Test
+  public void testGetStudy() {
+    Study study = new Study();
+    study.setStudyId(1);
+    study.setName("asdfasdfasdfasdfasdfasdf");
+    when(datasetService.findStudyById(1)).thenReturn(study);
```
fboulnois: I think it's also worth adding a test that returns mock datasets so you can ensure every line is covered.

fboulnois (approved):
The feature is in good shape, one suggestion:

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -371,6 +372,34 @@ public Response getDataset(@PathParam("datasetId") Integer datasetId) {
...
+      // if datasets exist, get full datasets and add to study
+      Set<Integer> datasetIds = study.getDatasetIds();
+
+      if(Objects.nonNull(datasetIds)){
+        List<Integer> datasetIdList = new ArrayList<>(datasetIds);
+        List<Dataset> datasetSet = datasetService.findDatasetsByIds(datasetIdList);
+        for ( Dataset dataset : datasetSet ) {
+          study.addDataset(dataset);
+        }
+      }
```
fboulnois:
```suggestion
      if(Objects.isNull(datasetIds)){
        throw new NotFoundException("Study has no datasets.");
      }
      List<Integer> datasetIdList = new ArrayList<>(datasetIds);
      List<Dataset> datasetSet = datasetService.findDatasetsByIds(datasetIdList);
      for ( Dataset dataset : datasetSet ) {
        study.addDataset(dataset);
      }
```

I think studies always need to have datasets, so if `datasetIds` is null then it should error.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -1079,6 +1081,37 @@ public void testCreateDatasetRegistration_invalidFileName() {
...
+    initResource();
+    Response response = resource.getStudyById(1);
+    assertEquals(200, response.getStatus());
+    assertEquals(study, response.getEntity());
+  }
+
+  @Test
+  public void testGetStudyByIdWithDatasets() {
```
fboulnois, line 1096: This currently does not test whether datasets exist because `Set<Integer> datasetIds = study.getDatasetIds();` returns `null` and thus will skip adding the datasets to the study. As a result, these lines are not covered.

To resolve this, mock the response from `study.getDatasetIds();` to return a list of integers and `datasetService.findDatasetsByIds(datasetIdList);` to return a list of dummy datasets.

fboulnois (changes requested):
A few thoughts:

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
kaitlynoflaherty: Thank you!

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -1079,6 +1081,53 @@ public void testCreateDatasetRegistration_invalidFileName() {
...
+    List<Integer> datasetIds = new ArrayList<>(study.getDatasetIds());
+
+    when(datasetService.findStudyById(12345)).thenReturn(study);
+    when(datasetService.findDatasetsByIds(datasetIds)).thenReturn(datasets);
+
+    initResource();
+    Response response = resource.getStudyById(12345);
+    assertEquals(200, response.getStatus());
+    assertEquals(study.getDatasetIds().size(), datasets.size())
+    ;
```
fboulnois:
```suggestion
    assertEquals(study.getDatasetIds().size(), datasets.size());
```

fboulnois (approved):
minor nit, otherwise looks good!

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -371,6 +372,35 @@ public Response getDataset(@PathParam("datasetId") Integer datasetId) {
...
+      Set<Integer> datasetIds = study.getDatasetIds();
+
+      if(Objects.isNull(datasetIds)){
+        throw new NotFoundException("Study has no datasets.");
+      }
+      List<Integer> datasetIdList = new ArrayList<>(datasetIds);
+      List<Dataset> datasetSet = datasetService.findDatasetsByIds(datasetIdList);
+      for (Dataset dataset : datasetSet) {
+        study.addDataset(dataset);
+      }
```
rushtong: nit: separation of concerns. This should be in a service call as opposed to being here in the resource class. I would suggest either updating `DatasetService.findStudyById` to return a study with datasets, or create a new method that explicitly retrieves a study with datasets, i.e., something like `DatasetService.findStudyWithDatasetsById`

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -371,6 +372,35 @@ public Response getDataset(@PathParam("datasetId") Integer datasetId) {
...
+
+      // if datasets exist, get full datasets and add to study
+      Set<Integer> datasetIds = study.getDatasetIds();
+
+      if(Objects.isNull(datasetIds)){
+        throw new NotFoundException("Study has no datasets.");
+      }
+      List<Integer> datasetIdList = new ArrayList<>(datasetIds);
+      List<Dataset> datasetSet = datasetService.findDatasetsByIds(datasetIdList);
```
rushtong: In theory, there should always be dataset ids, but in practice, we often have incomplete data in dev. The underlying call passes that list to `datasetDAO.findDatasetsByIdList(datasetIds)` and if it's empty, you'll get a sql error. We should handle that by checking if the list is empty and not making an erroneous call in that edge case.

src/main/java/org/broadinstitute/consent/http/models/Study.java
```diff
@@ -16,6 +16,7 @@ public class Study {
   private String piName;
   private List<String> dataTypes;
   private Set<Integer> datasetIds;
+  private Set<Dataset> datasetSet;
```
rushtong: nit: naming. I think this would be more descriptive as `datasets`

src/main/resources/assets/schemas/Study.yaml
```diff
@@ -27,6 +27,11 @@ properties:
       type: integer
       format: int32
     description: The ids of all of the datasets related to this study.
+  datasets:
```
rushtong, line 30: 👍🏽  This is what I think the name should be. What I see when calling this is the field name defined in `Study` ...
```
...
  "datasetIds": [
    1427
  ],
  "datasetSet": [
    {
      "dataSetId": 1427,
...
```

rushtong (commented):
I think this looks pretty good and functions as expected. Just a few comments inline about naming/code organization.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -548,4 +548,13 @@ public List<Dataset> findDatasetsByIds(List<Integer> datasetIds) {
...
   }
 
+  public Study addDatasetsToStudy(Set<Integer> datasetIds, Study study) {
+    List<Integer> datasetIdList = new ArrayList<>(datasetIds);
+    List<Dataset> datasets = findDatasetsByIds(datasetIdList);
+    for (Dataset dataset : datasets) {
+      study.addDataset(dataset);
+    }
+    return study;
+  }
```
rushtong: I think maybe my previous feedback was not clear. The goal of the service class is to populate the object requested, but not necessarily split that up into different methods. The resource class should ask the service class to get the study once, along with its datasets, so that we don't need to do population logic inside the resource class. If I were refactoring this, I'd have one method here in the service class, `getStudyWithDatasetsById(Integer studyId)`:
```
  public Study getStudyWithDatasetsById(Integer studyId) {
    Study study = studyDAO.findStudyById(studyId);
    if (Objects.isNull(study)) {
      throw new NotFoundException("Study not found");
    }
    if (Objects.nonNull(study.getDatasetIds()) && !study.getDatasetIds().isEmpty()) {
      List<Dataset> datasets = findDatasetsByIds(new ArrayList<>(study.getDatasetIds()));
      study.addDatasets(datasets);
    }
    return study;
  }
```
I would also add a new method to `Study` to add all datasets at once instead of looping and adding them individually.

kaitlynoflaherty: That makes sense, thank you!

rushtong (approved):
Thank you for the changes! 👍🏽

PR #2065, opened 2023-06-09
[DUOS-2075][risk=no] Part 1: Add translated data use to dataset

connorlbark:
### Addresses

PARTIALLY ADDRESSES https://broadworkbench.atlassian.net/browse/DUOS-2075

### Summary

Adds translated data use to the dataset, and prepopulates it with the already existing `translated_use_restriction` from consents.

----

- Verify all tests go green

### DIFF
--- a/src/main/resources/changesets/changelog-consent-2023-06-08-add-translated-data-use.xml
+++ b/src/main/resources/changesets/changelog-consent-2023-06-08-add-translated-data-use.xml
@@ -1,0 +1,16 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+  <changeSet id="2023-06-08-add-translated-data-use" author="connorlbark">
+    <addColumn tableName="dataset">
+      <column name="translated_data_use" type="text"/>
+    </addColumn>
+    <sql>
+      UPDATE dataset
+      SET translated_data_use = (SELECT c.translated_use_restriction AS translated_data_use
+                                 FROM consents c
+                                        INNER JOIN consent_associations ca ON ca.consent_id = c.consent_id
+                                 WHERE ca.dataset_id = dataset_id)
+    </sql>
+  </changeSet>
+</databaseChangeLog>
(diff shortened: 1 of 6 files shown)


src/main/resources/changesets/changelog-consent-2023-06-08-add-translated-data-use.xml
```diff
@@ -0,0 +1,17 @@
...
+      <column name="translated_data_use" type="text"/>
+    </addColumn>
+    <sql>
+      UPDATE dataset
+      SET translated_data_use = (SELECT c.translated_use_restriction AS translated_data_use
+                                 FROM consents c
+                                        INNER JOIN consent_associations ca ON ca.consent_id = c.consent_id
+                                 WHERE ca.dataset_id = dataset_id
+                                 ORDER BY c.create_date DESC LIMIT 1)
+    </sql>
```
connorlbark, line 15: One concern I have - the value that is stored by `translated_use_restriction` and thus now `translated_data_use` is just free-form text. Why not format it like `/translate/summary` like we do in elastic search? It would make it easier to parse into objects while still have the descriptions available to display to users, and give more options on how to display the data use to users.

Then, the value on the dataset could be a `DataUseSummary` instead of a `String`

rushtong: This is an interesting idea, but not backwards compatible with existing string translations. Let's discuss.

connorlbark: True - But I think it's pretty easy to recreate the original format if you do something like: `tdu.primary.map((res) => res.code + ":" + res.description) + tdu.secondary.map((res.code + ":" + res.description))`. Kinda ugly, but I don't think it would be a huge lift.

rushtong: We can leave it as a string here and do some work in the reducer to determine if it can be a `DataUseSummary` object. When populating it in future PRs, we can query Ontology for the correct value.

rushtong (approved):
👍🏽

src/main/resources/changesets/changelog-consent-2023-06-08-add-translated-data-use.xml
connorlbark: I'm not sure I follow - keep it as a string for now and in the future replace it? If so, I'm ok with that, it might be awkward in the intermediary period, though

rushtong: We won't need to replace anything in this column, we can parse the existing string value into whatever we want it to be. If it parses as a `DataUseSummary` (which would be the case for new or re-processed datasets), we can make that available to the UI. If it doesn't, then we can't - clients will need to handle that explicitly.

fboulnois (approved):
looks good 👍

PR #2069, opened 2023-06-13
[DUOS-2554] Edit Match Endpoint to Return V3 Result

kaitlynoflaherty:
### Addresses
https://broadworkbench.atlassian.net/jira/software/c/projects/DUOS/boards/123?modal=detail&selectedIssue=DUOS-2554

### Summary

----

- Verify all tests go green

### DIFF
--- a/src/main/resources/assets/api-docs.yaml
+++ b/src/main/resources/assets/api-docs.yaml
@@ -1550,6 +1550,9 @@ components:
         match:
           type: boolean
           description: True if the entities match, false otherwise.
+        abstain:
+          type: boolean
+          description: Nullable boolean will return true if the decision is abstain
         failed:
           type: boolean
           description: True if the matching process failed, for any reason.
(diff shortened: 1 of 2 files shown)


src/main/resources/assets/api-docs.yaml
```diff
@@ -1550,6 +1550,9 @@ components:
         match:
           type: boolean
           description: True if the entities match, false otherwise.
+        abstain:
+          type: boolean
+          description: Nullable boolean will return true if the decision is abstain
```
rushtong, line 1555: Interesting that we need this in two places, we should probably combine. That can be a separate PR though.

rushtong (approved):
👍🏽

src/main/resources/assets/api-docs.yaml
kaitlynoflaherty: Yeah, I can make a quick ticket for that as well!

fboulnois (approved):
lgtm

PR #2072, opened 2023-06-14
[DUOS-2532][risk=no] Update Study by Id

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2532

### Summary
Updates a study by id. Includes synchronization of both study and dataset properties and file uploads.

----

- Verify all tests go green


src/main/resources/assets/paths/studyById.yaml
```diff
@@ -21,4 +21,51 @@ get:
...
+      data sharing information is required in the registration schema.
+    content:
+      multipart/form-data:
+        schema:
+          type: object
+          properties:
+            dataset:
+              description: Required dataset registration json schema instance
+              type: object
+              $ref: '../schemas/DatasetRegistrationSchemaV1.yaml'
```
rushtong, line 52: Technically, `$ref` is not allowed here but this does work in practice and is necessary to fully document this API

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
```diff
@@ -14,48 +14,138 @@
...
+              datasetInserts.add(insert);
+            } catch (Exception e) {
+              logException(e);
+            }
+          }
+        });
+
+    List<StudyProperty> studyProps = convertRegistrationToStudyProperties(registration);
+    DatasetServiceDAO.StudyUpdate studyUpdate = new StudyUpdate(
+        registration.getStudyName(),
```
kaitlynoflaherty, line 136: I think there may be some discrepancy between the GET study endpoint and this PUT API. This endpoint is looking for fields on the study `studyName`, `consentGroups` and `studyDescription` where those fields are not the same on the `getStudyById` endpoint. I'm not sure if this problem is within the scope of this ticket but I am getting a bad request while testing with the GET. However, I also know that the data is corrupted in the studies so it may be something to do with that as well.

rushtong: The GET and the PUT are intentionally different. The GET should give you a study as defined by our Study class, but the PUT is expecting and instance of dataset registration schema which is very different from a Study

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -203,6 +203,49 @@ public Response createDatasetRegistration(
...
+      }
+
+      DatasetRegistrationSchemaV1 registration = jsonSchemaUtil.deserializeDatasetRegistration(
+          json);
+      User user = userService.findUserByEmail(authUser.getEmail());
+
+      // key: field name (not file name), value: file body part
+      Map<String, FormDataBodyPart> files = extractFilesFromMultiPart(multipart);
+
+      // Update study from registration
```
rushtong, line 262: I think I need something here to validate that we're not trying to delete any consent groups in the incoming payload.

kaitlynoflaherty (approved):
Looks great!!

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -203,6 +206,89 @@ public Response createDatasetRegistration(
...
+                  Objects.nonNull(cg.getNmds()) ||
+                  Objects.nonNull(cg.getGso()) ||
+                  Objects.nonNull(cg.getPub()) ||
+                  Objects.nonNull(cg.getCol()) ||
+                  Objects.nonNull(cg.getIrb()) ||
+                  Objects.nonNull(cg.getGs()) ||
+                  Objects.nonNull(cg.getMor()) ||
+                  Objects.nonNull(cg.getMorDate()) ||
+                  Objects.nonNull(cg.getNpu()) ||
+                  Objects.nonNull(cg.getOtherSecondary()))
```
fboulnois: nit: could we have a method on the consent group to determine whether any of these are filled?

fboulnois (approved):
one minor suggestion, otherwise lgtm:

PR #2110, opened 2023-07-27
[DUOS-2617][risk=no] Add abstain and rationales to match inserts

rushtong:
### Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DUOS-2617

### Summary
Existing code was not saving the `abstain` value returned from Ontology when a V3 match was made so when queried from the UI, they were always null. This PR ensures that we save that value when we get it back from Ontology. It also renames failure reasons to rationales to capture all returned match explanations.

See also: https://github.com/DataBiosphere/duos-ui/pull/2296

Example response:
```
  {
    "id": 1355,
    "consent": "DUOS-000013",
    "purpose": "7ca0f5a9-afdd-41e7-b226-2aad825f8079",
    "match": false,
    "abstain": true,
    "failed": false,
    "createDate": "Jul 27, 2023",
    "algorithmVersion": "v3",
    "rationales": [
      "The proposed population, origins, and/or ancestry research is within the bounds of the population, origins, and/or ancestry use permissions of the dataset(s)",
      "Methods research is permitted on controlled-access data so long as it is not expressly prohibited",
      "The Commercial Use Research Purpose does not match the No Commercial Use data use limitation.",
      "The Research Purpose does not result in DUOS Decision."
    ]
  },
```

----

- Verify all tests go green

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -37,7 +37,7 @@
         </value>
       </option>
     </AndroidXmlCodeStyleSettings>
-    <JSCodeStyleSettings>
+    <JSCodeStyleSettings version="0">
       <option name="INDENT_CHAINED_CALLS" value="false" />
     </JSCodeStyleSettings>
     <JavaCodeStyleSettings>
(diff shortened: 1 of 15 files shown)


fboulnois (approved):
looks great 👍

kaitlynoflaherty (approved):
👍

PR #2131, opened 2023-08-28
[DUOS-1442][risk=no] Remove deprecated APIs and related code.

rushtong:
### Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DUOS-1442

### Summary
Lots of cleanup in this PR. The primary impact is to remove the older, unused dataset APIs that made use of the deprecated `DatasetDTO` class.

There are two functional changes to look at in `DarCollectionService` - finding DAR Collection Summaries for Chairs and Members. In both cases, I replaced the old DAO call with an existing, similar one based on dac ids instead of the dto.

APIs removed:
* `GET /api/dataset`
* `GET /api/dataset/sample`
* `GET /api/dataset/dictionary`
* `GET /api/dataset/autocomplete/{partial}`

### Notes
I suggest hiding whitespace changes when reviewing - there are a lot of minor line formatting changes introduced by auto-reformatting.

----

- Verify all tests go green


src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
```diff
@@ -59,7 +60,7 @@
 import org.junit.jupiter.api.Test;
 import org.mockito.Mock;
 
-public class DarCollectionServiceTest {
+class DarCollectionServiceTest {
```
pshapiro4broad, line 63: In a future PR I think you should migrate to using `MockitoExtension` which IMO is a better way to create mocks in junit 5 (vs using `openMocks()`).

It makes the tests slightly stricter which means that you no longer need to `verify()` in most cases and can identify unused mocks in your test setup.

src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
```diff
@@ -910,7 +920,7 @@ public void testProcessDarCollectionSummariesForChair() {
     when(darCollectionSummaryDAO.getDarCollectionSummariesForDAC(any(), any()))
         .thenReturn(
             List.of(summaryOne, summaryTwo, summaryThree, summaryFour, summaryFive, summarySix));
-    when(datasetDAO.findDatasetDTOsByUserId(any())).thenReturn(Set.of());
+    when(datasetDAO.findDatasetListByDacIds(any())).thenReturn(List.of());
 
     initService();
```
pshapiro4broad: If all tests call `initService()` you can do it in your `BeforeEach` instead.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -350,17 +360,27 @@ public DarCollectionSummary getSummaryForRoleNameByCollectionId(User user, Strin
...
         case MEMBER:
-          datasetIds = datasetDAO.findDatasetDTOsByUserId(userId).stream()
-              .map(d -> d.getDataSetId())
-              .collect(Collectors.toList());
+          List<Integer> memberDacIds = user.getRoles().stream()
+              .filter(ur -> ur.getRoleId().equals(UserRoles.MEMBER.getRoleId()))
+              .map(UserRole::getDacId)
+              .filter(Objects::nonNull)
+              .toList();
+          datasetIds = memberDacIds.isEmpty() ? List.of() : datasetDAO.findDatasetListByDacIds(memberDacIds).stream()
```
pshapiro4broad: If you wanted to, you could remove the ternary op and use a stream, something like:
```
datasetIds = Stream.of(memberDacIds).filter(List::isEmpty).flatMap(datasetDAO.findDatasetListByDacIds(memberDacIds).stream()).map(...).toList();
```

pshapiro4broad: It looks like this block is duplicated four times, the only difference is the `UserRoles` value. Could it be refactored into a common method?
```
  private List<Integer> getDatasetIdsForUser(User user, UserRoles role) {
    List<Integer> chairDacIds = user.getRoles().stream()
        .filter(ur -> ur.getRoleId().equals(role.getRoleId()))
        .map(UserRole::getDacId)
        .filter(Objects::nonNull)
        .toList();
    return chairDacIds.isEmpty() ? List.of() : datasetDAO.findDatasetListByDacIds(chairDacIds).stream()
            .map(Dataset::getDataSetId)
            .toList();
  }
```

pshapiro4broad (approved):
This looks OK as far as I can tell. I have a few minor suggestions for code cleanups.

It's hard to tell if everything that needed to be removed was removed, and nothing that didn't need to be wasn't, especially since the api yaml isn't used to generate server code.

I see the build generates coverage data but I don't see a way to view it in the PR, so I don't know if coverage was changed by this PR. I'd recommend adding sonar support when you have a chance, which would give you the coverage information on PRs.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rushtong: Yes, this can be pulled out, thank you!

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -335,6 +322,20 @@ public List<DarCollectionSummary> getSummariesForRoleName(User user, String user
...
   }
 
+  private List<Integer> getDatasetIdsForUserAndRole(User user, UserRoles role) {
+    List<Integer> roleDacIds = user.getRoles().stream()
+        .filter(ur -> ur.getRoleId().equals(role.getRoleId()))
+        .map(UserRole::getDacId)
+        .filter(Objects::nonNull)
+        .toList();
+    return Stream.of(roleDacIds)
+        .filter(arr -> !arr.isEmpty())
```
pshapiro4broad: Whoops, guess I got the logic backward in my version. Up to you which way is clearer but using `Predicate.not()` should work here too
```suggestion
        .filter(Predicate.not(List::isEmpty))
```

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -324,6 +322,20 @@ public List<DarCollectionSummary> getSummariesForRoleName(User user, String user
     return summaries;
   }
 
+  private List<Integer> getDatasetIdsForUserAndRole(User user, UserRoles role) {
+    List<Integer> roleDacIds = user.getRoles().stream()
+        .filter(ur -> ur.getRoleId().equals(role.getRoleId()))
```
nmalfroy: Do you ever have to worry about `getRoleId()` returning a null?

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -324,6 +322,20 @@ public List<DarCollectionSummary> getSummariesForRoleName(User user, String user
...
+        .map(UserRole::getDacId)
+        .filter(Objects::nonNull)
+        .toList();
+    return Stream.of(roleDacIds)
```
nmalfroy, line 333: I'm curious about the choice of a flatMap here.  Seems like this could all be simplified by having `datasetDAO::findDatasetListByDacIds` just return an empty list if the input is an empty list no?

nmalfroy (approved):
Looks good!  Just a couple of small comments

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rushtong: Our DAO calls are just interface methods (jdbi) that just execute a simple statement. If the list is empty, we get a sql error, so I tend to err on the side of safety and prevent that in most places.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rushtong: Not in this case, we're passing in a pre-constructed role, but I could simplify this to accept a roleId instead.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rushtong: I like that, will give it a shot

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
pshapiro4broad: Yeah, my first inclination was to make the DAO layer check for an empty list and return early. But since you have an auto-generated DAO it makes more sense to address it in the caller, as long as there's only one caller.

PR #2135, opened 2023-09-01
[DUOS-2636][risk=no] Remove active and needs approval props

rushtong:
### Addresses
Back end work for https://broadworkbench.atlassian.net/browse/DUOS-2636

### Summary
See also: https://github.com/DataBiosphere/duos-ui/pull/2323

* Remove all usages of `active` and `needsApproval` on both `Dataset` and `DatasetDTO`
* Remove the now unused `DELETE /api/dataset/disable/{datasetId}/{active}` endpoint
* Remove the unused and undocumented `PUT /api/dataset` endpoint
* Removes all associated and unused code as fallout
* Minor fix for the older update dataset code path. The API was looking for property keys where the UI was not sending them, so we moved to names instead. Also removed the delete prop part since we should not be doing that in the UI at all.

----

- Verify all tests go green


fboulnois (approved):
looks great 👍

hams7504 (approved):
thanks for cleaning up everything!

PR #2140, opened 2023-09-07
[DUOS-2660][risk=no] Upgrade authentication tests to MockitoExtension

rushtong:
### Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DUOS-2660

### Summary
* Migrate some tests to using `MockitoExtension`
* Remove redundant `public` modifiers
* Packages addressed:
  * `org.broadinstitute.consent.http.authentication`
  * `org.broadinstitute.consent.http.cloudstore`
  * `org.broadinstitute.consent.http.models`
  * `org.broadinstitute.consent.http.util`

Future PRs will address other packages.

----

- Verify all tests go green


pshapiro4broad (approved):
LGTM. Nice to see unused mocks being removed!

When changing over to strict stubs, I sometimes run into cases where not all tests need all mocks, but some tests need the same mocks. In those cases I usually refactor the mock out to a separate method, although another approach is to use a nested test class, which lets you share one `BeforeEach` with a subset of tests. Not something I saw a need for here, but it could come in handy in the future.  For example: https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java#L379

hams7504 (approved):
Looks good!

PR #2144, opened 2023-09-12
[DUOS-2641][risk=no] Bug fix and new fields for dataset index terms

rushtong:
### Addresses
Support for https://broadworkbench.atlassian.net/browse/DUOS-2641

### Summary
* Add the translated data use to the reducer similar to what we do in the mapper. This was missed on a previous PR.
* Add the create user id to the dataset term index
* Add the dac name to the term index
* Add dac approval to the term index
* Add dac name to the term index

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetReducer.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetReducer.java
@@ -29,6 +29,9 @@ public class DatasetReducer implements LinkedHashMapRowReducer<Integer, Dataset>
       dataset.setDataUse(
           DataUse.parseDataUse(rowView.getColumn("data_use", String.class)).orElse(null));
     }
+    if (hasColumn(rowView, "translated_data_use", String.class)) {
+      dataset.setTranslatedDataUse(rowView.getColumn("translated_data_use", String.class));
+    }
     if (hasColumn(rowView, "in_use", Integer.class)) {
       Integer dsIdInUse = rowView.getColumn("in_use", Integer.class);
       dataset.setDeletable(Objects.isNull(dsIdInUse));
--- a/src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetReducer.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetReducer.java
@@ -29,9 +29,8 @@ public class DatasetReducer implements LinkedHashMapRowReducer<Integer, Dataset>
       dataset.setDataUse(
           DataUse.parseDataUse(rowView.getColumn("data_use", String.class)).orElse(null));
     }
-    if (hasColumn(rowView, "translated_data_use", String.class)) {
-      dataset.setTranslatedDataUse(rowView.getColumn("translated_data_use", String.class));
-    }
+    hasOptionalColumn(rowView, "translated_data_use", String.class)
+      .ifPresent(dataset::setTranslatedDataUse);
     if (hasColumn(rowView, "in_use", Integer.class)) {
       Integer dsIdInUse = rowView.getColumn("in_use", Integer.class);
       dataset.setDeletable(Objects.isNull(dsIdInUse));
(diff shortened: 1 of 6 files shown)


src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetReducer.java
```diff
@@ -29,6 +29,9 @@ public void accumulate(Map<Integer, Dataset> map, RowView rowView) {
       dataset.setDataUse(
           DataUse.parseDataUse(rowView.getColumn("data_use", String.class)).orElse(null));
     }
+    if (hasColumn(rowView, "translated_data_use", String.class)) {
+      dataset.setTranslatedDataUse(rowView.getColumn("translated_data_use", String.class));
+    }
```
pshapiro4broad: On the whole I'm not a huge fan of `Optional<T>` but in this case I think it would be a good improvement. The code currently calls `getColumn()` twice and has to repeat itself when specifying the name and type of the column. Adding a `default` method to `RowMapperHelper` that returns an `Optional` would avoid both of these issues and then this code would look like:
```
    getColumn(rowView, "translated_data_use", String.class).ifPresent(dataset::setTranslatedDataUse);
```

This method should work:
```
  default <T> Optional<T> getColumn(RowView rowView, String columnName, Class<T> clazz) {
    try {
      return Optional.of(rowView.getColumn(columnName, clazz));
    } catch (Exception e) {
      log.debug("RowView does not contain column " + columnName);
      return Optional.empty();
    }
  }
```

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -205,13 +210,26 @@ public DatasetTerm toDatasetTerm(Dataset dataset) {
     DatasetTerm term = new DatasetTerm();
 
     term.setDatasetId(dataset.getDataSetId());
+    if (Objects.nonNull(dataset.getCreateUserId())) {
```
pshapiro4broad: This is the convention of this code, but I'm wondering why the code doesn't compare against `null` directly. To me using `==` is easier to read than a function call but that is a subjective difference.
```suggestion
    if (dataset.getCreateUserId() != null) {
```

Alternatively, you can avoid repeating the expression `dataset.getCreateUserId()` by using `Optional`. That would look like:
```
    Optional.ofNullable(dataset.getCreateUserId()).ifPresent(
        userId -> {
          User user = userDAO.findUserById(userId);
          term.setCreateUserId(dataset.getCreateUserId());
          term.setCreateUserDisplayName(user.getDisplayName());
        }
    );
```

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -205,13 +210,26 @@ public DatasetTerm toDatasetTerm(Dataset dataset) {
...
+      User user = userDAO.findUserById(dataset.getCreateUserId());
+      term.setCreateUserId(dataset.getCreateUserId());
+      term.setCreateUserDisplayName(user.getDisplayName());
```
pshapiro4broad, line 216: I'm guessing it's OK to do this because a user's display name can't change over time? Or is there a step that recreates the index when a field like this changes?

pshapiro4broad (approved):
Looks OK, I have a few comments about possible future improvements.

It looks like there could be more test coverage for the changed lines. The ES test only tests one of the new fields added, and I didn't see a test for the changes to `DatasetReducer`.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
rushtong: This code is essentially the place where we recreate the index. There are some hooks that call this when a DUOS dataset is changed or when a DAR is approved/denied - all of those hooks flow through this method. We don't allow for users to change their display name, however, so that's always going to be whatever it was when they originally signed in.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
rushtong: I do like the optional approach in this case, thanks for the tip!

src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetReducer.java
rushtong: Yes, I like that - will take another stab at it.

hams7504 (approved):
Looks great!

PR #2153, opened 2023-09-25
[DUOS-1684][risk=no] Migrate automation scala tests to java

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1684

### Summary
Migrate from scala automation to java based junit tests
* Remove all previous scala automation code
* Runs post merge to develop branch
* New scenarios for status, liveness, version in java
* New smoke test action that:
  * Spins up a custom BEE
  * Run JUnit tests against BEE
  * Report results to BQ/Grafana
  * Spin down BEE
  * Notify Slack

See also: https://github.com/broadinstitute/terraform-ap-deployments/pull/1211

This is the view in Grafana:
Screenshot 2023-09-29 at 8 57 23 AM

Slack Notification:
Screenshot 2023-09-29 at 10 37 57 AM

----

- Verify all tests go green


.github/workflows/smoke-tests.yaml
```diff
@@ -0,0 +1,49 @@
...
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      - name: Bee Create
+        uses: broadinstitute/workflow-dispatch@v3
+        with:
+          workflow: bee-create
+          repo: broadinstitute/terra-github-workflows
+          ref: refs/heads/main
+          token: ${{ secrets.BROADBOT_TOKEN}}
+          inputs: '{ "bee-name": "${{ github.event.repository.name }}-${{ github.run_id }}-dev", "bee-template-name": "duos", "version-template": "dev" }'
```
pshapiro4broad: I wonder if there's a way to create a named binding for an expression so you don't need to repeat it everywhere, like `"${{ github.event.repository.name }}-${{ github.run_id }}-dev"`. You could do it using a step or job output that just created the bee name, not sure how much extra code that would be though.

pshapiro4broad (commented):
FYI there is a https://github.com/DataBiosphere/terra-test-runner which does a similar thing to gatling. Although, gatling may be nicer/easier to use and I'm not sure how much QA is recommending new projects to use the test-runner. If you're interested, the terra template project shows one way to use it, see https://github.com/DataBiosphere/terra-java-project-template/tree/main/integration

.github/workflows/smoke-tests.yaml
rushtong: There definitely is ... just haven't gotten there yet. Essentially you set it as the output and it's then a reference-able var

rushtong:
> FYI there is a https://github.com/DataBiosphere/terra-test-runner which does a similar thing to gatling. Although, gatling may be nicer/easier to use and I'm not sure how much QA is recommending new projects to use the test-runner. If you're interested, the terra template project shows one way to use it, see https://github.com/DataBiosphere/terra-java-project-template/tree/main/integration

Thanks for the links! I'm not sure I want to add gradle to our tooling. The goal of this work is to remove another, separate, test-running structure which is gatling+scala+jenkins ... I'm really hoping to slim down the additional overhead.

.github/workflows/smoke-tests.yaml
```diff
@@ -0,0 +1,80 @@
...
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: 17
+      - uses: actions/cache@v3
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
```
fboulnois:
```suggestion
          java-version: 17
          cache: 'maven'
```

I don't think we need the `actions/cache` step at all, the `setup-java` step should already cache all Maven outputs if you use the `cache: 'maven'` key: https://github.com/actions/setup-java#caching-maven-dependencies

fboulnois (changes requested):
Code cleanup looks good, one suggestion for the Maven cache:

fboulnois (approved):
👍

hams7504 (approved):
looks great! 👍

PR #2159, opened 2023-10-04
[DUOS-2679][risk=no] Return a DS instance from Post

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2679

### Summary
This PR returns a full instance of the Dataset Registration Schema after submission
It includes builder classes to instantiate a schema from a provided study and optional number of datasets.

----

- Verify all tests go green


fboulnois (approved):
This code lgtm 👍 , the only comment I have (and that I've had before) is we need to manually ensure we have all the properties. It would be nice if we could do this automatically, but I don't know of a good way to do this without reflection.

rushtong:
> This code lgtm 👍 , the only comment I have (and that I've had before) is we need to manually ensure we have all the properties. It would be nice if we could do this automatically, but I don't know of a good way to do this without reflection.

I have the same concern, but also don't know of a better way to do this more cleanly.

hams7504 (approved):
👍

PR #2164, opened 2023-10-12
[DUOS-2707][risk=no] Show null values in registration

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2707

### Summary
Ensure that null values are returned in APIs that return a registration schema

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/util/gson/GsonUtil.java
+++ b/src/main/java/org/broadinstitute/consent/http/util/gson/GsonUtil.java
@@ -23,6 +23,10 @@ public class GsonUtil {
     return gsonBuilderWithAdapters().create();
   }
 
+  public static Gson buildGsonNullSerializer() {
+    return gsonBuilderWithAdapters().serializeNulls().create();
+  }
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/consent/http/util/gson/GsonUtil.java
```diff
@@ -23,6 +23,10 @@ public static Gson buildGson() {
     return gsonBuilderWithAdapters().create();
   }
 
+  public static Gson buildGsonNullSerializer() {
+    return gsonBuilderWithAdapters().serializeNulls().create();
```
fboulnois, line 27: Interesting, this is good to know.

fboulnois (approved):
lgtm 👍

hams7504 (approved):
lgtm 👍

src/main/java/org/broadinstitute/consent/http/util/gson/GsonUtil.java
rushtong: Our default setting is to exclude null values so this should be a relatively infrequent use case.

PR #2196, opened 2023-11-22
DUOS-2778 - Update recipient for 'data-custodian-approval'

hams7504:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2778

### Summary

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
@@ -498,7 +498,7 @@ public class Dataset {
     return List.of();
   }
 
-  public List<String> getDataDepositorEmails() {
+  public List<String> getDataDepositors() {
     if (!(this.getProperties() == null)) {
       return this.getProperties()
           .stream()
--- a/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Dataset.java
@@ -1,5 +1,7 @@
 package org.broadinstitute.consent.http.models;
 
+import com.google.gson.Gson;
+import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashSet;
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/consent/http/models/Dataset.java
```diff
@@ -471,4 +477,37 @@ public User getCreateUser() {
...
+            return stringList;
+          })
+          .flatMap(List::stream)
+          .distinct()
+          .toList();
+    }
+    return List.of();
+  }
+
+  public List<String> getDataDepositorEmails() {
```
rushtong: Technically, this is getting the string value of the `Data Depositor` field, which could be an email, or not. I suggest renaming it so it isn't confusing to the callers. Related to this, I think the email service handles the case where the string is NOT a valid email, but maybe an extra check in `VoteService` would prevent sending invalid data in the first place.

src/main/java/org/broadinstitute/consent/http/service/VoteService.java
```diff
@@ -362,22 +365,36 @@ protected void notifyCustodiansOfApprovedDatasets(List<Dataset> datasets, User r
...
         });
       }
+
+      // Data Depositors
+      List<String> depositors = d.getDataDepositors();
+      List<String> depositorEmails = new ArrayList<>();
+      String emailRegex = "^[a-zA-Z0-9_+&*-]+(?:\\."+
+          "[a-zA-Z0-9_+&*-]+)*@" +
+          "(?:[a-zA-Z0-9-]+\\.)+[a-z" +
+          "A-Z]{2,7}$";
```
rushtong: I am usually hesitant to implement our own regexes for this kind of thing. It's usually better to let a common library do this for you. I recommend using https://commons.apache.org/proper/commons-validator/apidocs/org/apache/commons/validator/routines/EmailValidator.html for this. You can find the library here: https://mvnrepository.com/artifact/commons-validator/commons-validator/1.7 and add it to the `pom.xml` file. Implementation code might look something like this:
```
EmailValidator validator = EmailValidator.getInstance();
List<String> validEmails = depositors.stream().filter(e -> validator.isValid(e)).toList();
```

rushtong (commented):
Code and tests look good - see suggestion inline.

rushtong (approved):
Thank you for the changes 👍🏽

src/main/java/org/broadinstitute/consent/http/service/VoteService.java
```diff
@@ -362,22 +366,29 @@ protected void notifyCustodiansOfApprovedDatasets(List<Dataset> datasets, User r
...
-          .filter(p ->
-              (Objects.nonNull(p.getSchemaProperty()) && p.getSchemaProperty()
-                  .equalsIgnoreCase("dataCustodianEmail")) ||
-                  p.getPropertyName().equalsIgnoreCase("Data Depositor"))
-          .map(DatasetProperty::getPropertyValueAsString)
-          .distinct()
-          .toList();
+
+      // Data Custodians
+      List<String> custodianEmails = d.getDataCustodianEmails();
```
fboulnois: Should we also validate the Data Custodian emails?

fboulnois (commented):
One question:

src/main/java/org/broadinstitute/consent/http/service/VoteService.java
hams7504: I'm not sure, but it might be good to add it just in case!

fboulnois (approved):
lgtm 👍

PR #2210, opened 2023-12-08
[DUOS-2783] Send emails to chairs when datasets are submitted to DAC

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-2783

### Summary

Send emails to chairs when datasets are submitted to DAC

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
@@ -21,6 +21,7 @@ import org.broadinstitute.consent.http.db.DatasetDAO;
 import org.broadinstitute.consent.http.db.StudyDAO;
 import org.broadinstitute.consent.http.enumeration.FileCategory;
 import org.broadinstitute.consent.http.enumeration.PropertyType;
+import org.broadinstitute.consent.http.models.Dac;
 import org.broadinstitute.consent.http.models.DataUse;
 import org.broadinstitute.consent.http.models.Dataset;
 import org.broadinstitute.consent.http.models.DatasetProperty;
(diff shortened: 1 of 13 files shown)


src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
```diff
@@ -187,6 +190,19 @@ public List<Dataset> createDatasetsFromRegistration(
             datasetInserts);
 
     List<Dataset> datasets = datasetDAO.findDatasetsByIdList(createdDatasetIds);
+    try {
```
fboulnois: Not sure if this is the best approach or what the best way to test this is.

rushtong: There are a couple ways to test. If your local email config is set to not send email (which should be the case), then no email will be sent BUT the content should be saved to the email table. There is also the `test-sendgrid-html.sh` script that might be helpful to see the final output.

fboulnois: What I am not sure about:

1. Whether this is the best approach to notify DAC chairs that a new dataset has been submitted
2. How to best add unit tests for this specific code chunk

All the other pieces are tested (including sending the template with the appropriate details)

rushtong: To answer specifically:

1. Looking more closely at this, I think it needs to go down one level into `DatasetServiceDAO.insertDatasetRegistration`. I say that because there are two code paths for dataset creation - one of them in this code path where we are creating a brand new dataset registration. The other happens during the Study update process where the user could be adding new consent groups (datasets) to a study. Both processes use `DatasetServiceDAO.insertDatasetRegistration`. The alternative is to add this block to both the create and update code blocks which might be a good idea if ... 
2. I would suggest pulling this out into a separate method so it can be tested independently of dataset creation. If this block were outside of this larger method, you can easily mock the dao and service calls and ensure that `sendDatasetSubmittedMessage` is called an appropriate number of times given the dataset and dac chairs. If it is a separate method, it can be called conveniently from both the create and the update code paths.

rushtong (approved):
Thank you for the changes 👍🏽

hams7504 (approved):
Looks good to me 👍

PR #2215, opened 2023-12-15
[DUOS-2781][DUOS-2782] DAC rejects + accepts datasets

hams7504:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2781
https://broadworkbench.atlassian.net/browse/DUOS-2782

### Summary
- Changed email content for the case where a DAC accepts or rejects a dataset

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/mail/freemarker/DatasetApprovedModel.java
+++ b/src/main/java/org/broadinstitute/consent/http/mail/freemarker/DatasetApprovedModel.java
@@ -23,4 +23,5 @@ public class DatasetApprovedModel {
   public String getDacName() {
     return dacName;
   }
+
 }
--- a/src/main/java/org/broadinstitute/consent/http/mail/freemarker/DatasetDeniedModel.java
+++ b/src/main/java/org/broadinstitute/consent/http/mail/freemarker/DatasetDeniedModel.java
@@ -5,11 +5,13 @@ public class DatasetDeniedModel {
   private String dataSubmitterName;
   private String datasetName;
   private String dacName;
+  private String dacEmail;
 
-  public DatasetDeniedModel(String dataSubmitterName, String datasetName, String dacName) {
+  public DatasetDeniedModel(String dataSubmitterName, String datasetName, String dacName, String dacEmail) {
     this.dataSubmitterName = dataSubmitterName;
     this.datasetName = datasetName;
     this.dacName = dacName;
+    this.dacEmail = dacEmail;
   }
 
   public String getDataSubmitterName() {
@@ -23,4 +25,9 @@ public class DatasetDeniedModel {
   public String getDacName() {
     return dacName;
   }
+
+  public String getDacEmail() {
+    return dacEmail;
+  }
+
 }
--- a/src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
@@ -39,11 +39,12 @@ import org.broadinstitute.consent.http.models.dataset_registration_v1.ConsentGro
 import org.broadinstitute.consent.http.models.dto.DatasetDTO;
 import org.broadinstitute.consent.http.models.dto.DatasetPropertyDTO;
 import org.broadinstitute.consent.http.service.dao.DatasetServiceDAO;
+import org.broadinstitute.consent.http.util.ConsentLogger;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 
-public class DatasetService {
+public class DatasetService implements ConsentLogger {
 
   private final Logger logger = LoggerFactory.getLogger(this.getClass());
   public static final String DATASET_NAME_KEY = "Dataset Name";
@@ -345,10 +346,17 @@ public class DatasetService {
           dac.getName(),
           dataset.getDatasetIdentifier());
     } else {
-      emailService.sendDatasetDeniedMessage(
-          user,
-          dac.getName(),
-          dataset.getDatasetIdentifier());
+      if (!Objects.isNull(dac.getEmail())) {
+        String dacEmail = dac.getEmail();
+        emailService.sendDatasetDeniedMessage(
+            user,
+            dac.getName(),
+            dataset.getDatasetIdentifier(),
+            dacEmail);
+      }
+      else {
+        logWarn("Unable to send dataset denied email to DAC: " + dac.getDacId());
+      }
     }
 
   }
--- a/src/main/resources/freemarker/dataset-approved.html
+++ b/src/main/resources/freemarker/dataset-approved.html
@@ -22,7 +22,7 @@
         <tr>
             <td style="padding: 15px 15px 0px; font-family: 'Montserrat', sans-serif; font-size: 16px; color: #1F3B50; text-align: justify; line-height: 25px;">
                 <p id="content" style="margin: 0;">
-                    Thank you for using DUOS to register your dataset(s). We’re happy to inform you that ${dacName} has accepted your dataset. The ${datasetName} dataset should now be visible in the DUOS dataset catalog. Researchers will now be able to submit data access requests through DUOS for access to your dataset.
+                  Your dataset, ${datasetName}, submitted to the ${dacName} by ${dataSubmitterName} for management of future data access requests has been <b>accepted</b> and can be found in the DUOS Data Library with DUOS ID: ${datasetName}
                 </p>
                 <p>
                     Please reach out to <a href="mailto:duos-support@broadinstitute.zendesk.com" style="text-decoration: none; font-family: 'Montserrat', sans-serif; color: #00609F;">duos-support@broadinstitute.zendesk.com</a> if you have any questions or concerns.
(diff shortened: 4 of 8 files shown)


rushtong (approved):
👍🏽

src/main/java/org/broadinstitute/consent/http/mail/freemarker/DatasetDeniedModel.java
```diff
@@ -5,11 +5,13 @@ public class DatasetDeniedModel {
   private String dataSubmitterName;
   private String datasetName;
   private String dacName;
+  private String dacEmail;
```
pshapiro4broad, line 8: It looks like these fields can be declared as `final`, since they're not modified once the object is constructed. If they were, then this could become a `record`:
```
public record DatasetDeniedModel(String dataSubmitterName, String datasetName, String dacName) {
}
```

src/main/java/org/broadinstitute/consent/http/mail/freemarker/DatasetApprovedModel.java
```diff
@@ -23,4 +23,5 @@ public String getDatasetName() {
   public String getDacName() {
     return dacName;
   }
+
```
rjohanek, line 26: nit: rm line unless this is intentional

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -345,10 +346,17 @@ private void sendDatasetApprovalNotificationEmail(Dataset dataset, User user, Bo
...
+      if (!Objects.isNull(dac.getEmail())) {
+        String dacEmail = dac.getEmail();
+        emailService.sendDatasetDeniedMessage(
+            user,
+            dac.getName(),
+            dataset.getDatasetIdentifier(),
+            dacEmail);
+      }
+      else {
+        logWarn("Unable to send dataset denied email to DAC: " + dac.getDacId());
```
rjohanek, line 358: I'm not sure, but would it be helpful to log the email?

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -345,10 +346,17 @@ private void sendDatasetApprovalNotificationEmail(Dataset dataset, User user, Bo
           dac.getName(),
           dataset.getDatasetIdentifier());
     } else {
-      emailService.sendDatasetDeniedMessage(
-          user,
-          dac.getName(),
-          dataset.getDatasetIdentifier());
+      if (!Objects.isNull(dac.getEmail())) {
```
pshapiro4broad, line 349: It looks this use of `Objects.isNull()` is common in this codebase, and it works, but it wasn't intended for this use. It was intended to be used as a predicate, such as for a streaming operation. See https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Objects.html#isNull(java.lang.Object). Personally I would compare against `null` directly instead:

```suggestion
      if (dac.getEmail() != null) {
```

If you do want to use an `Objects` method, I would use `nonNull()` instead as then you don't have to negate the result:

```suggestion
      if (Objects.nonNull(dac.getEmail())) {
```

src/test/java/org/broadinstitute/consent/http/service/DatasetServiceTest.java
```diff
@@ -767,10 +768,46 @@ void testApproveDataset_DenyDataset() throws Exception {
...
+    dac.setName("DAC NAME");
+    initService();
+    when(dacDAO.findById(3)).thenReturn(dac);
+
+    Dataset returnedDataset = datasetService.approveDataset(dataset, user, payloadBool);
+    assertEquals(dataset.getDataSetId(), returnedDataset.getDataSetId());
+    assertFalse(returnedDataset.getDacApproval());
+
+    // do not send denied email
+    verify(emailService, times(0)).sendDatasetDeniedMessage(
```
rjohanek, line 802: Is there any email validation besides the isNull check? Is the expected behavior of this test the logged warning?

rjohanek (approved):
Looks good! I have a question about the test but not a blocker

src/main/resources/freemarker/dataset-approved.html
```diff
@@ -22,7 +22,7 @@
         <tr>
             <td style="padding: 15px 15px 0px; font-family: 'Montserrat', sans-serif; font-size: 16px; color: #1F3B50; text-align: justify; line-height: 25px;">
                 <p id="content" style="margin: 0;">
-                    Thank you for using DUOS to register your dataset(s). We’re happy to inform you that ${dacName} has accepted your dataset. The ${datasetName} dataset should now be visible in the DUOS dataset catalog. Researchers will now be able to submit data access requests through DUOS for access to your dataset.
+                  Your dataset, ${datasetName}, submitted to the ${dacName} by ${dataSubmitterName} for management of future data access requests has been <b>accepted</b> and can be found in the DUOS Data Library with DUOS ID: ${datasetName}
```
pshapiro4broad, line 25: I think the phrase `submitted to the ${dacName} by ${dataSubmitterName} for management of future data access requests` is an appositive clause so it also should have commas around it, in addition to `datasetName` since it's also an appositive for `Your dataset`. Also maybe it should be split into two sentences with a period at the end.
```suggestion
                  Your dataset, ${datasetName}, submitted to the ${dacName} by ${dataSubmitterName} for management of future data access requests, has been <b>accepted</b>. It can be found in the DUOS Data Library with DUOS ID: ${datasetName}.
```

As I understand it, the rationale for separating the appositive clause like this is to make it easer for a reader to find the verb and direct object of a sentence.

src/test/java/org/broadinstitute/consent/http/service/DatasetServiceTest.java
```diff
@@ -767,10 +768,46 @@ void testApproveDataset_DenyDataset() throws Exception {
...
   }
 
+  @Test
+  void testApproveDataset_DenyDataset_WithNoDACEmail() throws Exception {
+    Dataset dataset = new Dataset();
+    dataset.setDataSetId(1);
+    User user = new User();
+    user.setUserId(1);
+    user.setEmail("asdf@gmail.com");
+    user.setDisplayName("John Doe");
```
pshapiro4broad, line 783: `User` has a constructor that can be used here
```suggestion
    User user = new User(1, "asdf@gmail.com, "John Doe", null);
```

src/test/java/org/broadinstitute/consent/http/service/DatasetServiceTest.java
pshapiro4broad, line 802: A nice alias for `times(0)` is `never()`
```suggestion
    verify(emailService, never()).sendDatasetDeniedMessage(
```

pshapiro4broad: I agree, I was expecting to see a test that would cause `emailService.sendDatasetDeniedMessage()` to be called as that's the new code here.

hams7504: Yes, in this test, we expect the error message to be logged. However, I don't think we verify logged messages in our other tests.

pshapiro4broad (commented):
Just a few comments, looks OK overall

src/main/java/org/broadinstitute/consent/http/mail/freemarker/DatasetDeniedModel.java
rushtong: I like the idea of moving to record classes, this is a pretty old part of the code base.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
rushtong: Thanks for the link to the doc, don't think I've ever read that. I've been suggesting the `Objects` methods for some time now.

src/main/resources/freemarker/dataset-approved.html
rushtong: The language is specified by product - I'd prefer to leave this decision to @solideoglori for language choice in these cases.

src/test/java/org/broadinstitute/consent/http/service/DatasetServiceTest.java
rushtong: Nice

rushtong: > I agree, I was expecting to see a test that would cause `emailService.sendDatasetDeniedMessage()` to be called as that's the new code here.

The positive test already [exists in the test previous](https://github.com/DataBiosphere/consent/blob/8c6af8372e0ba4c37b673478c077db31572451b7/src/test/java/org/broadinstitute/consent/http/service/DatasetServiceTest.java#L742) to this one.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
pshapiro4broad: If you'd rather not compare against `null` directly, a lot of people like using `Optional.ofNullable()` instead. So you'd write something like:
```
     Optional.ofNullable(dac.getEmail()).ifPresentOrElse(dacEmail ->
        emailService.sendDatasetDeniedMessage(
            user,
            dac.getName(),
            dataset.getDatasetIdentifier(),
            dacEmail)).orElse(),
        () -> logWarn("Unable to send dataset denied email to DAC: " + dac.getDacId());
```

PR #2220, opened 2023-12-21
[DUOS-2856][risk=no] List Dataset Performance Update

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2856

### Summary
This PR updates the performance of parsing `DataUse` objects when retrieving lists of `Datasets`. We do two things to improve the bottleneck seen here. 
1. Use Gson to do the parsing instead of Jackson's object mapper
2. Use a local cache to avoid reprocessing the same string values into `DataUse` objects

Example dev vs local comparison (using DAC datasets call for DAC 3) is  71512 ms -> 23390 ms which is roughly a 3-fold improvement. More work should still be done to improve performance, however, since a 23 second response is still fairly slow for the volume of data we're processing.

#### Before
Screenshot 2023-12-21 at 3 15 28 PM

#### After
Screenshot 2023-12-21 at 3 32 10 PM

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/mapper/DacWithDatasetsReducer.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/mapper/DacWithDatasetsReducer.java
@@ -4,19 +4,15 @@ import java.sql.Date;
 import java.sql.Timestamp;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
 import org.broadinstitute.consent.http.models.Dac;
-import org.broadinstitute.consent.http.models.DataUse;
 import org.broadinstitute.consent.http.models.Dataset;
+import org.broadinstitute.consent.http.util.ConsentLogger;
 import org.jdbi.v3.core.mapper.MappingException;
 import org.jdbi.v3.core.result.LinkedHashMapRowReducer;
 import org.jdbi.v3.core.result.RowView;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
-public class DacWithDatasetsReducer implements LinkedHashMapRowReducer<Integer, Dac> {
-
-  private final Logger logger = LoggerFactory.getLogger(this.getClass());
+public class DacWithDatasetsReducer implements LinkedHashMapRowReducer<Integer, Dac>,
+    ConsentLogger {
 
   private final DataUseParser dataUseParser = new DataUseParser();
 
@@ -34,23 +30,22 @@ public class DacWithDatasetsReducer implements LinkedHashMapRowReducer<Integer, 
 
         try {
           //aliased columns must be set directly
-
-          if (Objects.nonNull(rowView.getColumn("dataset_alias", String.class))) {
-            String dsAlias = rowView.getColumn("dataset_alias", String.class);
+          String dsAlias = rowView.getColumn("dataset_alias", String.class);
+          if (dsAlias != null) {
             try {
               dataset.setAlias(Integer.parseInt(dsAlias));
             } catch (Exception e) {
-              logger.error("Exception parsing dataset alias: " + dsAlias, e);
+              logException("Exception parsing dataset alias: " + dsAlias, e);
             }
           }
 
-          if (Objects.nonNull(rowView.getColumn("dataset_create_date", Date.class))) {
-            Date createDate = rowView.getColumn("dataset_create_date", Date.class);
+          Date createDate = rowView.getColumn("dataset_create_date", Date.class);
+          if (createDate != null) {
             dataset.setCreateDate(createDate);
           }
 
-          if (Objects.nonNull(rowView.getColumn("dataset_update_date", Timestamp.class))) {
-            Timestamp updateDate = rowView.getColumn("dataset_update_date", Timestamp.class);
+          Timestamp updateDate = rowView.getColumn("dataset_update_date", Timestamp.class);
+          if (updateDate != null) {
             dataset.setUpdateDate(updateDate);
           }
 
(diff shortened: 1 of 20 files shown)


fboulnois (approved):
nice work 👍

src/main/java/org/broadinstitute/consent/http/db/mapper/DataUseParser.java
```diff
@@ -0,0 +1,27 @@
...
+  public DataUse parseDataUse(String dataUseString) {
+    dataUseCache.computeIfAbsent(dataUseString, s -> {
+      try {
+        return gson.fromJson(dataUseString, DataUse.class);
+      } catch (Exception e) {
+        logWarn(String.format("Unable to parse data use string: '%s'", dataUseString));
+      }
+      return null;
+    });
+    return dataUseCache.get(dataUseString);
```
pshapiro4broad: `computeIfAbsent()` returns the value so you don't need the cal to `get()` afterwards
```suggestion
    return dataUseCache.computeIfAbsent(dataUseString, s -> {
      try {
        return gson.fromJson(dataUseString, DataUse.class);
      } catch (Exception e) {
        logWarn(String.format("Unable to parse data use string: '%s'", dataUseString));
      }
      return null;
    });
```

src/main/java/org/broadinstitute/consent/http/db/mapper/DataUseParser.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import org.broadinstitute.consent.http.models.DataUse;
+import org.broadinstitute.consent.http.util.ConsentLogger;
+import org.broadinstitute.consent.http.util.gson.GsonUtil;
+
+public class DataUseParser implements ConsentLogger {
+
+  private final Gson gson = GsonUtil.gsonBuilderWithAdapters().create();
+  private final Map<String, DataUse> dataUseCache = new HashMap<>();
```
pshapiro4broad, line 13: I assume that the lifetime of a jdbi row reducer is only during a single request. If so, then this only improves the cases where the same data use expression occurs in different rows in the result, correct? Vs making the cache global across all operations. That would require using a concurrent map though so maybe it wouldn't be much of an improvement.

src/main/java/org/broadinstitute/consent/http/db/mapper/DacWithDatasetsReducer.java
```diff
@@ -58,7 +60,7 @@ public void accumulate(Map<Integer, Dac> container, RowView rowView) {
 
         if (Objects.nonNull(rowView.getColumn("dataset_data_use", String.class))) {
           String duStr = rowView.getColumn("dataset_data_use", String.class);
-          Optional<DataUse> du = DataUse.parseDataUse(duStr);
+          Optional<DataUse> du = Optional.ofNullable(dataUseParser.parseDataUse(duStr));
           du.ifPresent(dataset::setDataUse);
```
pshapiro4broad: Is this check necessary? if the result is `null`, what is the harm of calling `setDataUse`?
```
          dataset.setDataUse(dataUseParser.parseDataUse(duStr));
```

src/main/java/org/broadinstitute/consent/http/db/mapper/DacWithDatasetsReducer.java
```diff
@@ -58,7 +60,7 @@ public void accumulate(Map<Integer, Dac> container, RowView rowView) {
 
         if (Objects.nonNull(rowView.getColumn("dataset_data_use", String.class))) {
           String duStr = rowView.getColumn("dataset_data_use", String.class);
```
pshapiro4broad: Note that this form makes two calls to `rowView.getColumn()`. Storing the result of the first call would avoid this extra fetch.
```suggestion
        String duStr = rowView.getColumn("dataset_data_use", String.class);
        if (duStr != null) {
```

pshapiro4broad (approved):
Looks OK, I had a few minor comments. I've heard that Jackson is slower than gson but hadn't seen any benchmarks about it; did you do a perf test with Jackson vs gson separate from the cache change? I'm wondering if cached Jackson is that much slower than cached gson.

rushtong:
> I've heard that Jackson is slower than gson but hadn't seen any benchmarks about it; did you do a perf test with Jackson vs gson separate from the cache change? I'm wondering if cached Jackson is that much slower than cached gson.

I didn't do a performance check with Jackson vs Gson but I've seen similar reviews. We originally used Jackson in quite a few parts of consent/ontology codebases and started moving to Gson quite some time ago. I don't have any desire to refactor everything back to Jackson hence the re-write here.

src/main/java/org/broadinstitute/consent/http/db/mapper/DataUseParser.java
rushtong: Thank you, nice suggestion

src/main/java/org/broadinstitute/consent/http/db/mapper/DataUseParser.java
rushtong: Yes, mappers/reducers are instantiated per query, so their lifetime is short as is the cache - which I'm fine with for now. We certainly could make this a larger cache refactor but as you mention, we'd need to measure the tradeoffs. What we find in practice is that a lot of datasets have the exact same data use so I think the bulk of the performance gain comes from that fact.

src/main/java/org/broadinstitute/consent/http/db/mapper/DacWithDatasetsReducer.java
rushtong: Great point, I think this was an artifact of the original return signature which was an optional - definitely not necessary any longer.

PR #2228, opened 2024-01-10
DUOS-2200 implemented oidc endpoints for b2c

dvoet:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2200
https://broadworkbench.atlassian.net/browse/DUOS-2463 (most of it anyway)

### Summary
implemented
* `/oauth2/authorize`
* `/oauth2/token`
* `/oauth2/configuration`

adds appropriate section to api-docs.yaml

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/SwaggerResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/SwaggerResource.java
@@ -133,12 +133,12 @@ public class SwaggerResource {
           });
           ui.initOAuth({
             clientId: "OAUTH_CLIENT_ID",
-            clientSecret: "",
             realm: "Broad Institute",
             appName: "Consent",
             scopeSeparator: " ",
-            scopes: "profile email",
-            additionalQueryStringParams: {}
+            scopes: "openid profile email",
+            additionalQueryStringParams: {},
+            usePkceWithAuthorizationCodeGrant: true
           });
         };
         """;
--- a/src/test/java/org/broadinstitute/consent/http/db/OidcAuthorityDAOTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/db/OidcAuthorityDAOTest.java
@@ -14,19 +14,17 @@ import org.apache.http.message.BasicNameValuePair;
 import org.broadinstitute.consent.http.WithMockServer;
 import org.broadinstitute.consent.http.configurations.OidcConfiguration;
 import org.broadinstitute.consent.http.configurations.ServicesConfiguration;
-import org.broadinstitute.consent.http.db.OidcAuthorityDAO;
 import org.broadinstitute.consent.http.models.OidcAuthorityConfiguration;
 import org.broadinstitute.consent.http.util.HttpClientUtil;
 import org.jetbrains.annotations.NotNull;
 import org.junit.jupiter.api.AfterAll;
-import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.mockserver.client.MockServerClient;
 import org.testcontainers.containers.MockServerContainer;
 
-public class OidcAuthorityDAOTest implements WithMockServer {
+class OidcAuthorityDAOTest implements WithMockServer {
 
   private OidcAuthorityDAO dao;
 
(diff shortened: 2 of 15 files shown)


src/main/java/org/broadinstitute/consent/http/resources/SwaggerResource.java
```diff
@@ -133,12 +133,12 @@ private String getInitializer() {
           });
           ui.initOAuth({
             clientId: "OAUTH_CLIENT_ID",
```
dvoet, line 135: the `clientId` here is going to be a problem. Out of the box, it looks like the swagger ui supports only 1 client id but we have 2. TDR and others have an [index.html](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/resources/templates/index.html) that adds `insertClientIdsPlugin` but consent has a later version of swagger ui and I did not explore how it might work.

src/test/java/org/broadinstitute/consent/http/db/OidcAuthorityDAOTest.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.http.message.BasicNameValuePair;
+import org.broadinstitute.consent.http.WithMockServer;
+import org.broadinstitute.consent.http.configurations.OidcConfiguration;
+import org.broadinstitute.consent.http.configurations.ServicesConfiguration;
+import org.broadinstitute.consent.http.db.OidcAuthorityDAO;
+import org.broadinstitute.consent.http.models.OidcAuthorityConfiguration;
+import org.broadinstitute.consent.http.util.HttpClientUtil;
+import org.jetbrains.annotations.NotNull;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Assertions;
```
rushtong: nit: unused

src/test/java/org/broadinstitute/consent/http/db/OidcAuthorityDAOTest.java
```diff
@@ -0,0 +1,113 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockserver.client.MockServerClient;
+import org.testcontainers.containers.MockServerContainer;
+
+public class OidcAuthorityDAOTest implements WithMockServer {
```
rushtong: nit: public modifiers for junit5 classes and tests are unnecessary

rushtong: Additionally, if we annotate this class with
```
```
then we don't need to call ` openMocks(this);` in the init method. We've been slowly migrating to this approach as we hit test classes. The primary benefit here is that a test will fail if it includes unnecessary mocking which helps keep tests cleaner.

src/test/java/org/broadinstitute/consent/http/db/OidcAuthorityDAOTest.java
```diff
@@ -0,0 +1,113 @@
...
+                .withHeader("Content-Type", "application/json")
+                .withBody(
+                    "{\"issuer\":\"" + expectedIssuer + "\",\"authorization_endpoint\":\""
+                        + expectedAuthorizationEndpoint
+                        + "\",\"token_endpoint\":\"" + expectedTokenEndpoint
+                        + "\",\"userinfo_endpoint\":\"" + expectedIssuer
+                        + "/oauth2/userinfo\",\"revocation_endpoint\":\"https://example.com/oauth2/revoke\",\"jwks_uri\":\"https://example.com/oauth2/keys\",\"response_types_supported\":[\"code\",\"token\",\"id_token\",\"code token\",\"code id_token\",\"token id_token\",\"code token id_token\",\"none\"],\"subject_types_supported\":[\"public\"],\"id_token_signing_alg_values_supported\":[\"RS256\"],\"scopes_supported\":[\"openid\",\"profile\",\"email\",\"address\",\"phone\"],\"token_endpoint_auth_methods_supported\":[\"client_secret_basic\",\"client_secret_post\"],\"claims_supported\":[\"sub\",\"iss\",\"email\",\"email_verified\",\"phone_number\",\"phone_number_verified\",\"address\",\"name\",\"client_id\"],\"code_challenge_methods_supported\":[\"plain\",\"S256\"]}"));
```
rushtong: nit: This is tough to read - I like text blocks with `String.format` for this kind of thing, something like
```
    String body = """
        {
          "issuer":"%s",
          "authorization_endpoint":"%s",
          "token_endpoint":"%s",
          "userinfo_endpoint":"%s/oauth2/userinfo",
          "revocation_endpoint":"https://example.com/oauth2/revoke",
          "jwks_uri":"https://example.com/oauth2/keys",
          "response_types_supported":["code","token","id_token","code token","code id_token","token id_token","code token id_token","none"],
          "subject_types_supported":["public"],
          "id_token_signing_alg_values_supported":["RS256"],
          "scopes_supported":["openid","profile","email","address","phone"],
          "token_endpoint_auth_methods_supported":["client_secret_basic","client_secret_post"],
          "claims_supported":["sub","iss","email","email_verified","phone_number","phone_number_verified","address","name","client_id"],
          "code_challenge_methods_supported":["plain","S256"]
        }
        """;
```
and then here on this line:
```
.withBody(String.format(body, expectedIssuer, expectedAuthorizationEndpoint, expectedTokenEndpoint, expectedIssuer)));
```

src/main/java/org/broadinstitute/consent/http/resources/SwaggerResource.java
rushtong: We'll file a ticket to address this in a separate PR - it shouldn't be too involved to migrate to our own index like TDR does.

rushtong (approved):
This looks great, thank you! Some minor nits inline - feel free to address as you see fit. https://github.com/broadinstitute/terra-helmfile/pull/4939 has been merged so there should be no trouble when this hits dev.

nmalfroy (approved):
looks great!  thanks for taking this on :)

PR #2237, opened 2024-01-26
DT-2991: GA4GH Passport User Info API

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-2991

## Summary
This PR represents our first effort to generate a compliant [GA4GH passport](https://ga4gh.github.io/data-security/ga4gh-passport) implementation. This PR implements the following features:

* Authenticated userinfo endpoint that provides a JSON Passport. This follows the implementation requirements specified in the [AAI OpenID spec](https://ga4gh.github.io/data-security/aai-openid-connect-profile#visas-provided-by-a-broker-via-userinfo-endpoint)
* Passport consisting a list of visas:
  * `ControlledAccessGrants` Visa per dataset the user has current access to.
  * `AffiliationAndRole` Visa describing the user's affiliation
  * `ResearcherStatus` Visa describing the user's status

## What this PR does NOT cover
This work should be slated for future implementation efforts.
* Provide a [passport scoped access token](https://ga4gh.github.io/data-security/aai-openid-connect-profile#passport-scoped-access-token)
  * Requires an additional scope that we do not currently support: `ga4gh_passport_v1`
* Provide a signed JWT 
  * Requires public and private keys
* Complete the currently unimplemented claim types:
  * AcceptedTermsAndPolicies
  * LinkedIdentities
* ResearcherStatus: We need to provide a [URL Claim to describe a researcher](https://github.com/ga4gh-duri/ga4gh-duri.github.io/blob/master/researcher_ids/ga4gh_passport_v1.md#researcherstatus)

----

- Verify all tests go green


rushtong:
@dependabot show ignore conditions

src/main/java/org/broadinstitute/consent/http/ConsentApplication.java
```diff
@@ -179,6 +180,7 @@ public void run(ConsentConfiguration config, Environment env) {
     env.jersey().register(injector.getInstance(NihAccountResource.class));
     env.jersey().register(injector.getInstance(OAuth2Resource.class));
     env.jersey().register(injector.getInstance(OntologyResource.class));
+    env.jersey().register(injector.getInstance(PassportResource.class));
```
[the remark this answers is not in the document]
rushtong: This is a common pattern we use, but it is incorrect that `PassportResource` fails. `PassportService` does indeed instantiate cleanly with the provided `@Inject` annotations.

src/test/java/org/broadinstitute/consent/http/resources/PassportResourceTest.java
```diff
@@ -0,0 +1,88 @@
...
+  void testGetPassportSuccess() {
+    User user = createUser();
+    authUser.setEmail(user.getEmail());
+    UserStatusInfo userStatusInfo = createUserStatusInfo(user);
+    DuosUser duosUser = new DuosUser(authUser, user);
+    duosUser.setUserStatusInfo(userStatusInfo);
+
+    PassportResource resource = new PassportResource(passportService);
+    Response response = resource.getPassport(duosUser);
+    assertEquals(Response.Status.OK.getStatusCode(), response.getStatus());
```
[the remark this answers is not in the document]
rushtong: PassportResource does not need to know about validity of the entity it returns, it only cares that it has one or fails. Other tests of the service class ensure entity validity.

src/main/java/org/broadinstitute/consent/http/service/passport/AffiliationAndRole.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  @Override
+  public Long asserted() {
+    var assertedDate =
+        Optional.ofNullable(user.getLibraryCard())
+            .map(LibraryCard::getCreateDate)
+            .orElse(user.getCreateDate());
+    // GA4GH requires Unix timestamps in seconds; Date#getTime() returns milliseconds.
+    return assertedDate.getTime() / 1000L;
```
otchet-broad: What's the tolerance for being off by a second?  Technically speaking we could be rounding up.  If it's inconsequential, I'm not concerned.  If it is, we might want to convert these Date objects to something like Instant and use the built in functions to convert to epoch seconds.

src/main/java/org/broadinstitute/consent/http/service/passport/ControlledAccessGrants.java
```diff
@@ -0,0 +1,55 @@
...
+    this.approvedDataset = approvedDataset;
+  }
+
+  @Override
+  public String type() {
+    return VisaClaimTypes.CONTROLLED_ACCESS_GRANTS.type;
+  }
+
+  @Override
+  public Long asserted() {
```
otchet-broad, line 25: What does this date represent?  How will a consumer use it?

src/main/java/org/broadinstitute/consent/http/service/passport/PassportService.java
```diff
@@ -0,0 +1,83 @@
...
+    List<Visa> allVisas =
+        Stream.of(grantVisas, List.of(roleVisa), List.of(researcherVisa))
+            .flatMap(List::stream)
+            .toList();
+    return new PassportClaim(allVisas);
+  }
+
+  protected List<Visa> buildControlledAccessGrants(
+      String userSubjectId, List<ApprovedDataset> approvedDatasets) {
+    return approvedDatasets.stream()
```
otchet-broad, line 62: Do we need to be more selective about the filter here?  Don't we want the ApprovedDataset with the longest runway?

src/main/java/org/broadinstitute/consent/http/service/passport/AffiliationAndRole.java
rushtong: We don't have a hard requirement for that level of precision. I would be amenable to a more centralized function to ensure we're always following the same pattern.

src/main/java/org/broadinstitute/consent/http/service/passport/ControlledAccessGrants.java
rushtong: This represents the seconds since the visa assertion is made: https://ga4gh.github.io/data-security/ga4gh-passport#visa-object-claims

src/main/java/org/broadinstitute/consent/http/service/passport/PassportService.java
rushtong: That's a good point. I should sort them by the expiration and then filter out duplicates so we get the ones that extend the furthest into the future.

src/main/resources/assets/schemas/Visa.yaml
```diff
@@ -0,0 +1,17 @@
...
+    description: The Issuer, part of the [Visa Identity](https://github.com/ga4gh-duri/ga4gh-duri.github.io/blob/master/researcher_ids/ga4gh_passport_v1.md#visa-identity)
+  sub:
+    type: string
+    description: The Subject, part of the [Visa Identity](https://github.com/ga4gh-duri/ga4gh-duri.github.io/blob/master/researcher_ids/ga4gh_passport_v1.md#visa-identity)
+  iat:
+    type: number
+    description: The Issued-At timestamp
+  exp:
+    type: number
+    description: The Expiration timestamp
```
otchet-broad, line 15: We might want to be clearer that making an access decision after the exp is reached is unsupported.

otchet-broad:
As I'm coming up to speed on the GA4GH Passport spec, I'm wondering if we should be creating durable logs for every passport we issue so that we can trace (with high fidelity) all of the clients that have requested these objects and know their expiration so that we can:
1) Inform overseers and data custodians about potential usage in other systems.
2) Maintain a clear audit trail for issuance.
3) Monitor for abuse.

src/main/resources/assets/schemas/Visa.yaml
rushtong: `ControlledAccessGrants` is just one type of visa which is what that language would be applicable to. I think it would be confusing with respect to the other visa types that exist, such as `ResearcherStatus` or `AffiliationAndRole`.

rushtong:
> As I'm coming up to speed on the GA4GH Passport spec, I'm wondering if we should be creating durable logs for every passport we issue so that we can trace (with high fidelity) all of the clients that have requested these objects and know their expiration so that we can:
> 
> 1. Inform overseers and data custodians about potential usage in other systems.
> 2. Maintain a clear audit trail for issuance.
> 3. Monitor for abuse.

Absolutely, these are great features. There are a number of other things we need to do to be fully compliant as noted in the PR description so I intend that this be iterated upon over time. Currently, there are no clients that will be using this in the near (1-3 month at a minimum) future.

src/main/java/org/broadinstitute/consent/http/service/passport/PassportService.java
rushtong: After digging into this some more, it doesn't matter which of the `ApprovedDataset` entities we return, only that we return one of them. IOW, if the user has access to two `ApprovedDataset` objects, one that expires tomorrow, one that expires next year, they have the same underlying dataset identifier.

src/main/resources/assets/schemas/Visa.yaml
rushtong: > `ControlledAccessGrants` is just one type of visa which is what that language would be applicable to. I think it would be confusing with respect to the other visa types that exist, such as `ResearcherStatus` or `AffiliationAndRole`.

I have to correct myself. `exp` applies equally to all Visas, and in the case of `ControlledAccessGrants` it does NOT refer to the expiration of the user's access to the dataset. It applies only to the state of the Visa such that the visa will always expire 1 hour from issuance.

fboulnois (approved):
As a first approach, this looks reasonable! 👍

otchet-broad:
Thanks for the harmonization around dates.  I think that's going to be helpful in the future.

I'm wondering if the resource should be updated so that it's not PermitAll, but rather Admin and maybe Service Account for now.  I think there are Expiration and Revocation cases that are important from a compliance standpoint that are missing from this that make me concerned we'll communicate a controlled access grant without having these other mechanisms in place.  [https://github.com/ga4gh-duri/ga4gh-duri.github.io/blob/master/researcher_ids/ga4gh_passport_v1.md#overview](https://github.com/ga4gh-duri/ga4gh-duri.github.io/blob/master/researcher_ids/ga4gh_passport_v1.md#overview)

Would that be an okay next step on this?

rushtong:
> Thanks for the harmonization around dates. I think that's going to be helpful in the future.
> 
> I'm wondering if the resource should be updated so that it's not PermitAll, but rather Admin and maybe Service Account for now. I think there are Expiration and Revocation cases that are important from a compliance standpoint that are missing from this that make me concerned we'll communicate a controlled access grant without having these other mechanisms in place. https://github.com/ga4gh-duri/ga4gh-duri.github.io/blob/master/researcher_ids/ga4gh_passport_v1.md#overview
> 
> Would that be an okay next step on this?

Sure thing. I'd prefer to limit it to Admin for now rather than SA since they are not an intended audience, general researchers are the current target. For comparison, this API is doing essentially the same thing that https://consent.dsde-prod.broadinstitute.org/#/User/get_api_user_me_researcher_datasets does ... retrieving the same entity for an authenticated user. That API is also `@PermitAll`.

src/main/java/org/broadinstitute/consent/http/service/passport/PassportService.java
otchet-broad: It's true that they'll have the same identifier, but I'm thinking about how we communicate **when** access to that dataset is no longer authorized.  I haven't gotten through enough of the spec yet to see how that unwinds (or if it's even supported).  What is clear to me is that the expiration (`exp`) of the Visa shouldn't be past when the **we** would remove access (when access is planned to expire or the user's LC is revoked).  Fun stuff we'll be able to work through in future PRs.  :-)

otchet-broad (approved):
👍
Nice work!

src/main/java/org/broadinstitute/consent/http/service/passport/PassportService.java
rushtong: > What is clear to me is that the expiration (exp) of the Visa shouldn't be past when the we would remove access (when access is planned to expire or the user's LC is revoked) 

Of these two conditions, we have control over the planned expiration, but not future LC revocation. I can see improvements on the expiration front at least.

PR #2259, opened 2024-03-13
[DUOS-1479][DUOS-1480][risk=no] Remove connect data owner functionality

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1480
https://broadworkbench.atlassian.net/browse/DUOS-1479

### Summary
* Remove previously deprecated case code
* Deprecate/remove data owner functionality
* See front end changes here: https://github.com/DataBiosphere/duos-ui/pull/2485

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
aarohinadkarni, line 1: nit: The removal of Data Owner as role 6 leaves the roles as 1, 2, 3, 4, 5, 7, 8, 9 with no role designated as role 6

aarohinadkarni (approved):
looks good with one very minor, optional suggestion 👍

src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
rushtong: Yes, unfortunately, that's required since the role id is stored in the database. I didn't want to make any db changes for this in the event that we need to find those values for some reason, but I'll file a new ticket to make additional db changes as fallout from this work.

fboulnois: I'm not sure we should move the role numbers around, it sounds like a lot of additional complexity. Instead perhaps we can mark it as deprecated or something similar in the database.

fboulnois (approved):
looks good 👍

src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
rushtong: I'm definitely not moving role ids around - I'm leaving everything in place as it is now. We don't have a concept of deprecated roles currently - I'll put in a ticket for that.

PR #2261, opened 2024-03-14
[DUOS-2910] Update Library Card model

aarohinadkarni:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2910

### Summary
Updates our existing Library Card model to include associations between Library Cards & Data Access Agreements. 

Note: changes in this PR include changes from the DAA Model PR. The DAA Model  PR should be merged before this one.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
@@ -81,11 +81,11 @@ public interface LibraryCardDAO extends Transactional<LibraryCardDAO> {
   @RegisterBeanMapper(value = LibraryCard.class)
   @UseRowReducer(LibraryCardReducer.class)
   @SqlQuery("""
-      SELECT library_Card.*,
+      SELECT library_card.*,
       ld.daa_id
-      FROM library_Card
-      LEFT JOIN lc_daa ld ON library_Card.id = ld.lc_id
-      WHERE library_Card.institution_id = :institutionId
+      FROM library_card
+      LEFT JOIN lc_daa ld ON library_card.id = ld.lc_id
+      WHERE library_card.institution_id = :institutionId
       """)
   List<LibraryCard> findLibraryCardsByInstitutionId(@Bind("institutionId") Integer institutionId);
 
--- a/src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
@@ -3,6 +3,7 @@ package org.broadinstitute.consent.http.db;
 import java.util.Date;
 import java.util.List;
 import org.broadinstitute.consent.http.db.mapper.LibraryCardReducer;
+import org.broadinstitute.consent.http.models.DataAccessAgreement;
 import org.broadinstitute.consent.http.models.Institution;
 import org.broadinstitute.consent.http.models.LibraryCard;
 import org.jdbi.v3.sqlobject.config.RegisterBeanMapper;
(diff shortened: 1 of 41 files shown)


src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
```diff
@@ -70,12 +71,17 @@ void updateLibraryCardById(@Bind("libraryCardId") Integer libraryCardId,
       " FROM library_card AS lc " +
       " LEFT JOIN institution " +
       " ON lc.institution_id = institution.institution_id" +
+      " LEFT JOIN lc_daa ld ON lc.id = ld.lc_id " +
+      " LEFT JOIN data_access_agreement daa ON daa.daa_id = ld.daa_id " +
```
rushtong: This correctly adds a join, but we also need to add `daa_id` to the list of select options so that the reducer class can populate them. Similar feedback in the other list queries.

src/main/java/org/broadinstitute/consent/http/models/LibraryCard.java
```diff
@@ -132,4 +138,26 @@ public boolean equals(Object libraryCard) {
...
+        .stream()
+        .noneMatch(d -> d.equals(daaId))) {
+      this.daaIds.add(daaId);
+    }
+  }
+
+  public void removeDaa(Integer daaId) {
+    if (this.daaIds == null) {
+      return; //not sure if this is appropriate but we should not be removing anything if nothing exists
+    }
```
rushtong: Correct. We can also ignore the empty case.

src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
```diff
@@ -66,16 +67,22 @@ void updateLibraryCardById(@Bind("libraryCardId") Integer libraryCardId,
...
   List<LibraryCard> findLibraryCardsByUserId(@Bind("userId") Integer userId);
 
   @RegisterBeanMapper(value = LibraryCard.class)
   @UseRowReducer(LibraryCardReducer.class)
-  @SqlQuery("SELECT * FROM library_Card WHERE institution_id = :institutionId")
+  @SqlQuery("SELECT library_Card.*, " +
+      "ld.daa_id " +
+      "FROM library_Card " +
+      "LEFT JOIN lc_daa ld ON library_Card.id = ld.lc_id " +
+      "WHERE library_Card.institution_id = :institutionId")
```
rushtong: We should make these text blocks instead of concatenated strings. Semgrep sometimes complains about this, and text blocks are easier to read, copy, paste, etc.

rushtong (commented):
See comment inline, otherwise 👍🏽

src/test/java/org/broadinstitute/consent/http/db/LibraryCardDAOTest.java
```diff
@@ -189,6 +193,54 @@ public void testDeleteLibraryCardByUserId() {
     assertNull(libraryCardDAO.findLibraryCardById(card.getId()));
   }
 
+  @Test
+  public void testCreateLibraryCardDaaAssociation() {
```
rushtong: The following tests don't have any assertions in them, we should likely try to check on what was created/deleted

rushtong: Secondarily, we should try to clean up our test classes for JUnit5 best practices. That would mean removing the public modifiers from the class and all test methods, as well as adding `@ExtendWith(MockitoExtension.class)` as a class annotation. Those details help us keep tests cleaner and more efficient.

src/test/java/org/broadinstitute/consent/http/service/LibraryCardServiceTest.java
```diff
@@ -350,6 +350,25 @@ public void testFindLibraryCardById() {
     assertEquals(result.getId(), libraryCard.getId());
   }
 
+  @Test
+  public void testAddDaaToLibraryCard() {
```
rushtong, line 355: Similar to above tests, these don't have any assertions. For the most part, these are just pass-throughs to the underlying query so we don't need very much, perhaps just a simple assertion:
```
assertDoesNotThrow(() -> service.addDaaToLibraryCard(libraryCard.getId(), 1));
```

rushtong (commented):
This looks pretty good, just some test cleanup that I think we can add.

rushtong (approved):
Thank you for the changes! 👍🏽

src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
```diff
@@ -58,42 +59,55 @@ void updateLibraryCardById(@Bind("libraryCardId") Integer libraryCardId,
...
   @RegisterBeanMapper(value = LibraryCard.class)
   @UseRowReducer(LibraryCardReducer.class)
-  @SqlQuery("SELECT * FROM library_Card WHERE institution_id = :institutionId")
+  @SqlQuery("""
+      SELECT library_Card.*,
+      ld.daa_id
+      FROM library_Card
+      LEFT JOIN lc_daa ld ON library_Card.id = ld.lc_id
+      WHERE library_Card.institution_id = :institutionId
+      """)
```
fboulnois:
```suggestion
  @SqlQuery("""
      SELECT library_card.*,
      ld.daa_id
      FROM library_card
      LEFT JOIN lc_daa ld ON library_card.id = ld.lc_id
      WHERE library_card.institution_id = :institutionId
      """)
```

nit: casing

src/main/java/org/broadinstitute/consent/http/models/LibraryCard.java
```diff
@@ -1,6 +1,8 @@
 package org.broadinstitute.consent.http.models;
```
fboulnois, line 1: I think it's worth testing this class in more detail, but we can push that into a separate PR.

fboulnois (approved):
looks good, one minor nit:

PR #2263, opened 2024-03-18
[DUOS-2949] Get by ID DAA API

aarohinadkarni:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2949

### Summary
Adds a new endpoint to get a Data Access Agreement by its ID along with API documentation

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DaaService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DaaService.java
@@ -1,0 +1,34 @@
+package org.broadinstitute.consent.http.service;
+
+import com.google.inject.Inject;
+import org.broadinstitute.consent.http.db.DaaDAO;
+import org.broadinstitute.consent.http.models.DataAccessAgreement;
+import org.broadinstitute.consent.http.models.FileStorageObject;
+import org.broadinstitute.consent.http.service.dao.DaaServiceDAO;
+
+public class DaaService {
+
+  private final DaaServiceDAO daaServiceDAO;
+  private final DaaDAO daaDAO;
+
+  @Inject
+  public DaaService(DaaServiceDAO daaServiceDAO, DaaDAO daaDAO) {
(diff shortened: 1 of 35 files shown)


src/main/resources/assets/api-docs.yaml
```diff
@@ -122,6 +122,8 @@ paths:
       responses:
         404:
           description: Deprecated endpoint. Use  `GET /api/tdr/{identifier}` instead.
+  /api/daa/{daaId}:
+    $ref: 'paths/daaByDaaId.yaml'
   /api/daa/{dacId}:
     $ref: 'paths/daaById.yaml'
```
rushtong: We can't actually do this ... we might need a refactor of `/api/daa/{dacId}:` to be something like `/api/daa/dac/{dacId}:` or something similar. The two paths (as defined) have the same signature (`/api/daa/<integer>`) so Jersey cannot distinguish between the two.

aarohinadkarni: For now, I've refactored the POST api signature.

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
```diff
@@ -84,4 +86,11 @@ public void removeDacFromDaa(Integer dacId, Integer daaId) {
     daaDAO.deleteDacDaaRelation(dacId, daaId);
   }
 
+  public DataAccessAgreement findById(Integer daaId) {
+    DataAccessAgreement daa = daaDAO.findById(daaId);
+    if (Objects.nonNull(daa)) {
```
rushtong: nit:
```suggestion
    if (daa != null) {
```

rushtong (approved):
One minor suggestion inline, but 👍🏽 overall!

fboulnois (approved):
looks good 👍

PR #2266, opened 2024-03-19
[DUOS-2919] Get DAA API

aarohinadkarni:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2919

### Summary
Adds a new endpoint to get all Data Access Agreements along with API documentation

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -77,4 +78,15 @@ public Response createDaaForDac(
     }
   }
 
+  @GET
+  @Produces(MediaType.APPLICATION_JSON)
+  @RolesAllowed({ADMIN, MEMBER, CHAIRPERSON, RESEARCHER})
```
rushtong: This could be `@PermitAll` - I don't see any reason to restrict the access here. SO's will definitely need to see them as well.

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
```diff
@@ -84,4 +86,11 @@ public void removeDacFromDaa(Integer dacId, Integer daaId) {
     daaDAO.deleteDacDaaRelation(dacId, daaId);
   }
 
+  public List<DataAccessAgreement> findAll() {
+    List<DataAccessAgreement> daas = daaDAO.findAll();
+    if (Objects.nonNull(daas)) {
```
rushtong: nit
```suggestion
    if (daas != null) {
```

src/main/resources/assets/api-docs.yaml
```diff
@@ -108,6 +108,8 @@ paths:
           description: Deprecated endpoint. Use  `GET /api/tdr/{identifier}` instead.
   /api/daa/{dacId}:
     $ref: 'paths/daaById.yaml'
+  /api/daa:
+    $ref: 'paths/findAllDaa.yaml'
```
rushtong: nit: naming. The path yaml should reflect the path, not necessarily the operation. For example, we could call GET or DELETE (a bad idea in this case, but a potential) and they would both be defined in the same endpoint, just different methods. Another more concrete example for `/api/daa/{dacId}` ... we could possibly call GET, PUT, DELETE on that endpoint so `daaById.yaml` would contain docs for all of those and therefore the naming makes more sense.

Perhaps `findAllDaa.yaml` -> `daa.yaml`

src/main/resources/assets/paths/findAllDaa.yaml
```diff
@@ -0,0 +1,19 @@
...
+    - DAA
+  responses:
+    200:
+      description: Fetched all Data Access Agreements
+      content:
+        application/json:
+          schema:
+            $ref: '../schemas/DataAccessAgreement.yaml'
+    403:
+      description: User not authorized to request all DAAs.
```
rushtong: I don't think we need this.

src/main/resources/assets/paths/findAllDaa.yaml
```diff
@@ -0,0 +1,19 @@
...
+    404:
+      description: DAA Not Found
```
rushtong: I don't think this is true ... if there are no DAAs, we return an empty list.

rushtong (commented):
See comments inline - I think it can use just a little bit of cleanup.

rushtong (approved):
Thank you for the changes! 👍🏽

fboulnois (approved):
lgtm, and tests cover what I would expect 👍

PR #2269, opened 2024-03-22
[DUOS-2953] PUT Library Card DAA Relationship API DAA

aarohinadkarni:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2953

### Summary
Adds a new endpoint to create a relationship between a user's Library Card & a DAA along with API documentation. 

**Note**: This PR depends on https://github.com/DataBiosphere/consent/pull/2275.

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
```diff
@@ -53,7 +53,13 @@ void updateLibraryCardById(@Bind("libraryCardId") Integer libraryCardId,
 
   @RegisterBeanMapper(value = LibraryCard.class)
   @UseRowReducer(LibraryCardReducer.class)
-  @SqlQuery("SELECT * FROM library_card WHERE id = :libraryCardId")
+  @SqlQuery("""
+      SELECT lc.*,
+      ld.daa_id
+      FROM library_card AS lc
+      LEFT JOIN lc_daa ld ON lc.id = ld.lc_id
+      WHERE id = :libraryCardId
```
rushtong: Should we have the alias here to be clear?
```suggestion
      WHERE lc.id = :libraryCardId
```

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -78,6 +85,39 @@ public Response createDaaForDac(
...
+      DataAccessAgreement daa = daaService.findById(daaId);
+      User user = userService.findUserByEmail(authUser.getEmail());
+      int userInstitutionId = user.getInstitutionId();
+      LibraryCard libraryCard = libraryCardService.findLibraryCardById(libraryCardId);
+      int libraryCardInstitutionId = libraryCard.getInstitutionId();
+      // Assert that the user has the correct institution permissions to add a DAA-LC relationship.
+      // Admins can add a DAA with any DAC, but signing officials can only create relationships for
+      // library cards associated with the same institution they are associated with.
+      if (!user.hasUserRole(UserRoles.ADMIN)) {
+        if (!(userInstitutionId == libraryCardInstitutionId)) {
```
rushtong: nit: simplify
```suggestion
        if (userInstitutionId != libraryCardInstitutionId) {
```

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -78,6 +85,39 @@ public Response createDaaForDac(
...
+          return Response.status(Status.FORBIDDEN).build();
+        }
+      }
+      libraryCardService.addDaaToLibraryCard(libraryCardId, daaId);
+      URI uri = info.getBaseUriBuilder()
+          .replacePath("api/libraryCards/{libraryCardId}")
+          .build(libraryCardId);
+      return Response.created(uri).entity(libraryCard).build();
```
rushtong: I don't think this is the right response, we didn't create a library card, we just associated an existing one to an existing DAA. I think this could be simplified to an OK response. The location header is fine, and in fact, we should do better about that in our endpoints.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -78,6 +85,39 @@ public Response createDaaForDac(
     }
   }
 
+  @POST
```
rushtong: The more I think about this, the more I think it should be a `PUT`, not a `POST`. We're not creating any new objects here, just modifying a relationship between two existing ones.

rushtong (commented):
See comments inline - minor updates requested. The one question I have is whether we should instead do a PUT vs POST - let's discuss in standup.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -78,6 +85,39 @@ public Response createDaaForDac(
...
+      @Context UriInfo info,
+      @Auth AuthUser authUser,
+      @PathParam("daaId") Integer daaId,
+      @PathParam("libraryCardId") Integer libraryCardId) {
+    try {
+      DataAccessAgreement daa = daaService.findById(daaId);
+      User user = userService.findUserByEmail(authUser.getEmail());
+      int userInstitutionId = user.getInstitutionId();
+      LibraryCard libraryCard = libraryCardService.findLibraryCardById(libraryCardId);
+      int libraryCardInstitutionId = libraryCard.getInstitutionId();
```
fboulnois: Do the user validation first, otherwise we are doing unnecessary (and potentially unsafe) operations beforehand.

src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
fboulnois: Also, should this be a separate method? I imagine there are cases where we want the daa and the library cards, and other times where we just want the library cards.

src/test/java/org/broadinstitute/consent/http/resources/DaaResourceTest.java
```diff
@@ -164,9 +169,110 @@ void testFindDaaByDaaId() {
...
+    when(userService.findUserByEmail(any())).thenReturn(admin);
+    when(libraryCardService.findLibraryCardById(any())).thenReturn(lc);
+
+
+    resource = new DaaResource(daaService, dacService, userService, libraryCardService);
+    Response response = resource.createLibraryCardDaaRelation(info, authUser, daa.getDaaId(),  lc.getId());
+    assert response.getStatus() == HttpStatus.SC_FORBIDDEN;
+  }
+
+  @Test
```
fboulnois, line 301: Full coverage of all the lines would be beneficial, for example with an invalid or null user and perhaps other test cases.

fboulnois (changes requested):
See below:

src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
rushtong: The user stories we are trying to support here are:
1. Researcher sees their LCs with DAA associations on their profile page
2. Signing Officials see a list of users along with which DAAs they are associated to (see table mocks) so the SO can select/unselect a user's DAA association
3. Researcher sees their DAA associations related to the datasets they have selected on a data access request (which comes from their LC). 
4. Researcher can request a DAA assignment from their SO via the profile page.

In all of these cases, we do need the DAA ids in some way. We don't always need the fleshed out DAA entity on the library card, but sometimes we will, so I think we do need the above method as it is, and we probably need another that fleshes out the actual DAAs on the library card.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -78,6 +85,39 @@ public Response createDaaForDac(
...
 
+  @POST
+  @Produces(MediaType.APPLICATION_JSON)
+  @RolesAllowed({ADMIN, SIGNINGOFFICIAL})
+  @Path("{daaId}/{libraryCardId}")
+  public Response createLibraryCardDaaRelation(
+      @Context UriInfo info,
+      @Auth AuthUser authUser,
+      @PathParam("daaId") Integer daaId,
+      @PathParam("libraryCardId") Integer libraryCardId) {
```
rushtong: We should change this to be the researcher's user id instead of the LC id.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
rushtong: Instead of comparing the LC institution id, we should compare the researcher's institution id and the authed user's institution id in the case that the authed user is a Signing Official and not an Admin.

src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
aarohinadkarni: I've created a [ticket](https://broadworkbench.atlassian.net/browse/DUOS-3030) for a separate method when we need all details of the DAAs on the library card.

src/main/resources/assets/api-docs.yaml
```diff
@@ -108,6 +108,8 @@ paths:
           description: Deprecated endpoint. Use  `GET /api/tdr/{identifier}` instead.
   /api/daa/{daaId}:
     $ref: 'paths/daaByDaaId.yaml'
+  /api/daa/{daaId}/{userId}:
+    $ref: 'paths/daaForUserId.yaml'
```
rushtong, line 112: This is duplicated on line 117 - we only need one. The swagger page will show these kinds of issues pretty quickly ... 
Screenshot 2024-04-02 at 8 26 29 AM

rushtong:
We should add to the PR description that this PR depends on https://github.com/DataBiosphere/consent/pull/2275

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -83,6 +87,46 @@ public Response createDaaForDac(
...
+          return Response.status(Status.FORBIDDEN).build();
+        }
+      }
+      List<LibraryCard> libraryCards = libraryCardService.findLibraryCardsByUserId(userId);
+      Optional<LibraryCard> matchingCard = libraryCards.stream()
+          .filter(card -> card.getInstitutionId() == authedUser.getInstitutionId().intValue())
+          .findFirst();
+      if (matchingCard.isEmpty()) {
+        return Response.status(Status.NOT_FOUND).build();
+      }
```
rushtong, line 118: Thinking out loud here ... one thing we could do here is to push this logic to the service class and in the case that there is no LC, we could create one. However... that means the OK response should really be a Created response ... so I don't think we should go down that route - let's keep the APIs distinct.

rushtong (commented):
Once https://github.com/DataBiosphere/consent/pull/2275 and the api doc fix is in place, this looks good!

rushtong (approved):
Thanks for the changes! 👍🏽

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -83,6 +87,46 @@ public Response createDaaForDac(
...
+      User user = userService.findUserById(userId);
+      int userInstitutionId = user.getInstitutionId();
+      // Assert that the user has the correct institution permissions to add a DAA-LC relationship.
+      // Admins can add a DAA with any DAC, but signing officials can only create relationships for
+      // library cards associated with the same institution they are associated with.
+      if (!authedUser.hasUserRole(UserRoles.ADMIN)) {
+        if (authedUserInstitutionId != userInstitutionId) {
+          return Response.status(Status.FORBIDDEN).build();
+        }
+      }
```
fboulnois: I think this might be more clearly expressed as:

```suggestion
      if (authedUser.hasUserRole(UserRoles.SIGNINGOFFICIAL) && authedUserInstitutionId != userInstitutionId) {
          return Response.status(Status.FORBIDDEN).build();
      }
```

The non-admin role could be researcher etc, but the endpoint is limited to admins and signing officials.

fboulnois (approved):
Another suggestion:

PR #2278, opened 2024-04-03
[DUOS-2835][DUOS-2878][risk=no] Use the url dataset property instead of the dbGap version

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2835
https://broadworkbench.atlassian.net/browse/DUOS-2878

## Summary
Use the url dataset property of instead of the dbGap version. The url is meant to replace what we previously stored in dbGap.

See also this UI PR: https://github.com/DataBiosphere/duos-ui/pull/2521
This PR is not blocked on ☝🏽 - both can be moved forward independently.
Production dataset->study conversions necessary for this PR have been completed.

----

- Verify all tests go green


src/main/resources/changesets/changelog-consent-2024-04-02-rm-old-dbgap-url.xml
```diff
@@ -0,0 +1,18 @@
...
+      -- Remove duplicate properties
+      DELETE
+      FROM dataset_property
+      WHERE property_key IN (SELECT key_id FROM dictionary d WHERE KEY = 'dbGap')
+    </sql>
+    <sql>
+      -- Remove deprecated property key from dictionary
+      DELETE
+      FROM dictionary
+      WHERE key = 'dbGap';
```
fboulnois: Do we have datasets that ONLY have dbgap and not url? If so, we should also ensure these are converted over.

fboulnois (approved):
Overall looks good, one question:

src/main/resources/changesets/changelog-consent-2024-04-02-rm-old-dbgap-url.xml
rushtong: On dev, we do. On Prod, we have 3 that need to be migrated. I'll add that detail to the ticket. We cannot move over the old dbgap values - they are not necessarily urls - they were generic strings that didn't go through any validation.

rushtong: I'll also add that we are not using this value anywhere in the UI any longer either - it's essentially dead data.

src/main/resources/changesets/changelog-consent-2024-04-02-rm-old-dbgap-url.xml
```diff
@@ -0,0 +1,18 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+  <changeSet id="changelog-consent-2024-03-28-dataset-dac-id-props.xml" author="grushton">
+    <sql>
+      -- Remove duplicate properties
+      DELETE
+      FROM dataset_property
+      WHERE property_key IN (SELECT key_id FROM dictionary d WHERE KEY = 'dbGap')
```
aarohinadkarni: Looking at what's actually in the database, does this actually need to be the following?
```suggestion
      WHERE property_key IN (SELECT key_id FROM dictionary d WHERE KEY = 'dbGAP')
```

src/main/resources/changesets/changelog-consent-2024-04-02-rm-old-dbgap-url.xml
aarohinadkarni: Same here:
```suggestion
      WHERE key = 'dbGAP';
```

src/main/resources/changesets/changelog-consent-2024-04-02-rm-old-dbgap-url.xml
rushtong: Nice catch, thanks for that!

aarohinadkarni (approved):
thanks for the change! 👍

PR #2283, opened 2024-04-05
[DUOS-2952] POST DAA request API

aarohinadkarni:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2952

### Summary
Adds a new endpoint to request a relationship between a user's Library Card & a DAA along with API documentation.

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/mail/message/DaaRequestMessage.java
```diff
@@ -0,0 +1,21 @@
...
+import com.sendgrid.helpers.mail.Mail;
+import java.io.Writer;
+import javax.mail.MessagingException;
+
+public class DaaRequestMessage extends MailMessage {
+  private final String NEW_DAA_LIBRARY_CARD_REQUEST = "New DAA-Library Card Relationship Request in DUOS";
+
+  public Mail newDaaRequestMessage(String toAddress, String fromAddress, Writer template)
+      throws MessagingException {
+    // not sure if referenceId should be null or not; could pass in the DAA ID?
```
rushtong: Yes, let's make this the DAA ID

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -194,4 +198,19 @@ public Response deleteDaaForUser(
...
+
+  @POST
+  @PermitAll
+  @Path("/request/{daaId}")
+  public Response sendDaaRequestMessage(
+      @Auth AuthUser authUser,
+      @PathParam("daaId") Integer daaId) {
+    try {
+      User user = userService.findUserByEmail(authUser.getEmail());
+      daaService.sendDaaRequestEmails(user, daaId);
```
rushtong: I think we have some room here to do some additional validation. For example:
1. We might want to know that the user doesn't have an SO to send emails to and return with an error.
2. We might want to check to see if they already have the association - therefore we don't need to send an email
3. I'd love to have a way to prevent spamming here ... probably best to file an additional ticket for that with some potential implementation ideas.

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
```diff
@@ -102,6 +109,30 @@ public DataAccessAgreement findById(Integer daaId) {
     throw new NotFoundException("Could not find DAA with the provided ID: " + daaId);
   }
 
+  public void sendDaaRequestEmails(User user, Integer daaId) throws Exception {
+    try {
+      List<Institution> institutions = institutionService.findAllInstitutions();
+      Institution institution = institutions.stream()
+            .filter(inst -> inst.getId().equals(user.getInstitutionId()))
+          .findFirst()
+          .orElseThrow();
```
rushtong: If we are adding the institution to the user, do we need to do this additional query/filter? We should have that info handy ... and if not, we can throw an error here.

src/test/java/org/broadinstitute/consent/http/mail/freemarker/FreeMarkerTemplateHelperTest.java
```diff
@@ -175,6 +175,35 @@ void testGetDatasetSubmittedTemplate() throws Exception {
     assertFalse(templateString.contains("${"));
   }
 
+  @Test
+  void testGetDaaRequestTemplate() throws Exception {
+    Writer template = helper.getDaaRequestTemplate("signingOfficialUserName", "userName",
+        "daaName",
+        "serverUrl");
```
rushtong: One thing I like to do that ensures that you're testing what you think you're testing is to abstract these values into random string variables, then checking to see if those variables exist as opposed to testing the same string value all the time.

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
aarohinadkarni: When we add the institution to the user, all of the SOs aren't getting populated on the institution. Those SO values are only getting populated in the InstitutionService findAllInstitutions method— I chose to do this additional query because to get the SOs on an institution we are doing a JOIN operation between the institution and user tables which I didn't think made sense to do in findUserByEmail UserDAO method we were working in.

src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
```diff
@@ -91,14 +91,17 @@ public interface UserDAO extends Transactional<UserDAO> {
 
   @RegisterBeanMapper(value = User.class, prefix = "u")
   @RegisterBeanMapper(value = UserRole.class, prefix = "ur")
+  @RegisterBeanMapper(value = Institution.class, prefix = "i")
   @UseRowReducer(UserWithRolesReducer.class)
   @SqlQuery("SELECT "
```
okotsopoulos, line 96: [FYI]

Sharing something that we've really liked to use in TDR lately for defining SQL strings: Java now supports text blocks for defining multi-line strings. https://www.baeldung.com/java-multiline-string

src/main/java/org/broadinstitute/consent/http/mail/freemarker/FreeMarkerTemplateHelper.java
```diff
@@ -208,4 +215,18 @@ private Writer generateDisabledDatasetsTemplate(String user, List<String> datase
...
   }
 
+  private Writer generateNewDAARequestTemplate(
+      String signingOfficialUserName,
+      String userName,
+      String daaName,
+      String serverUrl,
+      Template temp
+  ) throws IOException, TemplateException {
+    NewDaaRequestModel model = new NewDaaRequestModel(serverUrl, daaName, userName,
```
okotsopoulos, line 225: [General Consent question]

Just curious, does this repository have linting enabled (e.g. via Spotless)?  I ask because I've seen some formatting inconsistencies that Spotless may have picked up if it were running.

src/main/java/org/broadinstitute/consent/http/mail/freemarker/NewDaaRequestModel.java
```diff
@@ -0,0 +1,42 @@
...
+    return daaName;
+  }
+
+  public String getUserName() {
+    return userName;
+  }
+
+  public String getSigningOfficialUserName() {
+    return signingOfficialUserName;
+  }
```
okotsopoulos, line 35: [FYI]

I'm not sure if they're used in Consent, but this model looks like a good use case for definition via a Java record:

```suggestion
public record NewDaaRequestModel(String serverUrl, String daaName, String userName, String signingOfficialUserName) {}
```

src/main/java/org/broadinstitute/consent/http/mail/freemarker/NewDaaRequestModel.java
```diff
@@ -0,0 +1,42 @@
+package org.broadinstitute.consent.http.mail.freemarker;
+
+import java.util.List;
+import java.util.Map;
+import org.broadinstitute.consent.http.models.Dac;
+import org.broadinstitute.consent.http.models.Dataset;
```
okotsopoulos: These imports appear unused (I think this would have also been caught by a linter).

src/main/java/org/broadinstitute/consent/http/models/LibraryCard.java
```diff
@@ -31,7 +31,7 @@ public class LibraryCard {
 
   private Institution institution;
 
-  private List<Integer> daaIds;
+  private List<Integer> daaIds = new ArrayList<>();
```
okotsopoulos: I'd recommend initializing this instance variable in the constructor instead, like `createDate` is doing.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -194,4 +198,23 @@ public Response deleteDaaForUser(
...
+      User user = userService.findUserByEmail(authUser.getEmail());
+      if (user.getLibraryCards().stream()
+          .anyMatch(libraryCard -> libraryCard.getDaaIds().contains(daaId))) {
+        throw new IllegalArgumentException("User already has this DAA associated with their Library Card");
+      }
+      daaService.sendDaaRequestEmails(user, daaId);
+      return Response.ok().build();
+    } catch (Exception e) {
+      return createExceptionResponse(e);
+    }
```
okotsopoulos, line 223: [General Consent question]

Do you know if Consent has a global exception handler, something like [this](https://github.com/DataBiosphere/jade-data-repo/blob/800d4a58827cab52fdcd210260896be9266d0526/src/main/java/bio/terra/app/controller/GlobalExceptionHandler.java)?

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
```diff
@@ -102,6 +109,33 @@ public DataAccessAgreement findById(Integer daaId) {
     throw new NotFoundException("Could not find DAA with the provided ID: " + daaId);
   }
 
+  public void sendDaaRequestEmails(User user, Integer daaId) throws Exception {
+    try {
+      List<Institution> institutions = institutionService.findAllInstitutions();
+      Institution institution = institutions.stream()
+            .filter(inst -> inst.getId().equals(user.getInstitutionId()))
```
okotsopoulos: Could `inst.getId()` ever be null?  If so, this would yield an NPE, similar to @rushtong 's recent PR.  Is there a reason not to compare with `==`?

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
```diff
@@ -102,6 +109,33 @@ public DataAccessAgreement findById(Integer daaId) {
...
+          .orElseThrow();
+      List<SimplifiedUser> signingOfficials = institution.getSigningOfficials();
+      if (signingOfficials.isEmpty()) {
+        throw new NotFoundException("No signing officials found for user: " + user.getDisplayName());
+      }
+      int userId = user.getUserId();
+      String userName = user.getDisplayName();
+      for (SimplifiedUser signingOfficial : signingOfficials) {
+        String signingOfficialName = signingOfficial.displayName;
+        String signingOfficialEmail = signingOfficial.email;
```
okotsopoulos: It's up to you, but since these are referenced only once and are descriptive without being too verbose, I'd pass them directly to the method.

src/main/resources/assets/paths/daaRequestByDaaId.yaml
```diff
@@ -0,0 +1,20 @@
+post:
+  summary: Request that a Signing Official create a DAA-Library Card Relationship
+  operationId: Request a DAA-Library Card Relationship
```
okotsopoulos: Can you help direct me to the definition of this operation in your PR? I'm having a hard time finding it.

(I am also used to seeing `operationIds` matching up to the method name of an API controller, so no spaces and not necessarily matching the description.)

src/test/java/org/broadinstitute/consent/http/service/DaaServiceTest.java
```diff
@@ -103,6 +117,91 @@ void testRemoveDacFromDaa() {
...
+    when(institution.getSigningOfficials()).thenReturn(List.of());
+
+    when(institutionService.findAllInstitutions()).thenReturn(
+        Collections.singletonList(institution));
+
+    initService();
+
+    DaaService daaSpy = spy(service);
+    assertThrows(NotFoundException.class, () -> daaSpy.sendDaaRequestEmails(user, 1));
+    verify(daaSpy, times(1)).sendDaaRequestEmails(any(), any());
```
okotsopoulos: When I find myself using Spys in tests, my spidey senses tingle 😄 they often point to a need to reorganize my code, or they are not needed.

In this case and throughout these new tests, I don't think you need to Spy the `service`: you are explicitly calling the method that you are verifying, so I don't think the verification gets you anything extra.

okotsopoulos (commented):
Thanks for letting me review your PR, Aarohi!

My comments include some general FYIs and some high-level questions about Consent since I'm new to the repo.  I tried to tag those accordingly: please do not let them block your merge.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
rushtong: We don't have a global handler, but all api-level response errors are handled in `createExceptionResponse`: https://broadworkbench.atlassian.net/browse/DUOS-2981

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
rushtong: The id in this case is a database id, it should never be null here. If it is, there is a deeper problem we need to address

src/main/java/org/broadinstitute/consent/http/mail/freemarker/FreeMarkerTemplateHelper.java
aarohinadkarni: No, I don't think we do.

fboulnois: We have code styles and IntelliJ built-in rules checked into the repository, but we don't have a separate Spotless task.

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
aarohinadkarni: These two variables are reset upon iteration through all the signing officials for an institution, so I'm not sure if I would be able to pass them into the method unless I'm missing something.

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
```diff
@@ -102,6 +109,33 @@ public DataAccessAgreement findById(Integer daaId) {
...
+        DataAccessAgreement daa = findById(daaId);
+        String daaName = daa.getFile().getFileName();
+        emailService.sendDaaRequestMessage(signingOfficialName, signingOfficialEmail, userName,
+            daaName, daaId, userId);
```
okotsopoulos:
```suggestion
        DataAccessAgreement daa = findById(daaId);
        String daaName = daa.getFile().getFileName();
        emailService.sendDaaRequestMessage(signingOfficial.displayName, signingOfficial.email, userName,
            daaName, daaId, userId);
```

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
okotsopoulos: https://github.com/DataBiosphere/consent/pull/2283/files/ab23cd2eab405263105196c5aafd2c789b7ab2af#r1559889542 here is a concrete suggestion!

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
okotsopoulos: @rushtong is that the right link?

src/main/resources/assets/paths/daaRequestByDaaId.yaml
rushtong: I think we've not had a solid understanding of `operationId` at all. They will need to be unique, which can be hard to guarantee if we have similar API methods that do similar things. I suggest we remove them since they are optional
https://swagger.io/docs/specification/paths-and-operations/

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
aarohinadkarni: Ah, I completely misread your suggestion. That makes sense, thank you for the clarification!

src/test/java/org/broadinstitute/consent/http/service/DaaServiceTest.java
rushtong: In service tests, there are often conditionals that are embedded in the service logic such that a method call may or may not be made depending on input. In those kinds of situations, I find `spy`s useful to ensure that we ARE calling a method in one case and NOT calling it in a different case. In this situation, I agree with @okotsopoulos in that we're probably over testing this particular path.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
rushtong: Hah, no, it is not. This is the right one:
https://github.com/DataBiosphere/consent/blob/develop/src/main/java/org/broadinstitute/consent/http/resources/Resource.java#L64

src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
aarohinadkarni: In this case, do you think I should use a text block along with String.format to propagate our query field variables?

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
rushtong: Ok, I think I see where you're coming from. In this method, what I think should happen is the following:
1. Keep the user as an argument
2. Use `InstitutionDAO` instead of institution service to get the full institution. One thing we try to avoid is making services depend on other services. There are exceptions (i.e. `EmailService` and `GCSService`), but if we can, we try. In this case, I think we can avoid the service.
3. `InstitutionDAO` doesn't actually have the thing we want, which is institution with it's SOs. I suggest adding a new DAO method that looks like `findAllInstitutions` but just for a single institution id.
4. With the `SimplifiedUser` entities on your brand new method ☝🏽, you'll be able to avoid the large DB call to find ALL institutions.
5. Then you can send out emails to each simplified user.

rushtong: Untested, but this should be close to what you want:
```
  @RegisterBeanMapper(value = User.class, prefix = "u")
  @RegisterBeanMapper(value = SimplifiedUser.class, prefix = "so")
  @UseRowReducer(InstitutionWithUsersReducer.class)
  @SqlQuery("""
        SELECT i.*,
        u.user_id AS u_user_id, 
        u.email AS u_email,
        u.display_name AS u_display_name, 
        u.create_date AS u_create_date,
        u.email_preference AS u_email_preference,
        u.era_commons_id AS u_era_commons_id,
        u2.user_id AS u2_user_id, u2.email AS u2_email,
        u2.display_name AS u2_display_name, u2.create_date AS u2_create_date,
        u2.email_preference AS u2_email_preference,
        u2.era_commons_id AS u2_era_commons_id,
        so.so_user_id, so.so_email, so.so_display_name
        FROM institution i
        LEFT JOIN users u ON u.user_id = i.create_user
        LEFT JOIN users u2 ON u2.user_id = i.update_user
        LEFT JOIN
          (SELECT
            so.user_id AS so_user_id, so.email AS so_email,
            so.display_name AS so_display_name, so.institution_id AS so_institution_id
            FROM users so
            LEFT JOIN user_role ur ON ur.user_id = so.user_i
            WHERE ur.role_id = 7) so ON i.institution_id = so.so_institution_id
        WHERE i.institution_id = :institutionId
      """)
  Institution findInstitutionByIdWithSOs(Integer institutionId);
```

rushtong: An even simpler solution would be to query ONLY for simplified users by institution id which would essentially be the left join subquery above.

src/main/java/org/broadinstitute/consent/http/mail/freemarker/NewDaaRequestModel.java
aarohinadkarni: I'm getting FreeMarker errors when changing this definition to a Java record. I'll create a ticket to change all of our FreeMarkerTemplate models to be defined as Records separately.
[Ticket](https://broadworkbench.atlassian.net/browse/DUOS-3059)

src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
okotsopoulos: You can if you want to, but I looked at the rest of the DAO and this existing pattern is used throughout so it's fine to match that pattern here.  More sharing as a "this is a neat language feature that has made my life easier" thing :)

rushtong: We have been moving towards using text blocks ... there's just a ton of legacy code that we haven't gotten to yet.

okotsopoulos (approved):
Nice job persevering through the challenging GHA behavior!

src/main/java/org/broadinstitute/consent/http/db/InstitutionDAO.java
```diff
@@ -120,6 +120,38 @@ void updateInstitutionById(@Bind("institutionId") Integer institutionId,
...
+      FROM institution i
+      LEFT JOIN users u ON u.user_id = i.create_user
+      LEFT JOIN users u2 ON u2.user_id = i.update_user
+      LEFT JOIN
+           (SELECT
+               so.user_id AS so_user_id, so.email AS so_email,
+               so.display_name AS so_display_name, so.institution_id AS so_institution_id
+               FROM users so
+               LEFT JOIN user_role ur ON ur.user_id = so.user_id
+               WHERE ur.role_id = 7) so
```
rushtong, line 149: A more generic (and safer way) to do this is to query for the role table for a role with the SO name because usually, we don't always know the database ID of the thing we're looking for - it could be different in different environment databases. In this case, which is an exception, we DO know that the role id is always 7 based on the `UserRoles` enum. No changes necessary for this PR though - I wanted to point this out as it is an easily overlooked bug in other contexts.

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
```diff
@@ -102,6 +110,27 @@ public DataAccessAgreement findById(Integer daaId) {
     throw new NotFoundException("Could not find DAA with the provided ID: " + daaId);
   }
 
+  public void sendDaaRequestEmails(User user, Integer daaId) throws Exception {
+    try {
+      Institution institution = institutionDAO.findInstitutionWithSOById(user.getInstitutionId());
```
rushtong, line 116: Just thought of this. There is a corner case of a user not having an institution yet. We should error out in the case that the user doesn't have an institution. This can happen if a user signs in, goes to their profile page, and never chooses an institution. I don't anticipate the user ever seeing UI that would trigger this code path, but we should be protective both here and in the resource class as well.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -195,4 +199,23 @@ public Response deleteDaaForUser(
...
   }
+
+  @POST
+  @PermitAll
+  @Path("/request/{daaId}")
+  public Response sendDaaRequestMessage(
+      @Auth AuthUser authUser,
+      @PathParam("daaId") Integer daaId) {
+    try {
+      User user = userService.findUserByEmail(authUser.getEmail());
```
rushtong, line 211: Here is another place where we should check the user's institution and return an error. I'm debating on what that is, but 403 is technically the right answer here ... we understand the request, but are refusing to honor it for internal reasons. However, I think a 400 is more consistent with how we are handling this in other scenarios.

src/main/resources/assets/paths/daaRequestByDaaId.yaml
```diff
@@ -0,0 +1,19 @@
...
+      description: The DAA ID
+      required: true
+      schema:
+        type: integer
+  tags:
+    - DAA
+  responses:
+    200:
+      description: Successfully sent a request to a Signing Official to create a DAA-Library Card Relationship
+    404:
```
rushtong, line 18: See resource comment - we likely need an additional response code here

rushtong (commented):
This is looking good ... just one more corner case to cover and I think we're done!

rushtong (approved):
Thank you for the changes!

PR #2286, opened 2024-04-09
[DUOS-2932][risk=no] Use V4 algorithm for all matches

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2932

### Summary
* Remove usages of commercial use
* Replace with non profit use where appropriate
* Use v4 algorithm for all matching efforts
* Significant refactor of the `DataUse` class (and `DataUseBuilder`) to match the [v4 algorithm](https://github.com/DataBiosphere/consent-ontology/blob/develop/docs/Algorithms.md#version-4)

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/DataUse.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/DataUse.java
@@ -20,7 +20,7 @@ public class DataUse {
   private Boolean populationOriginsAncestry;
   @Deprecated
   private Boolean populationStructure;
-  private Boolean commercialUse;
+//  private Boolean commercialUse;
   private Boolean nonProfitUse;
   private Boolean methodsResearch;
   @Deprecated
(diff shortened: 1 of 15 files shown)


src/main/java/org/broadinstitute/consent/http/models/DataUse.java
```diff
@@ -1,79 +1,44 @@
...
 import com.google.gson.GsonBuilder;
 import java.util.List;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
 
-@SuppressWarnings({"unused", "SameParameterValue", "WeakerAccess"})
 @JsonInclude(JsonInclude.Include.NON_EMPTY)
 public class DataUse {
```
okotsopoulos, line 10: [General Consent question]

Does Consent have the capability to auto-generate these models as part of your API spec definition?

okotsopoulos (approved):

src/main/java/org/broadinstitute/consent/http/models/DataUse.java
rushtong: Actually, no, it doesn't, which reminds me that I need to update the swagger models to match this, thanks for the reminder!

rushtong: Turns out that `DataUse.yaml` is not affected by these changes, but does need to be rewritten. I'll tackle that separately.

aarohinadkarni (approved):
looks good 👍

PR #2289, opened 2024-04-10
[DUOS-2841][DUOS-3058][risk=no] Set a timeout for Sam

rushtong:
### Addresses
* https://broadworkbench.atlassian.net/browse/DUOS-2841
* https://broadworkbench.atlassian.net/browse/DUOS-3058

### Summary
* Minor update to our Sam calls to timeout fast instead of holding up other API calls
* Reduce logging to warn in the general case

One thing we're seeing is that users are often blocked on slow-responding API calls when Sam is not responding. We see a lot of these kinds of errors in the logs:
```
jsonPayload: {
   logger: "org.broadinstitute.consent.http.authentication.OAuthAuthenticator".
   message: "Exception retrieving Sam user info for 'user@gmail.com'Server Error (Connect timed out)"
   service-name: "consent"
   thread: "pool-33-thread-195 - GET /api/dac"
   timestamp: 1710444777149
}
```
This PR should also reduce that logging down to a warning since it is so frequent.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
@@ -38,17 +38,19 @@ public class SamDAO implements ConsentLogger {
   private final ExecutorService executorService;
   private final HttpClientUtil clientUtil;
   private final ServicesConfiguration configuration;
+  private final Integer timeoutMilliseconds;
 
   public SamDAO(HttpClientUtil clientUtil, ServicesConfiguration configuration) {
     this.executorService = Executors.newCachedThreadPool();
     this.clientUtil = clientUtil;
     this.configuration = configuration;
+    this.timeoutMilliseconds = configuration.getTimeoutSeconds() * 1000;
   }
 
   public List<ResourceType> getResourceTypes(AuthUser authUser) throws Exception {
     GenericUrl genericUrl = new GenericUrl(configuration.getV1ResourceTypesUrl());
     HttpRequest request = clientUtil.buildGetRequest(genericUrl, authUser);
-    HttpResponse response = clientUtil.handleHttpRequest(request);
+    HttpResponse response = executeRequest(request);
     if (!response.isSuccessStatusCode()) {
       logException("Error getting resource types from Sam: " + response.getStatusMessage(),
           new ServerErrorException(response.getStatusMessage(), response.getStatusCode()));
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
```diff
@@ -187,4 +189,18 @@ public EmailResponse getV1UserByEmail(AuthUser authUser, String email) throws Ex
     return new Gson().fromJson(body, EmailResponse.class);
   }
 
+  /**
+   * Private method to handle the general case of sending requests to Sam.
+   * We inject timeouts here to prevent Sam from impacting API performance.
+   * The default is 10 seconds which should be more than enough for Sam calls.
```
okotsopoulos, line 199: [FYI]

TDR has found that 10 seconds is not always enough, but the operations TDR facilitates may be more prone to longer response times.

Here's the approach we take, copied from a past PR comment:
> We don't really want to leave TDR's caller hanging for much longer than 60 seconds (that's the value of `SAM_OPERATION_TIMEOUT_SECONDS`). This PR gives more flexibility as to how those 60 seconds are spent: if our calls to Sam fail fast, then we'll be able to wait and try again 1-2 more times. But if our calls to Sam need more time, we'll take that time, up to 60 seconds.

More details here: https://github.com/DataBiosphere/jade-data-repo/pull/1410
And a follow-on PR, writing unit tests was difficult in my first PR which pointed to a need to refactor the code under test (this is an example of what I brought up in a previous PR review about Spys in tests often being a code smell): https://github.com/DataBiosphere/jade-data-repo/pull/1412

okotsopoulos (approved):
Looks good!

I was surprised that `SamDAOTest`s still worked without modifications given the changes to how the request is constructed.  Could you help me understand how those tests work?

rushtong:
@okotsopoulos 
> I was surprised that SamDAOTests still worked without modifications given the changes to how the request is constructed. Could you help me understand how those tests work?

We actually have two sets of Sam tests, our original set of mocked tests and a newer set of pact tests. Neither one should be impacted as the only change we're making here is to add connect and read timeouts to the request. Our mockserver mocks don't simulate timeouts, but they do simulate other kinds of errors.

src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
rushtong: Actually, I think setting the connect timeout is probably the better choice, but the read timeout could be longer. I'll move that to 60 seconds so we aren't dropping connections too soon.

aarohinadkarni (approved):
Looks good! 👍 However, I am seeing that the Coverage check failed with an error that is inconsistent with a local run of the test it is saying failed 🤔

rushtong:
> Looks good! 👍 However, I am seeing that the Coverage check failed with an error that is inconsistent with a local run of the test it is saying failed 🤔

I'm looking into that now ... tests pass locally so I'm working on making it more GH friendly.

PR #2292, opened 2024-04-11
[DUOS-1740][risk=no] Version updates for all actions

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1740

### Summary
Update for all actions
Adds [dependabot check](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot) for actions

----

- Verify all tests go green

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,11 @@ updates:
     schedule:
       # Check for updates to GitHub Actions every week
       interval: "weekly"
+    target-branch: develop
+    reviewers:
+      - "@DataBiosphere/DUOS"
+    commit-message:
+      prefix: "[DUOS-1740-dependabot]"
   - package-ecosystem: maven
     directory: "/"
     schedule:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,10 @@
 version: 2
 updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "weekly"
   - package-ecosystem: maven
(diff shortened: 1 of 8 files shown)


.github/dependabot.yml
```diff
@@ -1,5 +1,10 @@
 version: 2
 updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "weekly"
```
fboulnois, line 7: This is a great addition.

fboulnois (approved):
Since we're bumping major versions have we checked that these have complete compatibility with our current actions?

.github/dependabot.yml
okotsopoulos: Do we also want to set `open-pull-requests-limit`?

okotsopoulos (approved):
Thanks for doing this!  I think the only modified workflow that doesn't run on PRs is `smoke-tests.yaml`, so the rest have run on this PR with the new actions.

.github/dependabot.yml
rushtong: > Do we also want to set `open-pull-requests-limit`?

We can certainly add it, but it hasn't been a problem to date. Since we have grouped PRs, we generally get one or two a week. We almost always get one for general library updates and rarely we get one for docker. I wouldn't anticipate the action PRs piling up either - the version bumps there are pretty infrequent

okotsopoulos: All good then, in the unlikely event it poses an issue we can add it.

aarohinadkarni (approved):
looks good 👍

PR #2303, opened 2024-04-21
[DUOS-3070][risk=no] Bug Fix: revert to v3 of upload-artifact action

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-3070

### Summary
* Revert version to v3, v4 is incompatible with our usage

----

- Verify all tests go green

### DIFF
--- a/.github/workflows/smoke-tests.yaml
+++ b/.github/workflows/smoke-tests.yaml
@@ -36,7 +36,7 @@ jobs:
         run: |
           mvn clean test -P integration-tests -DbaseUrl=https://consent.${{ steps.setup.outputs.bee-name }}.bee.envs-terra.bio/
       - name: Store Test Result Artifact
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v3
         if: always()
         with:
           name: test-reports


fboulnois (approved):
👍

.github/workflows/smoke-tests.yaml
```diff
@@ -36,7 +36,7 @@ jobs:
         run: |
           mvn clean test -P integration-tests -DbaseUrl=https://consent.${{ steps.setup.outputs.bee-name }}.bee.envs-terra.bio/
       - name: Store Test Result Artifact
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v3
```
okotsopoulos, line 39: If we want to pin this to an earlier version, I think we'll also want to instruct Dependabot to ignore the updates, right?  I personally favor codifying these ignore rules so that we have the history behind them, ideally we might be able to remove them someday.

I'd also leave a comment indicating the need for compatibility with whatever version [dsp-reusable-workflows/.github/workflows/upload_test_results_to_biquery.yaml at main · broadinstitute/dsp-reusable-workflows](https://github.com/broadinstitute/dsp-reusable-workflows/blob/main/.github/workflows/upload_test_results_to_biquery.yaml#L80) is using to download.

Lastly, I asked DevOps whether they'd consider publishing their workflows as versioned actions here: https://broadinstitute.slack.com/archives/C029LTN5L80/p1713883830299009

okotsopoulos (changes requested):
Thanks for fixing, happy to take another look but wasn't sure about the Dependabot ignore.

.github/workflows/smoke-tests.yaml
rushtong: We can ignore them when they come in - I just ignored one this morning: https://github.com/DataBiosphere/consent-ontology/pull/949

okotsopoulos: Is there an easy way to see all of the dependencies ignored in this way (rather than ignored via [configuration](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#specifying-dependencies-and-versions-to-ignore))?  How do you remove one of these ignore rules?

okotsopoulos (approved):
Thank you!  I won't hold approval since this dependency is being ignored, but I'll note that I continued our convo in my comment thread.

.github/workflows/smoke-tests.yaml
rushtong: I don't know of any way to see the current ignore settings this way - [the docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore) don't seem to be helpful here. In this scenario, when we need to ignore something for a legitimate reason, I will ticket the details so we can follow up on it.

PR #2306, opened 2024-04-22
[DUOS-3073] Simplify creation of user roles in code and tests

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-3073

### Summary

This refactor simplifies assigning new roles to users in code and tests. It will help with my upcoming DAA work implementation.

The main implementation pieces are in `UserRoles.java` and `User.java`. Where it was safe to do so, I replaced calls to `new UserRole(...)` with `User.set<Role>`. If it was not safe to do so, I added the refactored role directly to the user instead. There may still be a few places in the codebase that use `new UserRole(...)`, in this case these use exotic constructions where I could not be sure that either case was safe to use.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
+++ b/src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
@@ -38,35 +38,35 @@ public enum UserRoles {
     this.roleId = roleId;
   }
 
-  public static UserRole AdminRole() {
+  public static UserRole Admin() {
     return new UserRole(ADMIN.getRoleId(), ADMIN.getRoleName());
   }
 
-  public static UserRole AlumniRole() {
+  public static UserRole Alumni() {
     return new UserRole(ALUMNI.getRoleId(), ALUMNI.getRoleName());
   }
 
-  public static UserRole ChairpersonRole() {
+  public static UserRole Chairperson() {
     return new UserRole(CHAIRPERSON.getRoleId(), CHAIRPERSON.getRoleName());
   }
 
-  public static UserRole DataSubmitterRole() {
+  public static UserRole DataSubmitter() {
     return new UserRole(DATASUBMITTER.getRoleId(), DATASUBMITTER.getRoleName());
   }
 
-  public static UserRole ITDirectorRole() {
+  public static UserRole ITDirector() {
     return new UserRole(ITDIRECTOR.getRoleId(), ITDIRECTOR.getRoleName());
   }
 
-  public static UserRole MemberRole() {
+  public static UserRole Member() {
     return new UserRole(MEMBER.getRoleId(), MEMBER.getRoleName());
   }
 
-  public static UserRole ResearcherRole() {
+  public static UserRole Researcher() {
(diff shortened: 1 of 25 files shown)


rushtong (approved):
Thank you, this is a fantastic cleanup 👍🏽

aarohinadkarni (approved):
looks awesome, thank you 👍

src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
```diff
@@ -37,6 +38,38 @@ public enum UserRoles {
     this.roleId = roleId;
   }
 
+  public static UserRole AdminRole() {
```
okotsopoulos: As these methods will be called with `UserRole` as the qualifier, I think you could elect to drop the `Role` suffix throughout, e.g. `UserRole.Admin()`.

src/main/java/org/broadinstitute/consent/http/models/User.java
okotsopoulos, line 1: These additions feel like they'd work nicely with a builder pattern.  But itt's not currently present in this class, so I'd understand a desire not to implement it partially.

Here's an example (which also allows users to hold multiple roles, I don't know if that would ever be desirable):
```
public User withAdminRole() {
  if (this.roles == null) {
    this.roles = new ArrayList<>();
  }
  this.roles.add(UserRoles.AdminRole());
  return this;
}
```

Example usage:
```
User admin = new User().withAdminRole();
```

src/main/java/org/broadinstitute/consent/http/models/User.java
```diff
@@ -188,6 +189,36 @@ public void setRoles(List<UserRole> roles) {
     this.roles = roles;
   }
 
+  public void setAdminRole() {
+    this.roles = Collections.singletonList(UserRoles.AdminRole());
```
okotsopoulos: In case you like it better, Java 9+ includes `List.of(..)` method, which has been my personal favorite for defining immutable lists, singleton or otherwise.

okotsopoulos (approved):
Looks like a great improvement.  I left a few thoughts, but none are blockers to merge.

src/main/java/org/broadinstitute/consent/http/models/User.java
rushtong: I'm generally in favor of `List.of(..)` as well. The one caveat I've had with refactoring to that has to do with tests that attempt list mutations.

fboulnois: Yes, I was looking into this earlier this morning. I'll switch it up.

src/main/java/org/broadinstitute/consent/http/models/User.java
fboulnois: I'll make this as a follow up refactor with the `EnumMap` too.

fboulnois: I created follow up tickets:

- https://broadworkbench.atlassian.net/browse/DCJ-153
- https://broadworkbench.atlassian.net/browse/DCJ-154

PR #2308, opened 2024-04-23
[DUOS-3052][DCJ-61] Add or remove DAA from a list of users

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DUOS-3052

### Summary

Allows DAA to be added to or removed from a list of users.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
@@ -225,7 +225,7 @@ public class DaaResource extends Resource implements ConsentLogger {
 
   @POST
   @Consumes(MediaType.APPLICATION_JSON)
-  @RolesAllowed({ADMIN, CHAIRPERSON})
+  @RolesAllowed({ADMIN, SIGNINGOFFICIAL})
   @Path("/bulk/{daaId}")
   public Response bulkAddUsersToDaa(
       @Auth AuthUser authUser,
@@ -253,7 +253,7 @@ public class DaaResource extends Resource implements ConsentLogger {
 
   @DELETE
   @Consumes(MediaType.APPLICATION_JSON)
-  @RolesAllowed({ADMIN, CHAIRPERSON})
+  @RolesAllowed({ADMIN, SIGNINGOFFICIAL})
   @Path("/bulk/{daaId}")
   public Response bulkRemoveUsersFromDaa(
       @Auth AuthUser authUser,
--- a/src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
@@ -22,6 +22,7 @@ import jakarta.ws.rs.core.UriInfo;
 import java.io.InputStream;
 import java.net.URI;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import org.broadinstitute.consent.http.enumeration.UserRoles;
 import org.broadinstitute.consent.http.models.AuthUser;
@@ -34,7 +35,6 @@ import org.broadinstitute.consent.http.service.DacService;
 import org.broadinstitute.consent.http.service.EmailService;
 import org.broadinstitute.consent.http.service.LibraryCardService;
 import org.broadinstitute.consent.http.service.UserService;
-import org.broadinstitute.consent.http.service.UserService.SimplifiedUser;
 import org.broadinstitute.consent.http.util.ConsentLogger;
 import org.glassfish.jersey.media.multipart.FormDataContentDisposition;
 import org.glassfish.jersey.media.multipart.FormDataParam;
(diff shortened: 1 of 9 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -222,4 +222,60 @@ public Response sendDaaRequestMessage(
...
+      String json) {
+    try {
+      User authedUser = userService.findUserByEmail(authUser.getEmail());
+      List<User> users = userService.findUsersInJsonArray(json, "users");
+      if (authedUser.hasUserRole(UserRoles.SIGNINGOFFICIAL)) {
+        for (User user : users) {
+          if (!Objects.equals(authedUser.getInstitutionId(), user.getInstitutionId())) {
+            return Response.status(Status.FORBIDDEN).build();
+          }
+        }
```
rushtong, line 242: The ticket isn't clear on this but I think the original intention for this was to query for all users in the SO's institution instead of accepting input of user ids. It would simplify the logic since we wouldn't need that extra parameter or check for an invalid request. This approach, however, does give us some additional flexibility in that it gives the UI the ability to select individual users and apply the DAA to only those selected.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -222,4 +222,60 @@ public Response sendDaaRequestMessage(
...
+      }
+      daaService.findById(daaId);
+      for (User user : users) {
+        libraryCardService.addDaaToUserLibraryCardByInstitution(user, authedUser.getInstitutionId(), daaId);
+      }
```
rushtong, line 247: There is a potential future optimization here. This looped service call ends up making looped database calls. It would be more efficient to bulk add in a single database operation but we can tackle that when we need to.

rushtong (approved):
This looks good, thank you! I'm fine with this approach, but feel free to examine suggestions in comments 👍🏽

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -222,4 +222,60 @@ public Response sendDaaRequestMessage(
       return createExceptionResponse(e);
     }
   }
+
+  @POST
+  @Consumes(MediaType.APPLICATION_JSON)
+  @RolesAllowed({ADMIN, CHAIRPERSON})
```
aarohinadkarni: Should this be Admin & Signing Official?

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -222,4 +222,60 @@ public Response sendDaaRequestMessage(
...
+      }
+      return Response.ok().build();
+    } catch (Exception e) {
+      return createExceptionResponse(e);
+    }
+  }
+
+  @DELETE
+  @Consumes(MediaType.APPLICATION_JSON)
+  @RolesAllowed({ADMIN, CHAIRPERSON})
```
aarohinadkarni: Should this be Admin & Signing Official?

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
rushtong: I missed that, nice catch!

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -222,4 +222,60 @@ public Response sendDaaRequestMessage(
...
+      if (authedUser.hasUserRole(UserRoles.SIGNINGOFFICIAL)) {
+        for (User user : users) {
+          if (!Objects.equals(authedUser.getInstitutionId(), user.getInstitutionId())) {
+            return Response.status(Status.FORBIDDEN).build();
+          }
+        }
+      }
+      daaService.findById(daaId);
+      for (User user : users) {
+        libraryCardService.addDaaToUserLibraryCardByInstitution(user, authedUser.getInstitutionId(), daaId);
```
aarohinadkarni, line 246: What is the expected behavior when one of these users already has the DAA-LC relationship on their Library Card?

src/main/resources/assets/paths/daaBulkByDaaId.yaml
```diff
@@ -0,0 +1,82 @@
...
+        type: integer
+  requestBody:
+    required: true
+    content:
+      application/json:
+        schema:
+          type: object
+          properties:
+            users:
+            type: array
```
aarohinadkarni: Right now, I'm getting a rendering error for this endpoint on the Swagger:

image (1)

I think you need to fix the spacing here:
```suggestion
              type: array
              items:
                type: integer
```

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
fboulnois: Whoops I missed this. My user has all the right permissions for testing 😂

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
fboulnois: I don't know what the behavior should be. I would imagine that it's a new relationship but doesn't necessarily replace the previous one.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
fboulnois: My understanding is that in the mocks, you can select by column or by row. But you can also manually uncheck these, so you could potentially modify all users except one.

src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
```diff
@@ -113,7 +113,11 @@ public List<LibraryCard> addDaaToUserLibraryCardByInstitution(User user, Integer
         .toList();
     // typically there should be one library card per user per institution
     for (LibraryCard libraryCard : matchingLibraryCards) {
-      addDaaToLibraryCard(libraryCard.getId(), daaId);
+      try {
+        addDaaToLibraryCard(libraryCard.getId(), daaId);
```
rushtong: Another thing we can do to tweak the sql is to do nothing if the join record exists. This method in LibraryCardDAO:
```
  @SqlUpdate("""
      INSERT INTO lc_daa (lc_id, daa_id)
      VALUES (:lcId, :daaId)
      """)
  void createLibraryCardDaaRelation(@Bind("lcId") Integer lcId, @Bind("daaId") Integer daaId);
```
Could possibly be:
```
  @SqlUpdate("""
      INSERT INTO lc_daa (lc_id, daa_id)
      VALUES (:lcId, :daaId)
      ON CONFLICT DO NOTHING
      """)
  void createLibraryCardDaaRelation(@Bind("lcId") Integer lcId, @Bind("daaId") Integer daaId);
```

fboulnois: That's probably a better solution. I'll do that instead.

fboulnois: Fixed in https://github.com/DataBiosphere/consent/pull/2308/commits/6a25bf822186af72d85568e30d9fa56d5d90a214

rushtong (approved):
👍🏽

aarohinadkarni (approved):
looks good, thank you for the changes! 👍

PR #2321, opened 2024-05-03
[DCJ-302][risk=no] Remove duplicate data submitter property

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-302

### Summary
Minor update to remove the duplicate data submitter properties. The data submitter is the create user for both the dataset and the study. Current code creates both a dataset and a study property with the same values as the create user, which is redundant.

----

- Verify all tests go green


fboulnois (approved):
thank you 👍

aarohinadkarni (approved):
looks good 👍

PR #2327, opened 2024-05-13
[DCJ-350] Modify bulk DAA endpoints to create LC if none on user(s)

aarohinadkarni:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-350

### Summary
Modifies service level methods for the bulk DAA-LC endpoints to now create a library card for a user/users if they don't have one before creating a DAA-LC relationship. This change was similarly done for the non-bulk endpoints. These changes also modify the corresponding tests.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
@@ -115,6 +115,9 @@ public class LibraryCardService {
   }
 
   public List<LibraryCard> addDaaToUserLibraryCardByInstitution(User user, User signingOfficial, Integer daaId) {
+    if (signingOfficial.getInstitutionId() == null) {
+      throw new BadRequestException("This signing official does not have an institution.");
+    }
     List<LibraryCard> libraryCards = findLibraryCardsByUserId(user.getUserId());
     List<LibraryCard> matchingLibraryCards = libraryCards.stream()
         .filter(card -> Objects.equals(card.getInstitutionId(), signingOfficial.getInstitutionId()))
--- a/src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
@@ -118,19 +118,16 @@ public class LibraryCardService {
     if (signingOfficial.getInstitutionId() == null) {
       throw new BadRequestException("This signing official does not have an institution.");
     }
-    List<LibraryCard> libraryCards = findLibraryCardsByUserId(user.getUserId());
-    List<LibraryCard> matchingLibraryCards = libraryCards.stream()
-        .filter(card -> Objects.equals(card.getInstitutionId(), signingOfficial.getInstitutionId()))
-        .collect(Collectors.toCollection(ArrayList::new));
-    if (matchingLibraryCards.isEmpty()) {
+    List<LibraryCard> libraryCards = new ArrayList<>(libraryCardDAO.findLibraryCardsByUserIdInstitutionId(user.getUserId(), signingOfficial.getInstitutionId()));
+    if (libraryCards.isEmpty()) {
       LibraryCard lc = createLibraryCardForSigningOfficial(user, signingOfficial);
-      matchingLibraryCards.add(lc);
+      libraryCards.add(lc);
     }
     // typically there should be one library card per user per institution
-    for (LibraryCard libraryCard : matchingLibraryCards) {
+    for (LibraryCard libraryCard : libraryCards) {
       addDaaToLibraryCard(libraryCard.getId(), daaId);
     }
-    return matchingLibraryCards;
+    return libraryCardDAO.findLibraryCardsByUserIdInstitutionId(user.getUserId(), signingOfficial.getInstitutionId());
   }
 
   public List<LibraryCard> removeDaaFromUserLibraryCardByInstitution(User user, Integer institutionId, Integer daaId) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
```diff
@@ -112,11 +114,15 @@ public void removeDaaFromLibraryCard(Integer libraryCardId, Integer daaId) {
     libraryCardDAO.deleteLibraryCardDaaRelation(libraryCardId, daaId);
   }
 
-  public List<LibraryCard> addDaaToUserLibraryCardByInstitution(User user, Integer institutionId, Integer daaId) {
+  public List<LibraryCard> addDaaToUserLibraryCardByInstitution(User user, User signingOfficial, Integer daaId) {
     List<LibraryCard> libraryCards = findLibraryCardsByUserId(user.getUserId());
```
rushtong: We should probably fail here if the `signingOfficial` doesn't have an institution.

src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
```diff
@@ -112,11 +114,15 @@ public void removeDaaFromLibraryCard(Integer libraryCardId, Integer daaId) {
...
     List<LibraryCard> matchingLibraryCards = libraryCards.stream()
-        .filter(card -> Objects.equals(card.getInstitutionId(), institutionId))
-        .toList();
+        .filter(card -> Objects.equals(card.getInstitutionId(), signingOfficial.getInstitutionId()))
+        .collect(Collectors.toCollection(ArrayList::new));
+    if (matchingLibraryCards.isEmpty()) {
+      LibraryCard lc = createLibraryCardForSigningOfficial(user, signingOfficial);
+      matchingLibraryCards.add(lc);
+    }
```
rushtong: I'm finding the overall logic for this method a little convoluted. We do the following things:
1. Find all user LCs
2. Filter them on an institution
3. If none, create new LC
4. For each LC, add DAA to it
5. Return the filtered LCs without having updated them with the new DAAs

What would make more sense to me is the following logical process
1. Find user LCs by user and institution ID (likely a new dao call)
  a. If none, then create new LC
2. Issue a call to a new bulk insert DAO call that adds DAA association for all instances of user id and institution id. See, for example, [this method for bulk inserting](https://github.com/DataBiosphere/consent/blob/8305d4748f6c6589fc68723de759c871c9a2283e/src/main/java/org/broadinstitute/consent/http/db/UserPropertyDAO.java#L29) user properties
3. Return a fresh DAO call of all LCs by user id and institution id so that the DAA ids are populated on the LCs.

rushtong (commented):
This looks pretty good as it is. See comments inline and we can discuss potential improvements.

src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
```diff
@@ -112,16 +114,20 @@ public void removeDaaFromLibraryCard(Integer libraryCardId, Integer daaId) {
...
   }
 
-  public List<LibraryCard> addDaaToUserLibraryCardByInstitution(User user, Integer institutionId, Integer daaId) {
-    List<LibraryCard> libraryCards = findLibraryCardsByUserId(user.getUserId());
-    List<LibraryCard> matchingLibraryCards = libraryCards.stream()
-        .filter(card -> Objects.equals(card.getInstitutionId(), institutionId))
-        .toList();
+  public List<LibraryCard> addDaaToUserLibraryCardByInstitution(User user, User signingOfficial, Integer daaId) {
+    if (signingOfficial.getInstitutionId() == null) {
+      throw new BadRequestException("This signing official does not have an institution.");
```
fboulnois, line 119: Could we add a test for these lines?

fboulnois (approved):
See below:

src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
aarohinadkarni: I believe there is one: https://github.com/DataBiosphere/consent/pull/2327/files#diff-4fea547504587c95ef46d7936f68c193192046e49d4349cb3834ff2c9c1cabe0R455

fboulnois: If you set a breakpoint on this line and run the test in debug mode it will skip these lines since the institution is set. It throws later, so I suspect the test is not running as expected.

To see coverage, click the button next to tests and select `"Run <test> with Coverage"`.

src/test/java/org/broadinstitute/consent/http/db/LibraryCardDAOTest.java
```diff
@@ -219,6 +219,20 @@ void testFindLibraryCardByInstitutionId() {
     assertTrue(cardsFromDAO.get(0).getDaaIds().isEmpty());
   }
 
+  @Test
+  void testFindLibraryCardByUserIdInstitutionId() {
+    LibraryCard libraryCard = createLibraryCard();
```
rushtong, line 226: To exercise the daa part of the new query, I suggest adding this code:
```
    int dacId = dacDAO.createDac(RandomStringUtils.randomAlphabetic(5), RandomStringUtils.randomAlphabetic(5), RandomStringUtils.randomAlphabetic(5), new Date());
    Instant now = Instant.now();
    int daaId = daaDAO.createDaa(libraryCard.getUserId(), now, libraryCard.getUserId(), now, dacId);
    daaDAO.createDacDaaRelation(dacId, daaId);
    libraryCardDAO.createLibraryCardDaaRelation(libraryCard.getId(), daaId);
```
and then ensuring that `cardsFromDAO.get(0).getDaaIds().isEmpty()` is false

rushtong (approved):
Thank you for the updates 👍🏽

PR #2328, opened 2024-05-13
[DCJ-235] Fully populate LC on users for  /api/user/role/{roleName} User endpoint

aarohinadkarni:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-235

### Summary
Fully populates the library card object & associated properties on a user now in the `getUsersAsRole` method exposed through the `/api/user/role/{roleName}` endpoint. Modifies existing tests to check for Library Cards.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/UserService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/UserService.java
@@ -269,17 +269,15 @@ public class UserService {
       case Resource.SIGNINGOFFICIAL:
         Integer institutionId = user.getInstitutionId();
         if (Objects.nonNull(user.getInstitutionId())) {
-          List<User> users = userDAO.getUsersFromInstitutionWithCards(institutionId);
-          populateLibraryCardsOnUsers(users);
-          return users;
+          return userDAO.getUsersFromInstitutionWithCards(institutionId);
         } else {
           throw new NotFoundException("Signing Official (user: " + user.getDisplayName()
               + ") is not associated with an Institution.");
         }
       case Resource.ADMIN:
-        List<User> users = userDAO.findUsersWithLCsAndInstitution();
-        populateLibraryCardsOnUsers(users);
-        return users;
+        return userDAO.findUsersWithLCsAndInstitution();
+      default:
+        // do nothing
     }
     return Collections.emptyList();
   }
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -456,6 +460,15 @@ private void addExistingLibraryCards(User user) {
     userDAO.updateUser(user.getDisplayName(), user.getUserId(), user.getInstitutionId());
   }
 
+  private void populateLibraryCardsOnUsers(List<User> users) {
+    for (User user: users) {
+      List<LibraryCard> cards = libraryCardDAO.findLibraryCardsByUserId(user.getUserId());
+      if (Objects.nonNull(cards) && !cards.isEmpty()) {
+        user.setLibraryCards(cards);
+      }
+    }
```
rushtong: This is a flag for me when I see a DAO call in a loop. Instead of 1 database call for users with LCs, we're now making N+1 database calls. Ideally, we should be doing the population in `userDAO.getUsersFromInstitutionWithCards()` and `userDAO.findUsersWithLCsAndInstitution()`. Those methods should be fully populating the DAA associations as part of the library card - their names suggest as much. They are a bit thorny, but I think it's worth it to refactor them into text block queries and ditch the query field construction pattern for both of them.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -269,13 +269,17 @@ public List<User> getUsersAsRole(User user, String roleName) {
...
+          return users;
         } else {
           throw new NotFoundException("Signing Official (user: " + user.getDisplayName()
               + ") is not associated with an Institution.");
         }
       case Resource.ADMIN:
-        return userDAO.findUsersWithLCsAndInstitution();
+        List<User> users = userDAO.findUsersWithLCsAndInstitution();
+        populateLibraryCardsOnUsers(users);
+        return users;
```
fboulnois: We should add a default case too, even if it does nothing.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -269,13 +269,17 @@ public List<User> getUsersAsRole(User user, String roleName) {
...
     }
     return Collections.emptyList();
```
fboulnois, line 282: Could we add a test for this line?

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -456,6 +460,15 @@ private void addExistingLibraryCards(User user) {
     userDAO.updateUser(user.getDisplayName(), user.getUserId(), user.getInstitutionId());
   }
 
+  private void populateLibraryCardsOnUsers(List<User> users) {
+    for (User user: users) {
+      List<LibraryCard> cards = libraryCardDAO.findLibraryCardsByUserId(user.getUserId());
+      if (Objects.nonNull(cards) && !cards.isEmpty()) {
```
fboulnois: We should also test the null case here.

fboulnois (approved):
I agree with Greg's comment, but also here are a few comments relating to tests:

rushtong (approved):
Thank you for the changes! 👍🏽

PR #2330, opened 2024-05-20
[DCJ-49][DCJ-50][risk=no] New APIs for DAA Enforced DAR Create/Update

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DCJ-49
https://broadworkbench.atlassian.net/browse/DCJ-50

## Summary
* New v2/v3 APIs for all DAR create, update operations
* Add a DAA check that ensures that a user has the following requirements for each dataset requested in a DAR:
  * User has a Library Card
  * User's Library Card is associated to a Data Access Agreement
  * Data Access Agreement is associated to a DAC
  * DAC is associated to the Dataset

This chain of restrictions ensures that a user must have an active LC<->DAA association set up by the user's Signing Official when the user is requesting access to a Dataset through a Data Access Request. This requirement is checked when a DAR is created, submitted, and when it is updated in draft form.

## Follow On Work
We'll need new UI work to handle the (future) failure cases (see [DCJ-390](https://broadworkbench.atlassian.net/browse/DCJ-390)):
* The Data Library "Apply For Access" feature
* The Data Access Request "Save" button
* The Data Access Request "Attest" button

### Dev Notes
* Please note that the new APIs are essentially copied from the previous version but with a single addition of a DAA enforcement check.
* Existing UI functions are essentially unchanged since they call the original APIs
* To test this locally, you'll need to modify the UI as follows:
  * Update the following `v2` PUT and POST requests to `v3` in [DAR.js](https://github.com/DataBiosphere/duos-ui/blob/develop/src/libs/ajax/DAR.js):
  * `updateDarDraft`
  * `postDarDraft`
  * `postDar`
* To see if enforcement is working, first try to create a DAR against datasets for which you have no DAC-DAA associations in the Signing Official Console. This attempt should fail.
* Next, create for your user a DAC-DAA association and repeat the process - it should succeed.

Failure example:
Screenshot 2024-05-22 at 8 58 03 AM

----

- Verify all tests go green

[DCJ-390]: https://broadworkbench.atlassian.net/browse/DCJ-390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/resources/assets/paths/darV3.yaml
```diff
@@ -0,0 +1,25 @@
+post:
+  summary: Create Data Access Request, version 2
```
fboulnois:
```suggestion
  summary: Create Data Access Request, version 3
```

src/test/java/org/broadinstitute/consent/http/db/DaaDAOTest.java
```diff
@@ -214,4 +224,96 @@ void testFindWithDacs() {
     assertEquals(3, daa.getDacs().size());
   }
 
+
+  @Test
+  void testFindDaaDatasetIdsByUserId() {
+    // Testing the case of a user requesting DAR access to a datataset.
```
fboulnois:
```suggestion
    // Testing the case of a user requesting DAR access to a dataset.
```

src/test/java/org/broadinstitute/consent/http/db/DaaDAOTest.java
```diff
@@ -214,4 +224,96 @@ void testFindWithDacs() {
     assertEquals(3, daa.getDacs().size());
   }
 
+
+  @Test
+  void testFindDaaDatasetIdsByUserId() {
```
fboulnois, line 228: Could we add a test for the NULL user case?

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -153,4 +153,72 @@ public Response createDraftDataAccessRequest(
...
+        DataAccessRequestData data = new DataAccessRequestData();
+        String referenceId = UUID.randomUUID().toString();
+        newDar.setReferenceId(referenceId);
+        data.setReferenceId(referenceId);
+        if (!Objects.isNull(projectTitle) && !projectTitle.isBlank()) {
+          data.setProjectTitle(projectTitle);
+        }
+        newDar.setData(data);
+        newDar.setDatasetIds(datasetIds);
+        DataAccessRequest result = darService.insertDraftDataAccessRequest(user, newDar);
```
fboulnois: I think most of this logic should live in the service.

src/test/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceTest.java
```diff
@@ -785,4 +786,127 @@ void getDraftDar_UserNotAllowed() {
     assertEquals(res.getStatus(), HttpStatusCodes.STATUS_CODE_FORBIDDEN);
   }
 
+  @Test
+  void testCreateDataAccessRequestWithDAARestrictions() {
```
fboulnois, line 790: This doesn't cover any of the lines of `createDataAccessRequestWithDAARestrictions`. Is this expected?

fboulnois (changes requested):
A few suggestions:

src/test/java/org/broadinstitute/consent/http/db/DaaDAOTest.java
rushtong: Added!

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
rushtong: Great point, I can do that. It will simplify this class quite a bit.

src/test/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceTest.java
rushtong: Great catch, no, this is an error

src/main/resources/assets/api-docs.yaml
```diff
@@ -275,14 +275,26 @@ paths:
...
+
   /api/dar/v2/{referenceId}:
     $ref: './paths/darV2ByReferenceId.yaml'
+
+  /api/dar/v3/{referenceId}:
+    $ref: './paths/darV3ByReferenceId.yaml'
+
   /api/dar/v2/draft:
     post:
+      deprecated: true
```
aarohinadkarni, line 290: Did we want to deprecate this now and not wait for the cleanup to completely switch over to the endpoints that enforce DAAs?

src/main/resources/assets/paths/darV2.yaml
```diff
@@ -1,6 +1,9 @@
 post:
-  summary: Create Data Access Request, version 2
-  description: Creates a Data Access Request
+  deprecated: true
```
aarohinadkarni, line 2: Again, did we want to deprecate this now and not wait for the cleanup to completely switch over to the `v3` endpoint?

src/main/resources/assets/paths/darV2ByReferenceId.yaml
```diff
@@ -24,8 +24,11 @@ get:
     500:
       description: Server Error
 put:
-  summary: Update a Data Access Request by Reference Id
-  description: Returns the updated Data Access Request.
+  deprecated: true
```
aarohinadkarni: Ditto.

src/main/resources/assets/paths/darV3.yaml
```diff
@@ -0,0 +1,25 @@
+post:
```
aarohinadkarni, line 1: Do we need to add a 400 response for the error thrown if the user does not have all of the DAAs on their LC for any of the datasets in this DAR?

src/main/resources/assets/paths/darV3ByReferenceId.yaml
```diff
@@ -0,0 +1,31 @@
...
+        type: string
+    - name: dar
+      in: body
+      description: The fields that represent a DAR in json format.
+      required: true
+      schema:
+        $ref: '../schemas/DataAccessRequest.yaml'
+  tags:
+    - Data Access Request
+  responses:
```
aarohinadkarni: Again, do we need to add a 400 response for the error thrown if the user does not have all of the DAAs on their LC for any of the datasets in this update to the DAR?

Also, I'm realizing that in the update case, we are going to make the dataset selection in the DAR form readonly after initial creation, meaning the dataset selection will never change in the update case. That work is yet to be implemented, but is covered in [this ticket](https://broadworkbench.atlassian.net/browse/DCJ-48). So, will we ever run into the case where we have to enforce the DAA restrictions in this endpoint?

src/main/resources/assets/paths/darV3Draft.yaml
```diff
@@ -0,0 +1,19 @@
...
+  parameters:
+    - name: dar
+      in: body
+      description: The fields that represent a Data Access Request, json format.
+      required: true
+      schema:
+        $ref: '../schemas/DataAccessRequest.yaml'
+  tags:
+    - Data Access Request
+  responses:
```
aarohinadkarni, line 15: Again, do we need to add a 400 response for the error thrown if the user does not have all of the DAAs on their LC for any of the datasets in this draft DAR creation?

src/main/resources/assets/paths/darV3ByReferenceId.yaml
aarohinadkarni: Also, I don't actually think we created this new endpoint in `DataAccessRequestResource`.

src/main/resources/assets/paths/darV3DraftByReferenceId.yaml
```diff
@@ -0,0 +1,27 @@
...
+        type: string
+    - name: dar
+      in: body
+      description: The fields that represent a Data Access Request, json format.
+      required: true
+      schema:
+        $ref: '../schemas/DataAccessRequest.yaml'
+  tags:
+    - Data Access Request
+  responses:
```
aarohinadkarni, line 21: Do we need to add the 400 response for the error thrown if the user does not have all of the DAAs on their LC for any of the datasets in this draft DAR update?

src/main/resources/assets/paths/tdrDraftDAR.yaml
```diff
@@ -1,6 +1,9 @@
 post:
-  summary: createDraftDataAccessRequest
-  description: Create Draft Data Access Request
+  deprecated: true
```
aarohinadkarni, line 2: Did we want to deprecate this now and not wait for the cleanup to completely switch over to the endpoints that enforce DAAs?

src/main/resources/assets/paths/tdrV2DraftDAR.yaml
```diff
@@ -0,0 +1,27 @@
...
+        type: string
+    - name: projectTitle
+      in: query
+      description: Project title for the new Data Access Request.
+      required: false
+      schema:
+        type: string
+  tags:
+    - TDR
+  responses:
```
aarohinadkarni, line 21: Do we need to add the 400 response for the error thrown if the user does not have all of the DAAs on their LC for any of the datasets in this draft DAR creation?

aarohinadkarni (commented):
Looks good! Just a few questions:

src/main/resources/assets/paths/tdrDraftDAR.yaml
rushtong: I would prefer this since we need to leave these older APIs in place for at least a release

src/main/resources/assets/paths/tdrV2DraftDAR.yaml
rushtong: We do!

src/main/resources/assets/paths/darV3ByReferenceId.yaml
rushtong: This is a great catch and points out something we sometimes hit. I'm going to remove this and file a ticket for removing the v2 version of this endpoint. This is a PUT on a submitted DAR which is not something we should be allowing clients to do. The v2 API is not in use in the UI.

rushtong: https://broadworkbench.atlassian.net/browse/DCJ-400

fboulnois (approved):
thank you 👍

aarohinadkarni (approved):
looks good, thank you for the changes! 👍

PR #2334, opened 2024-05-30
[DCJ-56] Adds endpoint to add a DAC to a DAA

aarohinadkarni:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-56

### Summary
Adds an endpoint to add a DAC to a Data Access Agreement with appropriate tests. 

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
@@ -210,6 +210,23 @@ public class DaaResource extends Resource implements ConsentLogger {
     }
   }
 
+  @DELETE
+  @Path("{daaId}")
+  @Produces("application/json")
+  @RolesAllowed({ADMIN})
+  public Response adminDeleteDaa(
+      @Auth AuthUser authUser,
+      @PathParam("daaId") Integer daaId) {
+    try {
+      daaService.findById(daaId);
+      User user = userService.findUserByEmail(authUser.getEmail());
+      daaService.deleteDaa(daaId);
+      return Response.ok().build();
+    } catch (Exception e) {
+      return createExceptionResponse(e);
+    }
+  }
+
   @POST
   @PermitAll
   @Path("/request/{daaId}")
--- a/src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
@@ -21,6 +21,7 @@ import jakarta.ws.rs.core.StreamingOutput;
 import jakarta.ws.rs.core.UriInfo;
 import java.io.InputStream;
 import java.net.URI;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
@@ -332,7 +333,7 @@ public class DaaResource extends Resource implements ConsentLogger {
   @PUT
   @Produces(MediaType.APPLICATION_JSON)
   @RolesAllowed({ADMIN, CHAIRPERSON})
-  @Path("{daaId}/dac/{dacId}")
+  @Path("/{daaId}/dac/{dacId}")
   public Response addDacToDaa(
       @Auth AuthUser authUser,
       @PathParam("daaId") Integer daaId,
@@ -345,7 +346,7 @@ public class DaaResource extends Resource implements ConsentLogger {
       // chairperson for.
       if (!user.hasUserRole(UserRoles.ADMIN)) {
         List<Integer> dacIds = user.getRoles().stream().map(UserRole::getDacId).toList();
-        if (!dacIds.contains(dacId)) {
+        if (!new HashSet<>(dacIds).contains(dacId)) {
           return Response.status(Status.FORBIDDEN).build();
         }
       }
--- a/src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
@@ -79,14 +79,8 @@ public class DaaResource extends Resource implements ConsentLogger {
       // Admins can add a DAA with any DAC, but chairpersons can only add DAAs for DACs they are a
       // chairperson for.
       if (!user.hasUserRole(UserRoles.ADMIN)) {
-        List<Integer> matchedChairpersonDacIds = user
-            .getRoles()
-            .stream()
-            .filter(r -> r.getRoleId().equals(UserRoles.Chairperson().getRoleId()))
-            .map(UserRole::getDacId)
-            .filter(id -> Objects.equals(id, dacId))
-            .toList();
-        if (matchedChairpersonDacIds.isEmpty()) {
+        List<Integer> dacIds = user.getRoles().stream().map(UserRole::getDacId).toList();
+        if (!dacIds.contains(dacId)) {
           return Response.status(Status.FORBIDDEN).build();
         }
       }
@@ -368,8 +362,14 @@ public class DaaResource extends Resource implements ConsentLogger {
       // Admins can add a DAC to a DAA with any DAC, but chairpersons can only add DACs to DAAs for DACs they are a
       // chairperson for.
       if (!user.hasUserRole(UserRoles.ADMIN)) {
-        List<Integer> dacIds = user.getRoles().stream().map(UserRole::getDacId).toList();
-        if (!new HashSet<>(dacIds).contains(dacId)) {
+        List<Integer> matchedChairpersonDacIds = user
+            .getRoles()
+            .stream()
+            .filter(r -> r.getRoleId().equals(UserRoles.Chairperson().getRoleId()))
+            .map(UserRole::getDacId)
+            .filter(id -> Objects.equals(id, dacId))
+            .toList();
+        if (matchedChairpersonDacIds.isEmpty()) {
           return Response.status(Status.FORBIDDEN).build();
         }
       }
--- a/src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
@@ -21,7 +21,6 @@ import jakarta.ws.rs.core.StreamingOutput;
 import jakarta.ws.rs.core.UriInfo;
 import java.io.InputStream;
 import java.net.URI;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
@@ -362,14 +361,11 @@ public class DaaResource extends Resource implements ConsentLogger {
       // Admins can add a DAC to a DAA with any DAC, but chairpersons can only add DACs to DAAs for DACs they are a
       // chairperson for.
       if (!user.hasUserRole(UserRoles.ADMIN)) {
-        List<Integer> matchedChairpersonDacIds = user
-            .getRoles()
+        if (user.getRoles()
             .stream()
             .filter(r -> r.getRoleId().equals(UserRoles.Chairperson().getRoleId()))
             .map(UserRole::getDacId)
-            .filter(id -> Objects.equals(id, dacId))
-            .toList();
-        if (matchedChairpersonDacIds.isEmpty()) {
+            .noneMatch(dacId::equals)) {
           return Response.status(Status.FORBIDDEN).build();
         }
       }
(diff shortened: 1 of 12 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -327,4 +328,43 @@ public Response bulkRemoveDAAsFromUser(
...
+      @PathParam("daaId") Integer daaId,
+      @PathParam("dacId") Integer dacId) {
+    try {
+      dacService.findById(dacId);
+      User user = userService.findUserByEmail(authUser.getEmail());
+      // Assert that the user has the correct DAC permissions to add a DAC to a DAA for the provided DacId.
+      // Admins can add a DAC to a DAA with any DAC, but chairpersons can only add DACs to DAAs for DACs they are a
+      // chairperson for.
+      if (!user.hasUserRole(UserRoles.ADMIN)) {
+        List<Integer> dacIds = user.getRoles().stream().map(UserRole::getDacId).toList();
```
rushtong: Very small performance nit: list contains is less than ideal since it iterates over the whole list. I generally cast to a hash set when calling contains.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -327,4 +328,43 @@ public Response bulkRemoveDAAsFromUser(
...
+      Optional<Dac> matchingDac = Optional.empty();
+      if (daa.getDacs() != null) {
+        matchingDac = daa.getDacs().stream()
+            .filter(dac -> Objects.equals(dac.getDacId(), dacId))
+            .findFirst();
+      }
+      if (matchingDac.isEmpty()) {
+        daaService.addDacToDaa(dacId,daaId);
+      } else {
+        daaService.updateDateDaa(daaId);
```
rushtong: Do we need to make this call? I think we could return a not-modified (304) response in the case where an association already exists.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -327,4 +328,43 @@ public Response bulkRemoveDAAsFromUser(
       return createExceptionResponse(e);
     }
   }
+
+  @PUT
+  @Produces(MediaType.APPLICATION_JSON)
+  @RolesAllowed({ADMIN, CHAIRPERSON})
+  @Path("{daaId}/dac/{dacId}")
```
rushtong:
```suggestion
  @Path("/{daaId}/dac/{dacId}")
```

src/main/resources/assets/paths/daaForDacId.yaml
```diff
@@ -0,0 +1,30 @@
...
+  operationId: addDacToDaa
+  description: Create a link between a Data Access Agreement and a DAC
+  parameters:
+    - name: daaId
+      in: path
+      description: The DAA ID
+      required: true
+      schema:
+        type: integer
+    - name: dacUd
```
rushtong:
```suggestion
    - name: dacId
```

src/main/resources/assets/paths/daaForDacId.yaml
```diff
@@ -0,0 +1,30 @@
...
+      in: path
+      description: The DAC ID
+      required: true
+      schema:
+        type: integer
+  requestBody:
+    required: false
```
rushtong: I don't think you need this.

rushtong (commented):
I had to make a couple changes to get this to work locally - see comments inline. I also think we can simplify it a bit if we ignore the case of an existing dac-daa association.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
aarohinadkarni: Right now, yes. Inherently, we are only calling this endpoint if we want to add a DAC to a DAA (assigning the default DAA to a DAC). If a DAC chair goes from using the default DAC, to uploading their own, to going back to using the default DAC we would want to re-establish that "adding the DAC to the default DAA." In the UI, I am currently distinguishing the most recently assigned DAA using the update date as the deciding factor (a later update date means that is the DAA we want to use)  So, if DAC-DAA link already exists, but we are changing "back" to the default DAA, we want to update the update date for the DAA so that we can appropriately distinguish this DAA as the most recently assigned DAA.

rushtong: I guess I'm still confused here. We're not modifying the DAA entity, we're just modifying a relationship between a DAA and a DAC. The end result of any association to any dac is a change in the DAA's update date which seems counterintuitive to me. If we need to track a date modification, then it should likely be on the join table, not the DAA entity itself.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -344,4 +346,47 @@ public Response bulkRemoveDAAsFromUser(
...
+        List<Integer> matchedChairpersonDacIds = user
+            .getRoles()
+            .stream()
+            .filter(r -> r.getRoleId().equals(UserRoles.Chairperson().getRoleId()))
+            .map(UserRole::getDacId)
+            .filter(id -> Objects.equals(id, dacId))
+            .toList();
+        if (matchedChairpersonDacIds.isEmpty()) {
+          return Response.status(Status.FORBIDDEN).build();
+        }
```
rushtong: I think this is an even cleaner way to do this:
```
        if (user.getRoles()
            .stream()
            .filter(r -> r.getRoleId().equals(UserRoles.Chairperson().getRoleId()))
            .map(UserRole::getDacId)
            .noneMatch(dacId::equals)) {
          return Response.status(Status.FORBIDDEN).build();
        }
```

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -21,11 +21,13 @@
 import jakarta.ws.rs.core.UriInfo;
 import java.io.InputStream;
 import java.net.URI;
+import java.util.HashSet;
```
rushtong: This looks unused

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
```diff
@@ -11,6 +11,7 @@
 import jakarta.ws.rs.core.MediaType;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Date;
```
rushtong: This also looks unused

src/main/resources/assets/paths/daaForDacId.yaml
```diff
@@ -0,0 +1,28 @@
+put:
+  summary: Add DAA to DAC
+  operationId: addDacToDaa
```
rushtong: I would favor removing this. We're not using it for anything since we don't use any code generators that would need it.

src/test/java/org/broadinstitute/consent/http/db/DaaDAOTest.java
```diff
@@ -1,6 +1,7 @@
 package org.broadinstitute.consent.http.db;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
```
rushtong: This looks unused

src/test/java/org/broadinstitute/consent/http/resources/DaaResourceTest.java
```diff
@@ -933,6 +933,144 @@ void testBulkRemoveDAAsFromUserNotFound() {
     assert response.getStatus() == HttpStatus.SC_NOT_FOUND;
   }
 
+  @Test
+  void testAddDacToDaaAdmin() throws Exception {
```
rushtong: This test doesn't throw

src/test/java/org/broadinstitute/consent/http/resources/DaaResourceTest.java
```diff
@@ -933,6 +933,144 @@ void testBulkRemoveDAAsFromUserNotFound() {
...
+    when(userService.findUserByEmail(any())).thenReturn(admin);
+
+    resource = new DaaResource(daaService, dacService, userService, libraryCardService, emailService);
+
+    Response response = resource.modifyDacDaaRelationship(authUser, daaId, dac.getDacId());
+    assertEquals(HttpStatus.SC_OK, response.getStatus());
+  }
+
+  @Test
+  void testAddDacToDaaChairperson() throws Exception {
```
rushtong: Similar issue here, no exception is thrown.

src/test/java/org/broadinstitute/consent/http/resources/DaaResourceTest.java
```diff
@@ -933,6 +933,144 @@ void testBulkRemoveDAAsFromUserNotFound() {
...
+
+    Response response = resource.modifyDacDaaRelationship(authUser, daaId, dac.getDacId());
+    assertEquals(HttpStatus.SC_OK, response.getStatus());
+  }
+
+  @Test
+  void testAddDacToDaaChairpersonNoMatchingDac() {
+    int daaId = RandomUtils.nextInt(10, 100);
+    DataAccessAgreement daa = new DataAccessAgreement();
+    daa.setDaaId(daaId);
```
rushtong: This is unused in the test

src/test/java/org/broadinstitute/consent/http/resources/DaaResourceTest.java
```diff
@@ -933,6 +933,144 @@ void testBulkRemoveDAAsFromUserNotFound() {
...
+
+    Response response = resource.modifyDacDaaRelationship(authUser, daaId, dac.getDacId());
+    assertEquals(HttpStatus.SC_FORBIDDEN, response.getStatus());
+  }
+
+  @Test
+  void testAddDacToDaaFromUserForbidden() {
+    int daaId = RandomUtils.nextInt(10, 100);
+    DataAccessAgreement daa = new DataAccessAgreement();
+    daa.setDaaId(daaId);
```
rushtong: Same thing here, this is unused.

src/test/java/org/broadinstitute/consent/http/resources/DaaResourceTest.java
```diff
@@ -933,6 +933,144 @@ void testBulkRemoveDAAsFromUserNotFound() {
...
+
+    Response response = resource.modifyDacDaaRelationship(authUser, daaId, dac.getDacId());
+    assertEquals(HttpStatus.SC_FORBIDDEN, response.getStatus());
+  }
+
+  @Test
+  void testAddDacToDaaDaaNotFound() {
+    int daaId = RandomUtils.nextInt(10, 100);
+    DataAccessAgreement daa = new DataAccessAgreement();
+    daa.setDaaId(daaId);
```
rushtong: Unused here too.

src/test/java/org/broadinstitute/consent/http/resources/DaaResourceTest.java
```diff
@@ -933,6 +933,144 @@ void testBulkRemoveDAAsFromUserNotFound() {
...
+
+    Response response = resource.modifyDacDaaRelationship(authUser, daaId, dac.getDacId());
+    assertEquals(HttpStatus.SC_NOT_FOUND, response.getStatus());
+  }
+
+  @Test
+  void testAddDacToDaaDacNotFound() {
+    int daaId = RandomUtils.nextInt(10, 100);
+    DataAccessAgreement daa = new DataAccessAgreement();
+    daa.setDaaId(daaId);
```
rushtong: And here as well

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -344,4 +345,44 @@ public Response bulkRemoveDAAsFromUser(
...
+          return Response.status(Status.FORBIDDEN).build();
+        }
+      }
+      DataAccessAgreement daa = daaService.findById(daaId);
+      Optional<Dac> matchingDac = Optional.empty();
+      if (daa.getDacs() != null) {
+        matchingDac = daa.getDacs().stream()
+            .filter(dac -> Objects.equals(dac.getDacId(), dacId))
+            .findFirst();
+      }
```
rushtong, line 378: We could use a test where the DAA has dacs and try to exercise this condition.

rushtong (approved):
This looks great! You've addressed most of the feedback, I think we just need one more test for daa-dac logic. After that, 👍🏽

fboulnois (approved):
looks great 👍

rushtong (approved):
👍🏽

PR #2338, opened 2024-06-06
[DCJ-56][risk=no] Update liquibase statement so that we can create the unique constraint

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-56

### Summary
Minor update to the sql to ensure that the constraint can be created. Note that we're deleting the duplicate records here which is fine since there is no existing production data, just dev/staging data.

----

- Verify all tests go green


okotsopoulos (approved):

aarohinadkarni (approved):
thank you! 👍

PR #2340, opened 2024-06-11
[DCJ-432] Consent: Add endpoint to alert SOs & Researchers of newly uploaded DAA

aarohinadkarni:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-432

### Summary
Adds new endpoint to alert appropriate Signing Officials and Researchers that a new DAA was uploaded if they were previously approved for a DAC through a previous DAA or they had issued authorizations to researchers under the previous DAA, along with appropriate tests.

Emails sent to Researchers & Signing Officials, respectively:
image
image

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/enumeration/EmailType.java
+++ b/src/main/java/org/broadinstitute/consent/http/enumeration/EmailType.java
@@ -17,7 +17,8 @@ public enum EmailType {
   RESEARCHER_APPROVED(13),
   NEW_DATASET(14),
   NEW_DAA_REQUEST(15),
-  NEW_DAA_UPLOAD(16);
+  NEW_DAA_UPLOAD_RESEARCHER(16),
+  NEW_DAA_UPLOAD_SO(17);
 
   private final Integer typeInt;
 
--- a/src/main/java/org/broadinstitute/consent/http/enumeration/EmailType.java
+++ b/src/main/java/org/broadinstitute/consent/http/enumeration/EmailType.java
@@ -16,7 +16,8 @@ public enum EmailType {
   NEW_RESEARCHER(12),
   RESEARCHER_APPROVED(13),
   NEW_DATASET(14),
-  NEW_DAA_REQUEST(15);
+  NEW_DAA_REQUEST(15),
+  NEW_DAA_UPLOAD(16);
 
   private final Integer typeInt;
 
--- a/src/main/java/org/broadinstitute/consent/http/mail/freemarker/NewDAAUploadResearcherModel.java
+++ b/src/main/java/org/broadinstitute/consent/http/mail/freemarker/NewDAAUploadResearcherModel.java
@@ -1,0 +1,44 @@
+package org.broadinstitute.consent.http.mail.freemarker;
+
+public class NewDAAUploadResearcherModel {
+
+  private final String serverUrl;
+  private final String dacName;
+  private final String researcherUserName;
+  private final String previousDaaName;
+  private final String newDaaName;
+  public NewDAAUploadResearcherModel(
+      String serverUrl,
+      String dacName,
+      String researcherUserName,
+      String previousDaaName,
+      String newDaaName
+  ) {
+    this.serverUrl = serverUrl;
+    this.dacName = dacName;
+    this.researcherUserName = researcherUserName;
+    this.previousDaaName = previousDaaName;
+    this.newDaaName = newDaaName;
+  }
+
+  public String getServerUrl() {
+    return serverUrl;
+  }
+
+  public String getDacName() {
+    return dacName;
+  }
+
+  public String getResearcherUserName() {
+    return researcherUserName;
+  }
+
+  public String getPreviousDaaName() {
+    return previousDaaName;
+  }
+
+  public String getNewDaaName() {
+    return newDaaName;
+  }
+
+}
--- a/src/main/java/org/broadinstitute/consent/http/mail/message/NewDAAUploadResearcherMessage.java
+++ b/src/main/java/org/broadinstitute/consent/http/mail/message/NewDAAUploadResearcherMessage.java
@@ -1,0 +1,20 @@
+package org.broadinstitute.consent.http.mail.message;
+
+import com.sendgrid.helpers.mail.Mail;
+import java.io.Writer;
+import javax.mail.MessagingException;
+
+public class NewDAAUploadResearcherMessage extends MailMessage {
+  private final String NEW_DAA_UPLOAD = "New DAA uploaded for DAC in DUOS";
+
+  public Mail newDAAUploadResearcherMessage(String toAddress, String fromAddress, Writer template, String dacName)
+      throws MessagingException {
+    return generateEmailMessage(toAddress, fromAddress, template, dacName, null);
+  }
+
+  @Override
+  String assignSubject(String referenceId, String type) {
+    return NEW_DAA_UPLOAD;
+  }
+
+}
--- a/src/main/java/org/broadinstitute/consent/http/mail/message/NewDAAUploadResearcherMessage.java
+++ b/src/main/java/org/broadinstitute/consent/http/mail/message/NewDAAUploadResearcherMessage.java
@@ -5,7 +5,7 @@ import java.io.Writer;
 import javax.mail.MessagingException;
 
 public class NewDAAUploadResearcherMessage extends MailMessage {
-  private final String NEW_DAA_UPLOAD = "New DAA uploaded for DAC in DUOS";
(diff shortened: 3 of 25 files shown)


src/main/java/org/broadinstitute/consent/http/enumeration/EmailType.java
```diff
@@ -16,7 +16,8 @@ public enum EmailType {
   NEW_RESEARCHER(12),
   RESEARCHER_APPROVED(13),
   NEW_DATASET(14),
-  NEW_DAA_REQUEST(15);
+  NEW_DAA_REQUEST(15),
+  NEW_DAA_UPLOAD(16);
```
fboulnois: Consider splitting this into `NEW_DAA_UPLOAD_RESEARCHER` and `NEW_DAA_UPLOAD_SO` to be able to more easily differentiate it when auditing emails or logs.

src/main/java/org/broadinstitute/consent/http/mail/message/NewDAAUploadResearcherMessage.java
```diff
@@ -0,0 +1,20 @@
+package org.broadinstitute.consent.http.mail.message;
+
+import com.sendgrid.helpers.mail.Mail;
+import java.io.Writer;
+import javax.mail.MessagingException;
+
+public class NewDAAUploadResearcherMessage extends MailMessage {
+  private final String NEW_DAA_UPLOAD = "New DAA uploaded for DAC in DUOS";
```
fboulnois:
```suggestion
  private final String NEW_DAA_UPLOAD = "New DAA uploaded and sent to researcher for DAC in DUOS";
```

Perhaps a message like this.

src/main/java/org/broadinstitute/consent/http/mail/message/NewDAAUploadSOMessage.java
```diff
@@ -0,0 +1,20 @@
+package org.broadinstitute.consent.http.mail.message;
+
+import com.sendgrid.helpers.mail.Mail;
+import java.io.Writer;
+import javax.mail.MessagingException;
+
+public class NewDAAUploadSOMessage extends MailMessage {
+  private final String NEW_DAA_UPLOAD = "New DAA uploaded for DAC in DUOS";
```
fboulnois:
```suggestion
  private final String NEW_DAA_UPLOAD = "New DAA uploaded and sent to SO for DAC in DUOS";
```

Ditto here.

src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
```diff
@@ -588,6 +594,107 @@ void testGetUsersAsRoleInvalidRole() {
...
+  void testGetUsersByDaaIdNoMatchingUsers() {
+    User u1 = generateUser();
+    int dacId = RandomUtils.nextInt(0,50);
+    Instant now = Instant.now();
+    LibraryCard card = generateLibraryCard(u1);
+    int daaId = daaDAO.createDaa(card.getUserId(), now, card.getUserId(), now, dacId);
+    DataAccessAgreement daa = new DataAccessAgreement();
+    daa.setDaaId(daaId);
+    when(daaDAO.findById(any())).thenReturn(daa);
+//    libraryCardDAO.createLibraryCardDaaRelation(card.getId(), daaId);
```
fboulnois: This should be removed or fixed.

fboulnois (commented):
A few suggestions, will review again when tests pass:

src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
```diff
@@ -679,7 +679,6 @@ void testGetUsersByDaaIdNoMatchingUsers() {
     DataAccessAgreement daa = new DataAccessAgreement();
     daa.setDaaId(daaId);
     when(daaDAO.findById(any())).thenReturn(daa);
-//    libraryCardDAO.createLibraryCardDaaRelation(card.getId(), daaId);
```
rushtong: nit: remove if not necessary

src/main/java/org/broadinstitute/consent/http/mail/freemarker/NewDAAUploadResearcherModel.java
```diff
@@ -0,0 +1,44 @@
+package org.broadinstitute.consent.http.mail.freemarker;
+
+public class NewDAAUploadResearcherModel {
```
rushtong, line 3: I think this and `NewDAAUploadSOModel`  are a great cases for a Record class.

aarohinadkarni: I remember on my previous email PR, Olivia had suggested this and when changing the Model class to a Record class, I ran into FreeMarker errors. Similarly, when I changed the `NewDAAUploadResearcherModel` and `NewDAAUploadSOModel` to Java records, I got the same FreeMarker errors. Back then, I had created [this ticket](https://broadworkbench.atlassian.net/browse/DCJ-441) in the backlog to change all of our FreeMarkerTemplate models to be defined as Records. I have moved the ticket created from the DUOS project to the DCJ project.

src/main/resources/assets/paths/daaUploadByDacId.yaml
```diff
@@ -0,0 +1,31 @@
...
+      schema:
+        type: integer
+    - name: newDaaName
+      in: path
+      description: The name of the newly uploaded DAA
+      required: true
+      schema:
+        type: string
+  tags:
+    - DAA
```
rushtong: I think this should also be tagged in `Mail`

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -430,4 +430,24 @@ public Response removeDacDaaRelationship(
       return createExceptionResponse(e);
     }
   }
+
+  @POST
+  @RolesAllowed({ADMIN, CHAIRPERSON})
+  @Path("{dacId}/upload/{oldDaaId}/{newDaaName}")
```
rushtong: I think I'd prefer to use IDs here instead of name strings. If we did use IDs, we should be able to get the name from the previous DAA.
```suggestion
  @Path("{dacId}/upload/{oldDaaId}/{newDaaId}")
```
I'm also on the fence for the word `upload`. In this scenario, the upload has already happened. What we're doing here is pure messaging. Maybe something like `migrated`, `updated`, or `moved`?

rushtong: We could also verify that the old and new DAAs really exist before sending the email.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -430,4 +430,24 @@ public Response removeDacDaaRelationship(
...
+      @Auth AuthUser authUser,
+      @PathParam("dacId") Integer dacId,
+      @PathParam("oldDaaId") Integer oldDaaId,
+      @PathParam("newDaaName") String newDaaName
+      ) {
+    try {
+      User user = userService.findUserByEmail(authUser.getEmail());
+      Dac dac = dacService.findById(dacId);
+      String dacName = dac.getName();
+      daaService.sendNewDaaEmails(user, oldDaaId, dacName, newDaaName);
```
rushtong, line 448: It would be good to try and fail fast here if the old or new DAA were not found

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
aarohinadkarni: My main concern in using the newDaaId is that when this email endpoint will be getting called is at the same time that the endpoint that is creating this new DAA. So, we wouldn't necessarily have the newDaaId at the time of calling this endpoint because all of those endpoints get called upon save of the update to the DAC. 

Additionally, I believe we _do_ verify that the old DAA exists before sending the email in the [DAAService method](https://github.com/DataBiosphere/consent/pull/2340/files#diff-9ca6884b52d6c055b2f5b979f6a62a840fddde450921c232b8f5ad622abdc902R146), but again, since the new DAA is not created yet, we won't be able to check if the new DAA exists before sending the email.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
aarohinadkarni: I can move the logic of checking if the old or new DAA exist from the service class to the resource class!

rushtong (approved):
Looks good! 👍🏽

fboulnois (approved):
looks good, thank you 👍

PR #2342, opened 2024-06-17
[DCJ-445] Associate DAA with DAC object

aarohinadkarni:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-445

### Summary
Populates the DAC object with its associated DAA through the creation of a new DacWithDaasReducer, along with modification of existing tests.

Note: this PR's tests will not pass until https://github.com/DataBiosphere/consent/pull/2347 is merged.

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/db/DacDAO.java
```diff
@@ -101,7 +105,33 @@ public interface DacDAO extends Transactional<DacDAO> {
...
+      + "fso.create_user_id AS create_user_id, "
+      + "fso.update_date AS update_date, "
+      + "fso.update_user_id AS update_user_id, "
+      + "fso.deleted AS deleted, "
+      + "fso.delete_user_id AS delete_user_id "
+      + "FROM dac "
+      + "LEFT JOIN dac_daa dd ON dac.dac_id = dd.dac_id "
+      + "LEFT JOIN data_access_agreement daa ON dd.daa_id = daa.daa_id "
+      + "LEFT JOIN file_storage_object fso ON daa.daa_id::text = fso.entity_id "
+      + "WHERE dac.dac_id = :dacId")
```
rushtong: Can we make this a text block?

src/main/java/org/broadinstitute/consent/http/db/mapper/DacWithDaasReducer.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Objects;
+import org.broadinstitute.consent.http.models.Dac;
+import org.broadinstitute.consent.http.models.DataAccessAgreement;
+import org.broadinstitute.consent.http.models.FileStorageObject;
+import org.broadinstitute.consent.http.util.ConsentLogger;
+import org.jdbi.v3.core.mapper.MappingException;
+import org.jdbi.v3.core.result.LinkedHashMapRowReducer;
+import org.jdbi.v3.core.result.RowView;
+
+public class DacWithDaasReducer implements LinkedHashMapRowReducer<Integer, Dac>,
```
rushtong: I can see this class merged with `DacWithDatasetsReducer` and then perhaps renamed to `DacReducer`.

src/main/java/org/broadinstitute/consent/http/db/mapper/DacWithDaasReducer.java
```diff
@@ -0,0 +1,53 @@
...
+              rowView.getColumn("dac_id", Integer.class), id -> rowView.getRow(Dac.class));
+
+      if (Objects.nonNull(rowView.getColumn("daa_daa_id", Integer.class))) {
+        DataAccessAgreement daa = new DataAccessAgreement();
+
+        try {
+          if (dac != null && rowView.getColumn("daa_daa_id", Integer.class) != null) {
+            daa = rowView.getRow(DataAccessAgreement.class);
+          }
+        } catch (MappingException e) {
```
rushtong: We should at least log these at the warn level so we can follow up. I see in other reducer class cases we're logging messages, but we should start logging the full stack trace to make debugging easier.

src/main/java/org/broadinstitute/consent/http/models/Dac.java
```diff
@@ -32,6 +32,8 @@ public class Dac {
 
   private String email;
 
+  private DataAccessAgreement associatedDaa;
```
rushtong, line 35: We should add this new field to the api docs as well: `src/main/resources/assets/schemas/Dac.yaml`

rushtong (commented):
This looks great, and works as expected! See minor comments inline.

src/main/java/org/broadinstitute/consent/http/db/mapper/DacReducer.java
```diff
@@ -25,6 +27,32 @@ public void accumulate(Map<Integer, Dac> container, RowView rowView) {
           container.computeIfAbsent(
               rowView.getColumn("dac_id", Integer.class), id -> rowView.getRow(Dac.class));
 
+      if (Objects.nonNull(rowView.getColumn("daa_daa_id", Integer.class))) {
+        DataAccessAgreement daa = new DataAccessAgreement();
+
+        try {
+          if (dac != null && rowView.getColumn("daa_daa_id", Integer.class) != null) {
```
fboulnois: Since this is already in the `nonNull` case, this condition doesn't need to be checked again.

src/main/java/org/broadinstitute/consent/http/db/mapper/DacReducer.java
```diff
@@ -25,6 +27,32 @@ public void accumulate(Map<Integer, Dac> container, RowView rowView) {
...
+            daa = rowView.getRow(DataAccessAgreement.class);
+          }
+        } catch (MappingException e) {
+          logException(e);
```
fboulnois: This `try ... catch` is already in a `try` condition, so perhaps we should simplify these so we don't have nested `try ... catch` conditions.

src/test/java/org/broadinstitute/consent/http/db/DacDAOTest.java
```diff
@@ -97,13 +98,37 @@ void testFindAllDACUsersBySearchString_case2() {
...
+        "Test_" + RandomStringUtils.random(20, true, true),
+        "Test_" + RandomStringUtils.random(20, true, true),
+        new Date());
+    User user = createUser();
+    Integer daaId = daaDAO.createDaa(user.getUserId(), new Date().toInstant(), user.getUserId(), new Date().toInstant(), id);
+    DataAccessAgreement daa = daaDAO.findById(daaId);
+    daaDAO.createDacDaaRelation(id, daaId);
     Dac dac = dacDAO.findById(id);
     assertEquals(id, dac.getDacId());
+    assertEquals(daa.getDaaId(), dac.getAssociatedDaa().getDaaId());
```
fboulnois: I recommend pulling out the `daa` here and then executing these assertions.

fboulnois (commented):
A few suggestions:

src/main/resources/assets/schemas/Dac.yaml
```diff
@@ -12,6 +12,9 @@ properties:
   email:
     type: string
     description: Optional DAC Email address
+  associatedDaa:
+    type: string
```
rushtong: I think this should be:
```suggestion
    $ref: './DataAccessAgreement.yaml'
```
See `LibraryCard.yaml` for a working example - we have some other examples that are unfortunately incorrect

src/main/java/org/broadinstitute/consent/http/db/mapper/DacReducer.java
```diff
@@ -0,0 +1,83 @@
...
+  @Override
+  public void accumulate(Map<Integer, Dac> container, RowView rowView) {
+
+    try {
+
+      Dac dac =
+          container.computeIfAbsent(
+              rowView.getColumn("dac_id", Integer.class), id -> rowView.getRow(Dac.class));
+
+      if (Objects.nonNull(rowView.getColumn("daa_daa_id", Integer.class))) {
```
fboulnois: Since the logic is a little different now, I think it's worth adding some unit tests to cover the new lines.

fboulnois (commented):
Another recommendation:

src/main/java/org/broadinstitute/consent/http/db/mapper/DacReducer.java
```diff
@@ -0,0 +1,83 @@
...
+          daa.setFile(fso);
+        }
+
+        if (daa.getDaaId() != null) {
+          dac.setAssociatedDaa(daa);
+        }
+
+      }
+
+      if (Objects.nonNull(rowView.getColumn("dataset_id", Integer.class))) {
```
rushtong: I realize that git isn't understanding that this is the same as the original version in `DacWithDatasetsReducer`, but we should try and migrate away from the `Objects.nonNull` pattern we have been using.

src/main/java/org/broadinstitute/consent/http/db/mapper/DacReducer.java
```diff
@@ -0,0 +1,83 @@
...
+        Dataset dataset = rowView.getRow(Dataset.class);
+
+        //aliased columns must be set directly
+        String dsAlias = rowView.getColumn("dataset_alias", String.class);
+        if (dsAlias != null) {
+            dataset.setAlias(Integer.parseInt(dsAlias));
+        }
```
rushtong: The original code had a try around the integer parsing which was necessary in the past due to invalid dev data. We should try and maintain that condition here:
```
          if (dsAlias != null) {
            try {
              dataset.setAlias(Integer.parseInt(dsAlias));
            } catch (Exception e) {
              logException("Exception parsing dataset alias: " + dsAlias, e);
            }
          }
```

src/main/resources/assets/schemas/Dac.yaml
```diff
@@ -12,6 +12,8 @@ properties:
   email:
     type: string
     description: Optional DAC Email address
+  associatedDaa:
+    $ref: './DataAccessAgreement.yaml'
```
rushtong: Something about this isn't working - all schema entities are showing up with no values. I'm digging into this currently

rushtong (commented):
Still looking into the swagger issue - see comments inline

src/main/resources/assets/schemas/Dac.yaml
rushtong: Ok, I think the correct magic here is to use the `allOf` syntax:
```
  associatedDaa:
    description: The Data Access Agreement selected for this DAC
    allOf:
      - $ref: './DataAccessAgreement.yaml'
```
The current syntax **_should_** work, but for some reason, breaks all of the other schemas. 

Additionally, we could use an entry in `api-docs.yaml` for the DAA:
```
    Data Access Agreement:
      $ref: './schemas/DataAccessAgreement.yaml'
```

aarohinadkarni: Hm, that's very weird— thank you for digging into that! I've implemented those changes.

rushtong (approved):
Thank you for the changes!

src/main/java/org/broadinstitute/consent/http/db/mapper/DacWithDaasReducer.java
aarohinadkarni: I've noticed now, when logging these at the warn level, the logs are getting _polluted_ with these warning messages when the getById endpoint is called. @rushtong I'm not sure if this is what we want...

https://github.com/DataBiosphere/consent/assets/68303030/490b74d9-336f-41e8-971a-dc630132c7f1

rushtong: I think I know what that is ... that's from this:
```
rowView.getColumn("daa_daa_id", Integer.class) != null
```
If there is no value in column `daa_daa_id`, then the value that comes back is `0`. What you need to do is check to see if it's also `> 0`. That's because jdbc casts null to `0` since it is a primitive `int` and can't be a proper `null`

src/main/java/org/broadinstitute/consent/http/db/mapper/DacReducer.java
aarohinadkarni: Would these unit tests be in the DacDAOTest class?

src/main/java/org/broadinstitute/consent/http/db/mapper/DacWithDaasReducer.java
aarohinadkarni: Ah, very interesting. That seemed to do the trick— thank you!

src/main/java/org/broadinstitute/consent/http/db/mapper/DacReducer.java
fboulnois: Yes, I think it would make sense to add in that class.

src/test/java/org/broadinstitute/consent/http/db/DacDAOTest.java
```diff
@@ -97,13 +98,38 @@ void testFindAllDACUsersBySearchString_case2() {
...
+    daaDAO.createDacDaaRelation(id, daaId);
     Dac dac = dacDAO.findById(id);
+    DataAccessAgreement dacDaa = dac.getAssociatedDaa();
     assertEquals(id, dac.getDacId());
+    assertEquals(daa.getDaaId(), dacDaa.getDaaId());
+    assertEquals(daa.getCreateUserId(), dacDaa.getCreateUserId());
+    assertEquals(daa.getCreateDate(), dacDaa.getCreateDate());
+    assertEquals(daa.getUpdateUserId(), dacDaa.getUpdateUserId());
+    assertEquals(daa.getUpdateDate(), dacDaa.getUpdateDate());
+    assertEquals(daa.getInitialDacId(), dacDaa.getInitialDacId());
```
okotsopoulos, line 196: Just curious, would an equality check between `daa` and `dacDaa` work here?  That way the assertions won't short-circuit on the first mismatch.

src/main/java/org/broadinstitute/consent/http/db/mapper/DacReducer.java
okotsopoulos, line 1: Formatting note: this file looks to contain formatting that would have been updated by a linter (e.g. additional newlines, early line breaks).  Is there a linter you can run to update the formatting to match style conventions?

okotsopoulos (approved):
Looks good! A few small questions, not blockers to merge.

src/test/java/org/broadinstitute/consent/http/db/DacDAOTest.java
aarohinadkarni: I don't think it would work here, given that we have not overridden the equals method in the `DataAccessAgreement` class.

src/main/java/org/broadinstitute/consent/http/db/mapper/DacReducer.java
aarohinadkarni: Yes, we do! I forgot to run it, but I'll run that & commit the changes now.

PR #2347, opened 2024-06-24
DCJ-445: Update Row Helper Conditional

rushtong:
### Addresses
Partial fix for https://broadworkbench.atlassian.net/browse/DCJ-445

### Summary
This updates minor logic checks for whether a column exists or not. Previous library functionality would have thrown an error, but current functionality does not. JDBI treats null integer values as 0's which are not valid as primary key ids. In those cases, ensure that the value is > 0 before trying to set them on objects.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/mapper/DaaMapper.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/mapper/DaaMapper.java
@@ -15,9 +15,9 @@ public class DaaMapper implements RowMapper<DataAccessAgreement>, RowMapperHelpe
   @Override
   public DataAccessAgreement map(ResultSet resultSet, StatementContext statementContext) throws SQLException {
     DataAccessAgreement daa;
-    Integer daaId = hasColumn(resultSet, "daa_id") && resultSet.getInt("daa_id") > 0 ? resultSet.getInt("daa_id") : null;
+    Integer daaId = hasNonZeroColumn(resultSet, "daa_id") ? resultSet.getInt("daa_id") : null;
     if (daaId == null) {
-      if (hasColumn(resultSet, "daa_daa_id")) {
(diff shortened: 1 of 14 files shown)


aarohinadkarni (approved):
looks good once tests pass 👍

src/main/java/org/broadinstitute/consent/http/db/mapper/DaaMapper.java
```diff
@@ -15,7 +15,7 @@ public class DaaMapper implements RowMapper<DataAccessAgreement>, RowMapperHelpe
   @Override
   public DataAccessAgreement map(ResultSet resultSet, StatementContext statementContext) throws SQLException {
     DataAccessAgreement daa;
-    Integer daaId = hasColumn(resultSet, "daa_id") ? resultSet.getInt("daa_id") : null;
+    Integer daaId = hasColumn(resultSet, "daa_id") && resultSet.getInt("daa_id") > 0 ? resultSet.getInt("daa_id") : null;
```
fboulnois: Suggestion: pull this out into a `hasNonZeroColumn` method that these mappers can use.

fboulnois (commented):
One suggestion since this conditional is repeated so many times:

src/main/java/org/broadinstitute/consent/http/db/mapper/DaaMapper.java
okotsopoulos: +1 this suggestion.

okotsopoulos (approved):
Looks good, ditto @fboulnois ' suggestion!

src/main/java/org/broadinstitute/consent/http/db/mapper/DaaMapper.java
rushtong: Great idea, thanks!

PR #2351, opened 2024-07-01
[DCJ-481] Add "Broad DAA" flag to appropriate DAA

aarohinadkarni:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-481

### Summary
Adds a flag to a DataAccessAgreement saying it is the `broadDaa` if it was created by the broad DAC. This is to be used in the UI so that a `daaId` that may vary among environments does not have to be used to help match the Broad DAA. Additionally, makes sure that the DAAs are getting flagged appropriately when queried by DAA or by DAC, along with appropriate tests.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/DataAccessAgreement.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/DataAccessAgreement.java
@@ -99,8 +99,4 @@ public class DataAccessAgreement {
       this.dacs.add(dac);
     }
   }
-
-
-
-
 }
--- a/src/main/java/org/broadinstitute/consent/http/models/DataAccessAgreement.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/DataAccessAgreement.java
@@ -14,6 +14,7 @@ public class DataAccessAgreement {
   private Integer initialDacId;
   private FileStorageObject file;
   private List<Dac> dacs;
+  private Boolean broadDaa = false;
 
   public Integer getDaaId() {
     return daaId;
@@ -79,6 +80,14 @@ public class DataAccessAgreement {
     this.dacs = dacs;
   }
 
+  public Boolean getBroadDaa() {
+    return broadDaa;
+  }
+
+  public void setBroadDaa(Boolean broadDaa) {
+    this.broadDaa = broadDaa;
+  }
+
   public void addDac(Dac dac) {
     if (this.dacs == null) {
       this.dacs = new ArrayList<>();
@@ -90,4 +99,8 @@ public class DataAccessAgreement {
       this.dacs.add(dac);
     }
   }
+
+
+
+
 }
--- a/src/main/java/org/broadinstitute/consent/http/service/DaaService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DaaService.java
@@ -107,6 +107,8 @@ public class DaaService implements ConsentLogger {
     daaDAO.deleteDacDaaRelation(dacId, daaId);
   }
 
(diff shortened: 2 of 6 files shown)


src/main/java/org/broadinstitute/consent/http/models/DataAccessAgreement.java
```diff
@@ -90,4 +99,8 @@ public void addDac(Dac dac) {
       this.dacs.add(dac);
     }
   }
+
+
+
+
```
rushtong: nit: remove extraneous empty line additions

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
```diff
@@ -101,11 +107,38 @@ public void removeDacFromDaa(Integer dacId, Integer daaId) {
     daaDAO.deleteDacDaaRelation(dacId, daaId);
   }
 
-  public List<DataAccessAgreement> findAll() {
-    List<DataAccessAgreement> daas = daaDAO.findAll();
-    if (daas != null) {
-      return daas;
+  public boolean isBroadDAA(int daaId, List<DataAccessAgreement> allDaas, List<Dac> allDacs) {
```
rushtong, line 112: I think we could use some additional documentation/TODO/ticket reference here to indicate that this is not the permanent solution to this problem.

rushtong (commented):
Logically, this looks good. See minor comments inline.

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
aarohinadkarni: Is it not the permanent solution to the problem? I know this isn't an ideal solution, but I don't think we had an idea for a more "permanent" solution if I'm not mistaken.

rushtong: Sorry I missed this response. A more permanent solution would be to add a database column to the DAC table that defines this and allow for admins to set this in the UI or by API. There are other potential implementations, but we shouldn't be relying on names/creation dates to infer this value.

fboulnois: I agree with @rushtong . Instead of a boolean I would recommend an `enum` column that can be set to Broad, and the code should check that instead.

fboulnois (changes requested):
I agree with the earlier comment.

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
aarohinadkarni: @rushtong @fboulnois Should I implement those changes in this PR itself or create a new ticket to include this additional scoping? I'm not sure what we want to do in alignment with our "v1" of this work.

rushtong: My suggestion is just to document the better approach. This approach is OK for now. The ideal solution requires more time and effort that we have available right now.

fboulnois: That seems reasonable. I will unblock approval to be able to move this PR forward.

fboulnois (approved):
👍 given our previous discussion and limitations.

rushtong (approved):
👍🏽

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
aarohinadkarni: Ticket: https://broadworkbench.atlassian.net/browse/DCJ-498

PR #2355, opened 2024-07-08
[DCJ-493] Allow SOs to delete a LC-DAA relationship

aarohinadkarni:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-493

### Summary
Fixes bug that wasn't allowing users who are strictly SOs to remove a DAA-LC relationship & adds additional test to confirm new logic.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
@@ -190,9 +190,6 @@ public class DaaResource extends Resource implements ConsentLogger {
       @PathParam("userId") Integer userId) {
     try {
       User user = userService.findUserByEmail(authUser.getEmail());
-      if (!user.hasUserRole(UserRoles.ADMIN) && !user.hasUserRole(UserRoles.SIGNINGOFFICIAL)) {
-        return Response.status(Status.FORBIDDEN).build();
-      }
       List<LibraryCard> libraryCards = libraryCardService.findLibraryCardsByUserId(userId);
       for (LibraryCard libraryCard : libraryCards) {
         libraryCardService.removeDaaFromLibraryCard(libraryCard.getId(), daaId);
--- a/src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
@@ -190,7 +190,7 @@ public class DaaResource extends Resource implements ConsentLogger {
       @PathParam("userId") Integer userId) {
     try {
       User user = userService.findUserByEmail(authUser.getEmail());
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -190,7 +190,7 @@ public Response deleteDaaForUser(
       @PathParam("userId") Integer userId) {
     try {
       User user = userService.findUserByEmail(authUser.getEmail());
-      if (!user.hasUserRole(UserRoles.ADMIN)) {
+      if (!user.hasUserRole(UserRoles.ADMIN) && !user.hasUserRole(UserRoles.SIGNINGOFFICIAL)) {
```
rushtong: I don't think we need this check at all. The `@RolesAllowed({ADMIN, SIGNINGOFFICIAL})` annotation does this for us already

aarohinadkarni: If I don't have this, the tests don't pass. Should I remove the tests as well then?

rushtong: I don't think the tests are accurate. They are trying to enforce role based restrictions **_inside_** the resource method, but role based restrictions are applied _**outside**_ the method.

src/test/java/org/broadinstitute/consent/http/resources/DaaResourceTest.java
```diff
@@ -498,7 +518,7 @@ void testDeleteDaaForUserForbidden() {
     libraryCard.setUserId(user.getUserId());
     libraryCard.setDaaIds(List.of(daaId));
 
-    when(userService.findUserByEmail(any())).thenReturn(user);
+    when(userService.findUserByEmail(any())).thenThrow(new ForbiddenException());
```
rushtong: At a technical level, this doesn't reflect what is actually happening. `userService.findUserByEmail()` isn't hit in the resource method that's called in the test. What's actually happening at runtime is that the method is never called in the case that the user doesn't have the correct role - that is controlled by the `UserAuthorizer` class. I think these tests are not necessary here. To actually test this case, I think this would need to run it in a container (see https://www.dropwizard.io/en/release-4.0.x/manual/testing.html#test-containers) but we haven't set up the framework for that yet so it's definitely out of scope for this PR.

aarohinadkarni: I was thinking something similar, but then I got confused on how I would be able to test these conditions. I can remove the related tests.

rushtong (approved):
👍🏽  Thank you!

fboulnois (approved):
Looks good 👍

PR #2356, opened 2024-07-10
DCJ-504 Bug fix for null alias in queries

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-504

### Summary
Fixes fallout from https://github.com/DataBiosphere/consent/pull/2347 where we explicitly did not map zero-value aliases. There are several in staging/prod so we need to maintain this functionality.

----

- Verify all tests go green


okotsopoulos (approved):
Thanks for investigating and fixing!

fboulnois (approved):
Good catch 👍

PR #2358, opened 2024-07-15
DCJ-515: Fix json parsing for dataset updates

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-515

### Summary
This PR updates the `Gson` parsing for dataset update json objects to ensure that integer values in json are not converted to floats.

### Testing Strategy
With a recent database update, I tested updating `DUOS-000866` (ID: `2368`) in the following way:

Visit [local swagger v3 PUT endpoint](https://local.dsde-dev.broadinstitute.org:27443/#/Dataset/put_api_dataset_v3__datasetId_) with the dataset id of 2368 and this payload: 
```
{
  "name": "AJ test controlled access dataset (GRU)",
  "dacId": 3,
  "properties": [
    {
      "propertyName": "Data Location",
      "propertyValue": "TDR Location"
    },
    {
      "propertyName": "File Types",
      "propertyValue": [
        {}
      ]
    },
    {
      "propertyName": "Access Management",
      "propertyValue": "controlled"
    },
    {
      "propertyName": "# of participants",
      "propertyValue": 12
    }
  ]
}
```
Repeated calls to the same endpoint should work (previously, subsequent updates would fail due to integer conversion problems). The response should include all properties as defined in the above payload. Calls to the [local GET endpoint](https://local.dsde-dev.broadinstitute.org:27443/#/Dataset/get_api_dataset_v2__id_) should now also return the updated values correctly.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/DatasetUpdate.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/DatasetUpdate.java
@@ -1,7 +1,6 @@
 package org.broadinstitute.consent.http.models;
 
 import com.google.gson.Gson;
-import com.google.gson.ToNumberPolicy;
 import java.util.List;
 import org.broadinstitute.consent.http.util.gson.GsonUtil;
 
@@ -11,8 +10,7 @@ public record DatasetUpdate(
     List<DatasetProperty> properties
 ) {
 
-  private static final Gson GSON = GsonUtil.gsonBuilderWithAdapters().setObjectToNumberStrategy(
-      ToNumberPolicy.LONG_OR_DOUBLE).create();
+  private static final Gson GSON = GsonUtil.gsonBuilderWithAdapters().create();
 
   public DatasetUpdate(String json) {
     this(GSON.fromJson(json, DatasetUpdate.class).getName(),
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/consent/http/models/DatasetUpdate.java
```diff
@@ -1,12 +1,24 @@
...
-public record DatasetUpdate (
+public record DatasetUpdate(
     String name,
     Integer dacId,
     List<DatasetProperty> properties
-){
+) {
+
+  private static final Gson GSON = GsonUtil.gsonBuilderWithAdapters().setObjectToNumberStrategy(
+      ToNumberPolicy.LONG_OR_DOUBLE).create();
```
fboulnois: Should we move `setObjectToNumberStrategy` into `gsonBuilderWithAdapters` globally? That way we do not need to remember to depend on this particular `GSON` object for any DatasetUpdate related methods. I also don't think that we want integers to be treated as floats in the general case, so the default seems reasonable to me.

fboulnois (approved):
One comment:

src/main/java/org/broadinstitute/consent/http/models/DatasetUpdate.java
rushtong: That's a great point, thank you!

aarohinadkarni (approved):
looks good to me! 👍

PR #2359, opened 2024-07-16
[DCJ-469] Delete DAA & DAA-DAC relationships within DAC deletion

aarohinadkarni:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-469

### Summary
These changes ensure that when a DAC is deleted, all DAAs that were created under this DAC get deleted along with the deletion of any existing relationships of this DAC with any DAAs, specifically the "Broad" DAA. Tests are also modified/included in this PR.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DacService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DacService.java
@@ -185,10 +185,8 @@ public class DacService {
 
   public void deleteDac(Integer dacId) throws IllegalArgumentException, SQLException {
     Dac fullDac = dacDAO.findById(dacId);
-    List<Dac> dacs = dacDAO.findAll();
-    Optional<Dac> broadDac = dacs.stream()
-        .filter(dac -> dac.getName().toLowerCase().contains("broad") && dac.getDacId().equals(dacId)).findFirst();
-    if (broadDac.isPresent()) {
+    // TODO: Broad DAC logic will be updated with DCJ-498 to not be reliant on name
+    if (fullDac.getName().toLowerCase().contains("broad")) {
       throw new IllegalArgumentException("This is the Broad DAC, which can not be deleted.");
     }
     try {
--- a/src/main/java/org/broadinstitute/consent/http/service/DacService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DacService.java
@@ -185,17 +185,14 @@ public class DacService {
 
   public void deleteDac(Integer dacId) throws IllegalArgumentException, SQLException {
     Dac fullDac = dacDAO.findById(dacId);
-    List<DataAccessAgreement> daas = daaDAO.findAll();
     List<Dac> dacs = dacDAO.findAll();
     Optional<Dac> broadDac = dacs.stream()
         .filter(dac -> dac.getName().toLowerCase().contains("broad") && dac.getDacId().equals(dacId)).findFirst();
     if (broadDac.isPresent()) {
       throw new IllegalArgumentException("This is the Broad DAC, which can not be deleted.");
     }
-    List<DataAccessAgreement> filteredDaas = daas.stream()
-        .filter(daa -> daa.getInitialDacId().equals(dacId)).toList();
     try {
-      dacServiceDAO.deleteDacAndDaas(fullDac, filteredDaas);
+      dacServiceDAO.deleteDacAndDaas(fullDac);
     } catch (IllegalArgumentException e) {
       throw new IllegalArgumentException("Could not find DAC with the provided id: " + dacId);
     }
--- a/src/main/java/org/broadinstitute/consent/http/service/DacService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DacService.java
@@ -5,6 +5,7 @@ import static java.util.stream.Collectors.groupingBy;
 import com.google.inject.Inject;
 import jakarta.ws.rs.BadRequestException;
 import jakarta.ws.rs.NotFoundException;
+import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
@@ -32,7 +33,7 @@ import org.broadinstitute.consent.http.models.Election;
 import org.broadinstitute.consent.http.models.Role;
 import org.broadinstitute.consent.http.models.User;
 import org.broadinstitute.consent.http.models.UserRole;
-
+import org.broadinstitute.consent.http.service.dao.DacServiceDAO;
 public class DacService {
 
   private final DacDAO dacDAO;
@@ -43,11 +44,13 @@ public class DacService {
   private final DaaDAO daaDAO;
   private final VoteService voteService;
   private final DaaService daaService;
+  private final DacServiceDAO dacServiceDAO;
 
   @Inject
   public DacService(DacDAO dacDAO, UserDAO userDAO, DatasetDAO dataSetDAO,
       ElectionDAO electionDAO, DataAccessRequestDAO dataAccessRequestDAO,
-      VoteService voteService, DaaService daaService, DaaDAO daaDAO) {
+      VoteService voteService, DaaService daaService, DaaDAO daaDAO,
+      DacServiceDAO dacServiceDAO) {
     this.dacDAO = dacDAO;
     this.userDAO = userDAO;
     this.dataSetDAO = dataSetDAO;
@@ -56,6 +59,7 @@ public class DacService {
     this.daaDAO = daaDAO;
     this.voteService = voteService;
     this.daaService = daaService;
+    this.dacServiceDAO = dacServiceDAO;
   }
 
   public List<Dac> findAll() {
@@ -179,18 +183,22 @@ public class DacService {
     dacDAO.updateDac(name, description, email, updateDate, dacId);
   }
 
-  public void deleteDac(Integer dacId) {
+  public void deleteDac(Integer dacId) throws IllegalArgumentException, SQLException {
     Dac fullDac = dacDAO.findById(dacId);
     List<DataAccessAgreement> daas = daaDAO.findAll();
     List<Dac> dacs = dacDAO.findAll();
-    daas.stream()
-        .filter(daa -> daa.getInitialDacId().equals(dacId) && ! daaService.isBroadDAA(daa.getDaaId(), daas, dacs))
-        .forEach(daa -> daaService.deleteDaa(daa.getDaaId()));
-    if (daaService.isBroadDAA(fullDac.getAssociatedDaa().getDaaId(), daas, dacs)) {
-      daaService.removeDacFromDaa(dacId, fullDac.getAssociatedDaa().getDaaId());
+    Optional<Dac> broadDac = dacs.stream()
+        .filter(dac -> dac.getName().toLowerCase().contains("broad") && dac.getDacId().equals(dacId)).findFirst();
+    if (broadDac.isPresent()) {
+      throw new IllegalArgumentException("This is the Broad DAC, which can not be deleted.");
     }
-    dacDAO.deleteDacMembers(dacId);
-    dacDAO.deleteDac(dacId);
+    List<DataAccessAgreement> filteredDaas = daas.stream()
+        .filter(daa -> daa.getInitialDacId().equals(dacId) && ! daaService.isBroadDAA(daa.getDaaId(), daas, dacs)).toList();
+    try {
+      dacServiceDAO.deleteDacAndDaas(fullDac, filteredDaas);
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException("Could not find DAC with the provided id: " + dacId);
+    }
   }
 
   public User findUserById(Integer id) throws IllegalArgumentException {
--- a/src/main/java/org/broadinstitute/consent/http/service/DacService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DacService.java
@@ -34,8 +34,13 @@ import org.broadinstitute.consent.http.models.Role;
 import org.broadinstitute.consent.http.models.User;
 import org.broadinstitute.consent.http.models.UserRole;
 import org.broadinstitute.consent.http.service.dao.DacServiceDAO;
-public class DacService {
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/consent/http/service/DacService.java
```diff
@@ -180,6 +180,15 @@ public void updateDac(String name, String description, String email, Integer dac
...
   public void deleteDac(Integer dacId) {
+    Dac fullDac = dacDAO.findById(dacId);
+    List<DataAccessAgreement> daas = daaDAO.findAll();
+    List<Dac> dacs = dacDAO.findAll();
+    daas.stream()
+        .filter(daa -> daa.getInitialDacId().equals(dacId) && ! daaService.isBroadDAA(daa.getDaaId(), daas, dacs))
+        .forEach(daa -> daaService.deleteDaa(daa.getDaaId()));
+    if (daaService.isBroadDAA(fullDac.getAssociatedDaa().getDaaId(), daas, dacs)) {
+      daaService.removeDacFromDaa(dacId, fullDac.getAssociatedDaa().getDaaId());
+    }
```
rushtong, line 203: This approach doesn't group all of the deletes into a logical batch, but runs them individually. Instead, we should be batching them up in a single transaction. For example, have a look at [NihServiceDAO](https://github.com/DataBiosphere/consent/blob/e03a530e474446dbc848797cde82116128280346/src/main/java/org/broadinstitute/consent/http/service/dao/NihServiceDAO.java#L26) for an approach that keeps all of the updates in a single transaction. In the case of `DacService`, we don't have a service dao, so this is a good chance to add one.

Another option is to wrap everything up in a single statement similarly to what we do with DAA deletes in [DaaDAO](https://github.com/DataBiosphere/consent/blob/e03a530e474446dbc848797cde82116128280346/src/main/java/org/broadinstitute/consent/http/db/DaaDAO.java#L196), but that might be a little less maintainable than the service DAO approach.

src/main/java/org/broadinstitute/consent/http/service/DacService.java
```diff
@@ -179,9 +183,22 @@ public void updateDac(String name, String description, String email, Integer dac
...
+    Dac fullDac = dacDAO.findById(dacId);
+    List<DataAccessAgreement> daas = daaDAO.findAll();
+    List<Dac> dacs = dacDAO.findAll();
+    Optional<Dac> broadDac = dacs.stream()
+        .filter(dac -> dac.getName().toLowerCase().contains("broad") && dac.getDacId().equals(dacId)).findFirst();
+    if (broadDac.isPresent()) {
+      throw new IllegalArgumentException("This is the Broad DAC, which can not be deleted.");
+    }
+    List<DataAccessAgreement> filteredDaas = daas.stream()
+        .filter(daa -> daa.getInitialDacId().equals(dacId) && ! daaService.isBroadDAA(daa.getDaaId(), daas, dacs)).toList();
```
rushtong: I think there is some duplicated work here. First, we grab all the dacs to find out if we have the Broad one.
Then, again, in the filter, we're checking for a similar thing with `daaService.isBroadDAA`. If we've errored out with a Broad DAC, then it stands to follow that we are only working with non-Broad DAAs.

src/main/java/org/broadinstitute/consent/http/service/DacService.java
```diff
@@ -179,9 +183,22 @@ public void updateDac(String name, String description, String email, Integer dac
...
+    try {
+      dacServiceDAO.deleteDacAndDaas(fullDac, filteredDaas);
```
rushtong: Due to the logic above, I don't think we need `filteredDaas` at all. We should be able to do all the necessary work with just the dac id.

src/main/java/org/broadinstitute/consent/http/service/dao/DacServiceDAO.java
```diff
@@ -0,0 +1,68 @@
...
+
+        Update dacDaaDeletion = handler.createUpdate(deleteFromDacDaa);
+        dacDaaDeletion.bind("dacId", dac.getDacId());
+        dacDaaDeletion.execute();
+
+        daas.forEach(id -> {
+          Update daaDeletion = handler.createUpdate(deleteFromDaa);
+          daaDeletion.bind("dacId", dac.getDacId());
+          daaDeletion.execute();
+        });
```
rushtong: I think this is incorrect. Imagine if DAAs were passed in with different DAC ids. We should not be deleting DAAs created by other DACs, we should only be deleting the DAAs created by the DAC we're focused on.

rushtong (commented):
See comments inline - I think we're really close!

src/main/java/org/broadinstitute/consent/http/service/dao/DacServiceDAO.java
aarohinadkarni: That's why we are passing in the `filteredDaas` to this method, only the DAAs that were created by this DAC. I was doing the filtering in the service method so that only the deletions would have to happen in the ServiceDAO method.

src/main/java/org/broadinstitute/consent/http/service/DacService.java
aarohinadkarni: I think it's just a question of whether we want to filter out the DAAs not made by this DAC here, or in the ServiceDAO method.

src/main/java/org/broadinstitute/consent/http/service/dao/DacServiceDAO.java
rushtong: That's why I think this is incorrect. This method has to stand on it's own and not be dependent on some external process to do pre-work before getting here.

src/main/java/org/broadinstitute/consent/http/service/dao/DacServiceDAO.java
```diff
@@ -0,0 +1,74 @@
...
+
+        Update dacDaaDeletion = handler.createUpdate(deleteFromDacDaa);
+        dacDaaDeletion.bind("dacId", dac.getDacId());
+        dacDaaDeletion.execute();
+
+        filteredDaas.forEach(id -> {
+          Update daaDeletion = handler.createUpdate(deleteFromDaa);
+          daaDeletion.bind("dacId", dac.getDacId());
+          daaDeletion.execute();
+        });
```
rushtong: This can be a single line delete. In this construction, we have N DAAs all with a dac id of, for example, 10. We will be making N delete calls that are exactly the same: `DELETE FROM dac_daa WHERE dac_id = :dacId`

src/main/java/org/broadinstitute/consent/http/service/dao/DacServiceDAO.java
```diff
@@ -0,0 +1,74 @@
...
+
+  public void deleteDacAndDaas(Dac dac)
+      throws IllegalArgumentException, SQLException {
+    // fail fast
+    if (Objects.isNull(dac)) {
+      throw new IllegalArgumentException("Invalid DAC");
+    }
+    List<DataAccessAgreement> daas = daaDAO.findAll();
+    List<DataAccessAgreement> filteredDaas = daas.stream()
+        .filter(daa -> daa.getInitialDacId().equals(dac.getDacId())).toList();
```
rushtong: See comment where `filteredDaas` is used - we don't need to filter them nor do we need to get all DAAs to populate this list.

src/main/java/org/broadinstitute/consent/http/service/dao/DacServiceDAO.java
```diff
@@ -0,0 +1,67 @@
...
+
+  @Inject
+  public DacServiceDAO(Jdbi jdbi) {
+    this.jdbi = jdbi;
+  }
+
+  public void deleteDacAndDaas(Dac dac)
+      throws IllegalArgumentException, SQLException {
+    // fail fast
+    if (Objects.isNull(dac)) {
```
rushtong: minor nit: 
```suggestion
    if (dac == null) {
```

src/test/java/org/broadinstitute/consent/http/service/dao/DacServiceDAOTest.java
```diff
@@ -0,0 +1,132 @@
...
+    List<Dac> dacListRemoved = dacDAO.findAll();
+    assertEquals(1, dacListRemoved.size());
+    List<DataAccessAgreement> daaListRemoved = daaDAO.findAll();
+    assertEquals(1, daaListRemoved.size());
+  }
+
+  /**
+   * @return A list of 5 dacs
+   */
+  private List<Dac> getDacs() {
```
rushtong: minor nit, naming. In the method calls, it looks like we're getting pre-existing DACs. Maybe something like `createMockDACs` would be more informative.

src/test/java/org/broadinstitute/consent/http/service/dao/DacServiceDAOTest.java
```diff
@@ -0,0 +1,132 @@
...
+      fail("Delete should not fail");
+    }
+    List<Dac> dacListRemoved = dacDAO.findAll();
+    assertEquals(1, dacListRemoved.size());
+    List<DataAccessAgreement> daaListRemoved = daaDAO.findAll();
+    assertEquals(1, daaListRemoved.size());
+  }
+
+  /**
+   * @return A list of 5 dacs
```
rushtong: This actually creates 4 DACs, not 5. The math didn't line up for me in the delete method where we generate 5, delete 1, and then have 3 left. Maybe if the number of DACs to create were an argument here, it would be a lot more clear about what should exist and shouldn't.

rushtong (commented):
Some interim comments while I keep reviewing...

src/test/java/org/broadinstitute/consent/http/service/dao/DacServiceDAOTest.java
```diff
@@ -0,0 +1,116 @@
...
+      // DAC
+      int dacId = dacDAO.createDac(RandomStringUtils.randomAlphabetic(10), RandomStringUtils.randomAlphabetic(10), RandomStringUtils.randomAlphabetic(10),  new Date());
+      // DAA
+      int daaId = daaDAO.createDaa(superUser.getUserId(), new Date().toInstant(), superUser.getUserId(), new Date().toInstant(), dacId);
+      // DAC-DAA Association
+      daaDAO.createDacDaaRelation(dacId, daaId);
+      // Library Card User with Institution
+      User lcUser = createUser();
+      int institutionId = institutionDAO.insertInstitution(
+          RandomStringUtils.randomAlphabetic(10),
```
fboulnois: I think we should name these variables something more meaningful so it's clearer what the test is testing.

fboulnois (approved):
Looks good, one suggestion:

src/test/java/org/broadinstitute/consent/http/service/dao/DacServiceDAOTest.java
rushtong: I increased the comments here due to the extensive nature of the interconnections between DACs and all of the entities that are updated.

fboulnois: The `RandomStringUtils.randomAlphabetic(10)` aren't very meaningful, it may be worthwhile replacing these with `itDirectorName`, `itDirectorEmail`, `institutionUrl`, and so on so we can see what these values represent.

src/main/java/org/broadinstitute/consent/http/service/DacService.java
```diff
@@ -179,9 +183,17 @@ public void updateDac(String name, String description, String email, Integer dac
...
-    dacDAO.deleteDac(dacId);
+  public void deleteDac(Integer dacId) throws IllegalArgumentException, SQLException {
+    Dac fullDac = dacDAO.findById(dacId);
+    // TODO: Broad DAC logic will be updated with DCJ-498 to not be reliant on name
+    if (fullDac.getName().toLowerCase().contains("broad")) {
+      throw new IllegalArgumentException("This is the Broad DAC, which can not be deleted.");
+    }
+    try {
+      dacServiceDAO.deleteDacAndDaas(fullDac);
+    } catch (IllegalArgumentException e) {
```
okotsopoulos, line 199: I see we swallow the original exception in full, is there any value in logging it for debugging even if we don't propagate it to the user?

src/test/java/org/broadinstitute/consent/http/service/DacServiceTest.java
```diff
@@ -229,6 +235,30 @@ void testDeleteDac() {
...
+
+    assertThrows(IllegalArgumentException.class, () -> {
+      service.deleteDac(dacs.get(0).getDacId());
+    });
+  }
+
+  @Test
+  void testDeleteDacBroadDac() {
+    List<Dac> dacs = getDacs();
+    dacs.get(0).setName("Broad DAC");
```
okotsopoulos: Would it be worth making this a parameterized test with a variety of DAC names that would trigger this condition?

src/test/java/org/broadinstitute/consent/http/service/DacServiceTest.java
```diff
@@ -229,6 +235,30 @@ void testDeleteDac() {
...
+    when(dacDAO.findById(any())).thenReturn(dacs.get(0));
+    initService();
+
+    assertThrows(IllegalArgumentException.class, () -> {
+      service.deleteDac(1);
```
okotsopoulos: I'm not following this call -- I see that we made the "zero"-th DAC the Broad DAC.  Should we be calling `service.deleteDac(0)` instead to test this condition?

okotsopoulos (approved):
Looks good -- a few small questions.

src/test/java/org/broadinstitute/consent/http/service/DacServiceTest.java
rushtong: I'll refactor this test to make it more clear about the which DAC this is.

src/main/java/org/broadinstitute/consent/http/service/DacService.java
rushtong: Yes, there is some value to that, thank you

PR #2368, opened 2024-07-25
DCJ-539: New API for returning dataset + study summaries

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-539

### Summary
This PR adds a v3 version of the get datasets API that is much more streamlined than the v2 version which is essentially unusable on dev. We don't use the v2 version any longer in the UI, so that should be deprecated.

#### Example Output
```
[
  {
    "dataset_id": 361,
    "dataset_name": "TOPMed Amish_HMB_IRB_MDS_phs000956.v2.p1.c2 V3",
    "identifier": "DUOS-000004"
  },
  ...
  {
    "dataset_id": 2582,
    "dataset_name": "All of Us (Controlled Tier) - Set 1",
    "identifier": "DUOS-001077",
    "study_id": 6076,
    "study_name": "All of Us (Controlled Tier)"
  },
  ...
]
```

----

- Verify all tests go green

### DIFF
--- a/src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
@@ -6,7 +6,6 @@ import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assertions.fail;
 
 import com.google.gson.JsonObject;
 import java.sql.Timestamp;
@@ -39,6 +38,7 @@ import org.broadinstitute.consent.http.models.DataUseBuilder;
 import org.broadinstitute.consent.http.models.Dataset;
 import org.broadinstitute.consent.http.models.DatasetAudit;
 import org.broadinstitute.consent.http.models.DatasetProperty;
+import org.broadinstitute.consent.http.models.DatasetStudySummary;
 import org.broadinstitute.consent.http.models.DatasetSummary;
 import org.broadinstitute.consent.http.models.Dictionary;
 import org.broadinstitute.consent.http.models.Election;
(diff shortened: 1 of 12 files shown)


fboulnois (approved):
looks good 👍

src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
```diff
@@ -56,6 +56,30 @@
...
 
+  @Test
+  void testFindAllDatasetStudySummariesDatasetAndStudy() {
+    Dataset dataset = insertDataset();
+    Study study = insertStudyWithProperties();
+    datasetDAO.updateStudyId(dataset.getDataSetId(), study.getStudyId());
+
+    List<DatasetStudySummary> summaries = datasetDAO.findAllDatasetStudySummaries();
+    assertFalse(summaries.isEmpty());
+    assertEquals( 1, summaries.size());
```
okotsopoulos: If you'd like, these two assertions could be consolidated thusly (ditto throughout):
```suggestion
    assertThat(summaries, hasSize(1));
```

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -808,6 +808,15 @@ void testFindAllDatasetsAvailableToUserAsCustodian() {
     assertEquals(HttpStatusCodes.STATUS_CODE_OK, response.getStatus());
   }
 
+  @Test
+  void testFindAllDatasetStudySummaries() {
+    when(userService.findUserByEmail(any())).thenReturn(user);
+    when(datasetService.findAllDatasetStudySummaries()).thenReturn(List.of());
+    initResource();
+    Response response = resource.findAllDatasetStudySummaries(authUser);
+    assertEquals(HttpStatusCodes.STATUS_CODE_OK, response.getStatus());
```
okotsopoulos, line 817: Is there any value in asserting that the expected dataset summaries are returned, perhaps worth exercising in a different test case than the empty summary list one?

okotsopoulos (approved):
Looks good! A few suggestions for tests.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
rushtong: Personally, I don't find value in adding that unless the called method, `findAllDatasetStudySummaries`, performed some kind of logical operation on the what the service layer returned. In this case, we're returning whatever the service layer provides regardless of what it is. Similarly, in this case, the service layer is also doing a plain pass-through to the DAO layer. That's where we really care about what the entities look like.

PR #2371, opened 2024-08-10
DCJ-602: Convert fetch all datasets API to a streamed output

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-602

### Summary
This PR resurrects the v2 dataset api in a streaming format. Instead of pulling ALL datasets into memory before converting to JSON for a response, this approach pulls datasets in batches of 50 and streams the output to the client. The current approach maxes out the application memory and crashes the running pod.

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -314,15 +315,11 @@ public Response updateDataset(@Auth AuthUser authUser, @Context UriInfo info,
   @Produces("application/json")
   @PermitAll
   @Path("/v2")
-  @Deprecated
-  public Response findAllDatasetsAvailableToUser(@Auth AuthUser authUser,
-      @QueryParam("asCustodian") Boolean asCustodian) {
```
rushtong, line 319: The `asCustodian` feature was an early experiment and never actually used in practice.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -62,6 +68,7 @@ public class DatasetService implements ConsentLogger {
   private final StudyDAO studyDAO;
   private final DatasetServiceDAO datasetServiceDAO;
   private final UserDAO userDAO;
+  public Integer DATASET_BATCH_SIZE = 50;
```
okotsopoulos: Since this isn't a static final, the [convention](https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names) is to use lowerCamelCase instead of SCREAMING_SNAKE_CASE.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -681,4 +711,9 @@ private Integer updateStudyFromConversion(User user, Dataset dataset,
     return studyId;
   }
 
+  @VisibleForTesting
+  public void setDatasetBatchSize(Integer DATASET_BATCH_SIZE) {
```
okotsopoulos: If this method is only used in testing, how about removing the `public` modifier and leveraging the default behavior?

src/main/resources/assets/paths/datasetV2.yaml
```diff
@@ -1,26 +1,12 @@
...
-  summary: Get Datasets
+  summary: Get Streamed Datasets
   description: | 
-    Returns Dataset list filtered by authenticated user role.
-    Admins can see all datasets, both active and inactive.
-    Chairpersons can see all active datasets and inactive ones 
-    associated to any of their Data Access Committees. 
-    All other users can see active datasets.
-    If the optional `asCustodian` parameter is provided, all datasets that the user has created
-    or is specified as a custodian for will be returned.
```
okotsopoulos, line 11: Is this initial context no longer relevant?  I know `asCustodian` has been removed but how about the rest?

src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
```diff
@@ -888,6 +888,25 @@ void testFindAllDatasets() {
...
+    List<Integer> datasetIds = datasetDAO.findAllDatasetIds();
+    assertFalse(datasetIds.isEmpty());
+    assertEquals(datasetList.size(), datasetIds.size());
+    // Assert that the two lists of ids have all of the same values
+    List<Integer> insertedDatasetIds = datasetList.stream().map(Dataset::getDataSetId).toList();
+    assertTrue(datasetIds.containsAll(insertedDatasetIds));
+    assertTrue(insertedDatasetIds.containsAll(datasetIds));
+    // Assert the list order was preserved in the query
+    assertEquals(datasetList.get(0).getDataSetId(), datasetIds.get(0));
+    assertEquals(datasetList.get(datasetList.size() - 1).getDataSetId(), datasetIds.get(datasetIds.size() - 1));
```
okotsopoulos: I believe all of these assertions could be consolidated thusly:

```suggestion
    List<Integer> insertedDatasetIds = IntStream.range(1, 5).mapToObj(i -> {
      Dataset dataset = insertDataset();
      return dataset.getDataSetId();
    }).toList();
    List<Integer> datasetIds = datasetDAO.findAllDatasetIds();

    assertThat(datasetIds, contains(insertedDatasetIds.toArray()));
```

This matcher checks that the two lists of IDs have exactly the same values in the same order, no more and no less.

Docs: https://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#contains(E...)

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -791,21 +795,26 @@ void testDownloadDatasetApprovedUsersError() {
...
+    Response response = resource.findAllDatasetsStreaming(authUser);
     assertEquals(HttpStatusCodes.STATUS_CODE_OK, response.getStatus());
+    var entity = (StreamingOutput) response.getEntity();
+    var baos = new ByteArrayOutputStream();
+    entity.write(baos);
+    var entityString = baos.toString();
+    Type listOfDatasetsType = new TypeToken<List<Dataset>>() {}.getType();
+    List<Dataset> returnedDatasets = gson.fromJson(entityString, listOfDatasetsType);
+    assertFalse(returnedDatasets.isEmpty());
+    assertEquals(1, returnedDatasets.size());
```
okotsopoulos: These assertions could be consolidated like so:
```suggestion
    assertThat(returnedDatasets, hasSize(1));
```

src/test/java/org/broadinstitute/consent/http/service/DatasetServiceTest.java
```diff
@@ -626,6 +629,32 @@ void testFindAllDatasetsByUser() {
     verify(datasetDAO, times(0)).findDatasetsByAuthUserEmail(any());
   }
 
+  @Test
+  void testFindAllDatasetsAsStreamingOutput() throws Exception {
+    var datasets = getDatasets(RandomUtils.nextInt(10, 20));
+    when(datasetDAO.findAllDatasetIds()).thenReturn(datasets.stream().map(Dataset::getDataSetId).toList());
+    datasets.forEach(d -> {
+      System.out.println(d);
```
okotsopoulos: Is this print statement left over from debugging?

src/test/java/org/broadinstitute/consent/http/service/DatasetServiceTest.java
```diff
@@ -626,6 +629,32 @@ void testFindAllDatasetsByUser() {
...
+    var output = datasetService.findAllDatasetsAsStreamingOutput();
+    var baos = new ByteArrayOutputStream();
+    output.write(baos);
+    var datasetsJson = baos.toString();
+    var listOfDatasetsType = new TypeToken<List<Dataset>>() {}.getType();
+    var gson = GsonUtil.buildGson();
+    List<Dataset> returnedDatasets = gson.fromJson(datasetsJson, listOfDatasetsType);
+    assertFalse(returnedDatasets.isEmpty());
+    assertEquals(datasets.size(), returnedDatasets.size());
+    datasets.forEach(d -> assertTrue(returnedDatasets.contains(d)));
```
okotsopoulos: I think that the `contains` matcher can also be useful here:
```suggestion
    assertThat(returnedDatasets, contains(datasets.toArray());
```

okotsopoulos (approved):
Looks good, I've left some questions and test suggestions.

I am curious how this endpoint is meant to be interacted with.  I haven't seen this streaming approach in other services.  Is there a reason you've favored streaming over pagination?

rushtong:
> Looks good, I've left some questions and test suggestions.
> 
> I am curious how this endpoint is meant to be interacted with. I haven't seen this streaming approach in other services. Is there a reason you've favored streaming over pagination?

The intention for this API is purely for developer usage. A paginated approach would be great for UX, but we've been moving our UX to use the datasets in the ES index instead of interacting with APIs that look up datasets directly. My goal here is mostly for debugging purposes - I need a way to manually reconcile queries from support about what should or should not be visible to users when they do interact with the ES index. Given that the current implementation is unusable, it makes quick lookups more painful.

src/main/resources/assets/paths/datasetV2.yaml
rushtong: Yes, all of these conditions are no longer valid.

src/test/java/org/broadinstitute/consent/http/db/DatasetDAOTest.java
rushtong: Great suggestion, thank you!

src/test/java/org/broadinstitute/consent/http/service/DatasetServiceTest.java
okotsopoulos: Or `containsInAnyOrder` if order is not guaranteed.

rushtong: In this case, we added an ID sort in the query, so I do actually care about order.

fboulnois (approved):
thank you 👍

PR #2375, opened 2024-08-16
DCJ-620: Add timing to TDR endpoints

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-620

### Summary
Add timing to interesting TDR specific endpoints to eventually track in https://grafana.dsp-devops.broadinstitute.org/d/8OxKprrSk/duos-dashboard?orgId=1

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
@@ -1,5 +1,6 @@
 package org.broadinstitute.consent.http.resources;
 
+import com.codahale.metrics.annotation.Timed;
 import com.google.inject.Inject;
 import io.dropwizard.auth.Auth;
 import jakarta.annotation.security.PermitAll;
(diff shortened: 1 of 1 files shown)


okotsopoulos (approved):
Thanks!

I know that Consent is not a Spring Boot project (if it were its REST endpoints would be instrumented for free) but just double-checking -- is there a way for timing to automatically be enabled for Consent's endpoints by default?

rushtong:
> Thanks!
> 
> I know that Consent is not a Spring Boot project (if it were its REST endpoints would be instrumented for free) but just double-checking -- is there a way for timing to automatically be enabled for Consent's endpoints by default?

Not that I'm aware of. Dropwizard does do a ton of instrumentation for other things like database calls, but not for endpoints, unfortunately. There could be a way to hack the jersey annotations that DW uses, but I haven't tried looking into that.

fboulnois (approved):
looks good 👍

PR #2377, opened 2024-08-19
DCJ-400: Update logback versions

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-400

### Summary
Logback changed the location of the access module which dependabot can't fix.
See also: https://github.com/DataBiosphere/consent-ontology/pull/980

----

- Verify all tests go green

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -15,7 +15,7 @@
     <maven.compiler.target>21</maven.compiler.target>
     <liquibase.version>4.29.1</liquibase.version>
     <dropwizard.version>4.0.4</dropwizard.version>
-    <logback.version>1.4.14</logback.version>
+    <logback.version>1.5.7</logback.version>
     <pact.version>4.6.13</pact.version>
     <postgres.version>42.7.3</postgres.version>
     <surefire.version>3.3.1</surefire.version>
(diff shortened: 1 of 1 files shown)


fboulnois (approved):
that's frustrating!

pom.xml
```diff
@@ -547,10 +547,11 @@
...
 
     <!-- Security update. See https://broadworkbench.atlassian.net/browse/DUOS-2826 -->
+    <!-- Logback Access module moved to new location. See https://logback.qos.ch/news.html -->
     <dependency>
-        <groupId>ch.qos.logback</groupId>
-        <artifactId>logback-access</artifactId>
-        <version>${logback.version}</version>
+        <groupId>ch.qos.logback.access</groupId>
+        <artifactId>common</artifactId>
+        <version>2.0.3</version>
```
s-rubenstein, line 554: How come this is hardcoded now, rather than referencing the variable?

rushtong: The versions are different now

s-rubenstein: Ah, okay, I see. So the specific module is referenced separately. That's really frustrating.

s-rubenstein (approved):

pom.xml
rushtong: Yeah, it's a very non-standard way to do things - not sure why they took this approach.

PR #2390, opened 2024-08-25
DCJ-639: Delete access_rp Table and Refactor Send Reminder Email

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-639

### Summary
* Refactor (remove) usages of the old `access_rp` table
* Minor changes to the Send Reminder code path which is primarily affected

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/mail/freemarker/TemplateModel.java
+++ b/src/main/java/org/broadinstitute/consent/http/mail/freemarker/TemplateModel.java
@@ -1,16 +1,11 @@
 package org.broadinstitute.consent.http.mail.freemarker;
 
+/* This model works for the Send Reminder Message template */
 public class TemplateModel {
-
-  /* This model works for templates: collect and reminder. The placeholders are all the same. */
-
-  private String userName;
-
-  private String electionType;
-
-  private String entityName;
-
-  private String serverUrl;
+  private final String userName;
+  private final String electionType;
+  private final String entityName;
+  private final String serverUrl;
 
   public TemplateModel(String user, String election, String entityName, String serverUrl) {
     this.userName = user;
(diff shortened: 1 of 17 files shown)


src/main/resources/freemarker/reminder.html
```diff
@@ -21,7 +21,7 @@
         </tr>
         <tr>
             <td style="padding: 15px 15px 0px; font-family: 'Montserrat', sans-serif; font-size: 16px; color: #1F3B50; text-align: justify; line-height: 25px;">
-                Please vote on ${electionType} Review case id ${entityName}. Click on the following link to log your vote:
+                Please vote on Data Access Request: ${entityName}. Click on the following link to log your vote:
```
fboulnois, line 24: If the `electionType` is no longer being used, should it be deprecated in the template and other locations?

fboulnois (commented):
See below:

src/main/resources/freemarker/reminder.html
rushtong: Great point, yes, I'll make that change too.

fboulnois (approved):
👍

src/main/java/org/broadinstitute/consent/http/mail/freemarker/TemplateModel.java
```diff
@@ -1,20 +1,13 @@
 package org.broadinstitute.consent.http.mail.freemarker;
 
+/* This model works for the Send Reminder Message template */
 public class TemplateModel {
```
okotsopoulos: If this model is now only single-use, would it be worth renaming to something more specific?

okotsopoulos (approved):
Looks good!

src/main/java/org/broadinstitute/consent/http/mail/freemarker/TemplateModel.java
rushtong: Great point. I struggle sometimes with the scope of changes. I'll bite the bullet and make it more specific. This class, by the way, goes back to the earliest days of the app ... the very first commit has this message:
```
/* This model works for every template we have: collect, new-case and reminder. The placeholders are all the same. */
```

PR #2396, opened 2024-08-28
DCJ-645: Add Data Submitter to API roles

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-645

### Summary
Updates to studies require access to getting the study's create user id. Currently, a user without admin, chair, or member roles will instead get a 403 response when the page tries to populate the study data, thereby preventing the page from rendering correctly.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
@@ -162,7 +162,7 @@ public class UserResource extends Resource {
   @GET
   @Path("/{userId}")
   @Produces("application/json")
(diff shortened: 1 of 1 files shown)


fboulnois (approved):
👍

okotsopoulos (approved):
Looks good -- is there any precedent for unit testing to verify allowed roles for Consent endpoints?

rushtong:
> Looks good -- is there any precedent for unit testing to verify allowed roles for Consent endpoints?

Unfortunately, no, we don't. To do so would be a different approach to testing that we haven't explored. We would need to spin up a version of the app with this specific resource class and then try to hit it with different user role combinations. I'm sure we could do that - it's just hasn't been a priority for us.

PR #2397, opened 2024-08-29
DCJ-570: Add Associated DAA and DAA File info to Get All DACs 

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-570

### Summary

- Minor update to the GET ALL DACs endpoint to populate the associated DAA and file storage object.
- Additionally, standardize the DAA/FSO query info across the two instances of the relevant queries in DaaDAO
- Additionally, update/simplfy tests for clarity

----

- Verify all tests go green


fboulnois (approved):
good improvements 👍

src/main/java/org/broadinstitute/consent/http/db/DacDAO.java
```diff
@@ -39,17 +39,53 @@ public interface DacDAO extends Transactional<DacDAO> {
...
+        fso.media_type AS fso_media_type,
+        fso.deleted AS fso_deleted,
+        fso.delete_user_id AS fso_delete_user_id,
+        fso.create_date AS fso_create_date,
+        fso.create_user_id AS fso_create_user_id,
+        fso.update_date AS fso_update_date,
+        fso.update_user_id AS fso_update_user_id
       FROM dac
       LEFT OUTER JOIN dataset d ON dac.dac_id = d.dac_id
+      LEFT JOIN dac_daa dd ON dac.dac_id = dd.dac_id
```
okotsopoulos: I think that `LEFT OUTER JOIN` and `LEFT JOIN` are equivalent -- does this repo favor one representation over the over?  Can we make the joins consistent within this query?

okotsopoulos (approved):
Looks reasonable, one comment below.

Are there any concerns with how these additional joins may impact query performance?  (You can tell where my head is at lately :) )

src/main/java/org/broadinstitute/consent/http/db/DacDAO.java
rushtong: They are equivalent, thank you - I'll make update them here.

rushtong:
> Are there any concerns with how these additional joins may impact query performance? (You can tell where my head is at lately :) )

I do have concerns about the join efficiency here - we've been bitten by this especially on dataset entity queries. I am somewhat confident that we won't hit any performance problems for some time given that we should have a fairly low number of DACs - I estimate they'll stay under 100 for the next several years.

PR #2403, opened 2024-09-23
DCJ-685: Fix bugs in the conversion of study datasets to registration schema consent groups.

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-685

### Summary
This fixes an issue where we were trying to get the DAC ID and `DataUse` values from non-existent `DatasetProperties` instead of from the `Dataset` directly.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/builder/ConsentGroupFromDataset.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/builder/ConsentGroupFromDataset.java
@@ -1,34 +1,19 @@
 package org.broadinstitute.consent.http.models.dataset_registration_v1.builder;
 
 import static org.broadinstitute.consent.http.models.dataset_registration_v1.builder.DatasetRegistrationSchemaV1Builder.accessManagement;
-import static org.broadinstitute.consent.http.models.dataset_registration_v1.builder.DatasetRegistrationSchemaV1Builder.col;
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/builder/ConsentGroupFromDataset.java
```diff
@@ -40,7 +25,7 @@
 public class ConsentGroupFromDataset {
 
   public ConsentGroup build(Dataset dataset) {
-    if (Objects.nonNull(dataset)) {
+    if (dataset != null) {
```
rjohanek, line 28: is there a reason for this change? there is a use of nonNull below that is unchanged. Do we want to avoid using nonNull?

rjohanek (approved):
change and testing changes look good

src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/builder/ConsentGroupFromDataset.java
rushtong: Yes, we've been moving in this direction for a while, originally based on Phil's observation that the use of `Objects.**` is meant for use in streaming operations.

fboulnois (approved):
core changes look good 👍

PR #2404, opened 2024-09-24
DCJ-688: Return default responses when Sam fails

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DCJ-688

## Summary
When Sam fails, downstream systems that rely on Sam data can fail. In the case that we cannot communicate with Sam, we would now return generic default response objects instead of failing outright. This allows us to be less tightly coupled and enables continued DUOS operations when Sam is unavailable. When Sam recovers, we would then be able to respond with real Sam responses.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
@@ -14,11 +14,15 @@ import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
 import jakarta.ws.rs.ServerErrorException;
 import jakarta.ws.rs.core.MediaType;
+import java.io.IOException;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import org.broadinstitute.consent.http.configurations.ServicesConfiguration;
 import org.broadinstitute.consent.http.exceptions.ConsentConflictException;
 import org.broadinstitute.consent.http.models.AuthUser;
@@ -30,8 +34,6 @@ import org.broadinstitute.consent.http.models.sam.UserStatusDiagnostics;
 import org.broadinstitute.consent.http.models.sam.UserStatusInfo;
 import org.broadinstitute.consent.http.util.ConsentLogger;
 import org.broadinstitute.consent.http.util.HttpClientUtil;
-import org.checkerframework.checker.nullness.qual.NonNull;
-import org.checkerframework.checker.nullness.qual.Nullable;
 
 public class SamDAO implements ConsentLogger {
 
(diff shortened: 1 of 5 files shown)


rjohanek (commented):
I have a couple questions. I'm wondering what this gets us. I like the idea of being less tightly coupled with Sam, but how far can a user get with these mocked responses? Do they completely unblock users when Sam is down? Additionally, I have some questions about the new error messages. I'm wondering if they give away unnecessary/unhelpful implementation details by including terms that represent internal concepts like,  UserStatusDiagnostics, for example. Finally, I have a concern about the maintenance burden of returning the "most recent Terms of Service Text" because then this text will need to be updated here every time the terms of service is updated. Maybe that is so infrequent this isn't a concern, or maybe we just don't want to promise it is most recent. Not sure, I'd love to hear your thoughts.

src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
```diff
@@ -33,7 +33,7 @@
 import org.checkerframework.checker.nullness.qual.NonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
-public class SamDAO implements ConsentLogger {
+public class SamDAO implements ConsentLogger, SamDefaults {
```
pshapiro4broad: In general I'd recommend using embedding over inheritance, as it's more flexible and doesn't necessarily change an object's external interface. So instead of `implements SamDefaults` I'd add a `SamDefaults` member which is initialized in the constructor.

src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
```diff
@@ -54,10 +54,17 @@ public SamDAO(HttpClientUtil clientUtil, ServicesConfiguration configuration) {
...
-      logException("Error getting resource types from Sam: " + response.getStatusMessage(),
-          new ServerErrorException(response.getStatusMessage(), response.getStatusCode()));
+    HttpResponse response;
+    try {
+      response = executeRequest(request);
+      if (!response.isSuccessStatusCode()) {
+        logException("Error getting resource types from Sam: " + response.getStatusMessage(),
+            new ServerErrorException(response.getStatusMessage(), response.getStatusCode()));
+      }
+    } catch (ServerErrorException e) {
```
pshapiro4broad: It looks like all `SamDAO` APIs are called from a single class, `SamService`, and its members have the same signature as the defaults class. Why not wrap the DAO calls in `SamService` instead of doing it here?

rushtong:
@raejohanek - Some answers below:

> how far can a user get with these mocked responses? 

Any user should be completely unblocked within DUOS. Anything outside of DUOS is not addressed so if a user is directed outside of DUOS, then they will be subject to the Sam dependencies of that destination.

> Do they completely unblock users when Sam is down? 

Yes.

> Additionally, I have some questions about the new error messages. I'm wondering if they give away unnecessary/unhelpful implementation details by including terms that represent internal concepts like, UserStatusDiagnostics, for example. 

Great points. I agree with you, they do expose too much. I'll revisit the defaults so that they are less informative.

> I have a concern about the maintenance burden of returning the "most recent Terms of Service Text" because then this text will need to be updated here every time the terms of service is updated. Maybe that is so infrequent this isn't a concern, or maybe we just don't want to promise it is most recent. Not sure, I'd love to hear your thoughts.

The update frequency here is almost non-existent, so this is not a concern for me. If there is an update, we would need to revisit this anyway since we create a custom DUOS version of the default.

src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
```diff
@@ -54,10 +54,17 @@ public SamDAO(HttpClientUtil clientUtil, ServicesConfiguration configuration) {
...
-    if (!response.isSuccessStatusCode()) {
-      logException("Error getting resource types from Sam: " + response.getStatusMessage(),
-          new ServerErrorException(response.getStatusMessage(), response.getStatusCode()));
+    HttpResponse response;
+    try {
+      response = executeRequest(request);
+      if (!response.isSuccessStatusCode()) {
+        logException("Error getting resource types from Sam: " + response.getStatusMessage(),
+            new ServerErrorException(response.getStatusMessage(), response.getStatusCode()));
+      }
```
pshapiro4broad: I believe this can be done using a generic wrapper API to avoid the code duplication. All the error-wrapped APIs are of type `Function<T, R>` except `getToSText()` which is `Supplier<T>`. Something like
```
  interface FunctionWithError<T, R> {
    R apply(T t) throws ServerErrorException;
  }

  private <T, R> R handleError(FunctionWithError<T, R> wrapped, Function<T, R> onError, T input,
      Supplier<String> logMessage) {
    try {
      return wrapped.apply(input);
    } catch (ServerErrorException e) {
      logException("Sam is down, returning %s".formatted(logMessage.get()), e);
      return onError.apply(input);
    }
  }
```

pshapiro4broad:
Regarding the overall goal of the PR, what about duos' dependance on services that also depend on Sam, such as TDR? (Maybe TDR is the only other dependency?)

rushtong:
> Regarding the overall goal of the PR, what about duos' dependance on services that also depend on Sam, such as TDR? (Maybe TDR is the only other dependency?)

TDR is the only external dependency we have that also relies on Sam. However, in the near future, we'll also be relying on ECM once it fully supports eRA Commons in the way that Shibboleth does.

src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
rushtong: Thanks for the suggestion, I'll dig into that

src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
rushtong: This separation style is something we've been doing in Terra services forever. I'm not inclined to tackle this as a refactor at this point. I'm not opposed to having a larger conversation around this.

pshapiro4broad: I don't think what I'm suggesting is a refactor of the code. To be clear (and to mix together all my suggestions) this is what I'm thinking.
```
public class SamDefaults {

  public List<ResourceType> getDefaultResourceTypes(AuthUser authUser) {
    return List.of();
  }

  public UserStatusInfo getDefaultUserStatusInfo(AuthUser authUser) {
...
```
```
public class SamService implements ConsentLogger {

  private final SamDAO samDAO;

  private final SamDefaults samDefaults = new SamDefaults();

  @Inject
  public SamService(SamDAO samDAO) {
    this.samDAO = samDAO;
  }

  interface FunctionWithError<T, R> {
    R apply(T t) throws Exception;
  }

  private <T, R> R handleError(FunctionWithError<T, R> wrapped, Function<T, R> onError, T input,
      Supplier<String> logMessage) throws Exception {
    try {
      return wrapped.apply(input);
    } catch (ServerErrorException e) {
      logException("Sam is down, returning %s".formatted(logMessage.get()), e);
      return onError.apply(input);
    }
  }
  public List<ResourceType> getResourceTypes(AuthUser authUser) throws Exception {
    return handleError(samDAO::getResourceTypes, samDefaults::getDefaultResourceTypes, authUser,
        () -> "default resource types");
  }
```

And then revert the changes to `SamDAO`. Returning the default values in `SamService` makes it easier to generically intercept each API call, since at that point the result is a java type, not a HTTP response object.

src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
rushtong: Let me look at that approach. I tend to use inheritance for trait-like behavior, which is what I'm doing in this case.

src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
rushtong: Thank you, that makes more sense than what I thought you were suggesting - it looks promising 👍🏽

src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
rushtong: There are three categories of functions in this service class:

1. <T> function(AuthUser user)
2. void function(AuthUser user)
3. <T> function()

In looking more closely at this, I'll need 3 different interfaces (one each for `Function`, `Consumer`, and `Supplier`) and 3 error handlers, one for each interface which just moves the code duplication to the service class instead of the DAO class.

src/main/java/org/broadinstitute/consent/http/db/SamDefaults.java
```diff
@@ -0,0 +1,81 @@
...
+    userStatusInfo.setUserEmail(authUser.getEmail());
+    userStatusInfo.setUserSubjectId(authUser.getEmail());
+    return userStatusInfo;
+  }
+
+  public UserStatusDiagnostics getDefaultUserStatusDiagnostics() {
+    UserStatusDiagnostics userStatusDiagnostics = new UserStatusDiagnostics();
+    userStatusDiagnostics.setAdminEnabled(false);
+    userStatusDiagnostics.setEnabled(true);
+    userStatusDiagnostics.setTosAccepted(true);
```
fboulnois, line 33: This suggests that a new user who hasn't accepted the TOS will have the diagnostics report back that they have accepted it. Is this expected?

fboulnois (commented):
I tend to agree with @rjohanek , I think it's unusual for a production service to return mocked responses when a dependent service is down. Would it be more desirable for downstream services to determine what they should do when the dependent service is down? Also, should downstream services be able to differentiate between mocked responses and true responses?

src/main/java/org/broadinstitute/consent/http/db/SamDefaults.java
rushtong: Yes, this is intentional. Without ToS, the user cannot proceed in DUOS. If Sam is down, there's no way to determine this state without replicating it on our side - which I am not in favor of doing. When Sam comes back, the user will eventually need to accept ToS.

rjohanek (approved):
if having users on DUOS who may not have accepted ToS is not a compliance issue, this looks good to me. If it is, is there a way to allow users who have already accepted ToS to use DUOS while blocking those who haven't, or there is no way to tell while Sam is down?

rushtong:
> If it is, is there a way to allow users who have already accepted ToS to use DUOS while blocking those who haven't, or there is no way to tell while Sam is down?

@raejohanek 
We can tell if Sam is down or not which is how we determine whether to send the default response objects or not. However, we don't mirror ToS acceptance in DUOS, we rely on Sam for that info, so if Sam is down, DUOS cannot determine if a user has or has not accepted ToS.

rushtong:
Closing this PR in favor of a different approach based on new requirements.

PR #2407, opened 2024-10-09
DCJ-705: Patch Dataset

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DCJ-705

### Summary
This PR adds a new PATCH endpoint that allows for Admins, Chairpersons, and Data Submitters to make updates to dataset objects in a very limited fashion. Updatable fields include the name, and the following dataset property fields:

* \# of participants
* URL
* Data Location

#### Validation Rules

1. \# of participants must be an integer value > 0
2. URL must be a valid URL
3. Data Location MUST be one of these values: AnVIL Workspace, Terra Workspace, TDR Location, or Not Determined

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
@@ -250,8 +250,8 @@ public class DatasetResource extends Resource {
                 HttpStatusCodes.STATUS_CODE_BAD_REQUEST))
             .build();
       }
-      Dataset patched = datasetRegistrationService.patchDataset(datasetId, user, patch);
-      return Response.noContent().entity(patched).build();
+      datasetRegistrationService.patchDataset(datasetId, user, patch);
+      return Response.noContent().build();
     } catch (Exception e) {
       return createExceptionResponse(e);
     }
--- a/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
@@ -233,12 +233,20 @@ public class DatasetResource extends Resource {
       if (!patch.isPatchable(existingDataset)) {
         return Response.notModified().entity(existingDataset).build();
       }
-      // Is new name unique?
+      // Validate DatasetPatch values
       List<String> existingNames = datasetService.findAllDatasetNames();
-      if (patch.name() != null && !patch.name().equals(existingDataset.getName()) && existingNames.contains(patch.name())) {
+      if (patch.name() != null && !patch.name().equals(existingDataset.getName())
+          && existingNames.contains(patch.name())) {
         return Response.status(HttpStatusCodes.STATUS_CODE_BAD_REQUEST)
             .entity(new Error(
                 "The new name for this dataset already exists: " + patch.name(),
+                HttpStatusCodes.STATUS_CODE_BAD_REQUEST))
+            .build();
+      }
+      if (!patch.validateProperties()) {
+        return Response.status(HttpStatusCodes.STATUS_CODE_BAD_REQUEST)
+            .entity(new Error(
+                "Properties are invalid",
                 HttpStatusCodes.STATUS_CODE_BAD_REQUEST))
             .build();
       }
--- a/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
@@ -13,6 +13,7 @@ import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.DELETE;
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.NotFoundException;
+import jakarta.ws.rs.PATCH;
 import jakarta.ws.rs.POST;
 import jakarta.ws.rs.PUT;
 import jakarta.ws.rs.Path;
@@ -40,6 +41,7 @@ import org.broadinstitute.consent.http.enumeration.UserRoles;
 import org.broadinstitute.consent.http.models.AuthUser;
 import org.broadinstitute.consent.http.models.DataUse;
 import org.broadinstitute.consent.http.models.Dataset;
+import org.broadinstitute.consent.http.models.DatasetPatch;
 import org.broadinstitute.consent.http.models.DatasetStudySummary;
 import org.broadinstitute.consent.http.models.DatasetSummary;
 import org.broadinstitute.consent.http.models.DatasetUpdate;
@@ -190,6 +192,46 @@ public class DatasetResource extends Resource {
     }
   }
 
+  /**
+   * This endpoint updates the dataset.
+   */
+  @PATCH
+  @Consumes({MediaType.APPLICATION_JSON})
+  @Produces({MediaType.APPLICATION_JSON})
+  @Path("/{datasetId}")
+  @RolesAllowed({ADMIN, CHAIRPERSON, DATASUBMITTER})
+  public Response patchByDatasetUpdate(
+      @Auth AuthUser authUser,
+      @PathParam("datasetId") Integer datasetId,
+      @FormDataParam("dataset") String json) {
+    try {
+      // TODO: Check permissions for chair, data submitter
+      if (json == null || json.isEmpty()) {
+        throw new BadRequestException("Dataset Patch is required");
+      }
+      Gson gson = GsonUtil.gsonBuilderWithAdapters().create();
+      DatasetPatch patch;
+      try {
+        patch = gson.fromJson(json, DatasetPatch.class);
+      } catch (Exception e) {
+        throw new BadRequestException("Unable to parse dataset patch: " + json);
+      }
+      Dataset existingDataset = datasetService.findDatasetById(datasetId);
+      if (existingDataset == null) {
+        throw new NotFoundException("Could not find the dataset with id: " + datasetId);
+      }
+      if (!patch.isPatchable(existingDataset)) {
+        return Response.notModified().entity(existingDataset).build();
+      }
+      User user = userService.findUserByEmail(authUser.getEmail());
+      Dataset patched = datasetRegistrationService.patchDataset(datasetId, user, patch);
+      return Response.noContent().entity(patched).build();
+    } catch (Exception e) {
+      return createExceptionResponse(e);
+    }
+  }
+
+  @Deprecated
   @PUT
   @Consumes("application/json")
   @Produces("application/json")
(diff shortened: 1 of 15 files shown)


src/main/resources/assets/schemas/DatasetPatch.yaml
```diff
@@ -0,0 +1,22 @@
...
+        schemaProperty: numberOfParticipants
+        propertyType: Number
+      - propertyName: URL
+        propertyValue: https://duos.org
+        schemaProperty: url
+        propertyType: String
+      - propertyName: Data Location
+        propertyValue: Terra Workspace
+        schemaProperty: dataLocation
+        propertyType: String
```
snf2ye, line 22: Would it be possible to make this an enum? From the swagger description, it looks like there is a know set of possible values.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -190,6 +198,66 @@ public Response updateByDatasetUpdate(
...
+      if (json == null || json.isEmpty()) {
+        throw new BadRequestException("Dataset Patch is required");
+      }
+      Gson gson = GsonUtil.getInstance();
+      DatasetPatch patch;
+      try {
+        patch = gson.fromJson(json, DatasetPatch.class);
+      } catch (Exception e) {
+        throw new BadRequestException("Unable to parse dataset patch: " + json);
+      }
```
snf2ye: I'm surprised you have to do this sort of validation - I would think there is a way in swagger to require a body is requested and that it matches the provided schema. From a quick test in TDR, we do get an automated error message, but possibly this is from spring validation? 

Example TDR error message:
```
{
  "message": "400 BAD_REQUEST - see error details",
  "errorDetail": [
    "Required request body is missing: public org.springframework.http.ResponseEntity<bio.terra.model.JobModel> bio.terra.app.controller.DatasetsApiController.createDataset(bio.terra.model.DatasetRequestModel)"
  ]
}
```

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -558,4 +626,41 @@ private void validateDatasetDacAccess(User user, Dataset dataset) {
...
   }
+
+  /**
+   * Determine if the user is a dataset/study creator, or if they are listed as a data custodian.
+   *
+   * @param user    User
+   * @param dataset Dataset
+   * @return User is a creator or custodian of the dataset.
+   */
+  protected boolean isCreatorOrCustodian(User user, Dataset dataset) {
```
snf2ye: What would you think about breaking this into two separate methods? Would we ever need to individually check if a user is a creator vs a custodian?

snf2ye (approved):
I have a few questions, but no blocking concerns.

src/main/resources/assets/schemas/DatasetPatch.yaml
rushtong: Great idea, yeah, I should have done that - we do the same thing in a lot of other endpoint descriptions.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
rushtong: We can't rely only on swagger for validation, we have to have it here as well. Nate hits a lot of the dataset APIs via scripts.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
rushtong: There are other places in the code where we look for custodians, but this is the only place where we care this specific scenario. I could move this to the `Dataset` object too, that would make it more useful if we did need it elsewhere in the future.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -190,6 +198,66 @@ public Response updateByDatasetUpdate(
...
+      }
+      Dataset existingDataset = datasetService.findDatasetById(datasetId);
+      if (existingDataset == null) {
+        throw new NotFoundException("Could not find the dataset with id: " + datasetId);
+      }
+      // Check permissions for non-admin roles.
+      User user = userService.findUserByEmail(authUser.getEmail());
+      if (!user.hasUserRole(UserRoles.ADMIN)) {
+        if (!isCreatorOrCustodian(user, existingDataset)) {
+          throw new ForbiddenException("User does not have permission to update this dataset");
```
rjohanek, line 215: should this happen before other validation?

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -190,6 +198,66 @@ public Response updateByDatasetUpdate(
...
+      }
+      if (!patch.validateProperties()) {
+        return Response.status(HttpStatusCodes.STATUS_CODE_BAD_REQUEST)
+            .entity(new Error(
+                "Properties are invalid",
+                HttpStatusCodes.STATUS_CODE_BAD_REQUEST))
+            .build();
+      }
+      datasetRegistrationService.patchDataset(datasetId, user, patch);
+      return Response.noContent().build();
```
rjohanek: why not return the patched dataset? Or if we don't want to return it, why have the service return it?

src/main/resources/assets/paths/datasetById.yaml
```diff
@@ -1,3 +1,56 @@
+patch:
+  summary: Patch Dataset
+  description: |
+    This API allows Admins, Chairpersons, and Data Submitters to patch a limited set of dataset
+    properties. Data Location must conform to the schema defined in
```
rjohanek: the properties and the dataset name can be updated too right?

rjohanek (approved):
looks good, just have a couple questions

src/test/java/org/broadinstitute/consent/http/models/DatasetPatchTest.java
```diff
@@ -0,0 +1,91 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.EnumSource;
+
+public class DatasetPatchTest {
+
+  @Test
+  void testInvalidPropertyKeys() {
+    DatasetProperty invalidProp = new DatasetProperty();
+    invalidProp.setPropertyName(RandomStringUtils.randomAlphanumeric(25));
+    invalidProp.setPropertyValue(RandomUtils.nextInt(1, 10));
```
pshapiro4broad, line 22: What is the purpose of randomized data in this test? In cases where tests are run in parallel and name collisions can occur, having random data can help. But in cases like this where no collision can occur, I don't see the benefit of using random data.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
pshapiro4broad: The main difference here is that TDR relies on openapi (swagger) code generation, which converts the openapi yaml into java code. The generated code enforces validation (and also generates the equivalent `@PATCH`, `@Path()`, etc annotations too). Since consent doesn't use openapi codegen, it must manually implement these features of the swagger yaml.

I think the biggest takeaway here is that, with this approach, there is no enforcement that the openapi yaml matches the java implementation. Because of this, most (all?) other Java terra services use openapi codegen to avoid code disagreement and duplication.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
rushtong: A `Response.noContent()` will not have an entity, even if you add one - it's a jakarta Response pattern that I discovered when testing.

src/main/resources/assets/paths/datasetById.yaml
rushtong: Yes, thank you, I'll add that to the description.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
rushtong: Good point. I don't think it hurts to do other validations first, but logically, yes, we should.

src/test/java/org/broadinstitute/consent/http/models/DatasetPatchTest.java
rushtong: We have seen collisions in the past so this is a general pattern we've used to avoid the potential for flaky tests.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
rushtong: > I think the biggest takeaway here is that, with this approach, there is no enforcement that the openapi yaml matches the java implementation. Because of this, most (all?) other Java terra services use openapi codegen to avoid code disagreement and duplication.

This is correct, and essentially an artifact of history. Swagger codegen tools were not around when Consent/Ontology were set up and it's never been a priority for us to do a conversion.

src/main/resources/assets/paths/datasetById.yaml
```diff
@@ -1,3 +1,56 @@
...
+    - Dataset
+  requestBody:
+    description: DatasetPatch object
+    required: true
+    content:
+      application/json:
+        schema:
+          $ref: '../schemas/DatasetPatch.yaml'
+  responses:
+    204:
```
pshapiro4broad: I think this should be `200` since this response has content

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
pshapiro4broad: Looking at the yaml this should return the dataset. Although the status is `204` no content; I would make it `200` and return the dataset here, unless there's a reason not to.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -190,6 +198,66 @@ public Response updateByDatasetUpdate(
...
+                "The new name for this dataset already exists: " + patch.name(),
+                HttpStatusCodes.STATUS_CODE_BAD_REQUEST))
+            .build();
+      }
+      if (!patch.validateProperties()) {
+        return Response.status(HttpStatusCodes.STATUS_CODE_BAD_REQUEST)
+            .entity(new Error(
+                "Properties are invalid",
+                HttpStatusCodes.STATUS_CODE_BAD_REQUEST))
+            .build();
```
pshapiro4broad: From what I read about `WebApplicationException`, you could throw here to avoid making a response. This should have the same effect:
```suggestion
        throw new BadRequestException("Properties are invalid");
```

Maybe a better pattern overall is to have `validateX()` throw instead of returning a status that must be checked.

src/main/resources/assets/schemas/DatasetPatch.yaml
rushtong: I looked more into this, and I don't think it's possible in open api. This is an example value, but the item itself is a `DatasetProperty`. That object can have anything as a value, it's not restricted to an enum. I can't find anything in the spec that allows an `example` property to be limited to enums.

snf2ye (approved):

rjohanek (approved):

PR #2408, opened 2024-10-10
DCJ-664: Add support for saving draft study/dataset submission requests.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DCJ-664](https://broadworkbench.atlassian.net/browse/DCJ-664)

### Summary

Adds support for draft documents to be submitted with an arbitrary number of attachments by Admins and Data Submitters.  Initially, this set of services will be used to provide users of the new Dataset Submission form to pause and resume their work during composition and then turn their saved work into a single submission.

A new DraftResource controller governs access to the feature.  This controller ensures only Administrators and Data Submitters are configured for access at this time.  New models and services have been created to support drafts and draft file attachment using approaches common to the consent backend.  

The exception to the standardized approach includes a new DraftInterface java interface to allow for reflective classes being instantiated from stored drafts.  This will allow future drafts to include a "draft type" filter that could handle any number of the draft constructs within DUOS.  Since that work is outside of the scope of the current work, the draft type filter isn't completed at this time. 

Work to convert a draft to a "Dataset Submission" is not yet complete.  The goal of getting this work across the finish line is to enable front end development to commence for both the process of composing the new, adaptive form and the initial prototype of choosing the work to resume from saved state.  I envision adding support for that feature once the UI is ready.

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/db/DAOContainer.java
```diff
@@ -165,4 +166,8 @@ public AcknowledgementDAO getAcknowledgementDAO() {
   public void setAcknowledgementDAO(AcknowledgementDAO acknowledgementDAO) {
     this.acknowledgementDAO = acknowledgementDAO;
   }
+
+  public void setDraftSubmissionDAO(DraftDAO draftDAO) {
+    this.draftDAO = draftDAO;
+  }
```
rushtong: This class is an anomaly and not a great pattern that I would encourage. We should only expand this usage if necessary. The only class that needs this is `DataAccessRequestService` which is not affected in this PR, so I think it can be reverted.

src/main/java/org/broadinstitute/consent/http/ConsentModule.java
```diff
@@ -168,6 +171,7 @@ public DAOContainer providesDAOContainer() {
     container.setInstitutionDAO(providesInstitutionDAO());
     container.setFileStorageObjectDAO(providesFileStorageObjectDAO());
     container.setAcknowledgementDAO(providesAcknowledgementDAO());
+    container.setDraftSubmissionDAO(providesDraftSubmissionDAO());
```
rushtong: See comment in `DAOContainer`

src/main/java/org/broadinstitute/consent/http/db/mapper/DraftSubmissionInterfaceMapper.java
```diff
@@ -0,0 +1,87 @@
...
+    if (hasColumn(rs, "cu_user_id")) {
+      User createUser = buildUserFromResult(rs.getInt("cu_user_id"),
+          rs.getString("cu_email"),
+          rs.getString("cu_display_name"),
+          new Date(rs.getTimestamp("cu_create_date").getTime()),
+          rs.getBoolean("cu_email_preference"),
+          rs.getInt("cu_institution_id"),
+          rs.getString("cu_era_commons_id"));
+      dsi.setCreateUser(createUser);
+    }
```
rushtong: JDBI has a way to make this slightly simpler in the reducer, similar to what you did for the FSO. I think it would only work for a single user object, however, not in this case where there are two instances of the same class. For example, in the DAO class, add
```
@RegisterRowMapper(User.class, prefix = "cu")
```
and in the reducer,
```
User createUser = rowView.getRow(User.class);
draftSubmission.setCreateUser(createUser);
```
However, you'd still need to manually handle the other user like you have here in the mapper, so not much of a win. It would be nice if there were a way to register the same class multiple times with different prefixes, but I don't think JDBI supports that.

src/main/java/org/broadinstitute/consent/http/models/Draft.java
```diff
@@ -0,0 +1,142 @@
...
+import com.google.gson.Gson;
+import java.util.Date;
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
+import org.broadinstitute.consent.http.models.dataset_registration_v1.DatasetRegistrationSchemaV1;
+
+/**
+ * DraftSubmission represents a Data Submitter's partial submission of 0 or more elements. This is
```
rushtong: Minor nit:
```suggestion
 * Draft represents a Data Submitter's partial submission of 0 or more elements. This is
```

src/main/java/org/broadinstitute/consent/http/resources/DraftResource.java
```diff
@@ -0,0 +1,241 @@
...
+  @Path("/v1")
+  @RolesAllowed({ADMIN, DATASUBMITTER})
+  public Response createDraftRegistration(@Auth AuthUser authUser, String json) {
+    try {
+      User user = userService.findUserByEmail(authUser.getEmail());
+      DraftInterface draft = new Draft(json, user);
+      draftService.insertDraft(draft);
+      URI uri = UriBuilder.fromPath(String.format("/api/draft/v1/%s", draft.getUUID().toString()))
+          .build();
+      return Response.created(uri).entity(json).build();
```
rushtong: This is an example of the generated uri:
```
location: https://local.dsde-dev.broadinstitute.org:27443^(?!/introspect/)(/)(.*)api/draft/v1/4c8fa3f5-56aa-4318-a84d-7d89c0a0857d 
```
I think we need to look into the right way to format this

src/main/java/org/broadinstitute/consent/http/resources/DraftResource.java
```diff
@@ -0,0 +1,241 @@
...
+  public DraftResource(UserService userService,
+      DraftService draftService) {
+    this.userService = userService;
+    this.draftService = draftService;
+  }
+
+  @GET
+  @Produces({MediaType.APPLICATION_JSON})
+  @Path("/v1")
+  @RolesAllowed({ADMIN, DATASUBMITTER})
```
rushtong, line 55: We need to loop back with Jonathan on the roles here. Currently, for datasets and studies, we also allow for Chairpersons to create them. However, given that this is intended to return any kind of draft, maybe it's better to use `@PermitAll`

src/main/java/org/broadinstitute/consent/http/resources/DraftResource.java
```diff
@@ -0,0 +1,241 @@
...
+      return createExceptionResponse(e);
+    }
+
+  }
+
+  @POST
+  @Consumes({MediaType.APPLICATION_JSON})
+  @Produces({MediaType.APPLICATION_JSON})
+  @Path("/v1")
+  @RolesAllowed({ADMIN, DATASUBMITTER})
```
rushtong, line 72: Here is where we need to make sure the UI for this is accurate. We don't have a data submitter console currently. Data Submissions are shown in the Researcher console (for lack of a better location) and Chairpersons have access to the "Add Dataset/Study" functionality.

src/main/java/org/broadinstitute/consent/http/resources/DraftResource.java
```diff
@@ -0,0 +1,241 @@
...
+  public Response updateDraft(@Auth AuthUser authUser, @PathParam("draftUUID") String draftUUID,
+      String json) {
+    try {
+      User user = userService.findUserByEmail(authUser.getEmail());
+      DraftInterface draft = draftService.getAuthorizedDraft(
+          validateUUID(draftUUID), user);
+      draft.setJson(json);
+      draftService.updateDraft(draft, user);
+      return Response.ok().entity(draftService.draftAsJson(draft)).build();
```
rushtong: In local testing, I'm finding that the GET and the PUT are returning slightly different user objects. GET shows a simplified user without roles and library cards for both create and update user. PUT shows the same thing for the create user, but a fully fleshed out update user with roles, library cards, and institution. For these operations, the simpler version would be more preferable.

### GET
```
{
  "document": {
    "random": "value 2"
  },
  "meta": {
    "createDate": 1729619182645,
    "updateDate": 1729619683116,
    "name": "Created Tue Oct 22 17:46:22 UTC 2024",
    "createUser": {
      "userId": 3351,
      "email": "grushton@broadinstitute.org",
      "displayName": "Greg Rushton Admin",
      "createDate": 1452101967000,
      "emailPreference": false
    },
    "updateUser": {
      "userId": 3351,
      "email": "grushton@broadinstitute.org",
      "displayName": "Greg Rushton Admin",
      "createDate": 1452101967000,
      "emailPreference": false
    },
    "uuid": "4c8fa3f5-56aa-4318-a84d-7d89c0a0857d"
  }
}
```

### PUT
```
{
  "document": {
    "random": "value 2"
  },
  "meta": {
    "createDate": 1729619182645,
    "updateDate": 1729619683116,
    "name": "Created Tue Oct 22 17:46:22 UTC 2024",
    "createUser": {
      "userId": 3351,
      "email": "grushton@broadinstitute.org",
      "displayName": "Greg Rushton Admin",
      "createDate": 1452101967000,
      "emailPreference": false
    },
    "updateUser": {
      "userId": 3351,
      "email": "grushton@broadinstitute.org",
      "displayName": "Greg Rushton Admin",
      "createDate": 1452101967000,
      "roles": [
        {
          "userRoleId": 9,
          "userId": 3351,
          "roleId": 3,
          "name": "Alumni"
        },
        {
          "userRoleId": 32,
          "userId": 3351,
          "roleId": 4,
          "name": "Admin"
        },
        {
          "userRoleId": 59,
          "userId": 3351,
          "roleId": 5,
          "name": "Researcher"
        },
        {
          "userRoleId": 130,
          "userId": 3351,
          "roleId": 6,
          "name": "DataOwner"
        },
        {
          "userRoleId": 405,
          "userId": 3351,
          "roleId": 7,
          "name": "SigningOfficial"
        },
        {
          "userRoleId": 633,
          "userId": 3351,
          "roleId": 8,
          "name": "DataSubmitter"
        },
        {
          "userRoleId": 831,
          "userId": 3351,
          "roleId": 2,
          "name": "Chairperson",
          "dacId": 3
        },
        {
          "userRoleId": 662,
          "userId": 3351,
          "roleId": 2,
          "name": "Chairperson",
          "dacId": 4
        }
      ],
      "emailPreference": false,
      "institutionId": 150,
      "eraCommonsId": "GREGORYRUSHTON",
      "institution": {
        "id": 150,
        "name": "The Broad Institute of MIT and Harvard",
        "createDate": 1617716382481,
        "updateDate": 1620133344596
      },
      "libraryCards": [
        {
          "id": 92,
          "userId": 3351,
          "institutionId": 150,
          "eraCommonsId": "GREGORYRUSHTON",
          "userName": "Gregory Rushton",
          "userEmail": "grushton@broadinstitute.org",
          "createDate": 1694797680476,
          "createUserId": 3351,
          "institution": {
            "id": 150,
            "name": "The Broad Institute of MIT and Harvard",
            "createDate": 1617716382481,
            "createUserId": 3392,
            "updateDate": 1620133344596,
            "updateUserId": 3351
          },
          "daaIds": [
            1,
            10,
            25,
            26,
            21,
            24,
            28,
            30,
            6,
            29,
            23,
            27,
            31,
            5,
            4,
            22
          ]
        }
      ]
    },
    "uuid": "4c8fa3f5-56aa-4318-a84d-7d89c0a0857d"
  }
}
```

rushtong (commented):
Still reviewing but here are some in-flight thoughts.

src/main/java/org/broadinstitute/consent/http/models/Draft.java
```diff
@@ -0,0 +1,142 @@
...
+ * an internal structure, not intended to be serialized back to the client in the entirety.
+ */
+public class Draft implements DraftInterface {
+
+  private Date createDate;
+  private Date updateDate;
+  private Set<FileStorageObject> storedFiles;
+  private transient String json;
```
pshapiro4broad: Is there another way in gson to declare a field as non-serialized? This use will flag a warning in sonar since it only expects `transient` inside a `Serializable` class.

src/main/java/org/broadinstitute/consent/http/db/mapper/DraftSubmissionInterfaceMapper.java
```diff
@@ -0,0 +1,87 @@
...
+
+    return dsi;
+  }
+
+  private User buildUserFromResult(Integer userId, String email, String displayName,
+      Date createDate, boolean emailPreference, Integer institutionId, String eraCommonsId) {
```
pshapiro4broad: Is there a reason why `institutionId` and `eraCommonsId` are not set on `user`?

src/main/java/org/broadinstitute/consent/http/db/mapper/DraftSubmissionInterfaceMapper.java
```diff
@@ -0,0 +1,87 @@
...
+import org.broadinstitute.consent.http.models.User;
+import org.jdbi.v3.core.mapper.RowMapper;
+import org.jdbi.v3.core.statement.StatementContext;
+
+public class DraftSubmissionInterfaceMapper implements RowMapper<DraftInterface>, RowMapperHelper {
+
+  @Override
+  public DraftInterface map(ResultSet rs, StatementContext ctx) throws SQLException, NoMatchingClassException {
+    if (!hasColumn(rs, "schema_class")) {
+      throw new NoMatchingClassException("Missing class name.");
```
pshapiro4broad: Under what conditions can this error occur? Is there a way to make this error occur in a test without bypassing the DAO and writing to the database directly?

src/main/java/org/broadinstitute/consent/http/db/mapper/DraftSubmissionInterfaceMapper.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    DraftInterface dsi;
+    String className = rs.getString("schema_class");
+
+    try {
+      dsi = (DraftInterface) Class.forName(className).getDeclaredConstructor().newInstance();
```
pshapiro4broad: I think that allowing arbitrary class instantiation here is unnecessarily general. Unless we end up supporting a system where new draft types can be added via linking a .jar to the project, adding a new draft requires a code change, and the list of known draft types are known at compile time (vs link time).

One way to write this without general support is to add a static `Map` that maps a draft name to its constructor. The map could be initialized either directly in the class that declares it, or each draft type could have a static initializer that adds an entry to the static map. Instead of a `Map` an enum could be used, as enums are also static maps of keys to a set of properties.

This instantiation system is a form of DI and there may be a way to use Dropwizard APIs to do this as it also supports DI.

src/main/java/org/broadinstitute/consent/http/db/mapper/DraftSubmissionInterfaceMapper.java
```diff
@@ -0,0 +1,87 @@
...
+    } catch (ClassNotFoundException | NoSuchMethodException | InvocationTargetException |
+             InstantiationException | IllegalAccessException classNotFoundException) {
+      throw new NoMatchingClassException(className);
+    }
+
+    if (hasColumn(rs, "name")) {
+      dsi.setName(rs.getString("name"));
+    }
```
pshapiro4broad: This is not new to this PR, but there code duplication in the rowmappers here. It looks like a common pattern is `if (hasColumn(rs, x)) { ...(rs.getString(x) ... }`. Combining these operations would let you write
```
  default void getString(ResultSet rs, String columnName, Consumer<String> useString) throws SQLException {
    if (hasColumn(rs, columnName)) {
      useString.accept(rs.getString(columnName));
    }
  }
...

    getString(rs, "name", dsi::setName);
```

We'd need to have a different method for each type, but it would end up being less code overall in the end, and avoid the need to duplicate the column name in each case.

src/main/java/org/broadinstitute/consent/http/db/mapper/DraftSubmissionInterfaceMapper.java
```diff
@@ -0,0 +1,87 @@
...
+      throw new NoMatchingClassException("Missing class name.");
+    }
+
+    DraftInterface dsi;
+    String className = rs.getString("schema_class");
+
+    try {
+      dsi = (DraftInterface) Class.forName(className).getDeclaredConstructor().newInstance();
+    } catch (ClassNotFoundException | NoSuchMethodException | InvocationTargetException |
+             InstantiationException | IllegalAccessException classNotFoundException) {
```
pshapiro4broad: Name `classNotFoundException` is confusing here as it could be any of these exceptions. 
```suggestion
             InstantiationException | IllegalAccessException e) {
```

src/main/java/org/broadinstitute/consent/http/db/mapper/DraftSubmissionInterfaceMapper.java
```diff
@@ -0,0 +1,87 @@
...
+      throw new NoMatchingClassException(className);
```
pshapiro4broad: Is there a reason to not include the original exception here? That would let the top level logger log the underlying exception and throw point.

src/main/resources/assets/schemas/DraftSummary.yaml
```diff
@@ -0,0 +1,17 @@
+type: object
+properties:
+  id:
+   type: string
+   format: uuid
+   description: The UUID of the draft.  See https://en.wikipedia.org/wiki/Universally_unique_identifier#:~:text=%5B2%5D-,Version%204%20(random),-%5Bedit%5D
```
fboulnois:
```suggestion
   description: The UUID of the draft.  See https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)
```

src/main/resources/assets/paths/draftDocumentIndex.yaml
```diff
@@ -0,0 +1,76 @@
...
+  tags:
+    - Draft
+  parameters:
+    - name: draftUUID
+      in: path
+      description: The UUID identifying the draft
+      required: true
+      schema:
+        type: string
+        format: UUID v4.
```
fboulnois:
```suggestion
        format: uuid
```

I think this is the only valid OpenAPI UUID format.

src/main/resources/assets/paths/draftAttachments.yaml
```diff
@@ -0,0 +1,61 @@
...
+  tags:
+    - Draft
+  parameters:
+    - name: draftUUID
+      in: path
+      description: The UUID identifying the draft
+      required: true
+      schema:
+        type: string
+        format: UUID v4.
```
fboulnois:
```suggestion
        format: uuid
```

ditto here

src/main/resources/assets/paths/draftAttachments.yaml
```diff
@@ -0,0 +1,61 @@
...
```
fboulnois:
```suggestion
        format: uuid
```

ditto

src/main/resources/assets/paths/draftAttachments.yaml
```diff
@@ -0,0 +1,61 @@
...
+  responses:
+    200:
+      description: File contents.
+      content:
+        application/octet-stream:
+          schema:
+            type: string
+            format: binary
+    401:
+      description: Unauthorized.  Did you remember to authenticate?
```
fboulnois:
```suggestion
      description: User not authorized to get file attachment
```

I would suggest switching this language to more neutral language.

src/main/resources/assets/paths/draftAttachmentsIndex.yaml
```diff
@@ -0,0 +1,66 @@
...
+  tags:
+    - Draft
+  parameters:
+    - name: draftUUID
+      in: path
+      description: The UUID identifying the draft
+      required: true
+      schema:
+        type: string
+        format: UUID v4.
```
fboulnois:
```suggestion
        format: uuid
```

src/main/resources/assets/paths/draftAttachments.yaml
```diff
@@ -0,0 +1,61 @@
...
+      in: path
+      description: The integer identifying the attachment.
+      required: true
+      schema:
+        type: integer
+  responses:
+    200:
+      description: Successful delete.
+    401:
+      description: Unauthorized.  Did you remember to authenticate?
```
fboulnois:
```suggestion
      description: User not authorized to delete file attachment
```

ditto

src/main/resources/assets/paths/draftAttachmentsIndex.yaml
```diff
@@ -0,0 +1,66 @@
...
+    200:
+      description: A list of attachments for a draft.
+      content:
+        application/json:
+          schema:
+            type: array
+            items:
+              $ref: '../schemas/FileStorageObject.yaml'
+    401:
+      description: Unauthorized.  Did you remember to authenticate?
```
fboulnois:
```suggestion
      description: User not authorized to get list of file attachments
```

ditto

src/main/resources/assets/paths/draftAttachmentsIndex.yaml
```diff
@@ -0,0 +1,66 @@
...
+  tags:
+    - Draft
+  parameters:
+    - name: draftUUID
+      in: path
+      description: The UUID identifying the draft
+      required: true
+      schema:
+        type: string
+        format: UUID v4.
```
fboulnois:
```suggestion
        format: uuid
```

ditto

src/main/resources/assets/paths/draftAttachmentsIndex.yaml
```diff
@@ -0,0 +1,66 @@
...
+    200:
+      description: The attachment(s) were uploaded successfully.
+      content:
+        application/json:
+          schema:
+            type: array
+            items:
+              $ref: '../schemas/FileStorageObject.yaml'
+    401:
+      description: Unauthorized.  Did you remember to authenticate?
```
fboulnois:
```suggestion
      description: User not authorized to upload file attachments
```

src/main/resources/assets/paths/draftDocumentIndex.yaml
```diff
@@ -0,0 +1,76 @@
...
+  tags:
+  - Draft
+  parameters:
+    - name: draftUUID
+      in: path
+      description: The UUID identifying the draft
+      required: true
+      schema:
+        type: string
+        format: UUID v4.
```
fboulnois:
```suggestion
        format: uuid
```

src/main/resources/assets/paths/draftDocumentIndex.yaml
```diff
@@ -0,0 +1,76 @@
...
+  responses:
+    200:
+     description: A draft with metadata.
+     content:
+       application/json:
+         schema:
+          $ref: '../schemas/DraftDetails.yaml'
+    401:
+      description: Unauthorized.  Did you remember to authenticate?
```
fboulnois:
```suggestion
      description: User not authorized to get draft
```

ditto

src/main/resources/assets/paths/draftDocumentIndex.yaml
```diff
@@ -0,0 +1,76 @@
...
+  tags:
+    - Draft
+  parameters:
+    - name: draftUUID
+      in: path
+      description: The UUID identifying the draft
+      required: true
+      schema:
+        type: string
+        format: UUID v4.
```
fboulnois:
```suggestion
        format: uuid
```

ditto

src/main/resources/assets/paths/draftDocumentIndex.yaml
```diff
@@ -0,0 +1,76 @@
...
+          type: object
+  responses:
+      200:
+        description: A draft with metadata.
+        content:
+          application/json:
+           schema:
+            $ref: '../schemas/DraftDetails.yaml'
+      401:
+        description: Unauthorized.  Did you remember to authenticate?
```
fboulnois:
```suggestion
        description: User not authorized to update draft
```

ditto

src/main/resources/assets/paths/draftDocumentIndex.yaml
```diff
@@ -0,0 +1,76 @@
...
+      description: The UUID identifying the draft
+      required: true
+      schema:
+        type: string
+        format: UUID v4.
+  responses:
+    200:
+      description: The delete succeeded.
+    401:
+      description: Unauthorized.  Did you remember to authenticate?
```
fboulnois:
```suggestion
      description: User not authorized to delete draft
```

ditto

src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
```diff
@@ -265,4 +274,23 @@ private DataAccessRequest createDataAccessRequest(Integer userId, Integer collec
...
+        });
+    return mapOfFiles;
+  }
+
+  private FormDataBodyPart getFormDataBodyPartMock(String name) {
+    FormDataBodyPart part = mock(FormDataBodyPart.class);
+    lenient().when(part.getName()).thenReturn(name);
+    lenient().when(part.getMediaType()).thenReturn(MediaType.MULTIPART_FORM_DATA_TYPE);
+    lenient().when(part.getValueAs(InputStream.class))
+        .thenReturn(new ByteArrayInputStream(EMPTY_JSON_DOCUMENT.getBytes()));
```
fboulnois: I think we should avoid the use of `lenient` to ensure that we have strict mocks.

src/test/java/org/broadinstitute/consent/http/service/DraftFileStorageServiceTest.java
```diff
@@ -0,0 +1,100 @@
...
+public class DraftFileStorageServiceTest extends DAOTestHelper {
+
+  private DraftFileStorageService draftFileStorageService;
+
+  @BeforeEach
+  public void setUp() throws IOException {
+    GCSService gcsService = mock(GCSService.class);
+    when(gcsService.storeDocument(any(), anyString(), any())).thenReturn(
+        BlobId.of(UUID.randomUUID().toString(), UUID.randomUUID().toString()));
+    lenient().when(gcsService.getDocument((BlobId) any()))
```
fboulnois: ditto here

src/test/java/org/broadinstitute/consent/http/service/DraftServiceTest.java
```diff
@@ -0,0 +1,192 @@
...
+
+@ExtendWith(MockitoExtension.class)
+public class DraftServiceTest extends DAOTestHelper {
+
+  private DraftService draftService;
+
+  @BeforeEach
+  public void setup() throws IOException {
+    GCSService gcsService = mock(GCSService.class);
+    lenient().when(gcsService.storeDocument(any(), anyString(), any())).thenReturn(
```
fboulnois: ditto here

src/main/java/org/broadinstitute/consent/http/service/DraftService.java
```diff
@@ -0,0 +1,173 @@
...
+        handle.rollback();
+      }
+      handle.commit();
+    });
+  }
+
+  public StreamingOutput draftAsJson(DraftInterface draft) {
+    Gson gson = GsonUtil.buildGson();
+    return output -> {
+      output.write("{ \"document\":".getBytes());
```
fboulnois, line 163: Can this be constructed using multiline string interpolation? This would make it more readable.

fboulnois (approved):
A few early suggestions:

src/main/java/org/broadinstitute/consent/http/resources/DraftResource.java
otchet-broad: Note sent to Jonathan and @rushtong for approving this control change.

src/main/java/org/broadinstitute/consent/http/resources/DraftResource.java
otchet-broad: Agreed.  Listing the summary (and connecting up basic CRUD operations) is the first front end component I'm planning to build so we can insert it into the larger UI wherever it makes sense.

src/main/java/org/broadinstitute/consent/http/models/Draft.java
otchet-broad: From the [GSON manual](https://google.github.io/gson/UserGuide.html#excluding-fields-from-serialization-and-deserialization), there are three supported ways to do it:
- If a field is marked transient, (by default) it is ignored and not included in the JSON serialization or deserialization.
- We can decorate all of the properties we want to @Expose and use a different pattern on drafts, users, filestorage and other classses we make use of.
- We can add a filter, add field decorators, and apply the filter  at the point of serialization.

If you think the second or third option is worth the squeeze, I'll implement.

src/main/java/org/broadinstitute/consent/http/db/mapper/DraftSubmissionInterfaceMapper.java
otchet-broad: Good catch, thank you.  I wanted to thin out the user and missed this.

src/main/resources/assets/paths/draftDocumentIndex.yaml
otchet-broad: From the [Swagger docs](https://swagger.io/docs/specification/v3_0/data-models/data-types/#:~:text=pattern%20is%20specified.-,String%20Formats,-An%20optional%20format), format is open and only "serves as a hint at the contents and format of the string."

I'm fine with updating it to uuid or UUID v4 throughout.  After having worked with different formats of UUIDs, I have a greater appreciation for hinting more specifics.

Let me know what you think!

src/main/java/org/broadinstitute/consent/http/resources/DraftResource.java
otchet-broad: The service's update method was fixed to return the updated representation.
Tests were added to confirm this thins the user as expected.

src/main/java/org/broadinstitute/consent/http/models/Draft.java
otchet-broad: Given that we're now thinking about this in a bit more of a generic fashion, I've updated the part of the comment you've highlighted and the predicate (by striking "Data Submitter's")

src/main/java/org/broadinstitute/consent/http/db/mapper/DraftSubmissionInterfaceMapper.java
otchet-broad: Good question.  Mockito to the rescue.  Test testNoSchemaColumnValue added to the [DraftDAOTest](src/test/java/org/broadinstitute/consent/http/db/DraftDAOTest.java) suite.

src/main/java/org/broadinstitute/consent/http/db/mapper/DraftSubmissionInterfaceMapper.java
otchet-broad: Agreed, but I think that's going to be a bit larger of a refactor than this is change is intending.  Could I make a ticket for this so the team can talk about this?

src/main/java/org/broadinstitute/consent/http/db/mapper/DraftSubmissionInterfaceMapper.java
otchet-broad: Renamed.

src/main/java/org/broadinstitute/consent/http/db/mapper/DraftSubmissionInterfaceMapper.java
otchet-broad: My thinking was generally around leaking internals to a bad actor in this particular error state.  I wanted to communicate the condition but not the details.  If there's no need to worry about that for this, then I think the way to resolve this would be to bubble up the exception.

src/main/resources/assets/paths/draftDocumentIndex.yaml
otchet-broad: @fboulnois - I think I came up with something that can satisfy both of us for this one.  I updated the format to uuid and added the (updated, thank you) link to the description so it's visible to anyone that might read the documentation.

src/main/java/org/broadinstitute/consent/http/db/DAOContainer.java
otchet-broad: Done.

src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
otchet-broad: They're gone!  I think we have one other spot related to library cards that are not part of this PR.  See: src/test/java/org/broadinstitute/consent/http/service/LibraryCardServiceTest.java (strictness = Strictness.LENIENT)

Should I make a ticket to refactor so this can be removed?

src/main/java/org/broadinstitute/consent/http/service/DraftService.java
otchet-broad: Does this work for you?
    String outputString =
        String.format(
        """
        { "document": %s
        , "meta": %s
        }
        """, draft.getJson(), gson.toJson(draft));
    return output -> {
        output.write(outputString.getBytes());

otchet-broad: (Deleted my last code.  I added tests to validate and it turns out that the quotes in the JSON are stripped with String.format()).  Going to leave as-is for now.

src/main/java/org/broadinstitute/consent/http/db/mapper/DraftSubmissionInterfaceMapper.java
otchet-broad: ...more like OBE with my next push

src/main/java/org/broadinstitute/consent/http/db/mapper/DraftSubmissionInterfaceMapper.java
rushtong: I agree, please do add another ticket - it's a nice utility we could pretty easily add where needed, but definitely orthogonal to this PR.

src/main/java/org/broadinstitute/consent/http/db/mapper/DraftSubmissionInterfaceMapper.java
otchet-broad: I've refactored this a bit, based on a quick chat I had with Greg.  It removes the reflection and adds a DraftType enum with a small utility to return the proper implementation based on the Draft Type.

src/main/java/org/broadinstitute/consent/http/db/mapper/DraftSubmissionInterfaceMapper.java
otchet-broad: https://broadworkbench.atlassian.net/browse/DT-939?atlOrigin=eyJpIjoiOGE2MjdkYTU5MDk3NDA3YmI4MmM1MGQ1NjJlZjU2OWMiLCJwIjoiaiJ9

src/main/java/org/broadinstitute/consent/http/db/DraftDAO.java
```diff
@@ -0,0 +1,128 @@
...
+      @Bind("draftType") String draftType);
+
+  @SqlUpdate("""
+      UPDATE draft
+      SET
+          update_date = :updateDate,
+          update_user_id = :updateUserId
+      WHERE uuid = :uuid
+      """)
+  void updateDraftByDraftUUID(
```
rjohanek, line 93: what is the point of this method? When would this be used compared to the other updateDraftByDraftUUID?

src/main/java/org/broadinstitute/consent/http/models/DraftStudyDataset.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.Set;
+import java.util.UUID;
+import org.broadinstitute.consent.http.enumeration.DraftType;
+import org.broadinstitute.consent.http.models.dataset_registration_v1.DatasetRegistrationSchemaV1;
+
+/**
+ * Draft represents a partial submission of 0 or more elements. This is
+ * an internal structure, not intended to be serialized back to the client in the entirety.
+ */
+public class DraftStudyDataset implements DraftInterface {
```
rjohanek, line 16: What does study dataset mean here? is it a draft study or a draft dataset or something else?

otchet-broad: There's a new form that combines both a Study and Dataset(s) being submitted.  This impl will be used to handle drafts for that form.  I expect I'll be adding to the impl and interface to convert the draft into the real consent objects when the user is ready to submit.

src/main/java/org/broadinstitute/consent/http/db/DraftDAO.java
otchet-broad: The point of this one is to update the parent object's updated date/user when an attachment is added or removed.

src/main/java/org/broadinstitute/consent/http/enumeration/DraftType.java
```diff
@@ -0,0 +1,36 @@
...
+import com.google.gson.annotations.SerializedName;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public enum DraftType {
+  @SerializedName("StudyDatasetSubmissionV1")
+  STUDY_DATASET_SUBMISSION_V1("StudyDatasetSubmissionV1");
```
rjohanek, line 12: although there is only one value now, is the thought process here that having an enum for type makes the code more extensible/ future-proof? Or is this useful now too? Do we have plans for future draft types? Does this draft interface type have anything to do with the DraftStudyDataset class? If we have different classes implementing the Draft Interface type, do we need to store draft type on them or can their class indicate that information?

otchet-broad: Good questions!  This was part of a change Phil asked for to remove the reflective way I was doing that (by storing class information) and to use something the IDE and build tools would hint more at.
The DraftInterface (src/main/java/org/broadinstitute/consent/http/models/DraftInterface.java) now requires implementations to support a getDraftType (DraftType).  The DraftStudyDataset(src/main/java/org/broadinstitute/consent/http/models/DraftStudyDataset.java) declares the DraftType as static final and implements the getDraftType returning this value.

The enum value is persisted in the database.  You see it used in the src/main/java/org/broadinstitute/consent/http/db/mapper/DraftInterfaceMapper.java to instantiate the implementation.

As far as other draft types, consent already has them.  In the future, we can migrate them over to this more generalized approach.

src/main/java/org/broadinstitute/consent/http/service/DraftService.java
```diff
@@ -0,0 +1,174 @@
...
+  public DraftService(Jdbi jdbi, DraftDAO draftDAO,
+      DraftFileStorageService draftFileStorageService) {
+    this.jdbi = jdbi;
+    this.draftDAO = draftDAO;
+    this.draftFileStorageService = draftFileStorageService;
+  }
+
+  public void insertDraft(DraftInterface draft)
+      throws SQLException, BadRequestException {
+    jdbi.useHandle(handle -> {
```
rjohanek, line 44: why have the jdbi.useHandle logic here in the service instead of in the DAO? It looks like it lives in the DAO for most other cases

rjohanek (approved):
I have a few questions, but this looks great! Thanks!

src/main/java/org/broadinstitute/consent/http/resources/DraftResource.java
rushtong: I'm re-thinking that this is fine for now and we can revisit as usage expands.

src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
```diff
@@ -265,4 +274,23 @@ private DataAccessRequest createDataAccessRequest(Integer userId, Integer collec
...
+  }
+
+  private FormDataBodyPart getFormDataBodyPartMock(String name) {
+    FormDataBodyPart part = mock(FormDataBodyPart.class);
+    when(part.getName()).thenReturn(name);
+    when(part.getMediaType()).thenReturn(MediaType.MULTIPART_FORM_DATA_TYPE);
+    when(part.getValueAs(InputStream.class))
+        .thenReturn(new ByteArrayInputStream(EMPTY_JSON_DOCUMENT.getBytes()));
+    return part;
+  }
```
rushtong: My preference is to add these as private methods to the test class(es) that use them. What we've found is that this helper class balloons with slight variations of entity creation methods. I'm less concerned about DRY in tests - my experience is that it is usually an anti-pattern for generic data objects. The one thing we do need to share here is the infrastructure setup/teardown that all DAO classes need.

src/test/java/org/broadinstitute/consent/http/service/DraftServiceTest.java
```diff
@@ -0,0 +1,222 @@
...
+  @BeforeEach
+  public void setup() throws IOException {
+    DraftFileStorageService draftFileStorageService = new DraftFileStorageService(jdbi, gcsService,
+        fileStorageObjectDAO);
+    this.draftService = new DraftService(jdbi, draftDAO,
+        draftFileStorageService);
+  }
+
+  @Test
+  public void testCreateDraft() throws SQLException, IOException {
```
rushtong: Here, as well in all tests, the `public` modifier is not necessary in JUnit5

rushtong (approved):
Thank you for the changes! See minor feedback inline, but otherwise 👍🏽

PR #2412, opened 2024-10-16
DCJ-708: Refactor query performance

rushtong:
### Addresses
Ticket: https://broadworkbench.atlassian.net/browse/DCJ-708

### Summary
In this PR, we fix a long-running query bug that caused the instance to crash. This manifested in a never-ending loading screen for the user and an eventual redirect back to their console with an error message. The root of the problem is that it is very expensive to populate a large list of datasets which caused OOMs and the pod to crash/restart.

Fixes:
* Rename service/dao methods to be more reflective of what they are actually doing
* Query only for the Dataset ID instead of the full dataset object
  * This allows the server to complete the response instead of OOM-ing and crashing the instance
* Update tests to match new signatures.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
@@ -420,7 +420,7 @@ public class DarCollectionService implements ConsentLogger {
    * @return List of Dataset IDs
    */
   public List<Integer> findDatasetIdsByDACUser(User user) {
-    return datasetDAO.findDatasetIdsByDACUserEmail(user.getEmail());
+    return datasetDAO.findDatasetIdsByDACUserId(user.getUserId());
   }
 
   public void deleteByCollectionId(User user, Integer collectionId)
@@ -624,7 +624,7 @@ public class DarCollectionService implements ConsentLogger {
    */
   public DarCollection cancelDarCollectionElectionsAsChair(DarCollection collection, User user) {
     // Find dataset ids the chairperson has access to:
-    List<Integer> datasetIds = datasetDAO.findDatasetIdsByDACUserEmail(user.getEmail());
+    List<Integer> datasetIds = datasetDAO.findDatasetIdsByDACUserId(user.getUserId());
 
     // Filter the list of DARs we can operate on by the datasets accessible to this chairperson
     List<DataAccessRequest> dars = collection.getDars().values().stream()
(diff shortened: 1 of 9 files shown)


src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
```diff
@@ -238,7 +238,7 @@ void testCancelDarCollectionAsAdmin() {
   @Test
   void testCancelDarCollectionAsChair_ChairHasDatasets() {
     User user = new User();
-    user.setEmail("email");
+    user.setUserId(RandomUtils.nextInt(1, 10));
```
rjohanek, line 241: why a random user id for testing?

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -414,11 +414,13 @@ public DarCollectionSummary updateCollectionToDraftStatus(DarCollection sourceCo
...
-        .stream()
-        .map(Dataset::getDataSetId)
-        .collect(Collectors.toList());
+  /**
+   * Find all dataset ids by the DAC User. Will return ids for Chairpersons or Members
+   * @param user The DAC User
+   * @return List of Dataset IDs
+   */
+  public List<Integer> findDatasetIdsByDACUser(User user) {
+    return datasetDAO.findDatasetIdsByDACUserId(user.getUserId());
```
rjohanek, line 423: why change from user email to user id?

rjohanek (approved):
this looks great! so much more efficient!

src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
rushtong: > why a random user id for testing?

The only reason is to have something to query on. The actual value is irrelevant.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rushtong: The original code was using string comparison which can sometimes be flaky based on casing, etc. In theory, both should work, but I feel like the integer primary key is more stable than than the email.

fboulnois: Yes, typically you want to query by the integer primary key because it will be indexed, which makes for fast searches.

fboulnois (commented):
looks good, it might be worth describing the performance increase in the description 👍

fboulnois (approved):

PR #2414, opened 2024-10-22
[DT-698]Handle Sam User azureB2cId error

rjohanek:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-698

## Summary
The error case discussed here is when a user has an email that is both a gmail and a microsoft account and they've authenticated in a different way than they have previously. Example: Originally the user signed in authed as a google account so their user was created with google, then they auth as a microsoft account. We want this case to fail and surface the error. In the future, we want to improve the error message so it is more informative and instructional.

### What
- Propagate Sam errors in registering a user instead of just logging the error or catching and throwing a different error.

### Why
- We want to pass along this error to the duos ui instead of having it fail silently.

### Testing
- Added unit tests for error cases
- Replicated error manually by testing using the GET user/me endpoint and the POST user endpoint using an email that is both a google and microsoft account. After registering the user while authed using google, then sign in using microsoft and attempt to get the current user. Since the OAuth for microsoft is broken for consent, I copied the microsoft token from authing with Sam and used it in a curl command to my local consent. For the get me endpoint, the previous behavior was catching and logging the specific error and throwing a 404. New behavior is returning the specific Sam error. For example:

<b> New response </b>
```
curl -X 'GET' \ 
  'https://local.dsde-dev.broadinstitute.org:27443/api/user/me' \
  -H 'accept: application/json' \
  -H 'Authorization: Bearer <microsoft token>'  

{"code":500,"message":"Error posting user registration information. Email: <email>. Error message: Cannot update azureB2cId for user <user id number> because user does not exist or the azureB2cId has already been set for this user."}
```

Old response: fail with 404
```
{"message":"Unable to find user with email: <email>","code":404}
```

<b> New response </b>

```
curl -X 'POST' \
  'https://local.dsde-dev.broadinstitute.org:27443/api/user' \
  -H 'accept: application/json' \
  -H 'Authorization: Bearer <microsoft token> \   
  -H 'Content-Type: application/json' \
  -d '{}'

{"code":500,"message":"Error posting user registration information. Email: <email>. Error message: Cannot update azureB2cId for user <user id number> because user does not exist or the azureB2cId has already been set for this user."}
```

Old response: fail silently, return user without status information

```
{"userId":<num>,"email":<email>,"displayName":"unknown","createDate":1729798128501,"roles":[{"userRoleId":<num>,"userId":<num>,"roleId":<num>,"name":"Researcher"}],"emailPreference":true}
```

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
```diff
@@ -97,17 +97,13 @@ public UserStatus postRegistrationInfo(AuthUser authUser) throws Exception {
     HttpResponse response = executeRequest(request);
     String body = response.parseAsString();
     if (!response.isSuccessStatusCode()) {
-      if (HttpStatusCodes.STATUS_CODE_CONFLICT == response.getStatusCode()) {
```
rjohanek, line 100: removed this as a separate case because it was being caught and generalized in the OAuthAuthenticator anyways

src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
```diff
@@ -97,17 +97,13 @@ public UserStatus postRegistrationInfo(AuthUser authUser) throws Exception {
...
-        throw new ConsentConflictException("User exists in Sam: " + authUser.getEmail());
-      } else {
-        String errorMsg = String.format("Error posting user registration information to Sam. Email [%s]. Status Code [%s]; Status Message [%s];  ",
+      String message = JsonParser.parseString(body).getAsJsonObject().get("message").getAsString();
+      String errorMsg = String.format("Error posting user registration information. Email: %s. Error message: %s.",
```
rjohanek, line 103: removed "to Sam" because I don't think users need to know about Sam. I'm open to workshopping this message more. And I also want to improve the error message from Sam.

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
```diff
@@ -102,9 +99,15 @@ private AuthUser getUserWithStatusInfo(AuthUser authUser) {
         } else {
           logWarn("Error posting to Sam, AuthUser not able to be registered: " + gson.toJson(authUser));
         }
-      } catch (Exception exc) {
-        logException("AuthUser not able to be registered: '" + gson.toJson(authUser), exc);
+        // if post response is not successful this will be thrown, propagate the error to the user
```
pshapiro4broad: Personally I would put the comment inside the `catch` instead of before it, so it's nested inside the block it's commenting on

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
```diff
@@ -102,9 +99,15 @@ private AuthUser getUserWithStatusInfo(AuthUser authUser) {
...
+      } catch (WebApplicationException ex) {
+        throw ex;
+        // if building the request or parsing the response fails, this will be thrown
+        // authenticationExceptions are caught and rethrown as a 500 error in AuthFilter
+      } catch (Exception ex2) {
```
pshapiro4broad: This could also be `ex` since its scope is separate from the previous `catch`
```suggestion
      } catch (Exception ex) {
```

src/test/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticatorTest.java
```diff
@@ -31,16 +34,17 @@ class OAuthAuthenticatorTest {
   private SamService samService;
   private OAuthAuthenticator oAuthAuthenticator;
   private final ClaimsCache headerCache = ClaimsCache.getInstance();
+  private final String bearerToken = RandomStringUtils.randomAlphabetic(100);
+  private final MultivaluedMap<String, String> headerMap = new MultivaluedHashMap<>();
 
   @BeforeEach
   void setUp() {
     headerCache.cache.invalidateAll();
+    headerMap.clear();
```
pshapiro4broad: This is unnecessary as a new map is created for each test (method) run. If the map was `static` then this would be needed.

In the case of `headerCache`, `invalidateAll()` is needed because `ClaimsCache` has static state.

src/test/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticatorTest.java
```diff
@@ -99,13 +97,44 @@ void testAuthenticateGetUserWithStatusInfoFailurePostUserSuccess() throws Except
...
+  void testAuthenticateGetUserWithStatusInfoFailurePostUserFailureAuthEx() throws Exception {
+    headerMap.put(ClaimsCache.OAUTH2_CLAIM_email, List.of("email"));
+    headerMap.put(ClaimsCache.OAUTH2_CLAIM_name, List.of("name"));
+    headerCache.loadCache(bearerToken, headerMap);
+    when(samService.getRegistrationInfo(any())).thenThrow(new NotFoundException());
+    when(samService.postRegistrationInfo(any())).thenThrow(new AuthenticationException("errorMessage"));
+    oAuthAuthenticator = new OAuthAuthenticator(samService);
+
+    AuthenticationException ex = assertThrows(AuthenticationException.class, () -> oAuthAuthenticator.authenticate(bearerToken));
+    assertEquals("AuthUser not able to be registered: '{\"email\":\"email\",\"name\":\"name\"}", ex.getMessage());
```
pshapiro4broad: A text block can be used here
```suggestion
    assertEquals("""
        AuthUser not able to be registered: '{"email":"email","name":"name"}""", ex.getMessage());
```

pshapiro4broad (approved):

src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
```diff
@@ -97,17 +97,13 @@ public UserStatus postRegistrationInfo(AuthUser authUser) throws Exception {
     HttpResponse response = executeRequest(request);
     String body = response.parseAsString();
     if (!response.isSuccessStatusCode()) {
-      if (HttpStatusCodes.STATUS_CODE_CONFLICT == response.getStatusCode()) {
-        throw new ConsentConflictException("User exists in Sam: " + authUser.getEmail());
-      } else {
-        String errorMsg = String.format("Error posting user registration information to Sam. Email [%s]. Status Code [%s]; Status Message [%s];  ",
+      String message = JsonParser.parseString(body).getAsJsonObject().get("message").getAsString();
```
rushtong: This looks brittle to me - we're expecting that ALL non-successful cases will always have a `message` object. I think it will be safer to pass through the whole body instead of trying to parse it.

src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
rushtong: I'm fine with this. This was helpful mostly just for me when looking at consent logs.

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
```diff
@@ -102,9 +99,15 @@ private AuthUser getUserWithStatusInfo(AuthUser authUser) {
...
+        throw new AuthenticationException("AuthUser not able to be registered: '" + gson.toJson(authUser), ex2);
```
rushtong: In this scenario, the AuthUser HAS authenticated successfully, so it doesn't make sense to me that we throw an `AuthenticationException`. The error is a Sam specific registration error. Can we create a new exception case for this, maybe something like `RegistrationException extends WebApplicationException` or something similar? Open to other suggestions on this.

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
```diff
@@ -28,8 +30,7 @@ public OAuthAuthenticator(SamService samService) {
   }
 
   @Override
-  public Optional<AuthUser> authenticate(String bearer) {
-    try {
+  public Optional<AuthUser> authenticate(String bearer) throws AuthenticationException {
```
rushtong: Minor nit ... after removing the try/catch, the inner code is no longer indented correctly. Try reformatting the code to apply the project code style.

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
rjohanek: I think that could be a good idea- I am open to changing this because AuthenticationExceptions get caught and rethrown as general 500 errors. But, just to clarify - Has the user authed successfully in this case? If getting the registration info fails we try to post the registration info, and this is the case that posting also failed either because the request couldn't be built or the response couldn't be parsed. I guess what is the difference between failing to register a user and failing to auth?

src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
rjohanek: okay true, I think it would be preferable to have just the message so maybe I could try to get the message and if it doesn't exist I could return the whole body

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
rushtong: > Has the user authed successfully in this case? If getting the registration info fails we try to post the registration info, and this is the case that posting also failed either because the request couldn't be built or the response couldn't be parsed. I guess what is the difference between failing to register a user and failing to auth?

An authed user has gone through the authentication dance with the selected identity provider, Google or MS. We have a valid auth token and we know there is a real identity. A registered Sam user is an identity that Sam knows about and has information about - it's a completely separate entity.

rushtong (approved):
Thanks for the changes!

PR #2415, opened 2024-10-24
[DT-697] DUOS: B2C new MS login user display name is "unknown"

cinyecai:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-697
### Summary
If `OAUTH2_CLAIM_name=unknown` default the display name to the auth claim's email instead.

----

- Verify all tests go green


s-rubenstein (approved):
Edit: I missed what Repo this was in - please ignore me

pshapiro4broad (approved):
Looks good to me. Are the unit tests for this class? If so, could this be tested there?

cinyecai:
> Could you make it `name === "unknown"` rather than `name.equals("unknown")`? I know we have the null check before, but in case it is undefined I want to make sure we are using the safer check.

Can you explain a bit more about the distinction, just for my understanding? I don't think "===" can be used in Java.

s-rubenstein:
I misunderstood where this PR is - ignore me :|

cinyecai:
> Looks good to me. Are the unit tests for this class? If so, could this be tested there?

Yes, there are tests for this class, I made sure they all still pass. I can add some testing to test this change too!

rjohanek (approved):
with an additional unit test, I think this looks good!

PR #2417, opened 2024-10-29
[DT-841]Update error message for Sam AzureB2cId error in consent

rjohanek:
### Addresses
Ticket: https://broadworkbench.atlassian.net/browse/DT-841
Provide user instructions to make the error message actionable

### Summary
Make helper function to get the error message for post registration and provide user instructions if the error is the azureB2cId error.
Add tests of the new error message functionality

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
@@ -98,16 +98,24 @@ public class SamDAO implements ConsentLogger {
     HttpResponse response = executeRequest(request);
     String body = response.parseAsString();
     if (!response.isSuccessStatusCode()) {
-      JsonElement messageElement = JsonParser.parseString(body).getAsJsonObject().get("message");
-      String message = messageElement != null ? messageElement.getAsString() : body;
-      String errorMsg = String.format("Error posting user registration information. Email: %s. Error message: %s.",
-          authUser.getEmail(),
-          message);
+      var errorMsg = getErrorMessage(authUser, body);
       Exception e = new WebApplicationException(errorMsg, response.getStatusCode());
       logException(errorMsg, new Exception(body));
       throw e;
     }
     return new Gson().fromJson(body, UserStatus.class);
+  }
+
+  public static String getErrorMessage(AuthUser authUser, String body) {
+    JsonElement messageElement = JsonParser.parseString(body).getAsJsonObject().get("message");
+    String message = messageElement != null ? messageElement.getAsString() : body;
+    var errorMsg = String.format("Error posting user registration information. Email: %s. %s.",
+        authUser.getEmail(),
+        message);
+    if (message.contains("Cannot update azureB2cId")) {
+      errorMsg = String.format("Email: %s. You may have previously signed in with a different authentication provider (Google or Microsoft). Please sign in with that provider. For more information visit: https://support.terra.bio/hc/en-us/community/posts/24089648317467-Cannot-update-azureB2cId-for-user", authUser.getEmail());
+    }
+    return errorMsg;
   }
 
   public void asyncPostRegistrationInfo(AuthUser authUser) {
--- a/src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
@@ -12,6 +12,7 @@ import com.google.common.util.concurrent.MoreExecutors;
 import com.google.gson.Gson;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonParser;
+import com.google.gson.JsonSyntaxException;
 import com.google.gson.reflect.TypeToken;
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
```diff
@@ -98,18 +98,26 @@ public UserStatus postRegistrationInfo(AuthUser authUser) throws Exception {
...
   }
 
+  public static String getErrorMessage(AuthUser authUser, String body) {
+    JsonElement messageElement = JsonParser.parseString(body).getAsJsonObject().get("message");
+    String message = messageElement != null ? messageElement.getAsString() : body;
+    var errorMsg = String.format("Error posting user registration information. Email: %s. %s.",
+        authUser.getEmail(),
+        message);
+    if (message.contains("Cannot update azureB2cId")) {
+      errorMsg = String.format("Email: %s. You may have previously signed in with a different authentication provider (Google or Microsoft). Please sign in with that provider. For more information visit: https://support.terra.bio/hc/en-us/community/posts/24089648317467-Cannot-update-azureB2cId-for-user", authUser.getEmail());
```
cinyecai: Nice 👍

cinyecai (approved):
Lgtm, thanks for adding the Terra support link

src/test/java/org/broadinstitute/consent/http/service/dao/SamDAOTest.java
```diff
@@ -338,4 +339,32 @@ void testReadTimeout() {
...
+        getErrorMessage(authUser, body));
+  }
+
+  @Test
+  void testGetErrorMessageNoMessage() {
+    when(authUser.getEmail()).thenReturn("email@email.com");
+    String body = """
+        {"code":500}""";
+    assertEquals("""
+        Error posting user registration information. Email: email@email.com. {"code":500}.""",
```
fboulnois, line 366: This looks unusual. Is it the case that the SAM API returns a string concatenated with a JSON blob?

fboulnois (commented):
One comment:

src/test/java/org/broadinstitute/consent/http/service/dao/SamDAOTest.java
rjohanek: In the case where the error from Sam does not have a message, the entire body of the error is concatenated to this string. This is represented by line 111 in SamDAO. I made this change in [my previous PR](https://github.com/DataBiosphere/consent/pull/2414) in response to [a comment greg made ](https://github.com/DataBiosphere/consent/pull/2414#discussion_r1816513459)about just extracting the message being brittle.

rushtong: In my opinion, there are a lot of assumptions in this code - this was what I was trying to convey in my earlier feedback. We are expecting that Sam will _ALWAYS_ send an error object that contains a `code` and a `message` object. I think that expectation is brittle. Sam certainly can respond that way, but a 502 error, for example, will not contain those json elements. I think the error handling should be very flexible. If we can do the parsing, do the parsing. Otherwise, return whatever it is that Sam gives us.

src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
```diff
@@ -98,18 +98,26 @@ public UserStatus postRegistrationInfo(AuthUser authUser) throws Exception {
...
+      var errorMsg = getErrorMessage(authUser, body);
       Exception e = new WebApplicationException(errorMsg, response.getStatusCode());
       logException(errorMsg, new Exception(body));
       throw e;
     }
     return new Gson().fromJson(body, UserStatus.class);
   }
 
+  public static String getErrorMessage(AuthUser authUser, String body) {
+    JsonElement messageElement = JsonParser.parseString(body).getAsJsonObject().get("message");
```
rushtong: This is where I think the source of the brittleness is. I would love to see tests that exercise the case where `body` is null, is a non-json string, or some other random value.

src/test/java/org/broadinstitute/consent/http/service/dao/SamDAOTest.java
rjohanek: I think the only assumption being made here is that Sam will always return a json string with any number of elements, including none. Is that a safe assumption? Or does Sam return non json string objects/null sometimes? Just to clarify, The code checks if there is a "message", and if there is not it returns the entire json string. So, it doesn't assume there is a "message", it also makes so assumptions about there being a "code". I just used that as an example of a json string without a message for the test.

src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
rjohanek: Okay, if we think that we cannot rely on Sam to return a json string, then I think that makes sense. I'll add handling for those cases.

rushtong (approved):
Thank you 👍🏽

PR #2420, opened 2024-10-30
DT-945: Handle Update Study Registration Better

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-945

### Summary
This PR makes some changes to the Update Study API to fix a few bugs. As a result of updating a production study (see [DT-941](https://broadworkbench.atlassian.net/browse/DT-941)), there was a non-trivial amount of hand-editing the JSON to get around incorrect validation rules. This PR fixes that to make updates a lot more smoother in the future.

* Phenotype is not a required field
* Data Custodian Email  is not a required field
* If the consent group (i.e. dataset) is being updated:
  * All data use related fields are ignored
  * Access management changes are ignored
* If the consent group (i.e. dataset) is a new entity, all data use and access management values are preserved.
* Everything else provided in the study is valid for update.

----

- Verify all tests go green

[DT-941]: https://broadworkbench.atlassian.net/browse/DT-941?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/DatasetRegistrationSchemaV1UpdateValidator.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/DatasetRegistrationSchemaV1UpdateValidator.java
@@ -1,5 +1,8 @@
 package org.broadinstitute.consent.http.models.dataset_registration_v1;
 
+import com.google.gson.ExclusionStrategy;
+import com.google.gson.FieldAttributes;
+import com.google.gson.Gson;
 import jakarta.ws.rs.BadRequestException;
 import java.util.HashSet;
 import java.util.List;
@@ -11,6 +14,7 @@ import org.broadinstitute.consent.http.models.Dataset;
 import org.broadinstitute.consent.http.models.Study;
 import org.broadinstitute.consent.http.models.dataset_registration_v1.DatasetRegistrationSchemaV1.NihAnvilUse;
 import org.broadinstitute.consent.http.service.DatasetService;
+import org.broadinstitute.consent.http.util.gson.GsonUtil;
 
 public class DatasetRegistrationSchemaV1UpdateValidator {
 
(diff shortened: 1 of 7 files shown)


src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/DatasetRegistrationSchemaV1UpdateValidator.java
```diff
@@ -20,6 +27,89 @@ public DatasetRegistrationSchemaV1UpdateValidator(DatasetService datasetService)
...
+    public boolean shouldSkipClass(Class<?> aClass) {
+      return aClass.getSimpleName().equalsIgnoreCase("ConsentGroup");
+    }
+  };
+
+  private final ExclusionStrategy consentGroupExclusionStrategy = new ExclusionStrategy() {
+    @Override
+    public boolean shouldSkipField(FieldAttributes fieldAttributes) {
+      final HashSet<String> exclusions = new HashSet<>(List.of(
+          "accessManagement",
```
rjohanek: are data use fields an enum somewhere?

rjohanek (approved):
this seems like a great improvement to make updating studies easier for users in the future!

src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/DatasetRegistrationSchemaV1UpdateValidator.java
rushtong: Aha ... they are defined as static strings in `DatasetRegistrationSchemaV1Builder`, thank you! Will make that update.

snf2ye (approved):
I'm having a hard time wrapping my head around using GSON to convert between json and java objects, but as far as I can tell, LGTM!

rushtong:
> I'm having a hard time wrapping my head around using GSON to convert between json and java objects, but as far as I can tell, LGTM!

To be honest, this was super painful. If I were to re-implement this, I might choose a very different set of updates to support.

PR #2421, opened 2024-10-31
DT-950: Handle NPE in delete study API

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-950

### Summary
The DELETE study api is expecting that all studies have datasets although this is not enforced in the code. This PR allows the delete to continue if there are no datasets in the study. It also adds tests to complete the coverage of this endpoint.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/StudyResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/StudyResource.java
@@ -148,7 +148,7 @@ public class StudyResource extends Resource {
         throw new NotFoundException("Study not found");
       }
 
-      boolean deletable = study.getDatasets()
+      boolean deletable = (study.getDatasets() == null || study.getDatasets().isEmpty()) || study.getDatasets()
           .stream()
           .allMatch(Dataset::getDeletable);
       if (!deletable) {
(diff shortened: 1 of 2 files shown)


rjohanek (approved):
looks good. do we need a test for the case where datasetIds is null too?

src/main/java/org/broadinstitute/consent/http/resources/StudyResource.java
```diff
@@ -148,7 +148,7 @@ public Response deleteStudyById(@Auth AuthUser authUser, @PathParam("studyId") I
         throw new NotFoundException("Study not found");
       }
 
-      boolean deletable = study.getDatasets()
+      boolean deletable = (study.getDatasets() == null || study.getDatasets().isEmpty()) || study.getDatasets()
```
fboulnois, line 151: Just double checking the intent here: if the study is null or empty, then deletable is true (implying that the study is deletable). Is this correct?

fboulnois (approved):
Looks good 👍

rushtong:
> looks good. do we need a test for the case where datasetIds is null too?

Good question @raejohanek, thanks. I added a null check test which uncovered another potential NPE on dataset ids, which is also now handled.

PR #2423, opened 2024-11-05
[DT-970] Draft renaming + returning draft type in summary list.

otchet-broad:
### Addresses

[DT-970](https://broadworkbench.atlassian.net/browse/DT-970)

### Summary
Adds API to edit a draft name.
Returns draft type in DraftSummary entry

----

- Verify all tests go green

[DT-970]: https://broadworkbench.atlassian.net/browse/DT-970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DraftResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DraftResource.java
@@ -124,6 +124,7 @@ public class DraftResource extends Resource {
   @Produces({MediaType.APPLICATION_JSON})
   @Consumes({MediaType.TEXT_PLAIN})
   @Path("/v1/{draftUUID}")
+  @RolesAllowed({ADMIN, DATASUBMITTER})
   public Response patchDraftName(@Auth AuthUser authUser, @PathParam("draftUUID") String draftUUID, String name) {
     try {
       User user = userService.findUserByEmail(authUser.getEmail());
--- a/src/main/java/org/broadinstitute/consent/http/resources/DraftResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DraftResource.java
@@ -9,6 +9,7 @@ import jakarta.ws.rs.DELETE;
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.InternalServerErrorException;
 import jakarta.ws.rs.NotFoundException;
+import jakarta.ws.rs.PATCH;
 import jakarta.ws.rs.POST;
 import jakarta.ws.rs.PUT;
 import jakarta.ws.rs.Path;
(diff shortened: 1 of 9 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DraftResource.java
```diff
@@ -119,6 +120,23 @@ public Response updateDraft(@Auth AuthUser authUser, @PathParam("draftUUID") Str
...
 
+  @PATCH
+  @Produces({MediaType.APPLICATION_JSON})
+  @Consumes({MediaType.TEXT_PLAIN})
+  @Path("/v1/{draftUUID}")
+  @RolesAllowed({ADMIN, DATASUBMITTER})
+  public Response patchDraftName(@Auth AuthUser authUser, @PathParam("draftUUID") String draftUUID, String name) {
+    try {
+      User user = userService.findUserByEmail(authUser.getEmail());
+      DraftInterface draft = draftService.getAuthorizedDraft(validateUUID(draftUUID), user);
```
rushtong, line 131: General comment here, no action needed unless we do some refactoring. In `draftService.getAuthorizedDraft`, we check for user permissions on the draft. In general, we've been applying ACL-like conditions here in the resource layer, not in the business layer.

rushtong (approved):
👍🏽 Looks good, thank you!

src/main/java/org/broadinstitute/consent/http/resources/DraftResource.java
otchet-broad: I've seen different approaches to this over time and can certainly incorporate that in the future.

src/main/java/org/broadinstitute/consent/http/resources/DraftResource.java
```diff
@@ -119,6 +120,23 @@ public Response updateDraft(@Auth AuthUser authUser, @PathParam("draftUUID") Str
     }
   }
 
+  @PATCH
+  @Produces({MediaType.APPLICATION_JSON})
+  @Consumes({MediaType.TEXT_PLAIN})
+  @Path("/v1/{draftUUID}")
+  @RolesAllowed({ADMIN, DATASUBMITTER})
+  public Response patchDraftName(@Auth AuthUser authUser, @PathParam("draftUUID") String draftUUID, String name) {
```
pshapiro4broad, line 128: Is this the only field in a draft that can be updated? I'm wondering why there's a patch API for name rather than a general API that can update any draft field.

pshapiro4broad (approved):
Looks OK but it's hard to evaluate this code without context. Is there a design doc that describes how this feature works, what APIs are needed to implement it, etc?

src/main/java/org/broadinstitute/consent/http/resources/DraftResource.java
otchet-broad: Effectively, yes - it's the only metadata field that can be updated.  The created/updated when and by fields are computed by the service.  The payload (e.g. whatever the frontend wants to store as a draft) is handled by the POST endpoint as a replacement for the contents.

otchet-broad:
@pshapiro4broad  - We've got some documentation [here](https://docs.google.com/document/d/1BV-TPoPZF-ik0lYssyX3cXfPxZXzPzs0h5XlYcqY7M4) and [here](https://docs.google.com/document/d/1KPGK-EqY51w57UFYqF7smFyaOneCEKA5rnyL9zJmV2k/edit?pli=1&tab=t.0) that I'm working from/updating.

PR #2424, opened 2024-11-06
DT-978: Fix patch name bug

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-978

### Summary
Previous patch functionality would clear out the dataset name if not provided. This PR fixes it to make it truly optional.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
@@ -397,16 +397,19 @@ public class DatasetServiceDAO implements ConsentLogger {
       String datasetName,
       Integer userId,
       List<DatasetProperty> properties) {
-    if (datasetName == null || datasetName.isBlank()) {
-      Dataset d = datasetDAO.findDatasetById(datasetId);
-      datasetName = d.getDatasetName();
-    }
     // update dataset
-    datasetDAO.updateDatasetNameWithUpdateUser(
-        datasetId,
-        datasetName,
-        new Timestamp(new Date().getTime()),
-        userId);
+    if (datasetName == null || datasetName.isBlank()) {
+      datasetDAO.updateDatasetUpdateUser(
+          datasetId,
+          new Timestamp(new Date().getTime()),
+          userId);
+    } else {
+      datasetDAO.updateDatasetNameWithUpdateUser(
+          datasetId,
+          datasetName,
+          new Timestamp(new Date().getTime()),
+          userId);
+    }
     // insert properties
     executeSynchronizeDatasetProperties(handle, datasetId, properties, false);
 
--- a/src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
@@ -397,6 +397,10 @@ public class DatasetServiceDAO implements ConsentLogger {
       String datasetName,
       Integer userId,
       List<DatasetProperty> properties) {
+    if (datasetName == null || datasetName.isBlank()) {
+      Dataset d = datasetDAO.findDatasetById(datasetId);
+      datasetName = d.getDatasetName();
+    }
     // update dataset
     datasetDAO.updateDatasetNameWithUpdateUser(
         datasetId,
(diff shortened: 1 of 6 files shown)


fboulnois (approved):
nice catch and tests look good 👍

src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
```diff
@@ -398,11 +398,18 @@ public void executePatchDataset(Handle handle,
       Integer userId,
       List<DatasetProperty> properties) {
     // update dataset
-    datasetDAO.updateDatasetNameWithUpdateUser(
-        datasetId,
-        datasetName,
-        new Timestamp(new Date().getTime()),
-        userId);
+    if (datasetName == null || datasetName.isBlank()) {
```
pshapiro4broad, line 401: When I see validation like `isBlank()` in a DAO, I wonder if it's actually possible for a blank string to get here, and if so, if it could be validated further up the call chain. My general approach is to validate data at the outermost layer (here it would be in `DatasetResource`). This can help avoid duplication of validation at different downstream locations. The validation could replace blank with `null`, or it could overwrite blank and null with the current dataset name.

src/test/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAOTest.java
```diff
@@ -725,6 +725,30 @@ void testPatchDataset() throws Exception {
...
+
+    // Validate that the name is NOT updated to a null value
+    assertEquals(dataset.getName(), patched.getDatasetName());
+  }
+
+  @Test
+  void testPatchDatasetWithEmptyName() throws Exception {
+    Dataset dataset = createDataset();
+    User user = userDAO.findUserById(dataset.getCreateUserId());
+    DatasetPatch patch = new DatasetPatch("", List.of());
```
pshapiro4broad, line 744: Maybe it should contain whitespace to test `isBlank()`? Otherwise it's only testing the `isEmpty()` case.
```suggestion
    DatasetPatch patch = new DatasetPatch("    ", List.of());
```

src/test/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAOTest.java
```diff
@@ -725,6 +725,30 @@ void testPatchDataset() throws Exception {
     assertEquals(3, patched.getProperties().size());
   }
 
+  @Test
+  void testPatchDatasetWithNullName() throws Exception {
+    Dataset dataset = createDataset();
+    User user = userDAO.findUserById(dataset.getCreateUserId());
+    DatasetPatch patch = new DatasetPatch(null, List.of());
+    serviceDAO.patchDataset(dataset.getDatasetId(), user, patch);
+    Dataset patched = datasetDAO.findDatasetById(dataset.getDatasetId());
```
pshapiro4broad, line 734: Have you thought about making a fully mocked test for this class? It seems like you could mock the DAOs and JDBI so a test doesn't need to DB fetch to verify the code. That could also help narrow the scope of a test.

pshapiro4broad (approved):
Looks OK, just a few questions about the code base in general

src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
rushtong: In this case, we have validation on the patch object itself. I'll add some more there for additional coverage.

src/test/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAOTest.java
rushtong: TBH, no, we haven't tried that. I find more bugs by testing actual DB integration than in our other testing layers.

src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
rushtong: And to be clear, the patch validation methods are called from `DatasetResource`

pshapiro4broad: The problem I see here is that the validation isn't normalizing the data, so this code needs to check for two equivalent cases instead of one case. If the validation folded equivalent cases into a single canonical case then you'd only need to check one case here.

PR #2429, opened 2024-11-21
DT-1060: Add auditing to DatasetServiceDAO

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1060

### Summary

- Adds dataset audits to the insert, update, and patch operations.
- Significant test updates to ensure the audits happen where appropriate
- Test updates to move from the deprecated random string utility methods we used for random values

Note to reviewers, there are a lot of formatting changes so I recommend hiding white space

----
Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

PR #2433, opened 2024-12-03
DT-1078: Fix Study Update bug that cleared out dataset name

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-1078

## Summary
Add logic around the dataset update block to ensure we're not clearing out the dataset name. 

The existing Study Update code provides a dataset registration object that is conditionally populated with ConsentGroup objects. If the user creates a new consent group, the full data is provided. If the user edits an existing consent group, then only partial data is provided that may or may not include the consent group name. This update allows the API to function correctly in both cases where the name is or is not provided by the UI.

This could be handled in the UI as well. However, a significant amount of validation refactoring would be required AND this UI is being deprecated in favor of a new study registration process. For those reasons, it is easier and quicker to handle this edge case in server code.

## Testing Strategy
1. Run this code locally
2. Run a local UI against this instance
3. As a Chairperson, navigate to the My DAC's Datasets page
4. Choose a dataset that has a pencil icon which indicates that it is editable
5. That should direct you to the edit study page
6. Choose something to edit on the study and submit: there should be no errors
7. Repeat the process and choose to edit a consent group and submit: there should be no errors

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
@@ -368,23 +368,17 @@ public class DatasetServiceDAO implements ConsentLogger {
       boolean executeDeletes) {
 
     // Don't update the name if it isn't provided
-    if (StringUtils.isBlank(datasetName)) {
-      Dataset dataset = datasetDAO.findDatasetById(datasetId);
-      addAuditRecord(datasetId, dataset.getName(), userId, AuditActions.UPDATE);
-      datasetDAO.updateDatasetUpdateUser(datasetId, new Timestamp(new Date().getTime()),
-          userId
-      );
-    } else {
-      addAuditRecord(datasetId, datasetName, userId, AuditActions.UPDATE);
-      // update dataset
-      datasetDAO.updateDatasetByDatasetId(
-          datasetId,
-          datasetName,
-          new Timestamp(new Date().getTime()),
-          userId,
-          dacId
-      );
-    }
+    Dataset dataset = datasetDAO.findDatasetById(datasetId);
+    String updateName = StringUtils.isBlank(datasetName) ? dataset.getName() : datasetName;
+    addAuditRecord(datasetId, updateName, userId, AuditActions.UPDATE);
+    // update dataset
+    datasetDAO.updateDatasetByDatasetId(
+        datasetId,
+        updateName,
+        new Timestamp(new Date().getTime()),
+        userId,
+        dacId
+    );
 
     // insert properties
     executeSynchronizeDatasetProperties(handle, datasetId, properties, executeDeletes);
--- a/src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
@@ -12,6 +12,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
+import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.consent.http.db.DatasetDAO;
 import org.broadinstitute.consent.http.db.FileStorageObjectDAO;
 import org.broadinstitute.consent.http.db.StudyDAO;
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
```diff
@@ -365,15 +366,25 @@ public void executeUpdateDatasetWithFiles(Handle handle,
...
-        new Timestamp(new Date().getTime()),
-        userId,
-        dacId
-    );
+
+    // Don't update the name if it isn't provided
+    if (StringUtils.isBlank(datasetName)) {
+      Dataset dataset = datasetDAO.findDatasetById(datasetId);
+      addAuditRecord(datasetId, dataset.getName(), userId, AuditActions.UPDATE);
+      datasetDAO.updateDatasetUpdateUser(datasetId, new Timestamp(new Date().getTime()),
```
pshapiro4broad: This omits `dacId` vs the call below, is that correct?

src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
```diff
@@ -365,15 +366,25 @@ public void executeUpdateDatasetWithFiles(Handle handle,
...
-        datasetId,
-        datasetName,
-        new Timestamp(new Date().getTime()),
-        userId,
-        dacId
-    );
+
+    // Don't update the name if it isn't provided
+    if (StringUtils.isBlank(datasetName)) {
+      Dataset dataset = datasetDAO.findDatasetById(datasetId);
```
pshapiro4broad: Instead of a different code path here, what about
```suggestion
    if (StringUtils.isBlank(datasetName)) {
      datasetName = datasetDAO.findDatasetById(datasetId).datasetName;
    }
    addAuditRecord(datasetId, datasetName, userId, AuditActions.UPDATE);
    // update dataset
    datasetDAO.updateDatasetByDatasetId(
    ...
```

src/test/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAOTest.java
```diff
@@ -729,6 +730,76 @@ void testDeleteStudy() throws Exception {
 
   }
 
+  @Test
+  void testExecuteUpdateDatasetWithNullName() throws Exception {
+    // This creates a study with a single dataset:
+    Study study = createStudy(List.of());
+    List<Dataset> datasets = datasetDAO.findDatasetsByIdList(study.getDatasetIds().stream().toList());
```
pshapiro4broad: A better way to convert a set to a list is `copyOf()`
```suggestion
    List<Dataset> datasets = datasetDAO.findDatasetsByIdList(List.copyOf(study.getDatasetIds());
```
this creates a read-only view of the set and doesn't actually copy any data.

Although, I think an even better fix here is to change `findDatasetsByIdList()` to accept `Collection<Integer>` instead of `List<Integer>`. The docs for `@BindList` says the parameter must be an array or an Iterable so `Collection` should work.

src/test/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAOTest.java
```diff
@@ -729,6 +730,76 @@ void testDeleteStudy() throws Exception {
...
+    assertFalse(audits.isEmpty());
+  }
+
+  @Test
+  void testExecuteUpdateDatasetWithNewName() throws Exception {
+    // This creates a study with a single dataset:
+    Study study = createStudy(List.of());
+    List<Dataset> datasets = datasetDAO.findDatasetsByIdList(study.getDatasetIds().stream().toList());
+    Dataset dataset = datasets.get(0);
+    String newName = RandomStringUtils.secureStrong().nextAlphabetic(dataset.getName().length() + 10);
```
pshapiro4broad: What is the purpose of using a random string here? Is it possible for this test to collide with another test?

If you do need a random string here, why use `secureStrong()` instead of `insecure()`?

If the real need is to avoid collisions, I would use a name based on a timestamp instead of a random string.

src/test/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAOTest.java
```diff
@@ -729,6 +730,76 @@ void testDeleteStudy() throws Exception {
...
+        null,
+        study.getCreateUserId(),
+        dataset.getDacId(),
+        List.of(),
+        List.of(),
+        false)
+    );
+    Dataset updatedDataset = datasetDAO.findDatasetById(dataset.getDatasetId());
+    assertNotNull(updatedDataset);
+    assertNotNull(updatedDataset.getName());
```
pshapiro4broad: In this case I would omit the first `assertNotNull` since it will throw a NPE if it's null and the test will fail anyway.

pshapiro4broad (approved):
Looks OK, a few comments

src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
rushtong: Great catch - no, we do need to handle that case

src/test/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAOTest.java
rushtong: Using random strings is a general testing pattern that we have used because we have seen flaky tests with non-random strings. With the update to org.apache.commons.lang3.RandomStringUtils, the old way we did this was deprecated, so we moved to `secureStrong` which was recommended by Elliot. I'll update this PR to make use of what we did in https://github.com/DataBiosphere/consent/pull/2429/files#diff-e141aaccd3526037b702f172e86b30f409866db7d35266121ab09ef38adf3693R182

fboulnois (approved):
👍

PR #2444, opened 2025-01-07
DT-982: Update models, add attachment functionality

rushtong:
## Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DT-982

This work will enable the replacement of UI code (see [Support.js](https://github.com/DataBiosphere/duos-ui/blob/2b5d3f777019de681d2fb808193a1ac79fbad53e/src/libs/ajax/Support.js#L6)) with Consent APIs.

## Summary
This PR covers the following work:
* Update/add service layer code to enable zendesk ticket/upload functionality
* Replace home grown ticket models with supported third-party library models
* Add file upload service code
* Add an integration test that should **_only_** be run manually.
* Expand unit tests for file uploads

## Notes
The third-party library has a zendesk http client, but it requires authenticated usage. Our existing front-end usage does not make authenticated calls, so to recreate existing functionality, we need to not use that library and use our existing http infrastructure. On the plus side, we get http metrics for free. On the minus side, API usage could theoretically be abused. We'll look into usage throttling when we add client support for accessing this service code.

### Testing
If you wish to test this, modify `SupportRequestServiceIntegrationTest` to enable the test and run it. This will create a real ticket in our system, so please use this sparingly. 

Example ticket:
Screenshot 2025-01-07 at 9 57 32 AM

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/models/support/TicketFactory.java
```diff
@@ -0,0 +1,84 @@
...
+    if (url == null) {
+      throw new IllegalArgumentException("Support ticket url is required");
+    }
+
+    Ticket ticket = new Ticket(
+        new Ticket.Requester(name, email),
+        subject,
+        createComment(description, url, uploads));
+    ticket.setCustomFields(createCustomFields(name, type, email, description));
+    ticket.setTicketFormId(360000669472L);
```
s-rubenstein, line 63: What are these constants for? Could we assign them to a variable with a name or put a comment?

s-rubenstein (approved):
Looks reasonable

src/main/java/org/broadinstitute/consent/http/models/support/TicketFactory.java
rushtong: I'll add some comments here, they come from the original UI implementation: https://github.com/DataBiosphere/duos-ui/blob/2b5d3f777019de681d2fb808193a1ac79fbad53e/src/libs/ajax/Support.js#L6

snf2ye (approved):

PR #2448, opened 2025-01-14
DT-1143: Passthrough APIs for support requests and content uploads

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1143

### Summary
This PR adds two APIs that act as passthroughs to Zendesk's [Request API](https://developer.zendesk.com/api-reference/ticketing/tickets/ticket-requests) service. The intention here is to approximate what the UI expects in terms of response codes and objects. There are some minor changes to reduce the scope of what the UI needs to do in terms of parsing responses, but in general, it hews fairly close to what Zendesk returns.

See also: https://github.com/DataBiosphere/duos-ui/pull/2768

### Testing
To test sending a real ticket to Zendesk, update your local `consent.yaml` config value:
```
services:
  ...
  activateSupportNotifications: true
```
The `Support` APIs are not authenticated so can be easily tested in swagger.

----

- Verify all tests go green


snf2ye (approved):

src/main/java/org/broadinstitute/consent/http/service/SupportRequestService.java
```diff
@@ -57,13 +58,12 @@ public String postAttachmentToSupport(byte[] content) throws Exception {
...
         if (uploadObj != null && uploadObj.get("token") != null) {
-          return uploadObj.get("token").getAsString();
+          return uploadObj.get("token").getAsJsonObject();
         }
       }
-    } else {
-      logDebug("Not configured to send support attachments");
     }
-    return null;
+    logDebug("Not configured to send support attachments");
```
pshapiro4broad: If you throw, do you need to log as well? Usually when throwing you'd expect to log in the catch clause. Logging when throwing can result in the same error being logged twice.

src/main/java/org/broadinstitute/consent/http/resources/SupportResource.java
```diff
@@ -0,0 +1,67 @@
...
+import org.broadinstitute.consent.http.service.SupportRequestService;
+import org.broadinstitute.consent.http.util.gson.GsonUtil;
+import org.owasp.fileio.FileValidator;
+import org.zendesk.client.v2.model.Request;
+
+@Path("support")
+public class SupportResource extends Resource {
+
+  private final SupportRequestService supportRequestService;
+  private final static TicketFactory ticketFactory = new TicketFactory();
```
pshapiro4broad: `TicketFactory` has no data, is there a reason its methods aren't `static`? That way you wouldn't need to instantiate it here.

src/main/java/org/broadinstitute/consent/http/resources/SupportResource.java
```diff
@@ -0,0 +1,67 @@
...
+  private final static Gson gson = GsonUtil.getInstance();
+  private final static FileValidator validator = new FileValidator();
```
pshapiro4broad: `FileValidator` has mutable data so I don't think it's correct to use `static` to reference it here, since it will be shared across all instances of `SupportResource`.
```suggestion
  private final FileValidator validator = new FileValidator();
```

src/main/java/org/broadinstitute/consent/http/resources/SupportResource.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+  }
+
+  @POST
+  @Path("upload")
+  @Consumes("application/binary")
+  @Produces(MediaType.APPLICATION_JSON)
+  public Response postUpload(byte[] content) {
+    try {
+      if (content.length > validator.getMaxFileUploadSize()) {
```
pshapiro4broad: How is `FileValidator.getMaxFileUploadSize()` ever modified? If it's not modified, why not get its value during class construction and store that in a field?
```
    static final long MAX_FILE_UPLOAD_SIZE = new FileValidator.getMaxFileUploadSize();
```

This is the only usage of `validator` in this class

src/main/java/org/broadinstitute/consent/http/resources/SupportResource.java
```diff
@@ -0,0 +1,67 @@
...
+import org.broadinstitute.consent.http.util.gson.GsonUtil;
+import org.owasp.fileio.FileValidator;
+import org.zendesk.client.v2.model.Request;
+
+@Path("support")
+public class SupportResource extends Resource {
+
+  private final SupportRequestService supportRequestService;
+  private final static TicketFactory ticketFactory = new TicketFactory();
+  private final static Gson gson = GsonUtil.getInstance();
```
pshapiro4broad: The canonical ordering of modifiers is `static final`
```suggestion
  private static final Gson gson = GsonUtil.getInstance();
```

src/main/java/org/broadinstitute/consent/http/resources/SupportResource.java
```diff
@@ -0,0 +1,67 @@
...
+
+  @POST
+  @Path("request")
+  @Consumes(MediaType.APPLICATION_JSON)
+  @Produces(MediaType.APPLICATION_JSON)
+  public Response postRequest(String body) {
+    try {
+      TicketFields ticketFields = gson.fromJson(body, TicketFields.class);
+      DuosTicket ticket = ticketFactory.createTicket(ticketFields);
+      logInfo("Support Request Ticket: " + ticket.toString());
```
pshapiro4broad: `toString()` is implicit in a string concatenation expression
```suggestion
      logInfo("Support Request Ticket: " + ticket);
```

src/test/java/org/broadinstitute/consent/http/resources/SupportResourceTest.java
```diff
@@ -0,0 +1,189 @@
...
+          "subject": "Test Subject",
+          "description": "Test Description",
+          "type": "%s",
+          "url": "https://example.com",
+          "uploads": [
+            "token1",
+            "token2",
+          ]
+        }
+        """.formatted(type.name());
```
pshapiro4broad: `.name()` is redundant here
```suggestion
        """.formatted(type));
```

s-rubenstein (approved):
LGTM

src/test/java/org/broadinstitute/consent/http/resources/SupportResourceTest.java
```diff
@@ -0,0 +1,189 @@
...
+          "description": "Test Description",
+          "type": "QUESTION",
+          "uploads": [
+            "token1",
+            "token2",
+          ]
+        }
+        """;
+    try (Response response = supportResource.postRequest(body)) {
+      assertEquals(HttpStatusCodes.STATUS_CODE_BAD_REQUEST, response.getStatus());
```
pshapiro4broad, line 130: The test `body`s are different but the assert is the same for each. Using a parameterized test would make that clearer. I think this is equivalent
```
  private static Stream<Arguments> testPostRequestInvalidFields() {
    return Stream.of(
        Arguments.of("""
        {
          "email": "test.user@example.com",
          "subject": "Test Subject",
          "description": "Test Description",
          "type": "QUESTION",
          "url": "https://example.com",
          "uploads": ["token1", "token2"]
        }
        """),
        Arguments.of("""
        {
          "name": "Test User",
          "subject": "Test Subject",
          "description": "Test Description",
          "type": "QUESTION",
          "url": "https://example.com",
          "uploads": ["token1", "token2"]
        }
        """),
        Arguments.of("""
        {
          "name": "Test User",
          "email": "test.user@example.com",
          "description": "Test Description",
          "type": "QUESTION",
          "url": "https://example.com",
          "uploads": ["token1", "token2"]
        }
        """),
        Arguments.of("""
        {
          "name": "Test User",
          "email": "test.user@example.com",
          "subject": "Test Subject",
          "type": "QUESTION",
          "url": "https://example.com",
          "uploads": ["token1", "token2"]
        }
        """),
        Arguments.of("""
        {
          "name": "Test User",
          "email": "test.user@example.com",
          "subject": "Test Subject",
          "description": "Test Description",
          "url": "https://example.com",
          "uploads": ["token1", "token2"]
        }
        """),
        Arguments.of("""
        {
          "name": "Test User",
          "email": "test.user@example.com",
          "subject": "Test Subject",
          "description": "Test Description",
          "type": "QUESTION",
          "uploads": ["token1", "token2"]
        }
        """)
    );
  }

  @ParameterizedTest
  @MethodSource
  void testPostRequestInvalidFields(String body) {
    try (Response response = supportResource.postRequest(body)) {
      assertEquals(HttpStatusCodes.STATUS_CODE_BAD_REQUEST, response.getStatus());
    }
  }
```

src/test/java/org/broadinstitute/consent/http/resources/SupportResourceTest.java
```diff
@@ -0,0 +1,189 @@
...
+    }
+  }
+
+  @Test
+  public void testPostUpload() throws Exception {
```
pshapiro4broad: `public` is unnecessary here
```suggestion
  void testPostUpload() throws Exception {
```

src/main/java/org/broadinstitute/consent/http/models/support/TicketFields.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+    if (description() == null) {
+      throw new IllegalArgumentException("Description is required");
+    }
+    if (type() == null) {
+      throw new IllegalArgumentException("Type is required");
+    }
+    if (url() == null) {
+      throw new IllegalArgumentException("URL is required");
+    }
```
pshapiro4broad: References to fields from inside a `record` don't need to use method calls. Also the common code here could be refactoreed
```suggestion
  private static void validate(boolean condition, String prefix) {
    if (!condition) {
      throw new IllegalArgumentException(prefix + " is required");
    }
  }

  public void validate() {
    validate(name == null || email == null, "Name and email of user requesting support");
    validate(subject == null, "Subject");
    validate(description == null, "Description");
    validate(type == null, "Type");
    validate(url == null, "URL");
  }
```

src/main/java/org/broadinstitute/consent/http/models/support/TicketFactory.java
```diff
@@ -28,37 +28,15 @@ public Request parseRequestResponse(String response) {
...
+  public DuosTicket createTicket(TicketFields ticketFields) {
+    ticketFields.validate();
     Ticket ticket = new Ticket(
-        new Ticket.Requester(name, email),
-        subject,
-        createComment(description, url, uploads));
-    ticket.setCustomFields(createCustomFields(name, type, email, description));
+        new Ticket.Requester(ticketFields.name(), ticketFields.email()),
+        ticketFields.subject(),
+        createComment(ticketFields.description(), ticketFields.url(), ticketFields.uploads()));
```
pshapiro4broad: When you're mapping from one object to another, I think it's best to put the mapping code inside either the target or source, if you have control over that code. In this case we have control over the source object, so this could be moved into `TicketFields`:
```
public record TicketFields(

...

  Ticket toTicket() {
    return new Ticket(new Ticket.Requester(name, email), subject, createComment());
  }

  private Comment createComment() {
    Comment comment = new Comment();
    comment.setBody(description + "\n\n------------------\nSubmitted from: " + url);
    if (uploads != null && !uploads.isEmpty()) {
      comment.setUploads(uploads);
    }
    return comment;
  }
```

src/main/java/org/broadinstitute/consent/http/models/support/TicketFactory.java
```diff
@@ -28,37 +28,15 @@ public Request parseRequestResponse(String response) {
...
+    ticket.setCustomFields(createCustomFields(ticketFields.name(), ticketFields.type(), ticketFields.email(), ticketFields.description()));
```
pshapiro4broad: This could be moved into `TicketFields` too

pshapiro4broad (approved):
Looks OK, with a few comments

src/main/java/org/broadinstitute/consent/http/service/SupportRequestService.java
rushtong: Thanks, these were leftover debugging lines I didn't intend to keep around

PR #2449, opened 2025-01-16
[DT-97]Update DAR schema to allow Progress Report entities

rjohanek:
### Addresses
Ticket:https://broadworkbench.atlassian.net/browse/DT-97

### Summary

Add missing fields that represent progress report entities to the DataAccessRequest schema. These fields already exist in the backend in [DataAccessRequestData](https://github.com/DataBiosphere/consent/blob/develop/src/main/java/org/broadinstitute/consent/http/models/DataAccessRequestData.java).
This work just coverts the models into yaml schemas and adds them to DataAccessRequest.yaml

----

- Verify all tests go green


fboulnois (commented):
The intent of the original ticket was to also be able to submit these entities, although the acceptance criteria is not explicit about that piece. We could add that to a separate ticket if you'd like, or we could make it part of this ticket.

I think there may need to be some code added to our submit data access request endpoint to support this new functionality.

We should also add some unit tests for the API submission of a progress report if they don't already exist.

rjohanek:
> The intent of the original ticket was to also be able to submit these entities, although the acceptance criteria is not explicit about that piece. We could add that to a separate ticket if you'd like, or we could make it part of this ticket.
> 
> I think there may need to be some code added to our submit data access request endpoint to support this new functionality.
> 
> We should also add some unit tests for the API submission of a progress report if they don't already exist.

Oh okay, yeah the ticket just said to update the schema. Maybe we could talk more about the original intent on Tuesday and when I understand more I can either update this work or make another ticket

src/main/resources/assets/schemas/DataAccessRequest.yaml
```diff
@@ -203,3 +203,28 @@ properties:
...
+    items:
+      $ref: './Publication.yaml'
+  presentations:
+    type: array
+    description: An array of presentations.
+    items:
+      $ref: './Presentation.yaml'
+  dataManagementIncident:
+    $ref: './DataManagementIncident.yaml'
+    description: A Data Management Incident.
```
rushtong: There is a warning here that is a little confusing at first. The root cause is that if there is a `$ref`, then anything after it, i.e. description, is ignored. The idea is that all information should in theory be inside the `$ref`, not here in the property. Same thing applies to `closeOutSupplement`

rushtong: Example of an incorrect intellij warning:

Screenshot 2025-01-21 at 7 48 07 AM

rushtong (approved):
This looks correct to me after manually comparing against the existing java classes. See minor comment inline. I'd be happy with this moving forward and a separate PR for ensuring that the existing DAR crud operations work with these objects as defined here 👍🏽

fboulnois (approved):
👍

PR #2451, opened 2025-01-23
[DT-824] Return 400 for invalid byte sequence

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-824

### Summary

- A wait strategy was added for alternate container runtimes like `colima`, which take longer to bind the testcontainer database port. This caused tests to crash.
- A null character (`\0`) anywhere in a UTF-8 string is a valid UTF-8 byte sequence, however Postgres rejects these null characters in UTF-8 strings, likely because Postgres uses null-terminated strings internally ("C-style" strings) and these can cause security issues. The error is changed such that if an invalid byte sequence is detected it returns a 400 instead.
- Tests are added to assert that invalid names throw an error.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/Resource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/Resource.java
@@ -15,12 +15,12 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.sql.SQLException;
 import java.sql.SQLSyntaxErrorException;
-import java.util.AbstractMap;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.broadinstitute.consent.http.enumeration.UserRoles;
 import org.broadinstitute.consent.http.exceptions.ConsentConflictException;
 import org.broadinstitute.consent.http.exceptions.UnknownIdentifierException;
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/consent/http/resources/Resource.java
```diff
@@ -159,9 +159,15 @@ private interface ExceptionHandler {
         Resource::unableToExecuteExceptionHandler);
     dispatch.put(PSQLException.class,
         Resource::unableToExecuteExceptionHandler);
-    dispatch.put(SQLSyntaxErrorException.class, e ->
-        errorLoggedExceptionHandler(e,
-            new Error("Database Error", Response.Status.INTERNAL_SERVER_ERROR.getStatusCode())));
+    dispatch.put(SQLSyntaxErrorException.class, e -> {
+      if (e.getMessage().contains("invalid byte sequence")) {
+        return errorLoggedExceptionHandler(e,
+            new Error("Invalid byte sequence", Response.Status.BAD_REQUEST.getStatusCode()));
```
fboulnois: How can we test this specific error?

rushtong: In any resource test, if you mock the underlying service method to throw the exception, that should exercise this condition.

rushtong (approved):
Nice work 👍🏽

s-rubenstein (approved):

PR #2454, opened 2025-02-03
[DT-1172][risk=no] Draft Attachment POST response should only return refs for objects posted.

otchet-broad:
### Addresses 

[DT-1172](https://broadworkbench.atlassian.net/browse/DT-1172?atlOrigin=eyJpIjoiYzQ0ZTI0NzQxYTM1NDA4YTgyMWE3NDUxNDU3MjE5MWEiLCJwIjoiaiJ9) 

### Summary

DT-1172 requires building a reusable front-end component that can be leveraged multiple times on a single page for file upload.  When building out the component, it became impossible to know which files were successfully uploaded in the request when more than one component was loaded on the page.  This was because the response payload of the post included attachments that had been uploaded in prior requests.  This change updates the POST response to only include the references for the attachments uploaded in the request.  The full list of attachments will be obtained in the GET on the draft/attachments API call or when retrieving a draft.

----

- Verify all tests go green

[DT-1172]: https://broadworkbench.atlassian.net/browse/DT-1172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/test/java/org/broadinstitute/consent/http/resources/DraftResourceTest.java
```diff
@@ -272,12 +272,12 @@ void testUploadAttachmentSuccess() throws SQLException {
     when(draftService.getAuthorizedDraft(any(), any())).thenReturn(new DraftStudyDataset("{}", user));
     DraftInterface draftWithAttachment = new DraftStudyDataset("{}", user);
     draftWithAttachment.addStoredFile(mock(FileStorageObject.class));
-    when(draftService.addAttachments(any(), any(), any())).thenReturn(draftWithAttachment);
+    when(draftService.addAttachments(any(), any(), any())).thenReturn(draftWithAttachment.getStoredFiles().stream().toList());
```
pshapiro4broad: I would use `List.copyOf()` to convert a `Set` to a `List`, unless you need a mutable copy of the set.
```suggestion
    when(draftService.addAttachments(any(), any(), any())).thenReturn(List.copyOf(draftWithAttachment.getStoredFiles()));
```

src/test/java/org/broadinstitute/consent/http/resources/DraftResourceTest.java
```diff
@@ -272,12 +272,12 @@ void testUploadAttachmentSuccess() throws SQLException {
...
     initResource();
     Response response = resource.addAttachments(authUser, UUID.randomUUID().toString(), mock(
         FormDataMultiPart.class));
     assertEquals(HttpStatusCodes.STATUS_CODE_OK, response.getStatus());
-    assertEquals(draftWithAttachment.getStoredFiles(), response.getEntity());
+    assertEquals(draftWithAttachment.getStoredFiles().stream().toList(), response.getEntity());
```
pshapiro4broad:
```suggestion
    assertEquals(List.copyOf(draftWithAttachment.getStoredFiles()), response.getEntity());
```

rushtong (approved):
Looks reasonable to me, thank you 👍🏽

rjohanek (approved):
looks good to me, thanks!

src/test/java/org/broadinstitute/consent/http/service/DraftServiceTest.java
```diff
@@ -147,6 +148,19 @@ void testDeleteDraftsForUser() throws SQLException, IOException {
     draftService.deleteDraftsByUser(user2);
   }
 
+  @Test
+  void testAddAttachmentToDraft() throws SQLException, IOException {
+    when(gcsService.storeDocument(any(), anyString(), any())).thenReturn(
+        BlobId.of(UUID.randomUUID().toString(), UUID.randomUUID().toString()));
```
pshapiro4broad: This is OK but in general I think it's better to mock the layers directly beneath the one being tested.

In this case, instead of mocking `GCSService` and creating a `DraftFileStorageService`, you'd mock `DraftFileStorageService` and only mock `GCSService` when testing `DraftFileStorageService`.

pshapiro4broad (approved):
Looks ok, just a few minor comments

fboulnois (approved):
good suggestions from @pshapiro4broad , otherwise lgtm! 👍

src/test/java/org/broadinstitute/consent/http/service/DraftServiceTest.java
otchet-broad: Thanks for your suggestion @pshapiro4broad.  I appreciate your perspective and think this begs a bit of a larger conversation about the scope of this projects tests in general (are they unit, are they integration, should we be running these tests without an underlying database, etc.) and will hold off making the change until we can have that discussion.

src/test/java/org/broadinstitute/consent/http/service/DraftServiceTest.java
```diff
@@ -1,210 +1,78 @@
...
-
-  @NotNull
-  private DraftInterface createDraft(User user, Integer numberOfFiles)
-      throws SQLException {
-    DraftStudyDataset draft = new DraftStudyDataset("{}", user);
-    draftService.insertDraft(draft);
-    Map<String, FormDataBodyPart> mapOfFiles = getRandomFiles(numberOfFiles);
-    return draftService.addAttachments(draft, user, mapOfFiles);
+  void testGetDraftFile() throws IOException, SQLException {
+    when(gcsService.getDocument((BlobId) any()))
```
otchet-broad: @pshapiro4broad - Mocking the direct service input now.

rushtong (approved):
Thanks for the separation, this looks good 👍🏽

fboulnois (approved):
👍

src/main/java/org/broadinstitute/consent/http/ConsentApplication.java
```diff
@@ -84,6 +84,7 @@
 import org.broadinstitute.consent.http.service.DatasetRegistrationService;
 import org.broadinstitute.consent.http.service.DatasetService;
 import org.broadinstitute.consent.http.service.DraftService;
+import org.broadinstitute.consent.http.service.dao.DraftServiceDAO;
```
pshapiro4broad: This import appears to be unused, can it be removed?

src/main/java/org/broadinstitute/consent/http/service/DraftService.java
```diff
@@ -2,173 +2,75 @@
...
-  public DraftInterface addAttachments(DraftInterface draft, User user,
-      Map<String, FormDataBodyPart> files) throws SQLException {
-    draftFileStorageService.storeDraftFiles(draft.getUUID(), user, files);
-    draftDAO.updateDraftByDraftUUID(draft.getUUID(),
-        new Date().toInstant(), user.getUserId());
-    return findDraftByDraftUUID(draft.getUUID());
+  public StreamingOutput draftAsJson(DraftInterface draft) {
+      Gson gson = GsonUtil.buildGson();
+      return output -> {
+        output.write("{ \"document\":".getBytes());
```
pshapiro4broad: Is it typical in consent to build a json object by hand like this? Instead I would expect creating a POJO and then using `gson.toJson()`, or creating the Json object directly. Something like
```
    Gson gson = GsonUtil.buildGson();
    return output -> {
      output.write(gson.toJson(
          Map.of("document", gson.fromJson(draft.getJson(), Object.class), 
              "meta", draft)).getBytes());
    };
```

Although this does require converting `draft` to an object and back, maybe that's what this code is trying to avoid?

src/main/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAO.java
```diff
@@ -0,0 +1,156 @@
...
+    jdbi.useHandle(handle -> {
+      handle.getConnection().setAutoCommit(false);
+      try {
+        draftDAO.insert(draft.getName(), draft.getCreateDate().toInstant(),
+            draft.getCreateUser().getUserId(), draft.getJson(), draft.getUUID(),
+            draft.getType().getValue());
+      } catch (Exception e) {
+        handle.rollback();
+        throw new BadRequestException(
+            "Error submitting draft.  Drafts require valid json to be submitted.");
```
pshapiro4broad: Is there a reason to not include `e` in this exception? Leaving it out means that the underlying exception is lost, and it may contain useful information.

src/main/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAO.java
```diff
@@ -0,0 +1,156 @@
...
+    draft.setUpdateUser(user);
+    draft.setUpdateDate(new Date());
+    jdbi.useHandle(handle -> {
+      handle.getConnection().setAutoCommit(false);
+      try {
+        draftDAO.updateDraftByDraftUUID(draft.getName(),
+            draft.getUpdateDate().toInstant(), draft.getUpdateUser().getUserId(), draft.getJson(),
+            draft.getUUID(), draft.getType().getValue());
+      } catch (Exception e) {
+        handle.rollback();
```
pshapiro4broad, line 63: If the draft fails to update, wouldn't we want to let the user know? I would expect to see an exception thrown here.

src/main/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAO.java
```diff
@@ -0,0 +1,156 @@
...
+      }
+      handle.commit();
```
pshapiro4broad, line 67: Not sure how often in consent this pattern comes up, but it might be worth adding a wrapper around `handle` that abstracts this code pattern. something like
```
  private void jdbiTransaction(Consumer<Handle> operation) throws SQLException {
    jdbi.useHandle(handle -> {
      handle.getConnection().setAutoCommit(false);
      try {
        operation.accept(handle);
        handle.commit();
      } catch (Exception e) {
        handle.rollback();
        throw new SQLException("Transaction failed and was rolled back.", e);
      }
    });
  }

  public DraftInterface updateDraft(DraftInterface draft, User user) throws SQLException {
    draft.setUpdateUser(user);
    draft.setUpdateDate(new Date());
    jdbiTransaction(handle ->
      draftDAO.updateDraftByDraftUUID(draft.getName(),
          draft.getUpdateDate().toInstant(), draft.getUpdateUser().getUserId(), draft.getJson(),
          draft.getUUID(), draft.getType().getValue()));
    return getAuthorizedDraft(draft.getUUID(), user);
  }
```

src/main/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAO.java
```diff
@@ -0,0 +1,156 @@
...
+
+  public DraftInterface getAuthorizedDraft(UUID draftUUID, User user) {
+    DraftInterface draft;
+    try {
+      draft = findDraftByDraftUUID(draftUUID);
+    } catch (SQLException e) {
+      throw new NotFoundException(
+          String.format("Draft with UUID %s not found.", draftUUID.toString()));
+    }
+    if (Objects.isNull(draft)) {
```
pshapiro4broad: Usually `Objects.isNull()` is used for stream operations (like `.filter(Objects::isNull)`). In this case I would compare against `null` directly
```suggestion
    if (draft == null) {
```

src/main/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAO.java
```diff
@@ -0,0 +1,156 @@
...
+    return getAuthorizedDraft(draft.getUUID(), user);
+  }
+
+  public DraftInterface getAuthorizedDraft(UUID draftUUID, User user) {
+    DraftInterface draft;
+    try {
+      draft = findDraftByDraftUUID(draftUUID);
+    } catch (SQLException e) {
+      throw new NotFoundException(
+          String.format("Draft with UUID %s not found.", draftUUID.toString()));
```
pshapiro4broad: arguments to `format()` already undergo `toString()` so it's unnecessary here
```suggestion
          String.format("Draft with UUID %s not found.", draftUUID));
```

src/main/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAO.java
```diff
@@ -0,0 +1,156 @@
...
+      throw new NotAuthorizedException("User not authorized to modify resource.");
+    }
+    return draft;
+  }
+
+  public void deleteDraftsByUser(User user) {
+    Set<DraftInterface> userDrafts = findDraftsForUser(user);
+    for (DraftInterface draft : userDrafts) {
+      deleteDraft(draft, user);
+    }
```
pshapiro4broad: In general I would avoid using `Set` unless the code requires it, in case the collection type changes in the future. You could use `Collection<DraftInterface` here or use `forEach()` to write it functionally
```suggestion
    findDraftsForUser(user).forEach(draft -> deleteDraft(draft, user));
```

Another way to solve this is to use `var`.

src/main/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAO.java
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  public Set<DraftSummary> findDraftSummariesForUser(User user) {
+    return draftDAO.findDraftSummariesByUserId(user.getUserId());
+  }
```
pshapiro4broad: The caller here is `DraftService`, the code looks like
```
  public Set<DraftSummary> findDraftSummariesForUser(User user) {
    return draftServiceDAO.findDraftSummariesForUser(user);
  }
```
Why wouldn't `DraftService` call `DraftDAO` directly in this case instead of calling `DraftServiceDAO`?

src/main/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAO.java
```diff
@@ -0,0 +1,156 @@
...
+
+  public Set<DraftInterface> findDraftsForUser(User user) {
```
pshapiro4broad: This code returns `Set<DraftInterface>` but the only concrete implementation of `DraftInterface` is `DraftStudyDataset` which doesn't define `equals()` and `hashCode()`. Without those, what use is a `Set`?

pshapiro4broad (commented):
A few comments so far

src/main/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAO.java
```diff
@@ -0,0 +1,156 @@
...
+    return draftDAO.findDraftsByUserId(user.getUserId());
+  }
+
+  private DraftInterface findDraftByDraftUUID(
+      UUID draftUUID) throws SQLException {
```
pshapiro4broad: It doesn't look like this method can throw `SQLException`
```suggestion
      UUID draftUUID) {
```

I think jdbi annotated methods can throw `JdbiException` (which is unchecked), not `SqlException`.

src/main/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAO.java
```diff
@@ -0,0 +1,156 @@
...
+          new Date().toInstant(), user.getUserId());
+    } else {
+      throw new NotFoundException(
+          String.format("Draft attachment is not found.  Draft: %s, Attachment: %d",
+              draft.getUUID(), fileId));
+    }
+  }
+
+  public void deleteDraft(DraftInterface draft, User user)
+      throws RuntimeException {
```
pshapiro4broad: Some methods here throw `SqlException` and some throw unchecked exceptions, like this. Do consent DAOs typically return a mix of exception types? I would except the exception response to be the same across all methods so it's easier for the endpoints to have consistent behavior.

src/main/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAO.java
```diff
@@ -0,0 +1,156 @@
...
+            try {
+              draftFileStorageServiceDAO.deleteStoredFile(fileStorageObject, user);
+            } catch (SQLException e) {
+              throw new RuntimeException(e);
+            }
+          });
+        });
+      } catch (Exception e) {
+        handle.rollback();
+        throw new RuntimeException(e);
```
pshapiro4broad: Is there a more specific exception type that can be used here? When we report it to the user, will it be a client or server error http code?

src/main/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAO.java
```diff
@@ -0,0 +1,156 @@
...
+      try {
+        handle.useTransaction(handler -> {
+          draftDAO.deleteDraftByUUIDList(List.of(draft.getUUID()));
+          draft.getStoredFiles().forEach(fileStorageObject -> {
+            try {
+              draftFileStorageServiceDAO.deleteStoredFile(fileStorageObject, user);
+            } catch (SQLException e) {
+              throw new RuntimeException(e);
+            }
+          });
```
pshapiro4broad: It looks like the purpose of this try/catch is to allow the use of `forEach()`. If so, I think it would be clearer to write this as a loop and avoid `forEach()`.
```suggestion
          for (FileStorageObject fileStorageObject : draft.getStoredFiles()) {
            draftFileStorageServiceDAO.deleteStoredFile(fileStorageObject, user);
          }
```

src/test/java/org/broadinstitute/consent/http/service/DraftServiceTest.java
```diff
@@ -1,210 +1,78 @@
...
-    assertEquals(storedDraft.getUUID(), draft.getUUID());
-  }
-
-  @Test
-  void testCreateDraftWithInvalidJson() {
-    User user = createUser();
-    DraftStudyDataset draft = new DraftStudyDataset("Hello world!",user);
-    assertThrows(BadRequestException.class, ()-> draftService.insertDraft(draft));
-  }
+public class DraftServiceTest {
```
pshapiro4broad: In JUnit 5, you no longer need to declare a test class or its methods `public`
```suggestion
class DraftServiceTest {
```

src/test/java/org/broadinstitute/consent/http/service/DraftServiceTest.java
```diff
@@ -1,210 +1,78 @@
...
-    DraftStudyDataset draft = new DraftStudyDataset("{}", user);
-    draftService.insertDraft(draft);
-    Map<String, FormDataBodyPart> mapOfFiles = getRandomFiles(numberOfFiles);
-    return draftService.addAttachments(draft, user, mapOfFiles);
+  void testGetDraftFile() throws IOException, SQLException {
+    when(gcsService.getDocument((BlobId) any()))
+        .thenAnswer(inputStream -> new ByteArrayInputStream("{}".getBytes(StandardCharsets.UTF_8)) {
+        });
+    DraftService draftService = new DraftService(draftServiceDAO, gcsService);
+      FileStorageObject fileStorageObject = new FileStorageObject();
```
pshapiro4broad: indents appear off here

src/test/java/org/broadinstitute/consent/http/service/DraftServiceTest.java
```diff
@@ -1,210 +1,78 @@
...
+    DraftService draftService = new DraftService(draftServiceDAO, gcsService);
+    assertThrows(BadRequestException.class, ()-> draftService.insertDraft(null));
   }
 
   @Test
-  void testStreamingOutput() throws SQLException, IOException {
-    User user = createUser();
-    DraftInterface draft = createDraft(user, 1);
-    StreamingOutput output = draftService.draftAsJson(draft);
+  public void testStreamingOutput() throws SQLException, IOException {
```
pshapiro4broad: I think the `throws` clause here is wrong; the called methods don't throw `SQLException`. For test code, i think it's easier to use `throws Exception` since a test method doesn't care what exceptions are thrown; any will cause the test to fail.
```suggestion
  public void testStreamingOutput() throws Exception {
```

src/test/java/org/broadinstitute/consent/http/service/DraftServiceTest.java
pshapiro4broad: You can avoid using `any()` here by moving this line after you set up `fileStorageObject`:
```
    when(gcsService.getDocument(fileStorageObject.getBlobId()))
        .thenAnswer(inputStream -> new ByteArrayInputStream("{}".getBytes()) {
        });
```
It doesn't matter where this line goes with respect to constructing `DraftService` as the mock object is the same in either case. In fact, since `draftService` is used in all test methods, you could initialize it once in a `BeforeEach` method:
```
  private DraftService draftService;
  
  @BeforeEach
  void beforeEach() {
    draftService = new DraftService(draftServiceDAO, gcsService);
  }
```

src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
```diff
@@ -95,15 +96,15 @@ class UserServiceTest {
...
+  private DraftServiceDAO draftServiceDAO;
 
 
   private UserService service;
 
   private void initService() {
     service = new UserService(userDAO, userPropertyDAO, userRoleDAO, voteDAO, institutionDAO,
         libraryCardDAO, acknowledgementDAO, fileStorageObjectDAO, samDAO, userServiceDAO, daaDAO,
-        emailService, draftService);
+        emailService, draftServiceDAO);
```
pshapiro4broad, line 107: If you wanted to get coverage on the new code in `UserService`, you could add a line to `testDeleteUser()` below like
```
    verify(draftServiceDAO).deleteDraftsByUser(u);
```

src/test/java/org/broadinstitute/consent/http/service/DraftServiceTest.java
```diff
@@ -217,32 +85,4 @@ public StreamingDeserializer(String document, DraftStudyDataset meta) {
...
-    when(part.getName()).thenReturn(name);
-    when(part.getMediaType()).thenReturn(MediaType.MULTIPART_FORM_DATA_TYPE);
-    when(part.getValueAs(InputStream.class))
-        .thenReturn(new ByteArrayInputStream(EMPTY_JSON_DOCUMENT.getBytes()));
-    return part;
-  }
-
-
-}
+}
```
pshapiro4broad, line 197: It's true that most of `DraftService` are passthrough methods but note that this test doesn't cover most of the methods (only 60% line coverage). I usually like to add tests for trivial code in case more business logic is added in the future, to make it easier to extend the tests.

src/test/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAOTest.java
```diff
@@ -0,0 +1,242 @@
...
+    DraftInterface storedDraft = storedDrafts.iterator().next();
+    assertThat(storedDraft.getStoredFiles(), hasSize(3));
+    assertEquals(storedDraft.getUUID(), draft.getUUID());
+  }
+
+  @Test
+  void testCreateDraftWithInvalidJson() {
+    User user = createUser();
+    DraftStudyDataset draft = new DraftStudyDataset("Hello world!",user);
+    assertThrows(BadRequestException.class, ()-> draftServiceDAO.insertDraft(draft));
```
pshapiro4broad: will be nice once we get spotless working in this repo
```suggestion
    assertThrows(BadRequestException.class, () -> draftServiceDAO.insertDraft(draft));
```

src/test/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAOTest.java
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  @Test
+  void testThinUserIsReturnedFromDraft() throws SQLException, IOException {
+    when(gcsService.storeDocument(any(), anyString(), any())).thenReturn(
+        BlobId.of(UUID.randomUUID().toString(), UUID.randomUUID().toString()));
+    User user = createUser();
+    DraftInterface draft = createDraft(user, 3);
+    assertThat(user.getRoles(), hasSize(greaterThan(0)));
```
pshapiro4broad, line 92: What is the purpose of this assert? Is it testing a method that `createUser()` is calling?

src/test/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAOTest.java
```diff
@@ -0,0 +1,242 @@
...
+    Map<String, FormDataBodyPart> mapOfFiles = getRandomFiles(numberOfFiles);
+    draftServiceDAO.addAttachments(draft, user, mapOfFiles);
+    return draftServiceDAO.getAuthorizedDraft(draft.getUUID(), user);
+  }
+
+  private void assertThinUser(User user) {
+    assertThat(user.getRoles(), is(nullValue()));
+    assertThat(user.getUserId(), is(notNullValue()));
+    assertThat(user.getInstitutionId(), is(nullValue()));
+    assertThat(user.getEraCommonsId(), is(nullValue()));
```
pshapiro4broad: `is()`  is unnecessary here. But maybe it does make it "read" better?
```suggestion
    assertThat(user.getRoles(), nullValue());
    assertThat(user.getUserId(), notNullValue());
    assertThat(user.getInstitutionId(), nullValue());
    assertThat(user.getEraCommonsId(), nullValue());
```

src/test/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAOTest.java
```diff
@@ -0,0 +1,242 @@
...
+    DraftInterface updatedDraft = draftServiceDAO.updateDraft(draft, user);
+    assertEquals(draft.getUUID(), updatedDraft.getUUID());
+    assertEquals(newDraftName, updatedDraft.getName());
+    assertEquals(updatedJson, updatedDraft.getJson());
+    assertThinUser(updatedDraft.getCreateUser());
+    assertThinUser(updatedDraft.getUpdateUser());
+  }
+
+  @NotNull
+  private DraftInterface createDraft(User user, Integer numberOfFiles)
```
pshapiro4broad: Unless you need to store `null`, I would avoid using boxed types like `Integer`. Using `int` makes it clear to the reader that the value can never be `null`.
```suggestion
  private DraftInterface createDraft(User user, int numberOfFiles)
```

src/test/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAOTest.java
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  private Map<String, FormDataBodyPart> getRandomFiles(Integer count) {
+    Map<String, FormDataBodyPart> mapOfFiles = new HashMap<>();
+    IntStream.range(0, count)
+        .forEach(index -> {
+          String name = String.format("file%d", index);
+          mapOfFiles.put(name, getFormDataBodyPartMock(name));
+        });
+    return mapOfFiles;
```
pshapiro4broad: You can go directly from an `IntStream()` to a `Map`. (This version also uses `String.formatted()` instead of `String.format()`)
```suggestion
    return
        IntStream.range(0, count)
            .mapToObj("file%d"::formatted)
            .collect(Collectors.toMap(Function.identity(), this::getFormDataBodyPartMock));
```

pshapiro4broad (approved):
Looks OK. I have a few suggestions though, mostly minor changes.

src/test/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAOTest.java
```diff
@@ -0,0 +1,242 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import com.google.cloud.storage.BlobId;
+import com.google.gson.Gson;
```
pshapiro4broad: A number of imports in this file, like this one, are unused. Be sure to check for and remove all unused imports in the changed files. (Something else spotless would help with...)

src/main/java/org/broadinstitute/consent/http/service/DraftService.java
otchet-broad: It is.  We're actually doing something similar for similar reasons in another area of code.  We've discussed this one when I did the original PR back in September and think we'll leave as-is.

src/main/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAO.java
otchet-broad: Going to leave that for a future item.

src/main/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAO.java
otchet-broad: Surprised my IDE didn't hint this.  It has in the past.  I need to look at settings.  Good find.  Thanks!

src/main/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAO.java
otchet-broad: Moved to Collection to be consistent with other changes.

src/main/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAO.java
otchet-broad: It will soon be a Collection based on your other feedback.

In some projects I've been on the use of the Set without equivalence and/or hashCode is helpful when you only want the same object in memory added to the collection once.  The underlying default implementation looks at that as a test for equivalence.  Building the set can be a highly efficient test for that.  This code doesn't need that, so I'm perfectly happy moving to Collection.

src/main/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAO.java
otchet-broad: I think in some cases we mix the type based on the underlying system where the error may be SQL or GCS (as is in this case).  In this case we're trying to both delete the record in SQL and files in GCS as part of the same "transaction" and will rollback the operation if things in GCS go wonky.  Could we summarize the exception here?  Sure.  Doing so might cause a level of misdirection.  Something we can certainly harmonize in the future.

src/test/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAOTest.java
otchet-broad: To verify It was populated in the User object and now isn't returned in the "thin" user information included with a draft.  See: assertThinUser.

src/main/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAO.java
otchet-broad: Updated to throw either SQLException or NotFoundException.

src/test/java/org/broadinstitute/consent/http/service/DraftServiceTest.java
otchet-broad: Coverage added.

otchet-broad: image

PR #2459, opened 2025-02-19
[DT-1253] Check for empty and null arguments for `findVoteUsersByElectionReferenceIdList`

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1253

### Summary

Add tests and fix for when `findVoteUsersByElectionReferenceIdList` is empty or null

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
@@ -416,6 +416,7 @@ public class DarCollectionService implements ConsentLogger {
 
   /**
    * Find all dataset ids by the DAC User. Will return ids for Chairpersons or Members
+   *
    * @param user The DAC User
    * @return List of Dataset IDs
    */
@@ -657,13 +658,21 @@ public class DarCollectionService implements ConsentLogger {
    * @param collection The DarCollection
    * @return The updated DarCollection
    */
-  public DarCollection createElectionsForDarCollection(User user, DarCollection collection) {
+  public DarCollection createElectionsForDarCollection(User user, DarCollection collection)
+      throws Exception {
     try {
       List<String> createdElectionReferenceIds = collectionServiceDAO.createElectionsForDarCollection(
           user, collection);
-      List<User> voteUsers = voteDAO.findVoteUsersByElectionReferenceIdList(
-          createdElectionReferenceIds);
+      if (createdElectionReferenceIds == null || createdElectionReferenceIds.isEmpty()) {
+        var e = new IllegalStateException(
+            "No elections were created for DAR Collection: %s".formatted(
+                collection.getDarCode()));
+        logException(e);
+        throw e;
+      }
       try {
+        List<User> voteUsers = voteDAO.findVoteUsersByElectionReferenceIdList(
+            createdElectionReferenceIds);
         emailService.sendDarNewCollectionElectionMessage(voteUsers, collection);
       } catch (Exception e) {
         logException(
@@ -673,6 +682,7 @@ public class DarCollectionService implements ConsentLogger {
     } catch (Exception e) {
       logException("Exception creating elections and votes for collection: %s".formatted(
           collection.getDarCollectionId()), e);
+      throw e;
     }
     return darCollectionDAO.findDARCollectionByCollectionId(collection.getDarCollectionId());
   }
--- a/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
@@ -663,7 +663,7 @@ public class DarCollectionService implements ConsentLogger {
     try {
       List<String> createdElectionReferenceIds = collectionServiceDAO.createElectionsForDarCollection(
           user, collection);
-      if (createdElectionReferenceIds == null || createdElectionReferenceIds.isEmpty()) {
+      if (createdElectionReferenceIds.isEmpty()) {
         var e = new IllegalStateException(
             "No elections were created for DAR Collection: %s".formatted(
                 collection.getDarCode()));
--- a/src/test/java/org/broadinstitute/consent/http/db/VoteDAOTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/db/VoteDAOTest.java
@@ -4,6 +4,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.sql.Timestamp;
@@ -411,11 +412,25 @@ class VoteDAOTest extends DAOTestHelper {
   }
 
   @Test
-  void testFindVoteUsersByElectionReferenceIdList_Empty() {
+  void testFindVoteUsersByElectionReferenceIdList_Invalid() {
     // Empty case
     List<User> voteUsers = voteDAO.findVoteUsersByElectionReferenceIdList(
         List.of("invalid reference id"));
     assertTrue(voteUsers.isEmpty());
+  }
+
+  @Test
+  void testFindVoteUsersByElectionReferenceIdList_Empty() {
+    assertThrows(IllegalArgumentException.class, () -> {
+      voteDAO.findVoteUsersByElectionReferenceIdList(List.of());
+    });
+  }
+
+  @Test
+  void testFindVoteUsersByElectionReferenceIdList_Null() {
+    assertThrows(IllegalArgumentException.class, () -> {
+      voteDAO.findVoteUsersByElectionReferenceIdList(null);
+    });
   }
 
   @Test
--- a/src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
@@ -324,22 +324,6 @@ class DarCollectionServiceTest {
   }
 
   @Test
-  void testCreateElectionsForDarCollectionNull() throws Exception {
-    User user = new User();
-    user.setEmail("email");
-    DataAccessRequest dar = new DataAccessRequest();
-    dar.setReferenceId(UUID.randomUUID().toString());
-    DarCollection collection = createMockCollections(1).get(0);
-    collection.setDars(Map.of(dar.getReferenceId(), dar));
-    when(darCollectionServiceDAO.createElectionsForDarCollection(user, collection)).thenReturn(
-        null);
-
-    assertThrows(IllegalStateException.class, () -> {
-      service.createElectionsForDarCollection(user, collection);
-    });
-  }
-
-  @Test
   void testCreateElectionsForDarCollectionVoteUsersException() throws Exception {
     User user = new User();
     user.setEmail("email");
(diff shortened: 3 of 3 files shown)


src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -657,13 +658,21 @@ public DarCollection cancelDarCollectionElectionsAsChair(DarCollection collectio
...
-  public DarCollection createElectionsForDarCollection(User user, DarCollection collection) {
+  public DarCollection createElectionsForDarCollection(User user, DarCollection collection)
+      throws Exception {
     try {
       List<String> createdElectionReferenceIds = collectionServiceDAO.createElectionsForDarCollection(
           user, collection);
-      List<User> voteUsers = voteDAO.findVoteUsersByElectionReferenceIdList(
-          createdElectionReferenceIds);
+      if (createdElectionReferenceIds == null || createdElectionReferenceIds.isEmpty()) {
+        var e = new IllegalStateException(
```
rushtong, line 667: Nice

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -657,13 +658,21 @@ public DarCollection cancelDarCollectionElectionsAsChair(DarCollection collectio
...
    */
-  public DarCollection createElectionsForDarCollection(User user, DarCollection collection) {
+  public DarCollection createElectionsForDarCollection(User user, DarCollection collection)
+      throws Exception {
     try {
       List<String> createdElectionReferenceIds = collectionServiceDAO.createElectionsForDarCollection(
           user, collection);
-      List<User> voteUsers = voteDAO.findVoteUsersByElectionReferenceIdList(
-          createdElectionReferenceIds);
+      if (createdElectionReferenceIds == null || createdElectionReferenceIds.isEmpty()) {
```
pshapiro4broad: How can the result of `createElectionsForDarCollection()` ever be `null`? When calling an `internal` API that returns a collection type, we should be able to assert that we will never see `null`, only an empty collection.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -657,13 +658,21 @@ public DarCollection cancelDarCollectionElectionsAsChair(DarCollection collectio
...
+        var e = new IllegalStateException(
+            "No elections were created for DAR Collection: %s".formatted(
+                collection.getDarCode()));
+        logException(e);
+        throw e;
+      }
       try {
```
pshapiro4broad, line 673: Nesting `try`s should be avoided when possible. In this case, both the inner and outer `try` catch `Exception`, so the code should work the same if this nested `try` of moved after the outer `catch` to line 677 below

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -673,6 +682,7 @@ public DarCollection createElectionsForDarCollection(User user, DarCollection co
     } catch (Exception e) {
       logException("Exception creating elections and votes for collection: %s".formatted(
           collection.getDarCollectionId()), e);
+      throw e;
```
pshapiro4broad, line 685: This is a change in API behavior, is that part of this ticket? Was something broken by not throwing an exception which is now fixed?

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -657,13 +658,21 @@ public DarCollection cancelDarCollectionElectionsAsChair(DarCollection collectio
...
       List<String> createdElectionReferenceIds = collectionServiceDAO.createElectionsForDarCollection(
           user, collection);
-      List<User> voteUsers = voteDAO.findVoteUsersByElectionReferenceIdList(
-          createdElectionReferenceIds);
+      if (createdElectionReferenceIds == null || createdElectionReferenceIds.isEmpty()) {
+        var e = new IllegalStateException(
+            "No elections were created for DAR Collection: %s".formatted(
+                collection.getDarCode()));
+        logException(e);
+        throw e;
```
pshapiro4broad, line 671: I've mentioned this before and I don't know how common it is in the duos codebase, but in general if you `throw`, you shouldn't `log`, since any `catch` which doesn't `throw` should be logging already. Typically, logging before `throw` causes the exception to be logged twice.

rushtong (approved):
Looks reasonable to me, thank you 👍🏽

src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
```diff
@@ -312,6 +308,55 @@ void testCreateElectionsForDarCollection() throws Exception {
...
+
+  @Test
+  void testCreateElectionsForDarCollectionNull() throws Exception {
+    User user = new User();
+    user.setEmail("email");
+    DataAccessRequest dar = new DataAccessRequest();
+    dar.setReferenceId(UUID.randomUUID().toString());
+    DarCollection collection = createMockCollections(1).get(0);
+    collection.setDars(Map.of(dar.getReferenceId(), dar));
+    when(darCollectionServiceDAO.createElectionsForDarCollection(any(), any())).thenReturn(null);
```
pshapiro4broad: I believe this is the same as not mocking at all, since all mocked APIs return `null` by default. If you only want to verify that an API is called (which mocking will do if you have strict stubbing enabled), I would instead use `verify` after the API is called, to make the intent of the code clearer.

src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
```diff
@@ -312,6 +308,55 @@ void testCreateElectionsForDarCollection() throws Exception {
...
+    dar.setReferenceId(UUID.randomUUID().toString());
+    DarCollection collection = createMockCollections(1).get(0);
+    collection.setDars(Map.of(dar.getReferenceId(), dar));
+    when(darCollectionServiceDAO.createElectionsForDarCollection(any(), any())).thenReturn(
+        List.of("electionId"));
+    when(voteDAO.findVoteUsersByElectionReferenceIdList(any())).thenThrow(
+        IllegalArgumentException.class);
+
+    service.createElectionsForDarCollection(user, collection);
+    verify(darCollectionServiceDAO, times(1)).createElectionsForDarCollection(any(), any());
```
pshapiro4broad: `times(1)` is the default so you can omit it
```suggestion
    verify(darCollectionServiceDAO).createElectionsForDarCollection(any(), any());
```

Also I would encourage matching against more specific arguments than `any()` if possible. You can use `ArgumentCaptor` to assert on a subfield of an argument in cases where you can't match the entire argument.

src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
```diff
@@ -312,6 +308,55 @@ void testCreateElectionsForDarCollection() throws Exception {
...
+    verify(voteDAO, times(1)).findVoteUsersByElectionReferenceIdList(any());
+    verify(emailService, times(0)).sendDarNewCollectionElectionMessage(any(), any());
```
pshapiro4broad: If `emailService` isn't used at all, I would use `verifyNoInteractions()` instead
```suggestion
    verifyNoInteractions(emailService);
```

src/test/java/org/broadinstitute/consent/http/db/VoteDAOTest.java
```diff
@@ -411,13 +412,27 @@ void testUpdateRationaleByVoteIds() {
...
+  void testFindVoteUsersByElectionReferenceIdList_Empty() {
+    assertThrows(IllegalArgumentException.class, () -> {
+      voteDAO.findVoteUsersByElectionReferenceIdList(List.of());
+    });
+  }
+
+  @Test
+  void testFindVoteUsersByElectionReferenceIdList_Null() {
+    assertThrows(IllegalArgumentException.class, () -> {
+      voteDAO.findVoteUsersByElectionReferenceIdList(null);
```
pshapiro4broad, line 432: I suppose there's no harm in this, but why test a case which can't arise in the code base? I think it's better to only test paths which a user can encounter.

pshapiro4broad (approved):
I have a few comments. I think the fix is OK but I don't think it's helpful to code for cases which can't arise out of database state or user interactions.

src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
fboulnois: It seems to return an empty list by default, hence the separate test for `null`.

src/test/java/org/broadinstitute/consent/http/db/VoteDAOTest.java
fboulnois: This helped identify the problem in the first place: if the arguments to `findVoteUsersByElectionReferenceIdList` are `null` or empty, then this will throw the error that we saw from Sentry. This also helped me think through a solution to the issue.

In the future, it is possible that we refactor `createElectionsForDarCollection` and accidentally return a `null`, which would cause the subsequent call to `findVoteUsersByElectionReferenceIdList` to fail. By having explicitly added this test, I think we signal to future developers (e.g. myself) that this state cannot be valid and/or otherwise they need to also rethink how `findVoteUsersByElectionReferenceIdList` works.

In general, I like the idea of exhaustively testing the valid and invalid states of a method, even if they are not currently reachable. This encodes the assumptions that we are truly making about the code, and if those assumptions change and we refactor the code, then we also need to update the tests that broke as a result.

For these reasons I think it's worthwhile to keep this test.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
fboulnois: Yes, the error wasn't being percolated up to the caller and it continued on and called `darCollectionDAO.findDARCollectionByCollectionId`, which should not occur in this case.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
fboulnois: @rushtong suggested I scope `findVoteUsersByElectionReferenceIdList` to ignore email failures but not election creation errors. I otherwise didn't modify the logic here.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
fboulnois: If I comment out the `throw`s, this causes the `null` and `empty` tests to fail, which are the states we want to avoid.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
pshapiro4broad: Yes, my point is that the original code was nesting `try`s where it didn't need to do so.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
pshapiro4broad: There's no need to remove the `throw` if that's the behavior we want, my suggestion was to either `throw` or `log` but not both, as that will likely result in the exception being logged twice. But that's assuming that the consent service is logging in the root exception handler, which may not be the case.

Looking at `Resource.createExceptionResponse()`, unknown exceptions are logged, so it appears that double logging will occur. This code throws `IllegalStateException` but I don't see a custom handler for it in `Resource`, so it should end up handled by the `Exception.class` case.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
pshapiro4broad: OK, I think see. What I'd expect to see here is a catch that rethrows as a more specific exception and wrap the previous one as a cause. The log message would instead be the new exception's message. And finally if we wanted to log, we could do it in the top level catch handler.

But it looks like that's not how the consent service works right now, so I agree it's best to keep with the current process.

I do think nesting `try`s isn't great though, and now that this method `throws Exception` you no longer need the outer `catch`

src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
pshapiro4broad: You're right, I didn't know that about mockito. That makes sense though since I don't think collection APIs should ever return nulls in the first place..

PR #2462, opened 2025-02-26
DT-1282: Handle null/empty datasets when deleting studies

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1282

### Summary
Handle the case where there are no datasets in a study when deleting it. It is possible to delete datasets outside of the study context, so this code path needs to safely handle that case.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/Study.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Study.java
@@ -119,7 +119,7 @@ public class Study {
   }
 
   public Set<Dataset> getDatasets() {
-    return datasets;
+    return datasets == null? Set.of() : datasets;
   }
 
   public Date getCreateDate() {
--- a/src/main/java/org/broadinstitute/consent/http/models/Study.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Study.java
@@ -119,7 +119,7 @@ public class Study {
   }
 
   public Set<Dataset> getDatasets() {
-    return datasets == null? Set.of() : datasets;
+    return datasets == null ? Set.of() : datasets;
   }
 
   public Date getCreateDate() {
--- a/src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
@@ -67,15 +67,17 @@ public class DatasetServiceDAO implements ConsentLogger {
   public void deleteStudy(Study study, User user) throws Exception {
     jdbi.useHandle(handle -> {
       handle.getConnection().setAutoCommit(false);
-      study.getDatasets().forEach(d -> {
-        try {
-          deleteDataset(d, user.getUserId());
-        } catch (Exception e) {
-          handle.rollback();
-          logException(e);
-          throw new DatasetDeletionException(e);
-        }
-      });
+      if (study.getDatasets() != null) {
+        study.getDatasets().forEach(d -> {
+          try {
+            deleteDataset(d, user.getUserId());
+          } catch (Exception e) {
+            handle.rollback();
+            logException(e);
+            throw new DatasetDeletionException(e);
+          }
+        });
+      }
       try {
         studyDAO.deleteStudyByStudyId(study.getStudyId());
       } catch (Exception e) {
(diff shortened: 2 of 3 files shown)


src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
```diff
@@ -67,15 +67,17 @@ public void deleteDataset(Dataset dataset, Integer userId) throws Exception {
...
-      study.getDatasets().forEach(d -> {
-        try {
-          deleteDataset(d, user.getUserId());
-        } catch (Exception e) {
-          handle.rollback();
-          logException(e);
-          throw new DatasetDeletionException(e);
-        }
-      });
+      if (study.getDatasets() != null) {
```
snf2ye: Could study.getDatasets() alternatively return an empty list if there are no datasets? Not sure if that would be better or not.

snf2ye (approved):

src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
rushtong: It's theoretically possible, but the next `forEach` safely skips empty sets.

pshapiro4broad: Could you instead fix this by changing `Study` to initialize `datasets` so it's not `null`? That would avoid the need to check for `null` everywhere this getter is used.
```
  private Set<Dataset> datasets = Set.of();
```

src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
```diff
@@ -67,15 +67,17 @@ public void deleteDataset(Dataset dataset, Integer userId) throws Exception {
...
+        study.getDatasets().forEach(d -> {
+          try {
+            deleteDataset(d, user.getUserId());
+          } catch (Exception e) {
+            handle.rollback();
+            logException(e);
+            throw new DatasetDeletionException(e);
```
pshapiro4broad: This throws `DatasetDeletionException` but nothing catches it and it's not specially handled in the top level exception handler, as far as I can tell. If you didn't need to throw a different exception here could put this inside the `try`/`catch` below by using a `for` loop instead of the `forEach()` call.

src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
otchet-broad: Looks like @snf2ye 's suggestion could likely simplify some other checks where we're both null and empty list checking as well.
If time is of a concern, I'd be comfortable with this change as proposed and adding a ticket to consider the refactoring of this per the suggestion.

src/test/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAOTest.java
```diff
@@ -729,6 +729,21 @@ void testDeleteStudy() throws Exception {
...
+    Study study = createStudy(List.of());
+    // Delete any created datasets
+    study.getDatasetIds().forEach(id -> {
+      datasetDAO.deleteDatasetPropertiesByDatasetId(id);
+      datasetDAO.deleteDatasetById(id);
+    });
+    // Ensure that study deletion succeeds
+    serviceDAO.deleteStudy(study, createUser());
+    Study deletedStudy = studyDAO.findStudyById(study.getStudyId());
+    assertNull(deletedStudy);
```
otchet-broad, line 744: Nice!

otchet-broad (approved):
Nice, quick patching!  I think @snf2ye makes a good point.

pshapiro4broad (approved):

src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
rushtong: Ah ... I think I misunderstood the suggestion. Yes, we can modify the getter, that also works.

src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
rushtong: It bubbles up to the Resource class where eventually it gets rethrown as a 500 to the client. I ended up not touching this after using Shelby's suggestion, so will leave this as is for now.

src/main/java/org/broadinstitute/consent/http/models/Study.java
```diff
@@ -119,7 +119,7 @@ public void addDatasets(List<Dataset> datasetList) {
   }
 
   public Set<Dataset> getDatasets() {
-    return datasets;
+    return datasets == null ? Set.of() : datasets;
```
pshapiro4broad, line 122: This is fine, but unnecessary if the field is initialized on construction.

PR #2464, opened 2025-03-03
DT-1284: Improve dataset query performance

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1284

### Summary
The diff on this is confusing, I think, because there are large blocks of added/removed code. For example, I didn't make any changes to `findAllStudyNames` or `findAllDatasetNames`, but they show up as new additions and deletions in the diff.

This PR does several things
- Remove usages of unused/deprecated methods
  - Deprecated: `findAllDatasets()`
  - Unused: `getDatasets()`
  - Duplicative: `findDatasetWithDataUseByIdList()` -> `findDatasetsByIdList()`
- Remove study information from dataset collection queries
  - This allows for a simpler file storage object clause that reduces query time 10-fold
  - Still pull the study info when querying for a single dataset
- Since we're not populating study on dataset collections, populate the study id instead which should be enough for clients to follow up with.
- Add inner joins on create user as that is guaranteed to be non-null
- Update swagger docs to make it clear that study info is not returned with collection-based dataset endpoint.

### Testing Notes
APIs tested include:
* Get all datasets by ids: `GET /api/dataset/batch`
* Post all datasets to ES: `POST /api/dataset/index`
* Get registration: `GET /api/dataset/registration/{datasetIdentifier}`
* Get datasets v2: `GET /api/dataset/v2`
* Get datasets v3: `GET /api/dataset/v3`
* Get study: `GET /api/dataset/study/{studyId}`
  * For example, study id: 67  with 32 datasets: 
    * Prod: 42596 ms response
    * New: 1290 ms -> 33x better response
    * On prod, study id 62 has 66 datasets and crashes the running pod.
* Get study registration: `GET /api/dataset/study/registration/{studyId}`

### Explain Plans

Before
Screenshot 2025-03-05 at 2 54 56 PM

After
Screenshot 2025-03-05 at 2 56 20 PM

The thing to note between these two is the large removal of the Bitmap Heap Scan over the `file_storage_object` table. The original query was looking for FSOs associated to either the Dataset or the Study. When we processed the rows into objects, we never looked at the Study FSO results, just the Dataset FSO. By removing that OR condition, we significantly improve performance.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
@@ -141,6 +141,13 @@ public interface DatasetDAO extends Transactional<DatasetDAO> {
       """)
   Dataset findDatasetById(@Bind("datasetId") Integer datasetId);
 
+  /**
+   * Return a list of Dataset objects from a list of dataset ids. This explicitly does NOT populate
+   * study information due to performance issues.
+   *
+   * @param datasetIds List of dataset ids
+   * @return List of datasets
+   */
   @UseRowReducer(DatasetReducer.class)
   @SqlQuery("""
           SELECT d.dataset_id, d.name, d.create_date, d.create_user_id, d.update_date,
@@ -150,22 +157,6 @@ public interface DatasetDAO extends Transactional<DatasetDAO> {
               u.create_date AS u_create_date, u.email_preference AS u_email_preference,
               u.institution_id AS u_institution_id, u.era_commons_id AS u_era_commons_id,
               k.key, dp.property_value, dp.property_key, dp.property_type, dp.schema_property, dp.property_id,
-              s.study_id AS s_study_id,
-              s.name AS s_name,
-              s.description AS s_description,
-              s.data_types AS s_data_types,
-              s.pi_name AS s_pi_name,
-              s.create_user_id AS s_create_user_id,
-              s.create_date AS s_create_date,
-              s.update_user_id AS s_user_id,
-              s.update_date AS s_update_date,
-              s.public_visibility AS s_public_visibility,
-              s_dataset.dataset_id AS s_dataset_id,
-              sp.study_property_id AS sp_study_property_id,
-              sp.study_id AS sp_study_id,
-              sp.key AS sp_key,
-              sp.value AS sp_value,
-              sp.type AS sp_type,
               fso.file_storage_object_id AS fso_file_storage_object_id,
               fso.entity_id AS fso_entity_id,
               fso.file_name AS fso_file_name,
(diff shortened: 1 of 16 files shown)


src/test/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAOTest.java
```diff
@@ -348,7 +348,7 @@ void testInsertStudyWithAlternativeDataSharingFile() throws Exception {
 
     assertEquals(1, datasets.size());
 
-    Dataset dataset1 = datasets.get(0);
+    Dataset dataset1 = datasetDAO.findDatasetById(createdIds.get(0));
```
rjohanek, line 351: what is the purpose of this change? Is it just to test findDatasetById three times? Then, what is the point of the findDatasetsByIdList(createdIds) call on line 347 for example?

rjohanek (approved):
this looks good! super excited about these performance boosts

src/test/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAOTest.java
rushtong: If you scroll down a bit to line 353, there is an assertion that the study is non-null. Originally, `datasetDAO.findDatasetsByIdList()` would populate the study for every dataset. That's no longer true, so to minimize any effects on the test, I chose to re-query for the dataset with a DAO call that DOES populate the study.

fboulnois (approved):
looks good 👍

src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetMapper.java
```diff
@@ -47,6 +47,9 @@ public Dataset map(ResultSet r, StatementContext ctx) throws SQLException {
     if (hasColumn(r, "alias")) {
       dataset.setAlias(r.getInt("alias"));
     }
+    if (hasNonZeroColumn(r, "study_id")) {
+      dataset.setStudyId(r.getInt("study_id"));
+    }
```
pshapiro4broad, line 52: I think the pattern of `if (something(... "study_id")) { .. get("study_id") }` is error prone. Maybe in a future tech debt cleanup we could improve this. I think there are two general approaches

Allow `null` and pass through to setters. In most cases, the `r.get()` result is being passed to a setter which accepts `null` so there's no need to check before using its value, since if it's `null`, setting it has no effect. So
```
    if (hasNonZeroColumn(r, "study_id")) {
      dataset.setStudyId(r.getInt("study_id"));
    }
```
becomes
```
    dataset.setStudyId(getNonZeroColumn(r, "study_id")));
```

Or, if we want to avoid unnecessary setter calls and only use the value if it's present, we could have a `get` method that accepts a consumer argument.
```
   getNonZeroColumn(r, "study_id")), dataset::setStudyId);
```

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -530,7 +531,7 @@ public List<DarCollection> addDatasetsToCollections(List<DarCollection> collecti
       if (!filterDatasetIds.isEmpty()) {
         datasetIds.retainAll(filterDatasetIds);
       }
-      Set<Dataset> datasets = datasetDAO.findDatasetWithDataUseByIdList(datasetIds);
+      Set<Dataset> datasets = new HashSet<>(datasetDAO.findDatasetsByIdList(datasetIds));
       Map<Integer, Dataset> datasetMap = datasets.stream()
           .collect(Collectors.toMap(Dataset::getDatasetId, Function.identity()));
```
pshapiro4broad: The only use of `datasets` is to call `stream()`. If you want to remove duplicates from a `Stream` you can use `distinct()`.
```
      Set<Dataset> datasets = datasetDAO.findDatasetWithDataUseByIdList(datasetIds);
      Map<Integer, Dataset> datasetMap = datasets.stream()
          .distinct()
          .collect(Collectors.toMap(Dataset::getDatasetId, Function.identity()));
```

src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetMapper.java
rushtong: > I think the pattern of if (something(... "study_id")) { .. get("study_id") } is error prone. Maybe in a future tech debt cleanup we could improve this. I think there are two general approaches

I'm all for a clean up task in this area - it is cumbersome boilerplate.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rushtong: Ah - good point, I was trying to keep the diff as small as possible here, but we really didn't need the `Set`. I can clean this up some more.

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
```diff
@@ -272,45 +210,44 @@ LEFT JOIN file_storage_object fso ON (fso.entity_id = d.dataset_id::text OR fso.
...
   List<Dataset> findDatasetsAssociatedWithDac(@Bind("dacId") Integer dacId);
 
+  @SqlQuery("""
+          SELECT DISTINCT dp.property_value as name
+          FROM dataset_property dp
+          INNER JOIN dataset d ON dp.dataset_id = d.dataset_id
+          WHERE (dp.schema_property = 'studyName')
+          UNION DISTINCT
+          SELECT DISTINCT s.name as name
+          FROM study s
```
pshapiro4broad, line 235: Why `SELECT DISTINCT` if you use `UNION DISTINCT`? I'm guessing the query optimizer will make it work the same either way but it reads as redundant to me. Also I thought that `DISTINCT` was the default behavior of `UNION` and didn't need to be specified. But maybe that was specific to a SQL dialect?

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
```diff
@@ -272,45 +210,44 @@ LEFT JOIN file_storage_object fso ON (fso.entity_id = d.dataset_id::text OR fso.
...
+      """)
+  Set<String> findAllStudyNames();
+
+  @SqlQuery("""
+          SELECT DISTINCT d.name
```
pshapiro4broad, line 240: I thought `dataset.name` has a `UNIQUE` constraint, if so isn't `SELECT DISTINCT d.name` redundant?
```suggestion
          SELECT d.name
```

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
rushtong: This isn't new code, it's just re-arranged by the diff. I'm trying not to make changes where I don't need to.

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
rushtong: This is true, we don't need it, but this is also not new code, just an artifact of the diff, and I don't want to make changes where not necessary

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
rushtong: https://broadworkbench.atlassian.net/browse/DT-1332

src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
rushtong: https://broadworkbench.atlassian.net/browse/DT-1333

src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetMapper.java
rushtong: https://broadworkbench.atlassian.net/browse/DT-1334

PR #2466, opened 2025-03-12
[DT-1315]Incorrect ticket_form_id being recorded in zendesk

rjohanek:
### Addresses
The Form field on the Zendesk ticket was being incorrectly populated as Terra instead of DUOS. 

### Summary
The Zendesk Ticket class uses JsonProperty annotations to specify how the field names should be serialized into json. However, the toString() method on the DuosTicket (which has one field, a Zendesk Ticket) was using Gson to stringify the object. Gson and JsonProperty annotations are not compatible. We cannot edit the Zendesk ticket class to use SerializedName annotations which are compatible with Gson. So, instead we must change the toString() implementation to use Jackson which is compatible with the current JsonProperty annotations.

Jason updated the Zendesk process to correctly populate the Brand field based on the Form field. So, a follow on ticket for that work is not needed.

Tested by hitting the support APIs locally and looking at the form id field on the resulting zendesk ticket.

no security risk

----

- Verify all tests go green


fboulnois (approved):
looks good 👍

src/main/java/org/broadinstitute/consent/http/models/support/DuosTicket.java
```diff
@@ -19,26 +19,20 @@
...
 
   public DuosTicket(Ticket request) {
     this.request = request;
   }
 
   public String toString() {
-    return GsonUtil.gsonBuilderWithAdapters().addSerializationExclusionStrategy(strategy).create()
-        .toJson(this);
+    ObjectMapper mapper = new ObjectMapper();
+    mapper.setSerializationInclusion(Include.NON_NULL);
```
pshapiro4broad, line 32: Can you fix this problem without using Jackson? The rest of consent uses `Gson` and I think it would be better to only use a single json library in the code base if possible.

pshapiro4broad: If there's no way to fix it using Gson, add a comment explaining way an exception is being made here

pshapiro4broad (approved):

src/main/java/org/broadinstitute/consent/http/models/support/DuosTicket.java
rjohanek: Okay good idea, I will add a comment explaining why we cannot use Gson!

rushtong: I might look into this again, thank you for finding the root problem! Jackson has been problematic for us performance-wise, but in this case, I think that risk is super low.

PR #2468, opened 2025-03-18
DT-1418: Stop reporting study 404s to Sentry

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1418

### Summary
Minor update to stop sending study not found errors to Sentry.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
@@ -275,7 +275,6 @@ public class DatasetService implements ConsentLogger {
       logException(e);
       throw e;
     }
-
   }
 
   public List<ApprovedDataset> getApprovedDatasets(User user) {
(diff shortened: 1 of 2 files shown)


fboulnois (approved):
👍

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -269,11 +269,12 @@ public Study getStudyWithDatasetsById(Integer studyId) {
         study.addDatasets(datasets);
       }
       return study;
+    } catch (NotFoundException nfe) {
+      throw nfe;
```
rjohanek, line 273: another way you could do this is by including a conditional in the generic catch and only log if it isn't a 404 that might be a little clearer, but not a blocker

rjohanek (approved):

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
rushtong: I did try an `instanceof` approach but that triggered a sonarqube warning, so went back to this approach.

PR #2469, opened 2025-03-20
DT-1424: Add new indexed_date field to dataset

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1424

### Summary
This PR adds a new `indexed_date` field to `Dataset`. This will be used to indicate if a dataset has been indexed in ElasticSearch (currently). A future PR will make further use of the service method to ensure the updates are recorded when we both index and un-index a dataset.

Features:
* Liquibase addition
* Model updates
  * Note that this includes some drive-by fixes to `Dataset.yaml` - there were several missing fields and one incorrectly named field.
* New audit actions so we can track the history of un/indexing a dataset
* Query and mapper updates to ensure we're populating this field across all usages

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/enumeration/AuditActions.java
+++ b/src/main/java/org/broadinstitute/consent/http/enumeration/AuditActions.java
@@ -6,7 +6,9 @@ public enum AuditActions {
   DELETE("delete"),
   REMOVE("remove"),
   REPLACE("replace"),
-  UPDATE("update");
+  UPDATE("update"),
+  INDEXED("indexed"),
+  UN_INDEXED("un_indexed");
 
   private final String value;
 
--- a/src/main/java/org/broadinstitute/consent/http/enumeration/AuditActions.java
+++ b/src/main/java/org/broadinstitute/consent/http/enumeration/AuditActions.java
@@ -8,7 +8,7 @@ public enum AuditActions {
   REPLACE("replace"),
   UPDATE("update"),
   INDEXED("indexed"),
-  UN_INDEXED("un_indexed");
+  DEINDEXED("deindexed");
 
   private final String value;
 
--- a/src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
@@ -218,7 +218,7 @@ public class DatasetServiceDAO implements ConsentLogger {
     jdbi.useHandle(handle -> {
       handle.getConnection().setAutoCommit(false);
       Dataset dataset = datasetDAO.findDatasetById(datasetId);
-      AuditActions action = indexDate == null ? AuditActions.UN_INDEXED : AuditActions.INDEXED;
+      AuditActions action = indexDate == null ? AuditActions.DEINDEXED : AuditActions.INDEXED;
       String dsAuditName = dataset.getName() == null ? dataset.getDatasetIdentifier() : dataset.getName();
       try {
         datasetDAO.updateDatasetIndexedDate(dataset.getDatasetId(), indexDate);
--- a/src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
@@ -206,6 +206,32 @@ public class DatasetServiceDAO implements ConsentLogger {
     return datasetId;
   }
 
+  /**
+   * Updates the dataset index date and adds an audit record.
+   *
+   * @param datasetId The ID of the dataset to update.
+   * @param userId    The ID of the user performing the update.
+   * @param indexDate The new index date, or null to un-index the dataset.
+   * @throws SQLException if DB transaction fails.
+   */
+  public void updateDatasetIndex(Integer datasetId, Integer userId, Instant indexDate) throws SQLException {
+    jdbi.useHandle(handle -> {
+      handle.getConnection().setAutoCommit(false);
+      Dataset dataset = datasetDAO.findDatasetById(datasetId);
+      AuditActions action = indexDate == null ? AuditActions.UN_INDEXED : AuditActions.INDEXED;
+      String dsAuditName = dataset.getName() == null ? dataset.getDatasetIdentifier() : dataset.getName();
+      try {
+        datasetDAO.updateDatasetIndexedDate(dataset.getDatasetId(), indexDate);
+        addAuditRecord(dataset.getDatasetId(), dsAuditName, userId, action);
+      } catch (Exception e) {
+        handle.rollback();
+        logException(e);
+        throw e;
+      }
+      handle.commit();
+    });
+  }
+
   private Integer executeInsertStudy(Handle handle, StudyInsert insert) {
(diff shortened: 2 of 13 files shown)


src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
```diff
@@ -206,6 +206,32 @@ public Integer executeInsertDatasetWithFiles(Handle handle,
...
+  /**
+   * Updates the dataset index date and adds an audit record.
+   *
+   * @param datasetId The ID of the dataset to update.
+   * @param userId    The ID of the user performing the update.
+   * @param indexDate The new index date, or null to un-index the dataset.
+   * @throws SQLException if DB transaction fails.
+   */
+  public void updateDatasetIndex(Integer datasetId, Integer userId, Instant indexDate) throws SQLException {
+    jdbi.useHandle(handle -> {
```
rjohanek, line 218: just like a general question, there's the DatasetService and the DatasetDAO, but there's also this level that I forgot about, the DatasetServiceDAO. Can you explain this architecture/what is the purpose of this layer?  Why can't the work here be distributed between the Service and the DAO layer?

src/main/resources/assets/schemas/Dataset.yaml
```diff
@@ -69,7 +73,17 @@ properties:
   dacApproval:
     type: boolean
     description: DAC approval/disapproval for this dataset
-  nihCertificationFile:
+  nihInstitutionalCertificationFile:
```
rjohanek, line 76: there's a couple changes that look like drive by changes to bring the api schema in alignment with the Dataset model, is that right? Are these bugs/is the data currently ignored because the name of the field is incorrect? Same with the field removed below, I see it is not in the model.

src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
rushtong: Great question. Our `*DAO` classes are actually interfaces and are very good with singular transactions. The vast majority of our db code works well in this case.

Our `*ServiceDAO` classes are intended to handle complex cases where there are multiple db actions that might require rollbacks if previous db actions fail. We chose this architecture to distinguish between the two. 

We have a package-info doc that covers this: https://github.com/DataBiosphere/consent/blob/develop/src/main/java/org/broadinstitute/consent/http/service/dao/package-info.java

rjohanek (approved):
a couple questions for my own understanding, but looks good! Looking forward to seeing the second part of this work!

src/main/resources/assets/schemas/Dataset.yaml
rushtong: > there's a couple changes that look like drive by changes to bring the api schema in alignment with the Dataset model, is that right? 

Correct

> Are these bugs/is the data currently ignored because the name of the field is incorrect? Same with the field removed below, I see it is not in the model.

There are no known bugs due to these inconsistencies - this is purely documentation since there are no known tools that look at this for client code generation.

src/main/java/org/broadinstitute/consent/http/enumeration/AuditActions.java
```diff
@@ -6,7 +6,9 @@ public enum AuditActions {
   DELETE("delete"),
   REMOVE("remove"),
   REPLACE("replace"),
-  UPDATE("update");
+  UPDATE("update"),
+  INDEXED("indexed"),
+  UN_INDEXED("un_indexed");
```
fboulnois:
```suggestion
  DEINDEXED("deindexed");
```

src/main/resources/changelog-master.xml
```diff
@@ -166,4 +166,6 @@
     relativeToChangelogFile="true"/>
   <include file="changesets/changelog-consent-2024-09-17-draft.xml"
     relativeToChangelogFile="true"/>
+  <include file="changesets/changelog-consent-2025-05-19-indexed.xml"
```
fboulnois:
```suggestion
  <include file="changesets/changelog-consent-2025-03-19-indexed.xml"
```

src/main/resources/changesets/changelog-consent-2025-05-19-indexed.xml
fboulnois, line 1: This should be renamed to `2025-03-19`

src/main/resources/changesets/changelog-consent-2025-05-19-indexed.xml
```diff
@@ -0,0 +1,9 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+  <changeSet id="changelog-consent-2025-05-19-indexed.xml" author="grushton">
```
fboulnois:
```suggestion
  <changeSet id="changelog-consent-2025-03-19-indexed.xml" author="grushton">
```

src/test/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAOTest.java
```diff
@@ -961,6 +962,38 @@ void testPatchDatasetWithBlankName() throws Exception {
...
+    Dataset updatedDataset = datasetDAO.findDatasetById(dataset.getDatasetId());
+
+    // Validate that the indexed date is null
+    assertNull(updatedDataset.getIndexedDate());
+
+    // Validate that a UN_INDEXED audit record was added:
+    List<DatasetAudit> audits = datasetDAO.findAuditsByDatasetId(updatedDataset.getDatasetId());
+    assertFalse(audits.isEmpty());
+    assertTrue(
+        audits.stream().anyMatch(a -> a.getAction().equalsIgnoreCase(AuditActions.UN_INDEXED.name())));
```
fboulnois:
```suggestion
        audits.stream().anyMatch(a -> a.getAction().equalsIgnoreCase(AuditActions.DEINDEXED.name())));
```

src/test/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAOTest.java
```diff
@@ -961,6 +962,38 @@ void testPatchDatasetWithBlankName() throws Exception {
...
+  @Test
+  void testUpdateDatasetIndexWithNull() throws Exception {
+    Dataset dataset = createDataset();
+    serviceDAO.updateDatasetIndex(dataset.getDatasetId(), dataset.getCreateUserId(), null);
+    Dataset updatedDataset = datasetDAO.findDatasetById(dataset.getDatasetId());
+
+    // Validate that the indexed date is null
+    assertNull(updatedDataset.getIndexedDate());
+
+    // Validate that a UN_INDEXED audit record was added:
```
fboulnois:
```suggestion
    // Validate that a DEINDEXED audit record was added:
```

fboulnois (approved):
A few suggestions:

src/main/resources/changelog-master.xml
rushtong: Ugh ... thank you!

PR #2470, opened 2025-03-20
DT-1424: Mark datasets as indexed/deindexed when indexing operations are called

rushtong:
### Addresses
Final part of https://broadworkbench.atlassian.net/browse/DT-1424

### Summary
* Move the service call to `ElasticSearchService` since that's where we needed it.
* Significant method signature changes to pass the updating `User` down to the service calls
* Significant test updates to move away from older/deprecated patterns
* Handle several new cases where we should have been re-indexing, but were not:
  * Sync Data Use Translation
  * Convert Dataset to Study
  * Update Study Custodians
  * Patch Dataset via patchable properties

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/Study.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Study.java
@@ -78,8 +78,8 @@ public class Study {
     return properties;
   }
 
-  public void addProperties(Set<StudyProperty> properties) {
-    this.properties.addAll(properties);
+  public void addProperties(StudyProperty... properties) {
+    this.properties.addAll(Set.of(properties));
   }
 
   public void addProperty(StudyProperty prop) {
--- a/src/main/java/org/broadinstitute/consent/http/models/Study.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Study.java
@@ -3,7 +3,6 @@ package org.broadinstitute.consent.http.models;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
 
@@ -15,9 +14,9 @@ public class Study {
   private Boolean publicVisibility;
   private String piName;
   private List<String> dataTypes;
-  private Set<Integer> datasetIds = new HashSet<>();
+  private final Set<Integer> datasetIds = new HashSet<>();
   private Set<Dataset> datasets;
-  private Set<StudyProperty> properties = new HashSet<>();
+  private final Set<StudyProperty> properties = new HashSet<>();
   private FileStorageObject alternativeDataSharingPlan;
   private Date createDate;
   private String createUserEmail;
@@ -80,13 +79,10 @@ public class Study {
   }
 
   public void setProperties(Set<StudyProperty> properties) {
-    this.properties = properties;
+    this.properties.addAll(properties);
   }
 
   public void addProperty(StudyProperty prop) {
-    if (Objects.isNull(this.properties)) {
-      this.properties = new HashSet<>();
-    }
     this.properties.add(prop);
   }
 
@@ -103,14 +99,10 @@ public class Study {
   }
 
   public void setDatasetIds(Set<Integer> datasetIds) {
-    this.datasetIds = datasetIds;
+    this.datasetIds.addAll(datasetIds);
   }
 
   public void addDatasetId(Integer datasetId) {
-    if (Objects.isNull(this.datasetIds)) {
-      this.datasetIds = new HashSet<>();
-    }
-
     this.datasetIds.add(datasetId);
   }
 
--- a/src/main/java/org/broadinstitute/consent/http/models/Study.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Study.java
@@ -78,7 +78,7 @@ public class Study {
     return properties;
   }
 
-  public void setProperties(Set<StudyProperty> properties) {
+  public void addProperties(Set<StudyProperty> properties) {
     this.properties.addAll(properties);
   }
 
@@ -98,7 +98,7 @@ public class Study {
     return datasetIds;
   }
 
-  public void setDatasetIds(Set<Integer> datasetIds) {
+  public void addDatasetIds(Set<Integer> datasetIds) {
     this.datasetIds.addAll(datasetIds);
   }
 
--- a/src/main/java/org/broadinstitute/consent/http/models/Study.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Study.java
@@ -17,7 +17,7 @@ public class Study {
   private List<String> dataTypes;
   private Set<Integer> datasetIds = new HashSet<>();
   private Set<Dataset> datasets;
-  private Set<StudyProperty> properties;
+  private Set<StudyProperty> properties = new HashSet<>();
   private FileStorageObject alternativeDataSharingPlan;
   private Date createDate;
   private String createUserEmail;
--- a/src/main/java/org/broadinstitute/consent/http/models/Study.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/Study.java
@@ -15,7 +15,7 @@ public class Study {
   private Boolean publicVisibility;
   private String piName;
   private List<String> dataTypes;
-  private Set<Integer> datasetIds;
+  private Set<Integer> datasetIds = new HashSet<>();
   private Set<Dataset> datasets;
   private Set<StudyProperty> properties;
   private FileStorageObject alternativeDataSharingPlan;
--- a/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
@@ -1,6 +1,7 @@
 package org.broadinstitute.consent.http.resources;
 
 import com.codahale.metrics.annotation.Timed;
+import com.google.api.client.http.HttpStatusCodes;
 import com.google.gson.Gson;
 import com.google.gson.JsonSyntaxException;
 import com.google.inject.Inject;
@@ -27,10 +28,8 @@ import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.StreamingOutput;
 import jakarta.ws.rs.core.UriBuilder;
 import jakarta.ws.rs.core.UriInfo;
-import java.io.IOException;
 import java.net.URI;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -378,11 +377,10 @@ public class DatasetResource extends Resource {
         logException(e);
         return createExceptionResponse(e);
       }
-      try {
-        elasticSearchService.deleteIndex(datasetId, user.getUserId());
-      } catch (IOException e) {
-        logException(e);
-        return createExceptionResponse(e);
+      try (var deleteResponse = elasticSearchService.deleteIndex(datasetId, user.getUserId())) {
+        if (!HttpStatusCodes.isSuccess(deleteResponse.getStatus())) {
+          logWarn("Unable to delete index for dataset: " + datasetId);
+        }
       }
       return Response.ok().build();
     } catch (Exception e) {
--- a/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
@@ -379,7 +379,7 @@ public class DatasetResource extends Resource {
         return createExceptionResponse(e);
       }
       try {
-        elasticSearchService.deleteIndex(datasetId);
+        elasticSearchService.deleteIndex(datasetId, user);
       } catch (IOException e) {
         logException(e);
         return createExceptionResponse(e);
@@ -393,10 +393,11 @@ public class DatasetResource extends Resource {
   @POST
   @Path("/index")
   @RolesAllowed(ADMIN)
-  public Response indexDatasets() {
+  public Response indexDatasets(@Auth AuthUser authUser) {
     try {
+      User user = userService.findUserByEmail(authUser.getEmail());
       var datasetIds = datasetService.findAllDatasetIds();
-      StreamingOutput indexResponse = elasticSearchService.indexDatasetIds(datasetIds);
+      StreamingOutput indexResponse = elasticSearchService.indexDatasetIds(datasetIds, user);
       return Response.ok(indexResponse, MediaType.APPLICATION_JSON).build();
     } catch (Exception e) {
       return createExceptionResponse(e);
@@ -406,10 +407,11 @@ public class DatasetResource extends Resource {
   @POST
   @Path("/index/{datasetId}")
   @RolesAllowed(ADMIN)
-  public Response indexDataset(@PathParam("datasetId") Integer datasetId) {
+  public Response indexDataset(@Auth AuthUser authUser, @PathParam("datasetId") Integer datasetId) {
     try {
+      User user = userService.findUserByEmail(authUser.getEmail());
       var dataset = datasetService.findDatasetById(datasetId);
-      return elasticSearchService.indexDataset(dataset);
+      return elasticSearchService.indexDataset(dataset, user);
     } catch (Exception e) {
       return createExceptionResponse(e);
     }
@@ -418,9 +420,10 @@ public class DatasetResource extends Resource {
   @DELETE
   @Path("/index/{datasetId}")
   @RolesAllowed(ADMIN)
-  public Response deleteDatasetIndex(@PathParam("datasetId") Integer datasetId) {
+  public Response deleteDatasetIndex(@Auth AuthUser authUser, @PathParam("datasetId") Integer datasetId) {
     try {
-      return elasticSearchService.deleteIndex(datasetId);
+      User user = userService.findUserByEmail(authUser.getEmail());
+      return elasticSearchService.deleteIndex(datasetId, user);
     } catch (Exception e) {
       return createExceptionResponse(e);
     }
--- a/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
@@ -379,7 +379,7 @@ public class DatasetResource extends Resource {
         return createExceptionResponse(e);
       }
       try {
-        elasticSearchService.deleteIndex(datasetId, user);
+        elasticSearchService.deleteIndex(datasetId, user.getUserId());
       } catch (IOException e) {
         logException(e);
         return createExceptionResponse(e);
@@ -423,7 +423,7 @@ public class DatasetResource extends Resource {
   public Response deleteDatasetIndex(@Auth AuthUser authUser, @PathParam("datasetId") Integer datasetId) {
     try {
       User user = userService.findUserByEmail(authUser.getEmail());
-      return elasticSearchService.deleteIndex(datasetId, user);
+      return elasticSearchService.deleteIndex(datasetId, user.getUserId());
     } catch (Exception e) {
       return createExceptionResponse(e);
     }
--- a/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
@@ -209,14 +209,7 @@ public class DatasetResource extends Resource {
         throw new BadRequestException("Properties are invalid");
       }
       Dataset patched = datasetRegistrationService.patchDataset(datasetId, user, patch);
-      if (patched != null && patched.getIndexedDate() != null) {
-        // If the dataset was indexed, we need to re-index it
-        try (var indexResponse = elasticSearchService.indexDataset(patched, user)) {
-          if (!HttpStatusCodes.isSuccess(indexResponse.getStatus())) {
-            logWarn("Unable to index patched dataset: " + patched.getDatasetId());
-          }
-        }
-      }
+      elasticSearchService.synchronizeDatasetInESIndex(patched, user, false);
       return Response.ok(patched).build();
     } catch (Exception e) {
       return createExceptionResponse(e);
--- a/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
@@ -209,6 +209,14 @@ public class DatasetResource extends Resource {
         throw new BadRequestException("Properties are invalid");
       }
       Dataset patched = datasetRegistrationService.patchDataset(datasetId, user, patch);
+      if (patched != null && patched.getIndexedDate() != null) {
+        // If the dataset was indexed, we need to re-index it
+        try (var indexResponse = elasticSearchService.indexDataset(patched, user)) {
+          if (!HttpStatusCodes.isSuccess(indexResponse.getStatus())) {
+            logWarn("Unable to index patched dataset: " + patched.getDatasetId());
+          }
+        }
+      }
       return Response.ok(patched).build();
     } catch (Exception e) {
       return createExceptionResponse(e);
--- a/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
@@ -159,6 +159,9 @@ public class DatasetResource extends Resource {
 
       Dataset updatedDataset = datasetRegistrationService.updateDataset(datasetId, user, update,
           files);
+      if (updatedDataset.getIndexedDate() != null) {
+        elasticSearchService.indexDataset(updatedDataset, user);
+      }
       return Response.ok().entity(updatedDataset).build();
     } catch (Exception e) {
       return createExceptionResponse(e);
@@ -479,6 +482,10 @@ public class DatasetResource extends Resource {
         return Response.notModified().entity(originalDataset).build();
       }
       Dataset dataset = datasetService.updateDatasetDataUse(user, id, dataUse);
+      // Re-index if the dataset is already indexed
+      if (dataset.getIndexedDate() != null) {
+        elasticSearchService.indexDataset(dataset, user);
+      }
       return Response.ok().entity(dataset).build();
     } catch (JsonSyntaxException jse) {
       return createExceptionResponse(
@@ -494,7 +501,8 @@ public class DatasetResource extends Resource {
   @Path("/{id}/reprocess/datause")
   public Response syncDataUseTranslation(@Auth AuthUser authUser, @PathParam("id") Integer id) {
     try {
-      Dataset ds = datasetService.syncDatasetDataUseTranslation(id);
+      User user = userService.findUserByEmail(authUser.getEmail());
+      Dataset ds = datasetService.syncDatasetDataUseTranslation(id, user);
       return Response.ok(ds).build();
     } catch (Exception e) {
       return createExceptionResponse(e);
--- a/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
@@ -159,9 +159,6 @@ public class DatasetResource extends Resource {
 
       Dataset updatedDataset = datasetRegistrationService.updateDataset(datasetId, user, update,
           files);
-      if (updatedDataset.getIndexedDate() != null) {
-        elasticSearchService.indexDataset(updatedDataset, user);
-      }
       return Response.ok().entity(updatedDataset).build();
     } catch (Exception e) {
       return createExceptionResponse(e);
@@ -482,10 +479,6 @@ public class DatasetResource extends Resource {
         return Response.notModified().entity(originalDataset).build();
       }
       Dataset dataset = datasetService.updateDatasetDataUse(user, id, dataUse);
-      // Re-index if the dataset is already indexed
-      if (dataset.getIndexedDate() != null) {
-        elasticSearchService.indexDataset(dataset, user);
-      }
       return Response.ok().entity(dataset).build();
     } catch (JsonSyntaxException jse) {
       return createExceptionResponse(
--- a/src/main/java/org/broadinstitute/consent/http/resources/StudyResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/StudyResource.java
@@ -20,7 +20,6 @@ import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.Response.Status;
 import java.io.IOException;
 import java.lang.reflect.Type;
-import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
--- a/src/main/java/org/broadinstitute/consent/http/resources/StudyResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/StudyResource.java
@@ -159,7 +159,7 @@ public class StudyResource extends Resource {
       // Remove from ES index
       if (studyDatasetIds != null) {
         studyDatasetIds.forEach(id -> {
-          try (Response indexResponse = elasticSearchService.deleteIndex(id, user)) {
+          try (Response indexResponse = elasticSearchService.deleteIndex(id, user.getUserId())) {
             if (indexResponse.getStatus() >= Status.BAD_REQUEST.getStatusCode()) {
               logWarn("Non-OK response when deleting index for dataset with id: " + id);
             }
--- a/src/main/java/org/broadinstitute/consent/http/resources/StudyResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/StudyResource.java
@@ -20,8 +20,8 @@ import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.Response.Status;
 import java.io.IOException;
 import java.lang.reflect.Type;
+import java.sql.SQLException;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -160,8 +160,8 @@ public class StudyResource extends Resource {
       if (studyDatasetIds != null) {
         studyDatasetIds.forEach(id -> {
           try {
-            elasticSearchService.deleteIndex(id);
-          } catch (IOException e) {
+            elasticSearchService.deleteIndex(id, user);
+          } catch (IOException | SQLException e) {
             logException(e);
           }
         });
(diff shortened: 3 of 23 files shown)


src/test/java/org/broadinstitute/consent/http/service/DatasetServiceTest.java
```diff
@@ -519,12 +518,13 @@ void testEnforceDAARestrictions() {
   }
 
   @Test
-  void testUpdateDatasetIndex() throws SQLException {
+  void testUpdateIfIndexed() throws SQLException {
+    Dataset dataset = new Dataset();
+    dataset.setIndexedDate(new Date());
     doNothing().when(datasetServiceDAO).updateDatasetIndex(any(), any(), any());
```
pshapiro4broad: This is the default behavior for a mock, so it should be unnecessary. If you want to verify that a mock is used, I would use `verify()` rather than relying on strict stubbing (if that's what you're doing here).
```
    verify(datasetServiceDAO, times(2)).updateDatasetIndex(any(), any(), any());
```

src/test/java/org/broadinstitute/consent/http/service/DatasetServiceTest.java
```diff
@@ -519,12 +518,13 @@ void testEnforceDAARestrictions() {
...
-    when(datasetDAO.findDatasetById(any())).thenReturn(new Dataset());
     initService();
-    assertDoesNotThrow(() -> datasetService.updateDatasetIndex(1, 1, Instant.now()));
-    assertDoesNotThrow(() -> datasetService.updateDatasetIndex(1, 1, null));
+    assertDoesNotThrow(() -> datasetService.updateIfIndexed(dataset, 1, Instant.now()));
+    assertDoesNotThrow(() -> datasetService.updateIfIndexed(dataset, 1, null));
```
pshapiro4broad: Why do you `assertDoesNotThrow` here? If anything is thrown, the test will fail. It's nice to use it in cases where nothing else can be asserted, but if you use `verify()` then that should be sufficient.

Also you may want to verify the negative case here too. I think this would work for both
```
  void testUpdateIfIndexed() throws SQLException {
    Dataset dataset = new Dataset();
    dataset.setDatasetId(2);
    dataset.setIndexedDate(new Date());
    initService();
    datasetService.updateIfIndexed(dataset, 1, Instant.now());
    datasetService.updateIfIndexed(dataset, 1, null);
    verify(datasetServiceDAO, times(2)).updateDatasetIndex(eq(2), eq(1), any());

    reset(datasetServiceDAO);
    dataset.setIndexedDate(null);
    datasetService.updateIfIndexed(dataset, 1, null);
    verifyNoInteractions(datasetServiceDAO);
  }
```

src/test/java/org/broadinstitute/consent/http/resources/StudyResourceTest.java
```diff
@@ -222,11 +225,12 @@ void testDeleteStudyById() throws Exception {
     admin.setAdminRole();
     admin.setUserId(study.getCreateUserId());
     when(userService.findUserByEmail(any())).thenReturn(admin);
+    when(elasticSearchService.deleteIndex(any(), any())).thenReturn(response);
     initResource();
 
     try (Response response = resource.deleteStudyById(authUser, study.getStudyId())) {
       assertEquals(HttpStatusCodes.STATUS_CODE_OK, response.getStatus());
-      verify(elasticSearchService, atLeastOnce()).deleteIndex(any());
+      verify(elasticSearchService, atLeastOnce()).deleteIndex(any(), any());
```
pshapiro4broad: Why do you use `atLeastOnce()` here? Do we expect it to be called more than once?
```suggestion
      verify(elasticSearchService).deleteIndex(any(), any());
```

Also in this case since you use strict stubbing this `verify()` is redundant.

src/test/java/org/broadinstitute/consent/http/resources/StudyResourceTest.java
```diff
@@ -222,11 +225,12 @@ void testDeleteStudyById() throws Exception {
     admin.setAdminRole();
     admin.setUserId(study.getCreateUserId());
     when(userService.findUserByEmail(any())).thenReturn(admin);
+    when(elasticSearchService.deleteIndex(any(), any())).thenReturn(response);
```
pshapiro4broad, line 245: In general I think it's best to avoid `any()` when writing tests, if possible. In this case the test can be written without using `any()`:
```
  void testDeleteStudyById() throws Exception {
    Study study = createMockStudy();
    study.getDatasets().forEach(d -> d.setDeletable(true));
    when(datasetService.getStudyWithDatasetsById(study.getStudyId())).thenReturn(study);
    User admin = new User();
    admin.setAdminRole();
    admin.setUserId(study.getCreateUserId());
    when(userService.findUserByEmail(authUser.getEmail())).thenReturn(admin);
    when(elasticSearchService.deleteIndex(100, admin)).thenReturn(response);
    initResource();

    try (Response response = resource.deleteStudyById(authUser, study.getStudyId())) {
    }
  }
```

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -526,10 +534,14 @@ public void setDatasetBatchSize(Integer datasetBatchSize) {
     this.datasetBatchSize = datasetBatchSize;
   }
 
-  public Dataset updateDatasetIndex(Integer datasetId, Integer userId, Instant indexDate)
-      throws SQLException {
-    datasetServiceDAO.updateDatasetIndex(datasetId, userId, indexDate);
-    return datasetDAO.findDatasetById(datasetId);
+  protected void updateIfIndexed(Dataset dataset, Integer userId, Instant indexDate) {
```
pshapiro4broad: Every caller to this method passes `Instant.now()`. Instead you can omit the argument and use `Instant.now()` directly in the code. It looks like all other caller to `datasetServiceDAO.updateDatasetIndex()` also pass in `Instant.now()` except for the case where it's set to `null`. You could split the DAO layer into two methods, update and delete and get rid of the argument completely.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -243,30 +256,30 @@ public InstitutionTerm toInstitutionTerm(Institution institution) {
     return new InstitutionTerm(institution.getId(), institution.getName());
   }
 
-  public Response indexDataset(Dataset dataset) throws IOException {
-    return indexDatasetTerms(List.of(toDatasetTerm(dataset)));
+  public Response indexDataset(Dataset dataset, User user) throws IOException {
+    return indexDatasetTerms(List.of(toDatasetTerm(dataset)), user);
```
pshapiro4broad: You could remove the call to `toDatasetTerm()` if you delegated to the overloaded method
```suggestion
    return indexDatasets(List.of(dataset), user);
```

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
pshapiro4broad: Furthermore, I don't see why the date is stored at all. The only use is to compare it against `null`, so it's being treated like a boolean flag. Is there a future use case where the index date would be useful?

Although, maybe this should be part of a more general approach to indexing.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -418,9 +420,10 @@ public Response indexDataset(@PathParam("datasetId") Integer datasetId) {
   @DELETE
   @Path("/index/{datasetId}")
   @RolesAllowed(ADMIN)
-  public Response deleteDatasetIndex(@PathParam("datasetId") Integer datasetId) {
+  public Response deleteDatasetIndex(@Auth AuthUser authUser, @PathParam("datasetId") Integer datasetId) {
```
pshapiro4broad, line 422: What is the use case of this API? Since the code is using ad-hoc operations to index datasets, I don't see how this will prevent the index from being created on another operation.

src/main/java/org/broadinstitute/consent/http/resources/StudyResource.java
```diff
@@ -159,9 +159,11 @@ public Response deleteStudyById(@Auth AuthUser authUser, @PathParam("studyId") I
       // Remove from ES index
       if (studyDatasetIds != null) {
         studyDatasetIds.forEach(id -> {
-          try {
-            elasticSearchService.deleteIndex(id);
-          } catch (IOException e) {
+          try (Response indexResponse = elasticSearchService.deleteIndex(id, user)) {
```
pshapiro4broad: It looks like this will end up trying to update `dataset.indexed_date` on the dataset, but it was already deleted in `datasetService.deleteStudy()` above. I think there's only one case where the index needs to be set to `null`, when called from `DatasetResource.deleteDatasetIndex()`. The other two cases are on study or dataset deletion, in which case the dataset and study records will be deleted.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -380,6 +384,10 @@ public Study updateStudyCustodians(User user, Integer studyId, String custodians
     } else {
```
pshapiro4broad, line 395: It's not part of this PR, but the above code fragment
```
   Optional x = stream.filter(something).findFirst();
   if (x.isPresent) { ...
```
can be replaced with
```
  boolean b = stream.anyMatch(something);
  if (b) { ...
```
as long as `x` isn't needed, which is true in this case.

You can also replace the ternary op with `Objects.requireNonNullElse()` as long as you qualify the `Set.of()` type:
```
  Objects.requireNonNullElse(study.getProperties(), Set.<StudyProperty>of()).stream()
```

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -152,18 +153,20 @@ public Dataset syncDatasetDataUseTranslation(Integer datasetId) {
     String translation = ontologyService.translateDataUse(dataset.getDataUse(),
         DataUseTranslationType.DATASET);
     datasetDAO.updateDatasetTranslatedDataUse(datasetId, translation);
-
+    updateIfIndexed(dataset, user.getUserId(), Instant.now());
     return datasetDAO.findDatasetById(datasetId);
   }
 
   public void deleteDataset(Integer datasetId, Integer userId) throws Exception {
+    // TODO: Figure out how to remove from index
```
pshapiro4broad: I believe these are now addressed in this PR

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -526,10 +534,14 @@ public void setDatasetBatchSize(Integer datasetBatchSize) {
...
   }
 
-  public Dataset updateDatasetIndex(Integer datasetId, Integer userId, Instant indexDate)
-      throws SQLException {
-    datasetServiceDAO.updateDatasetIndex(datasetId, userId, indexDate);
-    return datasetDAO.findDatasetById(datasetId);
+  protected void updateIfIndexed(Dataset dataset, Integer userId, Instant indexDate) {
+    if (dataset.getIndexedDate() != null) {
+      try {
+        datasetServiceDAO.updateDatasetIndex(dataset.getDatasetId(), userId, indexDate);
```
pshapiro4broad: Is this fix incomplete? I would expect to see a call to elasticsearch to actually update the index (which would update the DB), rather than this updating the database record directly.

rushtong: This PR is still in draft - I'm still verifying all of the changes functionally, so this could very well be incomplete.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
rushtong: The use case here it to ensure that a specific dataset exists in the index. We have needed this as an admin when we wanted to re-index a dataset for some reason. This will NOT prevent the index from being updated in other operations, which is intentional - we explicitly DO need to update the index in other operations.

src/test/java/org/broadinstitute/consent/http/resources/StudyResourceTest.java
rushtong: On a conceptual level, I respectfully disagree with not using `any()`. If we took the rewrite as an example, then we're testing that the exact values dependent services are receiving are the same as the values in the mocks. I don't think this really adds any test value.
In this case, the resource method is testing the positive case of a deletion, so it _only_ cares that dependent service methods work successfully. Negative resource method tests will test the failure cases. Tests for the input and output of service methods are more appropriately tested in their respective service test rather than here in the resource test.

src/test/java/org/broadinstitute/consent/http/resources/StudyResourceTest.java
rushtong: I think `atLeastOnce()` is the wrong test here. We should be ensuring that we're only calling the delete once and only once.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
rushtong: >  Is there a future use case where the index date would be useful?

I think the potential use case is not necessarily for traditional user's to see, but for admins to see when it was last indexed separate from the most recent update date, which is something we also store, but is not modified by the indexing operations.

src/main/java/org/broadinstitute/consent/http/models/Study.java
```diff
@@ -15,9 +15,9 @@ public class Study {
   private Boolean publicVisibility;
   private String piName;
   private List<String> dataTypes;
-  private Set<Integer> datasetIds;
+  private Set<Integer> datasetIds = new HashSet<>();
   private Set<Dataset> datasets;
-  private Set<StudyProperty> properties;
+  private Set<StudyProperty> properties = new HashSet<>();
```
pshapiro4broad: I think this is a good opportunity to clean this up a bit more. I'd also make these fields final (so a caller can't overwrite them with a `null` value) and initialize `datasets` here and make it `final` too.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -210,6 +209,14 @@ public Response patchByDatasetUpdate(@Auth AuthUser authUser,
         throw new BadRequestException("Properties are invalid");
       }
       Dataset patched = datasetRegistrationService.patchDataset(datasetId, user, patch);
+      if (patched != null && patched.getIndexedDate() != null) {
```
pshapiro4broad: How can `patched` ever be `null`? If the DAO call fails, it will throw.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -210,6 +209,14 @@ public Response patchByDatasetUpdate(@Auth AuthUser authUser,
...
+        // If the dataset was indexed, we need to re-index it
+        try (var indexResponse = elasticSearchService.indexDataset(patched, user)) {
+          if (!HttpStatusCodes.isSuccess(indexResponse.getStatus())) {
```
pshapiro4broad: When this is called from `DatasetRegistrationService.updateDataset()`, the check is
```
        if (response.getStatus() >= 400) {
```

And in `DacResource.approveDataset()` the check is:
```
        if (indexResponse.getStatus() >= Status.BAD_REQUEST.getStatusCode()) {
```

I also see the same check done in `DatasetService.synchronizeDatasetInESIndex()`. Can all these cases be refactored into one method in `ElasticSearchService`?

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
pshapiro4broad: > ensure that a specific dataset exists in the index

Do you mean does not exist? This API deletes a dataset from the index.

My point before was that this wasn't setting last index date to null, so it wouldn't prevent it from being re-indexed. Now it does set it to null, so that works. But I'm still wondering why we'd want to delete a specific dataset from the index.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -393,10 +399,11 @@ public Response delete(@Auth AuthUser authUser, @PathParam("datasetId") Integer
   @POST
   @Path("/index")
   @RolesAllowed(ADMIN)
-  public Response indexDatasets() {
+  public Response indexDatasets(@Auth AuthUser authUser) {
     try {
+      User user = userService.findUserByEmail(authUser.getEmail());
```
pshapiro4broad, line 397: Something for a future PR; I think you can create an `Authenticator` to support injecting `User` directly. Conceptually it would inject `AuthUser` and then call `userService.findUserByEmail()` like this code does.

src/main/java/org/broadinstitute/consent/http/resources/StudyResource.java
```diff
@@ -159,9 +159,11 @@ public Response deleteStudyById(@Auth AuthUser authUser, @PathParam("studyId") I
       // Remove from ES index
       if (studyDatasetIds != null) {
         studyDatasetIds.forEach(id -> {
-          try {
-            elasticSearchService.deleteIndex(id);
-          } catch (IOException e) {
+          try (Response indexResponse = elasticSearchService.deleteIndex(id, user.getUserId())) {
```
pshapiro4broad: This looks like the same code that's in `DatasetService. deleteStudy()`

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -526,10 +546,14 @@ public void setDatasetBatchSize(Integer datasetBatchSize) {
     this.datasetBatchSize = datasetBatchSize;
   }
 
-  public Dataset updateDatasetIndex(Integer datasetId, Integer userId, Instant indexDate)
-      throws SQLException {
-    datasetServiceDAO.updateDatasetIndex(datasetId, userId, indexDate);
-    return datasetDAO.findDatasetById(datasetId);
+  private void synchronizeDatasetInESIndex(Dataset dataset, User user) {
```
pshapiro4broad: Doesn't this need to check `updatedDataset.getIndexedDate() != null`?

pshapiro4broad: If this was `public`, it could be used in other places where `indexDataset()` is called.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -363,23 +382,24 @@ public Study convertDatasetToStudy(User user, Dataset dataset, StudyConversion s
...
+    if (propPresent) {
+      studyDAO.updateStudyProperty(studyId, dataCustodianEmail, PropertyType.Json.toString(),
+          custodians);
     } else {
-      studyDAO.insertStudyProperty(studyId, dataCustodianEmail, PropertyType.Json.toString(), custodians);
+      studyDAO.insertStudyProperty(studyId, dataCustodianEmail, PropertyType.Json.toString(),
+          custodians);
     }
+    List<Dataset> datasets = study.getDatasetIds().isEmpty() ? List.of()
+        : datasetDAO.findDatasetsByIdList(study.getDatasetIds());
```
pshapiro4broad: Forcing this check into the caller will result in duplicate code. I believe JDBI will correctly handle empty lists, but if there's a concern, I would add a wrapper function to the DAO API so callers don't have to worry about this.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -95,12 +101,18 @@ private Response performRequest(Request request) throws IOException {
...
 
     datasets.forEach(dsTerm -> {
-      bulkApiCall.add(bulkHeader.formatted(dsTerm.getDatasetId()));
+      bulkApiCall.add(BULK_HEADER.formatted(dsTerm.getDatasetId()));
       bulkApiCall.add(GsonUtil.getInstance().toJson(dsTerm) + "\n");
+      try {
+        updateDatasetIndex(dsTerm.getDatasetId(), user.getUserId(), Instant.now());
+      } catch (SQLException e) {
+        // We don't want to send these to Sentry, but we do want to log them for follow up off cycle
+        logWarn("Error updating dataset indexed date for dataset id: %d ".formatted(dsTerm.getDatasetId()), e);
```
pshapiro4broad: Should this logging and try/catch be done in `updateDatasetIndex()`? There is one other call to this method where `null` is written, do we want that to also be logged?

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -383,6 +396,15 @@ public DatasetTerm toDatasetTerm(Dataset dataset) {
     return term;
   }
 
+  protected void updateDatasetIndex(Integer datasetId, Integer userId, Instant indexDate)
```
pshapiro4broad: The method name is confusing to me, since it doesn't update the index but updates the index date
```suggestion
  protected void updateDatasetIndexDate(Integer datasetId, Integer userId, Instant indexDate)
```

The same is true for `DatasetServiceDAO.updateDatasetIndex()`.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -314,12 +307,39 @@ void testPatchByDatasetUpdate_patchableNoName() {
...
+    when(authUser.getEmail()).thenReturn("test@test.com");
+    when(userService.findUserByEmail("test@test.com")).thenReturn(user);
+    when(user.getUserId()).thenReturn(randomInt(1, 100));
     dataset.setCreateUserId(user.getUserId());
+    when(elasticSearchService.indexDataset(dataset, user)).thenReturn(mockResponse);
 
     initResource();
     try (Response response = resource.patchByDatasetUpdate(authUser, dataset.getDatasetId(),
         gson.toJson(patch))) {
+      verify(elasticSearchService, times(1)).indexDataset(dataset, user);
```
pshapiro4broad: `times(1)` is the default so it can be omitted.

In this case the `verify()` call isn't necessary since you have a `when()` and strict stubbing is enabled. If you wanted to, you could remove the `when()` and leave the `verify()`, which would have the same effect.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -314,12 +307,39 @@ void testPatchByDatasetUpdate_patchableNoName() {
...
+    Dataset dataset = new Dataset();
+    dataset.setDatasetId(randomInt(1, 100));
+    dataset.setName(randomAlphabetic(10));
+    dataset.setIndexedDate(new Date());
+    when(datasetService.findDatasetById(any())).thenReturn(dataset);
+    String newName = randomAlphabetic(20);
+    DatasetPatch patch = new DatasetPatch(newName, List.of());
+
+    when(datasetRegistrationService.patchDataset(any(), any(), any())).thenReturn(dataset);
+    when(authUser.getEmail()).thenReturn("test@test.com");
```
pshapiro4broad: This could be avoided by not using a mock for `AuthUser`, since it's a POJO. You could create a test auth user that has its email set for all tests to use.
```
  private final AuthUser authUser = new AuthUser().setEmail("test@test.com");
```

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -366,50 +386,56 @@ void testValidateDatasetNameSuccess() {
     Dataset testDataset = new Dataset();
     when(datasetService.getDatasetByName("test")).thenReturn(testDataset);
     initResource();
-    Response response = resource.validateDatasetName("test");
-    assertEquals(HttpStatusCodes.STATUS_CODE_OK, response.getStatus());
+    try(var response = resource.validateDatasetName("test")) {
```
pshapiro4broad: whitespace
```suggestion
    try (var response = resource.validateDatasetName("test")) {
```

Is this change to satisfy a lint warning?

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -771,63 +811,69 @@ void testUpdateDatasetDataUse_NotModified() {
...
-    assertEquals(dataset.getDatasetId(), returnedDatasets.get(0).getDatasetId());
+    try (var response = resource.findAllDatasetsStreaming(authUser)) {
+      assertEquals(HttpStatusCodes.STATUS_CODE_OK, response.getStatus());
+      var entity = (StreamingOutput) response.getEntity();
+      var baos = new ByteArrayOutputStream();
+      entity.write(baos);
+      var entityString = baos.toString();
+      Type listOfDatasetsType = new TypeToken<List<Dataset>>() {
+      }.getType();
+      List<Dataset> returnedDatasets = gson.fromJson(entityString, listOfDatasetsType);
```
pshapiro4broad, line 823: The extra assignment isn't needed and using `TypeToken` as a parameter lets the compiler use type inference.
```suggestion
      List<Dataset> returnedDatasets = gson.fromJson(entityString, new TypeToken<>() {});
```

pshapiro4broad (commented):
Some comments so far

src/test/java/org/broadinstitute/consent/http/resources/StudyResourceTest.java
pshapiro4broad: Maybe the difference is in how we think about what a test is doing. My thinking is that the purpose of test for a method is to validate how it maps inputs to outputs (where "mapping" is whatever work the method applies to its inputs). In cases where a method calls another service, in addition to verifying that the correct service API is called, any parameters the method passes to that service are also outputs of the method. so they must be tested.

src/test/java/org/broadinstitute/consent/http/service/DatasetServiceTest.java
```diff
@@ -79,10 +80,14 @@ class DatasetServiceTest {
   private DatasetServiceDAO datasetServiceDAO;
   @Mock
   private UserDAO userDAO;
+  @Mock
+  private User user;
+  @Mock
+  private Response response;
 
   private void initService() {
```
pshapiro4broad: If this was a `BeforeEach` method you wouldn't need to call it from every test method
```suggestion
  @BeforeEach
  public void initService() {
```

src/main/java/org/broadinstitute/consent/http/models/Study.java
pshapiro4broad: With this change I think you can remove the code in `StudyReducer.reduceStudy()`:
```
    } else if (Objects.isNull(study.getProperties())) {
      study.setProperties(new HashSet<>());
    }
```

src/test/java/org/broadinstitute/consent/http/service/ElasticSearchServiceTest.java
```diff
@@ -632,16 +647,46 @@ void testIndexStudyWithDatasets() {
...
+    assertDoesNotThrow(() -> {
+      try (var response = service.indexStudy(1, user)) {
+        assertEquals(HttpStatusCodes.STATUS_CODE_NOT_FOUND, response.getStatus());
+      }
+    });
   }
+
+  @Test
+  void testUpdateDatasetIndexWithValue() {
+    initService();
```
pshapiro4broad: Same change here; using `BeforeEach` would let you avoid calling `initService()` in each test method.

src/test/java/org/broadinstitute/consent/http/service/ElasticSearchServiceTest.java
```diff
@@ -632,16 +647,46 @@ void testIndexStudyWithDatasets() {
...
+    assertDoesNotThrow(() -> service.updateDatasetIndex(1, 1, Instant.now()));
+  }
+
+  @Test
+  void testDeleteDatasetIndexWhenDatasetExists() throws Exception {
+    Dataset dataset = new Dataset();
+    when(datasetDAO.findDatasetById(any())).thenReturn(dataset);
+    initService();
+    assertDoesNotThrow(() -> service.updateDatasetIndex(1, 1, null));
+    verify(datasetServiceDAO, times(1)).updateDatasetIndex(any(), any(), any());
```
pshapiro4broad: `times(1)` is redundant here
```suggestion
    verify(datasetServiceDAO).updateDatasetIndex(any(), any(), any());
```

pshapiro4broad (approved):
A few comments but looks OK overall. I do think a general approach where we periodically scanned for stale datasets would be a better approach, but that would be a much bigger change.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
```diff
@@ -771,63 +811,69 @@ void testUpdateDatasetDataUse_NotModified() {
...
+      assertThat(returnedDatasets, hasSize(1));
+      assertEquals(dataset.getDatasetId(), returnedDatasets.get(0).getDatasetId());
```
pshapiro4broad: This is the same as
```suggestion
      assertThat(returnedDatasets, contains(dataset));
```

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
s-rubenstein: If so, I think this is the clearest pattern of those three.

src/test/java/org/broadinstitute/consent/http/resources/StudyResourceTest.java
s-rubenstein: To weigh in here, I think I lean towards thinking about whether we care if those parameters change suddenly or not. If we aren't worried about that, `any()` is okay, but my guess is we care if the parameters did suddenly change.

s-rubenstein (approved):
I think Phil had some great comments. I left a few of my own thoughts, but otherwise this looks reasonable.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
rushtong: > > ensure that a specific dataset exists in the index
> 
> Do you mean does not exist? This API deletes a dataset from the index.
> 

Thank you, yes, I did mean that it is intended to remove an entry from the index.

> But I'm still wondering why we'd want to delete a specific dataset from the index.

We have had requests from Jonathan to remove a specific dataset when the custodian has specifically requested it. Those are always one-off requests - eventually the dataset gets re-indexed when the custodian is ready for it.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
rushtong: The DAO call can return a null value from `return datasetDAO.findDatasetById(datasetId);` which does not throw a sql exception.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
rushtong: This is a good catch. There are two usages in this class, one is intended to update the index because we've already updated the index date and the second set of usages should be checking the date, but they are not - I need to refactor this.

src/test/java/org/broadinstitute/consent/http/resources/DatasetResourceTest.java
rushtong: > `times(1)` is the default so it can be omitted.

This works, thank you!

> In this case the `verify()` call isn't necessary since you have a `when()` and strict stubbing is enabled. If you wanted to, you could remove the `when()` and leave the `verify()`, which would have the same effect.

Removing `when` and leaving `verify` fails, I do need both to ensure that it is called.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
rushtong: Yes, I think I can do some re-organizing. There are two distinct cases, one where we want to ensure that we are always performing both operations (update index and indexed date), and one where we only want to do it when it has been previously indexed.

fboulnois (approved):
Changes look good after PR changes 👍

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
rushtong: JDBI interface DAO calls fail if there is an empty list passed into a call that tries to `@BindList` on the argument.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
rushtong: I can move the logging there, that would simplify this method a bit to avoid the try/catch

src/main/java/org/broadinstitute/consent/http/models/Study.java
```diff
@@ -79,14 +78,11 @@ public Set<StudyProperty> getProperties() {
     return properties;
   }
 
-  public void setProperties(Set<StudyProperty> properties) {
-    this.properties = properties;
+  public void addProperties(Set<StudyProperty> properties) {
+    this.properties.addAll(properties);
   }
```
pshapiro4broad: Given how this is used, you could make it easier on the caller and accept varargs
```suggestion
  public void addProperties(StudyProperty... properties) {
    this.properties.addAll(Set.of(properties));
  }
```

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -204,7 +222,7 @@ public Dataset approveDataset(Dataset dataset, User user, Boolean approval) {
   private void sendDatasetApprovalNotificationEmail(Dataset dataset, User user, Boolean approval)
       throws Exception {
     Dac dac = dacDAO.findById(dataset.getDacId());
-    if (approval) {
+    if (Boolean.TRUE.equals(approval)) {
```
pshapiro4broad: This form is only needed because `approval` is not of type `boolean`. Since we know it's not null (it's validated in `DacResource.approveDataset()` you can change its type to `boolean` here and in `approveDataset()` above so this isn't required.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
pshapiro4broad: I'm not that familiar with JDBI but it looks like the behavior is configurable. https://jdbi.org/apidocs/org/jdbi/v3/sqlobject/customizer/BindList.EmptyHandling.html.

pshapiro4broad (approved):
Looks OK just a few last comments

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
rushtong: This is nice, I didn't know about that convenience. There's a good amount of safety checks we can avoid with this.

PR #2471, opened 2025-03-20
[DT-643][risk=no] Optimize tests so they can run locally.

otchet-broad:
### Addresses
[https://broadworkbench.atlassian.net/browse/DT-643](https://broadworkbench.atlassian.net/browse/DT-643)
### Summary
Tests were failing with "Too Many Files Open" errors. 

----

- Verify all tests go green


src/test/java/org/broadinstitute/consent/http/WithMockServer.java
```diff
@@ -23,6 +23,24 @@ public interface WithMockServer {
   DockerImageName IMAGE =
       DockerImageName.parse("mockserver/mockserver:mockserver-" + getMockServerVersion());
 
+  static String getMockServerVersion() {
+    String version = "5.15.0";
```
otchet-broad, line 27: This version was being set at 5.12.0, but the testcontainer was pulling in v.5.15.0.

pom.xml
```diff
@@ -602,6 +607,13 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
```
otchet-broad, line 612: Added so we can now launch one database and one mockserver container that can be used across all tests.
We were tearing down each of these per test class and our construction was leaky.

src/test/java/org/broadinstitute/consent/http/MockServerTestHelper.java
```diff
@@ -0,0 +1,35 @@
+package org.broadinstitute.consent.http;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.platform.launcher.TestExecutionListener;
+import org.junit.platform.launcher.TestPlan;
+import org.mockserver.client.MockServerClient;
+import org.testcontainers.containers.MockServerContainer;
+import org.testcontainers.containers.wait.strategy.Wait;
+
+public class MockServerTestHelper extends AbstractTestHelper implements TestExecutionListener,
```
otchet-broad, line 10: Extends AbstractTestHelper so that a test that needs both the MockServerTestHelper and the AbstractTestHelper can easily extend a single class and be off to the races.

src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
```diff
@@ -31,17 +31,19 @@
...
 import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.platform.launcher.TestExecutionListener;
+import org.junit.platform.launcher.TestPlan;
 import org.testcontainers.containers.PostgreSQLContainer;
 import org.testcontainers.containers.wait.strategy.Wait;
 
-public class DAOTestHelper extends AbstractTestHelper {
+public class DAOTestHelper extends AbstractTestHelper implements TestExecutionListener {
```
otchet-broad, line 38: TestExecutionListener is the magic in Junit5 that lets us build objects that can live across our tests.

src/test/java/org/broadinstitute/consent/http/db/TestingDAO.java
```diff
@@ -3,97 +3,17 @@
...
-  void deleteAllDACDAAs();
-
-  @SqlUpdate("DELETE FROM lc_daa")
-  void deleteAllLibraryCardDAAs();
+  @SqlUpdate("""
+      DO $$ DECLARE
+          table_name text;
+      BEGIN
+          FOR table_name IN (SELECT tablename FROM pg_tables WHERE schemaname='public'
+                AND tablename NOT IN ('roles', 'dictionary')) LOOP
```
otchet-broad, line 13: We have reference data in roles and dictionary that needs to survive.  Everything else must go!

src/test/java/org/broadinstitute/consent/http/service/DataAccessReportsParserTest.java
```diff
@@ -77,41 +78,44 @@ void testDataAccessApprovedReport() throws IOException {
...
-        assertEquals(columns[4], HeaderDAR.DATA_REQUESTER_NAME.getValue());
-        assertEquals(columns[5], HeaderDAR.ORGANIZATION.getValue());
-        assertEquals(columns[6], HeaderDAR.CODED_VERSION_SDUL.getValue());
-        assertEquals(columns[7], HeaderDAR.CODED_VERSION_DAR.getValue());
-        assertEquals(columns[8], HeaderDAR.RESEARCH_PURPOSE.getValue());
-        assertEquals(columns[9], HeaderDAR.DATE_REQUEST_SUBMISSION.getValue());
-        assertEquals(columns[10], HeaderDAR.DATE_REQUEST_APPROVAL.getValue());
-        assertEquals(columns[11],
-            HeaderDAR.DATE_REQUEST_RE_ATTESTATION.getValue());
+    try (Stream<String> stream = Files.lines(Paths.get(file.getPath()))) {
```
otchet-broad: with-resources!

src/test/java/org/broadinstitute/consent/http/service/DataAccessReportsParserTest.java
```diff
@@ -124,35 +128,38 @@ void testDataAccessReviewedReport() throws IOException {
...
-        assertEquals(columns[7], HeaderDAR.APPROVED_DISAPPROVED.getValue());
-      }
-      if (i == 1) {
-        assertEquals(DAR_CODE, columns[0]);
-        assertEquals(NAME, columns[1]);
-        assertEquals(DS_IDENTIFIER, columns[2]);
-        assertEquals(CONSENT_NAME, columns[3]);
-        assertEquals(columns[4], sDUL.replace("\n", " "));
-        assertEquals("Yes", columns[7]);
+    try(Stream<String> stream = Files.lines(Paths.get(file.getPath()))){
```
otchet-broad: with-resources!

src/test/java/org/broadinstitute/consent/http/service/DataAccessReportsParserTest.java
```diff
@@ -166,35 +173,38 @@ void testDataAccessReviewedReportNullElectionDate() throws IOException {
...
```
otchet-broad: with-resources!

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
```diff
@@ -91,6 +92,12 @@ class EmailServiceTest {
   void setUp() {
   }
 
+  @AfterEach
```
otchet-broad: These were leaking.

src/test/java/org/broadinstitute/consent/http/service/OntologyServiceTest.java
```diff
@@ -10,46 +10,22 @@
...
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.junit.jupiter.MockitoExtension;
-import org.mockserver.client.MockServerClient;
 import org.mockserver.model.Header;
-import org.testcontainers.containers.MockServerContainer;
 
 @ExtendWith(MockitoExtension.class)
-class OntologyServiceTest implements WithMockServer {
```
otchet-broad, line 30: We get to remove lots of boilerplate!

src/test/java/org/broadinstitute/consent/http/WithMockServer.java
rushtong: Really nice catch here. My approach for swagger is to pull the version dynamically from `mvn.properties`. In my local value, I see:
```
...
mockserver.version=5.15.0
...
```
See https://github.com/DataBiosphere/consent/blob/f9dae18c771324472613b22cf3b0e7873f65a6f7/src/main/java/org/broadinstitute/consent/http/resources/SwaggerResource.java#L38 for how we're doing that.

Edit: Never mind ... looks like you found that code already.

src/test/java/org/broadinstitute/consent/http/service/DataAccessReportsParserTest.java
rushtong: Yes, this is a real problem in current tests. I've been tackling them on an as-needed basis but that's been very slow.

rushtong:
The current smoke test failures have to do with parsing one of the test xml results. Logs from pulling images are in the content and that's throwing off the python parser.
```
  <testcase name="status_page_OK" classname="org.broadinstitute.consent.integration.status.StatusTests" time="0.533">
    <system-out><![CDATA[22:49:05.877 [main] INFO  tc.testcontainers/ryuk:0.11.0 - Pulling docker image: testcontainers/ryuk:0.11.0. Please be patient; this may take some time but only needs to be done once.
22:49:06.208 [docker-java-stream--1295187651] INFO  tc.testcontainers/ryuk:0.11.0 - Starting to pull image
22:49:06.240 [docker-java-stream--1295187651] INFO  tc.testcontainers/ryuk:0.11.0 - Pulling image layers:  0 pending,  0 downloaded,  0 extracted, (0 bytes/0 bytes)

...

UPDATE SUMMARY
Run:                        224
Previously run:               0
Filtered out:                 0
-------------------------------
Total change sets:          224


]]></system-out>
```

pom.xml
```diff
@@ -83,6 +87,9 @@
             <version>${surefire.version}</version>
             <configuration>
               <includes>**/integration/**/*.java</includes>
+              <systemProperties>
+                <disableTestContainers>true</disableTestContainers>
```
otchet-broad: Here we set the property to turn off running containers for the integration tests because
1) They aren't needed for these tests.
2) The script that uploads the successful run can't handle CDATA tags in XML so the upload of the build results to bigquery fails.

src/test/java/org/broadinstitute/consent/http/AbstractTestHelper.java
```diff
@@ -21,4 +21,8 @@ public long randomLong() {
     return RandomUtils.secureStrong().randomLong();
   }
 
+public static boolean disableTestContainers() {
+    return System.getProperty("disableTestContainers", "false").equals("true");
```
otchet-broad: Read the property to see if we should turn off test containers.

src/test/java/org/broadinstitute/consent/http/MockServerTestHelper.java
```diff
@@ -0,0 +1,38 @@
...
+    WithMockServer {
+
+  public static final MockServerContainer container = new MockServerContainer(IMAGE).waitingFor(
+      Wait.forLogMessage(".*started on port:.*", 1));
+  public static MockServerClient mockServerClient;
+
+  static void startUp() {
+    if (disableTestContainers()) {
```
otchet-broad: If we should disable the containers, simply return.

src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
```diff
@@ -71,8 +71,19 @@ public class DAOTestHelper extends AbstractTestHelper {
   @SuppressWarnings("rawtypes")
   private static PostgreSQLContainer postgresContainer;
 
-  @BeforeAll
-  public static void startUp() throws Exception {
+   @Override
+  public void testPlanExecutionStarted(TestPlan testPlan) {
+     if (disableTestContainers()) {
```
otchet-broad: If we don't need containers, we're not starting the app so we can just return here.

pom.xml
```diff
@@ -397,6 +404,11 @@
       <artifactId>jdbi3-guava</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>org.jdbi</groupId>
+      <artifactId>jdbi3-testing</artifactId>
```
rushtong: Can we add the test scope here or is it necessary for full builds?

src/test/java/org/broadinstitute/consent/http/db/TestingDAO.java
rushtong: Nice cleanup! I didn't know that truncate would work across foreign key constraints.

pom.xml
fboulnois: Does this mean that each test that exercises the database will have content from the previous tests?

src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
rushtong: This is an important distinction - it would be nice to have some kind of commentary here. In other words, `disableTestContainers` indicates two things:
1. Test containers are not necessary
2. App startup is not possible because it relies on existing test containers (postgres + mockserver).

src/test/java/org/broadinstitute/consent/http/AbstractTestHelper.java
```diff
@@ -5,6 +5,10 @@
 
 public abstract class AbstractTestHelper {
 
+  public static boolean disableTestContainers() {
+    return System.getProperty("disableTestContainers", "false").equals("true");
+  }
```
fboulnois:
```suggestion
  public static boolean disableTestContainers() {
    String defaultProp = "false";
    return System.getProperty("disableTestContainers", defaultProp).equals("true");
  }
```

Clarifies that the second argument is a default for `getProperty`, otherwise I found it hard to read.

rushtong (approved):
Love the overall cleanup 👍🏽 Address comments as you see fit.

DEVNOTES.md
```diff
@@ -17,8 +17,20 @@ cd consent
...
+
+For example:
+```
+% brew install podman
+% podman machine
+% podman machine init --cpus 8 -m 4096 --now
+```
+Depending on your system configuration, you may need to set the DOCKER_HOST environment variable.
+Check the output of the `podman machine init` command.
+
```
fboulnois: Suggestion: Move the added text to the Docker section. Also, I'm not sure if we should specifically call out how to install `podman`, since different developers have different types of systems and this only applies to macOS. On the other hand, it may be worth emphasizing that running with `podman` provided a performance advantage for these tests.

fboulnois (commented):
A few suggestions:

DEVNOTES.md
otchet-broad: I'm inclined to leave the intro to the testing section, because to run the tests, you do need an OCI runtime (e.g., docker, podman, etc) running - but I like your other suggestions.  Let me know if my update works for you.

pom.xml
otchet-broad: This was from an earlier strategy.  I was able to remove this dependency.

pom.xml
otchet-broad: No, we've handled that differently.
With this new approach, we're building one postgres and one mockserver container.  I've hooked the @BeforeEach to truncate (remove) all of the data in postgres using an @BeforeEach to ensure we don't have data remaining between tests for test classes that extend the DAOTestHelper.  We did something similar before (with SQL statements curated to delete that suffered from cases where constraints changed and ordering would change for sequencing), but we would also stop and restart the container after each test class.  This was very inefficient - and we would leak a JDBI at each test class that extended the class.  We would make a similar leak with the MockServerClient and I've also hooked a reset in a @BeforeEach in the new class those tests extend.

DEVNOTES.md
fboulnois, line 1: 👍

fboulnois (approved):
Looks good!

pom.xml
pshapiro4broad: Could this be `enableTestContainers` instead? I think boolean properties should be expressed in the positive whenever possible. This helps avoid double negatives in the code, which can make it harder to understand.

I guess the tension here is that it's nice to think of the default/missing case as `false` rather than `true`. That could still work by enabling it for a directory instead of disabling it. Or, I think the `includes` form can use a negative regex.

src/test/java/org/broadinstitute/consent/http/MockServerTestHelper.java
```diff
@@ -0,0 +1,38 @@
...
+import org.testcontainers.containers.wait.strategy.Wait;
+
+public class MockServerTestHelper extends AbstractTestHelper implements TestExecutionListener,
+    WithMockServer {
+
+  public static final MockServerContainer container = new MockServerContainer(IMAGE).waitingFor(
+      Wait.forLogMessage(".*started on port:.*", 1));
+  public static MockServerClient mockServerClient;
+
+  static void startUp() {
```
pshapiro4broad: Does this need to be a separate method? its only use is in this file.
```suggestion
  private static void startUp() {
```

src/test/java/org/broadinstitute/consent/http/MockServerTestHelper.java
```diff
@@ -0,0 +1,38 @@
...
+import org.junit.platform.launcher.TestExecutionListener;
+import org.junit.platform.launcher.TestPlan;
+import org.mockserver.client.MockServerClient;
+import org.testcontainers.containers.MockServerContainer;
+import org.testcontainers.containers.wait.strategy.Wait;
+
+public class MockServerTestHelper extends AbstractTestHelper implements TestExecutionListener,
+    WithMockServer {
+
+  public static final MockServerContainer container = new MockServerContainer(IMAGE).waitingFor(
```
pshapiro4broad: Normally static final objects are uppercase
```suggestion
  public static final MockServerContainer CONTAINER = new MockServerContainer(IMAGE).waitingFor(
```

src/test/java/org/broadinstitute/consent/http/MockServerTestHelper.java
```diff
@@ -0,0 +1,38 @@
...
+    try {
+      startUp();
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  @BeforeEach
+  void beforeEach() {
+    mockServerClient.reset();
```
pshapiro4broad, line 32: Does this require a static `mockServerClient` or would it also work to create a new one each time? From looking at the code, my thinking was as long as `container` is static, the client wouldn't need to be.

src/test/java/org/broadinstitute/consent/http/WithMockServer.java
pshapiro4broad: Why hardcode a default value at all? If the value is missing from the `mvn.propeties` file, wouldn't it be better to fail the test? Having this here introduces a code duplication that must manually be kept in sync.

src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
```diff
@@ -71,8 +71,25 @@ public class DAOTestHelper extends AbstractTestHelper {
   @SuppressWarnings("rawtypes")
   private static PostgreSQLContainer postgresContainer;
 
-  @BeforeAll
-  public static void startUp() throws Exception {
+   @Override
```
pshapiro4broad: whitespace
```suggestion
  @Override
```

src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
```diff
@@ -71,8 +71,25 @@ public class DAOTestHelper extends AbstractTestHelper {
...
+  public void testPlanExecutionStarted(TestPlan testPlan) {
+     /* The tests that extend this class make the necessary assumption that the app will be started
+     and backed by a running database.  When we do not need the containers, we are also indicating
+     we do not need the application within this class, hence we can early return and skip resource
+     construction.
+     This became a necessary optimization because of command line generated output that could not
+     be properly handled by a CI/CD process. */
```
pshapiro4broad: In general the java terra code uses google code style https://google.github.io/styleguide/javaguide.html#s4.8.6-comments; Personally I would use `//` but `/*` is OK if it's formatted as in the style guide.
```suggestion
     // The tests that extend this class make the necessary assumption that the app will be started
     // and backed by a running database.  When we do not need the containers, we are also indicating
     // we do not need the application within this class, hence we can early return and skip resource
     // construction.
     //
     // This became a necessary optimization because of command line generated output that could not
     // be properly handled by a CI/CD process.
```

src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
pshapiro4broad: It would be nice to assert incorrect configurations in the case where the test's assumptions differ from the configuration. That could help avoid confusion when new tests are added that don't work as expected.

It looks like none of the tests that subclass `DAOTestHelper` are in the `integration` directory. Doesn't that mean that they all rely on test containers?

src/test/java/org/broadinstitute/consent/http/AbstractTestHelper.java
```diff
@@ -5,6 +5,11 @@
 
 public abstract class AbstractTestHelper {
 
+  public static boolean disableTestContainers() {
+    String defaultProp = "false";
+    return System.getProperty("disableTestContainers", defaultProp).equals("true");
```
pshapiro4broad: `Boolean.parseBoolean()` is equivalent to this form; it handles `null` and returns `false` for non `"true"` values.
```suggestion
    return Boolean.parseBoolean(System.getProperty("disableTestContainers"));
```

Note that this maps misspelled boolean values to `false`. It may be better to fail in that case.

src/test/java/org/broadinstitute/consent/http/service/DataAccessReportsParserTest.java
```diff
@@ -77,41 +78,44 @@ void testDataAccessApprovedReport() throws IOException {
...
-        assertEquals(CONSENT_NAME, columns[3]);
-        assertEquals(REQUESTER, columns[4]);
-        assertEquals(ORGANIZATION, columns[5]);
-        assertEquals(columns[6], sDUL.replace("\n", " "));
-        assertEquals(RUS_SUMMARY, columns[8]);
-      }
-      i++;
+      assertEquals(2, i);
+    } catch (Exception e) {
+      fail(e);
```
pshapiro4broad: try-with-resources doesn't need a `catch`. A thrown exception will fail the test so it's not necessary to `fail()` here.
```suggestion
```

src/test/java/org/broadinstitute/consent/http/service/DataAccessReportsParserTest.java
pshapiro4broad: Often `try-with-resources` is a great place to use `var`
```suggestion
    try (var stream = Files.lines(Paths.get(file.getPath()))) {
```

Although, from looking at this code, the input file is fixed length and only has two lines, so using a `Stream` is unnecessary.
```
    var lines = Files.readAllLines(Paths.get(file.getPath()))
```

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
pshapiro4broad: I don't understand why this is necessary. At most you'll have only one copy of each object at a time, since each call to `initServices()` will overwrite the previous object. Once the test is complete, `EmailServiceTest` should be gc'd, which should also cause these members to be gc'd. So I don't think nulling the fields is required here.

src/test/java/org/broadinstitute/consent/http/MockServerTestHelper.java
```diff
@@ -0,0 +1,38 @@
...
+  public void testPlanExecutionStarted(TestPlan testPlan) {
+    try {
+      startUp();
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  @BeforeEach
+  void beforeEach() {
```
pshapiro4broad, line 31: This is one reason I don't like to use inheritance for tests. If a subclass overrides this method, and doesn't call `super()`, this method will never be called. But if the method is not overridden, and a subclass has a `@BeforeEach` with a different name, then JUnit will call both `@BeforeEach` methods, and you don't want to call the base class method explicitly. This can lead to hard to track down issues in tests.

As an alternate strategy, I would prefer embedding over inheritance. Users would need to explicitly startup/teardown the embedded object, so it's a bit more code, but I find it easier to maintain. Although, the current code is already using inheritance so this is more a comment about future possible changes not something for this PR.

src/test/java/org/broadinstitute/consent/http/service/SupportRequestServiceTest.java
```diff
@@ -157,15 +135,17 @@ void testPostTicketToSupportUnableToParseResponse() {
             .withStatusCode(HttpStatusCodes.STATUS_CODE_CREATED)
             .withBody(expectedBody)
         );
-    assertThrows(ServerErrorException.class, () -> service.postAttachmentToSupport("Test".getBytes()));
+    assertThrows(ServerErrorException.class,
+        () -> service.postAttachmentToSupport("Test".getBytes()));
```
pshapiro4broad, line 139: it's a minor thing but sonar will likely flag all these cases as it prefers to only have a single method call inside an `assertThrows()`. E.g., if `getBytes()` throws a `ServerErrorException` then the test will pass but for the wrong reason.

src/test/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAOTest.java
```diff
@@ -41,31 +41,29 @@
...
-  @BeforeEach
-  void setup() throws IOException {
+  @BeforeAll
+  static void setup() throws IOException {
+    GCSService gcsService = Mockito.mock(GCSService.class);
     DraftFileStorageServiceDAO draftFileStorageServiceDAO = new DraftFileStorageServiceDAO(jdbi,
         gcsService, fileStorageObjectDAO);
-    this.draftServiceDAO = new DraftServiceDAO(jdbi, draftDAO, draftFileStorageServiceDAO);
+    when(gcsService.storeDocument(any(), anyString(), any())).thenReturn(
+        BlobId.of(UUID.randomUUID().toString(), UUID.randomUUID().toString()));
```
pshapiro4broad: Do you need to move the mocks into the `BeforeAll`? I would leave them in `BeforeEach`, so they benefit from strict stubbing, and can support the case where a test needs to change the mock behavior for that test only.

src/test/resources/META-INF/services/org.junit.platform.launcher.TestExecutionListener
```diff
@@ -0,0 +1,2 @@
+org.broadinstitute.consent.http.db.DAOTestHelper
```
pshapiro4broad, line 1: Adding metadata to a text file seems bad but I guess there's no better way to do this since we need the code to run before the test suite is executed.

pshapiro4broad (approved):
Overall this looks OK but I have a few comments.

Although, I was wondering if there was a way to remove the `disableTestContainers` property. If there are subclasses that only tests that need this feature will use,, e.g., `DAOTestHelper` is only used by tests that need test containers, does it need to be configured as well?

pom.xml
otchet-broad: Easy enough to rename the property, the method name in code, and the default value (from false to true).  Sound like a reasonable plan @pshapiro4broad ?

src/test/java/org/broadinstitute/consent/http/AbstractTestHelper.java
otchet-broad: @pshapiro4broad, Based on the plan for changing this to enableTestContainers, something like Boolean.parseBoolean(System.getProperty("enableTestContainers", "true"));
I'll also flip the conditionals in MockServerTestHelper and DAOTestHelper.
Reasonable?

src/test/java/org/broadinstitute/consent/http/MockServerTestHelper.java
otchet-broad: Refactor coming to eliminate the method.

src/test/java/org/broadinstitute/consent/http/MockServerTestHelper.java
otchet-broad: In the old pattern, we were creating (and leaking) these clients because underlying connections remained open (similar to JDBI) and held references. I wish both had working <destroy> methods that could be invoked.  Given the objects remained (with handles), I ended up making this static. Is it required, no.  IMHO, it helps explain how this actually works because calling reset on the client is just sending a command to the mock server for **the server** to reset.  The client remains connected.

src/test/java/org/broadinstitute/consent/http/WithMockServer.java
otchet-broad: Your point is well taken, @pshapiro4broad.  If we decide on that, maybe we make another ticket for this change - since it's beyond getting the tests to pass locally.  

The reason the vast majority of the code below is green is because of formatting changes that alphabetized the order of methods.

src/test/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAOTest.java
otchet-broad: The mocks weren't in the BeforeEach prior to this refactor, so they can't be be "left" in BeforeEach.  Instead, each test used the same stubbing.  I'm fine with adding them to only the specific tests that need it.  When added to the @.BeforeEach, I get the UnnecessaryStubbing exception.  The @.BeforeAll doesn't throw that.

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
otchet-broad: Just re-ran and can definitely remove these.  I wonder if there was something else happening that made it look like these weren't being released.

pom.xml
pshapiro4broad: Sounds good to me

src/test/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAOTest.java
pshapiro4broad: > When added to the @.BeforeEach, I get the UnnecessaryStubbing exception. The @.BeforeAll doesn't throw that.

This is working as intended; the idea is that any unused mock will cause a test to fail, effectively making all `when()` mocks into `verify()`s.

By mocking in BeforeAll, the mocks are now shared between all tests so any custom mocking or verification is no longer possible.

There are basically two approaches (assuming you want to take advantage of strict stubbing). One is to only mock what you need by moving mocking into shared methods that can be called from test methods. Another is to mock everything in BeforeEach and `reset()` any mocks that are unused in a test. The latter is nice for cases where almost every test uses the mocks and there are only a couple outliers that don't need the mocks. A third approach which is a refinement of the first one is to use nested test classes. That lets you have a set of tests that all share a BeforeEach without having to call it explicitly.

pshapiro4broad: Looking at this code, I think there's only one test that doesn't use the mock, so I would use the `reset()` approach.

src/test/java/org/broadinstitute/consent/http/service/SupportRequestServiceTest.java
otchet-broad: This my change for this was just formatting.  Should I create a new ticket for this?

pom.xml
otchet-broad: (@pshapiro4broad - Let me know what you think of the changes)

pshapiro4broad:
I checked out the branch and verified that I was able to run `mvn clean test package` locally without error. Nice Job!

src/test/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAOTest.java
otchet-broad: Got it.  Great suggestion.  Thanks!

src/test/java/org/broadinstitute/consent/http/service/SupportRequestServiceTest.java
pshapiro4broad: There are likely many sonar warnings so there's no need to call this out specifically. It will eventually get addressed when this code is worked on again.

PR #2476, opened 2025-04-08
DT-1447: Test Refactoring/cleanup

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1447

### Summary
Refactor the tests for:
1. Remove a test-only query
2. Fix all swapped assertion order problems
3. Migrate to non-deprecated random utility functions
4. Fix tests that expected an ordered list where no order existed in the db query 

----

- Verify all tests go green

### DIFF
--- a/src/test/java/org/broadinstitute/consent/http/db/DacDAOTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/db/DacDAOTest.java
@@ -1,5 +1,8 @@
 package org.broadinstitute.consent.http.db;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.hasSize;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
@@ -81,8 +84,7 @@ class DacDAOTest extends DAOTestHelper {
     dacDAO.addDacMember(UserRoles.CHAIRPERSON.getRoleId(), chair.getUserId(), dac.getDacId());
 
     Set<User> users = dacDAO.findAllDACUsersBySearchString(chair.getEmail());
-    assertFalse(users.isEmpty());
-    assertEquals(1, users.size());
+    assertThat(users, hasSize(1));
   }
 
   @Test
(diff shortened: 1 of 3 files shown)


src/test/java/org/broadinstitute/consent/http/db/DacDAOTest.java
```diff
@@ -96,12 +97,14 @@ void testFindAllNoDatasets() {
     Integer dacId2 = createRandomDAC();
     List<Dac> dacs = dacDAO.findAll();
 
-    Dac dac1 = dacs.get(0);
+    Dac dac1 = dacs.stream().filter(d -> d.getDacId().equals(dacId1)).findFirst().orElse(null);
+    assertNotNull(dac1);
```
pshapiro4broad: I would use `orElseThrow()` instead of `assertNotNull()`; either way the test fails with an exception. It's true that one is an NPE and one is an `AssertionError`.. I don't mind this difference though.
```suggestion
    Dac dac1 = dacs.stream().filter(d -> d.getDacId().equals(dacId1)).findFirst().orElseThrow();
```

src/test/java/org/broadinstitute/consent/http/db/DacDAOTest.java
```diff
@@ -258,7 +270,7 @@ void testFindMembersByDacId() {
     List<User> dacMembers = dacDAO.findMembersByDacId(dac.getDacId());
     assertNotNull(dacMembers);
     assertFalse(dacMembers.isEmpty());
-    assertEquals(dacMembers.size(), 4);
+    assertEquals(4, dacMembers.size());
```
pshapiro4broad: I know this test doesn't use hamcrest, but if it did, you could replace
```
    assertNotNull(dacMembers);
    assertFalse(dacMembers.isEmpty());
    assertEquals(4, dacMembers.size());
```

with
```
    assertThat(dacMembers, hasSize(4));
```

src/test/java/org/broadinstitute/consent/http/db/DacDAOTest.java
rushtong: Nice simplification. We're pretty sure the `get(0)` calls were the root of the flakiness since the DAO methods do not have an enforced order.

src/test/java/org/broadinstitute/consent/http/db/DacDAOTest.java
rushtong: I'm revamping this pretty heavily anyway ... might as well go all the way.

pshapiro4broad (approved):

snf2ye (approved):

PR #2477, opened 2025-04-10
DT-1496: Prevent update self from updating institution id

rushtong:
### Addresses
Back end work for https://broadworkbench.atlassian.net/browse/DT-1496

### Summary
Prevent users from updating their institution id via the self update endpoint.
If a user tries to do this, return a bad request response.
Relatively large change to resource test due to the removal of previous logic and some other minor cleanup efforts.
See https://github.com/DataBiosphere/duos-ui/pull/2821 for related front end work.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/UserUpdateFields.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/UserUpdateFields.java
@@ -18,7 +18,7 @@ public class UserUpdateFields {
   protected static final List<Integer> IGNORE_ROLE_IDS = List.of(UserRoles.CHAIRPERSON.getRoleId(),
       UserRoles.MEMBER.getRoleId());
   private static final List<Integer> VALID_ROLE_IDS = Arrays.stream(UserRoles.values())
-      .map(UserRoles::getRoleId).collect(Collectors.toList());
+      .map(UserRoles::getRoleId).toList();
   private String displayName;
   private Integer institutionId;
   private Boolean emailPreference;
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/consent/http/models/UserUpdateFields.java
```diff
@@ -177,6 +179,31 @@ public List<Integer> getRoleIdsToRemove(List<Integer> currentUserRoleIds) {
                   VALID_ROLE_IDS.contains(
                       id);                            // Only remove roles we know about
             })
-        .collect(Collectors.toList());
+        .toList();
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
rushtong, line 184: This was only really needed for testing.

pshapiro4broad: That's good because since this is a mutable object the results aren't valid over time. It would be better to make this immutable using `record` so you don't have write and maintain these methods for tests. But that would require a bigger change.

src/main/java/org/broadinstitute/consent/http/models/UserUpdateFields.java
```diff
@@ -1,5 +1,7 @@
 package org.broadinstitute.consent.http.models;
 
+import static com.google.common.base.Objects.equal;
```
pshapiro4broad: Even in this case, I wouldn't use a static import in non-test code. Also there is `Objects.equals()` and `Objects.hash()` in `java.util.Objects`. I would prefer using JDK APIs over google ones when possible.

src/main/resources/assets/paths/user.yaml
```diff
@@ -34,9 +34,6 @@ put:
             displayName:
               description: The display name for the user
               type: string
-            institutionId:
```
pshapiro4broad, line 37: If this was using openapi generated code, there would be a different object type for this request (vs the admin API) so we wouldn't need to add a specific validation for this field.

We could do this here too, e.g., have both `SelfUserUpdateFields` and `AdminUserUpdateFields`. Then the `400` would be thrown when `gson.fromJson()` is called.

pshapiro4broad (approved):
A few comments, looks OK overall

fboulnois (approved):
👍

PR #2478, opened 2025-04-10
[DT-1493] Add mapping from institution to domains

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1493

### Summary

Creates a new object which maps institutions to their domains.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/cloudstore/GCSService.java
+++ b/src/main/java/org/broadinstitute/consent/http/cloudstore/GCSService.java
@@ -8,12 +8,14 @@ import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageOptions;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.gson.Gson;
 import com.google.inject.Inject;
 import jakarta.ws.rs.NotFoundException;
 import java.io.ByteArrayInputStream;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -22,8 +24,11 @@ import java.util.UUID;
 import org.apache.commons.io.IOUtils;
 import org.broadinstitute.consent.http.configurations.StoreConfiguration;
 import org.broadinstitute.consent.http.util.ConsentLogger;
+import org.broadinstitute.consent.http.util.gson.GsonUtil;
 
 public class GCSService implements ConsentLogger {
+
+  private static final Gson GSON = GsonUtil.gsonBuilderWithAdapters().create();
 
   private StoreConfiguration config;
   private Storage storage;
@@ -135,6 +140,23 @@ public class GCSService implements ConsentLogger {
       return output;
     } else {
       throw new NotFoundException("Document Not Found: " + blobIds.toString());
+    }
+  }
+
+  /**
+   * Read JSON file from GCS bucket
+   *
+   * @param blobIdName String value of the document blob id name
+   * @param valueType  Class type of the object to be read
+   * @return Object of the specified type
+   */
+  public <T> T readJsonFileFromBucket(String blobIdName, Class<T> valueType) throws IOException {
+    try (InputStream is = getDocument(blobIdName)) {
+      InputStreamReader reader = new InputStreamReader(is);
+      return GSON.fromJson(reader, valueType);
+    } catch (Exception e) {
+      logException("Error reading JSON file from bucket: " + e.getMessage(), e);
+      throw new IOException("Failed to read JSON file from bucket: " + blobIdName, e);
     }
   }
 
--- a/src/main/java/org/broadinstitute/consent/http/models/InstitutionDomainMap.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/InstitutionDomainMap.java
@@ -1,0 +1,21 @@
+package org.broadinstitute.consent.http.models;
+
+import java.util.List;
+import java.util.Map;
+
+public class InstitutionDomainMap {
+
+  private Map<String, List<String>> institutionDomainMap;
+
+  public Map<String, List<String>> getInstitutionDomainMap() {
+    return institutionDomainMap;
+  }
+
+  public void setInstitutionDomainMap(Map<String, List<String>> institutionDomainMap) {
+    this.institutionDomainMap = institutionDomainMap;
+  }
+
+  public List<String> getDomainsForInstitution(String institution) {
+    return institutionDomainMap.get(institution);
+  }
+}
--- a/src/main/java/org/broadinstitute/consent/http/models/InstitutionDomainMap.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/InstitutionDomainMap.java
@@ -1,21 +1,21 @@
 package org.broadinstitute.consent.http.models;
 
-import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 public class InstitutionDomainMap {
 
-  private Map<String, List<String>> institutionDomainMap;
+  private Map<String, Set<String>> institutionDomainMap;
 
-  public Map<String, List<String>> getInstitutionDomainMap() {
+  public Map<String, Set<String>> getInstitutionDomainMap() {
     return institutionDomainMap;
   }
 
-  public void setInstitutionDomainMap(Map<String, List<String>> institutionDomainMap) {
+  public void setInstitutionDomainMap(Map<String, Set<String>> institutionDomainMap) {
     this.institutionDomainMap = institutionDomainMap;
   }
 
-  public List<String> getDomainsForInstitution(String institution) {
+  public Set<String> getDomainsForInstitution(String institution) {
     return institutionDomainMap.get(institution);
   }
 }
--- a/src/test/java/org/broadinstitute/consent/http/cloudstore/GCSServiceTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/cloudstore/GCSServiceTest.java
@@ -21,7 +21,6 @@ import java.util.UUID;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.RandomStringUtils;
 import org.broadinstitute.consent.http.configurations.StoreConfiguration;
-import org.broadinstitute.consent.http.models.InstitutionDomainMap;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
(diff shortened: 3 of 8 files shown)


src/main/java/org/broadinstitute/consent/http/models/InstitutionDomainMap.java
```diff
@@ -0,0 +1,21 @@
+package org.broadinstitute.consent.http.models;
+
+import java.util.List;
+import java.util.Map;
+
+public class InstitutionDomainMap {
+
+  private Map<String, List<String>> institutionDomainMap;
```
otchet-broad: Do we expect that order matters in the List<>?

fboulnois: I wasn't anticipating that it would.

otchet-broad: Would Set be more appropriate?

fboulnois: I had a version with a Set, but ultimately didn't see a tremendous advantage for operations like `contains` since the lists are likely to be relatively short. I can switch back, however, if you think that would clarify functionality.

src/main/java/org/broadinstitute/consent/http/models/InstitutionDomainMap.java
```diff
@@ -0,0 +1,21 @@
...
+  public Map<String, List<String>> getInstitutionDomainMap() {
+    return institutionDomainMap;
+  }
+
+  public void setInstitutionDomainMap(Map<String, List<String>> institutionDomainMap) {
+    this.institutionDomainMap = institutionDomainMap;
+  }
+
+  public List<String> getDomainsForInstitution(String institution) {
+    return institutionDomainMap.get(institution);
```
rjohanek, line 19: might be helpful to have a get Institution by domain method, since we will be deriving institution from a domain, but this could also just be added later when we are using it

rjohanek (commented):
this looks good, but where does the actual loading of the info on start up happen? Like where is readJsonFileFromBucket called and where is that information stored? Or we just call it every time we need to use that information?

src/main/java/org/broadinstitute/consent/http/models/InstitutionDomainMap.java
otchet-broad: `List` to me implies two things I want to pay attention to:

1. Order
2. Duplicates

It sounds like we won't have to worry about either.  IMHO, `Set` conveys that clearer than `List`.  That being said, what I've written in my own code as a placeholder is Map<String, Collection\<String\>> with the impl creating a HashSet<>().  I think that probably gives us the most flexibility here.

fboulnois:
> this looks good, but where does the actual loading of the info on start up happen? Like where is readJsonFileFromBucket called and where is that information stored? Or we just call it every time we need to use that information?

Yes, that's the other part that I'm not sure about. The ticket says to do this on startup, but maybe we should stick it in the service that is actually going to do the check instead?

src/main/java/org/broadinstitute/consent/http/models/InstitutionDomainMap.java
otchet-broad: This makes me wonder about the relationships in this data we might have to consider if we allow that mapping.  When were you thinking we would need to do the mapping of domain(s) to institution(s)?  Can a domain be mapped to more than one institution?

src/main/java/org/broadinstitute/consent/http/models/InstitutionDomainMap.java
fboulnois: Ok, I'll switch to `Set`

src/main/java/org/broadinstitute/consent/http/models/InstitutionDomainMap.java
fboulnois: The short answer is no, a domain CANNOT be mapped to more than one institution.

src/main/java/org/broadinstitute/consent/http/cloudstore/GCSService.java
```diff
@@ -138,6 +143,23 @@ public Map<BlobId, InputStream> getDocuments(List<BlobId> blobIds) throws NotFou
...
+   * Read JSON file from GCS bucket
+   *
+   * @param blobIdName String value of the document blob id name
+   * @param valueType  Class type of the object to be read
+   * @return Object of the specified type
+   */
+  public <T> T readJsonFileFromBucket(String blobIdName, Class<T> valueType) throws IOException {
+    try (InputStream is = getDocument(blobIdName)) {
+      InputStreamReader reader = new InputStreamReader(is);
+      return GSON.fromJson(reader, valueType);
```
pshapiro4broad, line 156: This is a little convoluted in that the actual document source is `byte[]` but `getDocument()` converts it to an `InputStream` which this code converts back to a GSON value. It would be more efficient to have a `getDocument()` API that returns `byte[]`
```suggestion
    try {
      byte[] bytes = getDocumentBytes(blobIdName);
      return GSON.fromJson(new String(bytes), valueType);
```

src/test/java/org/broadinstitute/consent/http/cloudstore/GCSServiceTest.java
```diff
@@ -139,4 +140,30 @@ void testDeleteDocument() {
...
+        {
+          "institutionDomainMap": {
+            "Broad Institute": ["broadinstitute.org", "broad.mit.edu"],
+            "Harvard": ["harvard.edu", "hms.harvard.edu"]
+          }
+        }
+        """;
+
+    String blobIdName = "institution-domain-mapping.json";
+    Blob blob = mock(Blob.class);
```
pshapiro4broad: `mock()` infers its return type from context so a parameter is unnecessary
```suggestion
    Blob blob = mock();
```

Although in this case there is already a mock `blob` field so I don't think this is necessary.

src/test/java/org/broadinstitute/consent/http/cloudstore/GCSServiceTest.java
```diff
@@ -139,4 +140,30 @@ void testDeleteDocument() {
...
+
+    initStore();
+
+    InstitutionDomainMap result = service.readJsonFileFromBucket(blobIdName,
+        InstitutionDomainMap.class);
+
+    assertNotNull(result);
+    assertEquals(2, result.getInstitutionDomainMap().size());
+    assertTrue(
+        result.getInstitutionDomainMap().get("Broad Institute").contains("broadinstitute.org"));
```
pshapiro4broad: You could use hamcrest here
```suggestion
    assertThat(result.getInstitutionDomainMap(),
        allOf(aMapWithSize(2), hasEntry("Broad Institute", Set.of("broadinstitute.org", "broad.mit.edu"))));
```

src/test/java/org/broadinstitute/consent/http/cloudstore/GCSServiceTest.java
```diff
@@ -139,4 +140,30 @@ void testDeleteDocument() {
...
+        """;
+
+    String blobIdName = "institution-domain-mapping.json";
+    Blob blob = mock(Blob.class);
+    when(blob.getContent()).thenReturn(testJson.getBytes());
+    when(storage.get(any(BlobId.class))).thenReturn(blob);
+
+    initStore();
+
+    InstitutionDomainMap result = service.readJsonFileFromBucket(blobIdName,
```
pshapiro4broad: This is probably OK but it does conflate testing the `InstitutionDomainMap`'s serialization with the GCS service testing. If the type changes, then this test may need to change. You could use a locally defined class instead if you wanted to isolate this test from `InstitutionDomainMap`

src/test/java/org/broadinstitute/consent/http/models/InstitutionDomainMapTest.java
```diff
@@ -0,0 +1,83 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import com.google.gson.Gson;
+import java.util.Map;
+import java.util.Set;
+import org.junit.jupiter.api.Test;
+
+public class InstitutionDomainMapTest {
```
pshapiro4broad, line 12: `public` is unnecessary for class and member functions in JUnit 5
```suggestion
class InstitutionDomainMapTest {
```

src/test/java/org/broadinstitute/consent/http/models/InstitutionDomainMapTest.java
```diff
@@ -0,0 +1,83 @@
...
+    // set a test map
+    Map<String, Set<String>> testMap = Map.of(
+        "Broad Institute", Set.of("broadinstitute.org", "broad.mit.edu"),
+        "Harvard", Set.of("harvard.edu", "hms.harvard.edu")
+    );
+    map.setInstitutionDomainMap(testMap);
+
+    assertEquals(testMap, map.getInstitutionDomainMap());
+    assertEquals(2, map.getInstitutionDomainMap().size());
+    assertEquals(2, map.getInstitutionDomainMap().get("Broad Institute").size());
```
pshapiro4broad, line 30: I find hamcrest better for this, but for some it can be harder to read (and if you're not familiar with it, harder to write), but you could instead write
```suggestion
    assertThat(map.getInstitutionDomainMap(), 
        allOf(aMapWithSize(2), hasEntry(is("Broad Institute"), hasSize(2))));
```

pshapiro4broad (approved):
Looks OK

src/main/java/org/broadinstitute/consent/http/cloudstore/GCSService.java
```diff
@@ -138,6 +143,23 @@ public Map<BlobId, InputStream> getDocuments(List<BlobId> blobIds) throws NotFou
...
 
+  /**
+   * Read JSON file from GCS bucket
+   *
+   * @param blobIdName String value of the document blob id name
+   * @param valueType  Class type of the object to be read
+   * @return Object of the specified type
+   */
+  public <T> T readJsonFileFromBucket(String blobIdName, Class<T> valueType) throws IOException {
+    try (InputStream is = getDocument(blobIdName)) {
```
pshapiro4broad, line 154: It's not part of this PR, but looking at the rest of this class, it's not using `Optional` correctly. It's building an `Optional` using `Optional.of()`, not `Optional.ofNullable()`. If you use `.of()`, the contents can't be `null` and are always present.

I don't think the GCS APIs will return `null` because if a blob isn't found an exception is thrown. If that's true, `Optional` can be removed from this class.

src/test/java/org/broadinstitute/consent/http/cloudstore/GCSServiceTest.java
fboulnois: The rest of the code uses the `junit` Assertions, so I will use the same style.

rushtong (approved):
Looks reasonable to me. I do think we need a generic tech-debt/cleanup ticket to cover incorrect `Optional.of()` usages since that is out of the scope of this ticket.

rjohanek (approved):
looks good thanks for talking through this!

PR #2479, opened 2025-04-11
[DT-1501]Remove NIH Intramural designation in DAR given RAS update

rjohanek:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1501
Check collaborator provides a loophole to avoid RAS authentication, we don't want that anymore.

### Summary
Remove unused fields from DataAccessRequestData. They are now unused in the UI, given https://github.com/DataBiosphere/duos-ui/pull/2825 

### Testing
Manually tested that old DARs can still be loaded in the UI and no longer contain this information (although it still exists in the database).
Unit tested serialization does not include removed fields

----

- Verify all tests go green


src/test/java/org/broadinstitute/consent/http/models/DataAccessRequestDataTest.java
```diff
@@ -0,0 +1,23 @@
...
+    String exampleDarData = """
+        {
+          "institution": "The Broad Institute of MIT and Harvard",
+          "projectTitle": "title",
+          "checkCollaborator": false,
+          "checkNihDataOnly": false
+        }""";
+    DataAccessRequestData resultingDarData = DataAccessRequestData.fromString(exampleDarData);
+    String expectedDarData = """
+        {"projectTitle":"title","checkNihDataOnly":false}""";
```
pshapiro4broad, line 20: Is the order of properties deterministic? If not, this may fail on different systems.

The [JSONAssert](https://github.com/skyscreamer/JSONassert) library is a good way to test JSON objects, but consent doesn't use it yet. It's bundled with spring-boot so It is used by other terra services for testing.

pshapiro4broad (approved):

s-rubenstein (approved):
Agreed with Phil's comment, otherwise LGTM

src/test/java/org/broadinstitute/consent/http/models/DataAccessRequestDataTest.java
rushtong: I'm pretty sure Jackson can do this too. Not sure which is better, but if we have the capability already, I'd prefer to use what we have.

PR #2481, opened 2025-04-15
[DT-1502] Enforce Library card rule on all DAR creation; return specific Error text when this rule is being enforced.

otchet-broad:
### Addresses
[https://broadworkbench.atlassian.net/browse/DT-1502](https://broadworkbench.atlassian.net/browse/DT-1502)

### Summary
Ensures DARs can only be created when a user has a library card
Removes unused endpoints that allowed creation.
Modifies tests to conform to new rule.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/exceptions/LibraryCardRequiredException.java
+++ b/src/main/java/org/broadinstitute/consent/http/exceptions/LibraryCardRequiredException.java
@@ -1,7 +1,7 @@
 package org.broadinstitute.consent.http.exceptions;
 
 public class LibraryCardRequiredException extends UnprocessableEntityException{
-  public static final String MESSAGE = "Please register in DUOS with your institution's email and obtain approval from your Signing Official in order to submit a data access request.";
+  public static final String MESSAGE = "Please register in DUOS with your institution's email and obtain a library card from your Signing Official in order to submit a data access request.";
   public LibraryCardRequiredException() {
     super(MESSAGE);
   }
--- a/src/main/java/org/broadinstitute/consent/http/exceptions/LibraryCardRequiredException.java
+++ b/src/main/java/org/broadinstitute/consent/http/exceptions/LibraryCardRequiredException.java
@@ -1,0 +1,8 @@
+package org.broadinstitute.consent.http.exceptions;
+
+public class LibraryCardRequiredException extends UnprocessableEntityException{
+  public static final String MESSAGE = "Please register in DUOS with your institution's email and obtain approval from your Signing Official in order to submit a data access request.";
+  public LibraryCardRequiredException() {
+    super(MESSAGE);
+  }
+}
--- a/src/main/java/org/broadinstitute/consent/http/exceptions/NIHComplianceRuleException.java
+++ b/src/main/java/org/broadinstitute/consent/http/exceptions/NIHComplianceRuleException.java
@@ -1,0 +1,13 @@
+package org.broadinstitute.consent.http.exceptions;
+
+import com.google.api.client.http.HttpStatusCodes;
+import jakarta.ws.rs.ClientErrorException;
+
+public class NIHComplianceRuleException extends ClientErrorException {
+
+  public static final String MESSAGE = "NIH has received, reviewed, and processed your data access request. This request is denied. Please see Guide Notice [NOT-OD-25-083](https://grants.nih.gov/grants/guide/notice-files/NOT-OD-25-083.html) for more information.";
+
+  public NIHComplianceRuleException() {
+    super(MESSAGE, HttpStatusCodes.STATUS_CODE_UNPROCESSABLE_ENTITY);
+  }
+}
--- a/src/main/java/org/broadinstitute/consent/http/models/User.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/User.java
@@ -283,6 +283,9 @@ public class User {
   }
 
   public List<LibraryCard> getLibraryCards() {
+    if (this.libraryCards == null) {
+      this.libraryCards = new ArrayList<>();
+    }
     return this.libraryCards;
   }
 
@@ -306,11 +309,11 @@ public class User {
   }
 
   public void addLibraryCard(LibraryCard card) {
-    if (Objects.isNull(this.getLibraryCards())) {
-      this.setLibraryCards(new ArrayList<>());
+    if (this.libraryCards == null) {
+      this.libraryCards = new ArrayList<>();
     }
-    if (!this.getLibraryCards().contains(card)) {
-      this.getLibraryCards().add(card);
+    if (!this.libraryCards.contains(card)) {
+      this.libraryCards.add(card);
     }
   }
 
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -418,7 +418,12 @@ public class DataAccessRequestResource extends Resource {
       @FormDataParam("ethicsApprovalRequiredFile") FormDataContentDisposition ethicsFileDetails) {
     User user = userService.findUserByEmail(authUser.getEmail());
     DataAccessRequest parentDar = dataAccessRequestService.findByReferenceId(parentReferenceId);
-    checkAuthorizedUpdateUser(user, parentDar);
+    try {
+      checkAuthorizedUpdateUser(user, parentDar);
+    } catch (Exception e) {
+      return createExceptionResponse(e);
+    }
+
     DataAccessRequest payload = populateDarFromJsonString(user, dar);
     DataAccessRequest childDar = dataAccessRequestService.createDataAccessRequest(user, payload);
 
@@ -585,6 +590,9 @@ public class DataAccessRequestResource extends Resource {
   private void checkAuthorizedUpdateUser(User user, DataAccessRequest dar) {
     if (!user.getUserId().equals(dar.getUserId())) {
       throw new ForbiddenException("User not authorized to update this Data Access Request");
+    }
+    if (Objects.isNull(user.getLibraryCards()) || user.getLibraryCards().isEmpty()) {
+      throw new NIHComplianceRuleException();
     }
   }
 
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -35,7 +35,9 @@ import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.consent.http.cloudstore.GCSService;
 import org.broadinstitute.consent.http.enumeration.DarDocumentType;
 import org.broadinstitute.consent.http.enumeration.UserRoles;
+import org.broadinstitute.consent.http.exceptions.LibraryCardRequiredException;
 import org.broadinstitute.consent.http.exceptions.NIHComplianceRuleException;
+import org.broadinstitute.consent.http.exceptions.UnprocessableEntityException;
 import org.broadinstitute.consent.http.models.AuthUser;
 import org.broadinstitute.consent.http.models.DataAccessAgreement;
 import org.broadinstitute.consent.http.models.DataAccessRequest;
@@ -592,7 +594,7 @@ public class DataAccessRequestResource extends Resource {
       throw new ForbiddenException("User not authorized to update this Data Access Request");
     }
     if (user.getLibraryCards().isEmpty()) {
-      throw new NIHComplianceRuleException();
+      throw new LibraryCardRequiredException();
     }
   }
 
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -106,7 +106,7 @@ public class DataAccessRequestResource extends Resource {
       @Auth AuthUser authUser, @Context UriInfo info, String dar) {
     try {
       User user = findUserByEmail(authUser.getEmail());
-      if (Objects.isNull(user.getLibraryCards()) || user.getLibraryCards().isEmpty()) {
+      if (user.getLibraryCards().isEmpty()) {
         throw new NIHComplianceRuleException();
       }
 
@@ -136,7 +136,7 @@ public class DataAccessRequestResource extends Resource {
       @Auth AuthUser authUser, @Context UriInfo info, String dar) {
     try {
       User user = findUserByEmail(authUser.getEmail());
-      if (Objects.isNull(user.getLibraryCards()) || user.getLibraryCards().isEmpty()) {
+      if (user.getLibraryCards().isEmpty()) {
         throw new NIHComplianceRuleException();
       }
       DataAccessRequest payload = populateDarFromJsonString(user, dar);
@@ -591,7 +591,7 @@ public class DataAccessRequestResource extends Resource {
     if (!user.getUserId().equals(dar.getUserId())) {
       throw new ForbiddenException("User not authorized to update this Data Access Request");
     }
-    if (Objects.isNull(user.getLibraryCards()) || user.getLibraryCards().isEmpty()) {
+    if (user.getLibraryCards().isEmpty()) {
       throw new NIHComplianceRuleException();
     }
   }
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -35,6 +35,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.consent.http.cloudstore.GCSService;
 import org.broadinstitute.consent.http.enumeration.DarDocumentType;
 import org.broadinstitute.consent.http.enumeration.UserRoles;
+import org.broadinstitute.consent.http.exceptions.NIHComplianceRuleException;
 import org.broadinstitute.consent.http.models.AuthUser;
 import org.broadinstitute.consent.http.models.DataAccessAgreement;
 import org.broadinstitute.consent.http.models.DataAccessRequest;
@@ -106,7 +107,7 @@ public class DataAccessRequestResource extends Resource {
     try {
       User user = findUserByEmail(authUser.getEmail());
       if (Objects.isNull(user.getLibraryCards()) || user.getLibraryCards().isEmpty()) {
-        throw new IllegalArgumentException("User must have a library card to create a DAR.");
+        throw new NIHComplianceRuleException();
       }
 
       DataAccessRequest payload = populateDarFromJsonString(user, dar);
@@ -136,7 +137,7 @@ public class DataAccessRequestResource extends Resource {
     try {
       User user = findUserByEmail(authUser.getEmail());
       if (Objects.isNull(user.getLibraryCards()) || user.getLibraryCards().isEmpty()) {
-        throw new IllegalArgumentException("User must have a library card to create a DAR.");
+        throw new NIHComplianceRuleException();
       }
       DataAccessRequest payload = populateDarFromJsonString(user, dar);
       // DAA Enforcement
--- a/src/main/java/org/broadinstitute/consent/http/resources/Resource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/Resource.java
@@ -24,6 +24,7 @@ import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.broadinstitute.consent.http.enumeration.UserRoles;
 import org.broadinstitute.consent.http.exceptions.ConsentConflictException;
+import org.broadinstitute.consent.http.exceptions.NIHComplianceRuleException;
 import org.broadinstitute.consent.http.exceptions.UnknownIdentifierException;
 import org.broadinstitute.consent.http.exceptions.UnprocessableEntityException;
 import org.broadinstitute.consent.http.models.Error;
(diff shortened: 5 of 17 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -417,7 +418,12 @@ public Response postProgressReport(
       @FormDataParam("ethicsApprovalRequiredFile") FormDataContentDisposition ethicsFileDetails) {
     User user = userService.findUserByEmail(authUser.getEmail());
     DataAccessRequest parentDar = dataAccessRequestService.findByReferenceId(parentReferenceId);
-    checkAuthorizedUpdateUser(user, parentDar);
+    try {
+      checkAuthorizedUpdateUser(user, parentDar);
+    } catch (Exception e) {
```
s-rubenstein, line 425: Can we be more specific than just `Exception`?

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -585,6 +591,9 @@ private void checkAuthorizedUpdateUser(User user, DataAccessRequest dar) {
     if (!user.getUserId().equals(dar.getUserId())) {
       throw new ForbiddenException("User not authorized to update this Data Access Request");
     }
+    if (Objects.isNull(user.getLibraryCards()) || user.getLibraryCards().isEmpty()) {
```
s-rubenstein: Out of curiosity, why `Objects.isNull` rather than `== null`?

pshapiro4broad: Or, could this field be made non-null so the caller doesn't need to check?

otchet-broad: Consistency.  Since we do a good bit of `Collection Stream` processing, it seems like we've standardized on this instead of == null even in `if` conditions.  They are effectively the same.  See: [https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html)

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
otchet-broad: We can, but we should be really wrapping the larger method in this pattern (like the other methods in this class). Since it isn't currently used, I punted on the bigger refactor for now and am catching the generic Exception as my tell.  I'm adding a note to the ticket here: https://broadworkbench.atlassian.net/browse/DT-1167 that we probably need to revisit this.

src/main/java/org/broadinstitute/consent/http/resources/Resource.java
```diff
@@ -68,67 +69,22 @@ abstract public class Resource implements ConsentLogger {
...
-              + " MB");
-    }
-  }
-
-  private interface ExceptionHandler {
-
-    Response handle(Exception e);
-  }
-
   private static final Map<Class<? extends Throwable>, ExceptionHandler> dispatch = new HashMap<>();
```
pshapiro4broad: Not sure why this is not using `Map.of()` vs a `static` block. Does it need to be mutable? Also since it's `static final` it should be all caps.

src/main/java/org/broadinstitute/consent/http/resources/Resource.java
```diff
@@ -217,6 +173,52 @@ protected static Response unableToExecuteExceptionHandler(Exception e) {
         .build();
   }
 
+  protected Response createExceptionResponse(Exception e) {
```
pshapiro4broad, line 182: Did you mean to move these? It makes it harder to tell if anything was changed.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -85,59 +70,4 @@ public Response getDatasetByIdentifier(@Auth AuthUser authUser,
     }
   }
 
-  @POST
-  @Consumes("application/json")
-  @Produces("application/json")
-  @Path("/dar/draft")
-  @PermitAll
-  @Timed
-  public Response createDraftDataAccessRequest(
```
pshapiro4broad, line 94: If you delete these, I think you also need to delete `tdrDraftDAR.yaml` and `tdrV2DraftDAR.yaml` and update `api-docs.yaml`.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -111,6 +112,9 @@ public DataAccessRequest insertDraftDataAccessRequest(User user, DataAccessReque
         || Objects.isNull(dar.getData())) {
       throw new IllegalArgumentException("User and DataAccessRequest are required");
     }
+    if (Objects.isNull(user.getLibraryCards()) || user.getLibraryCards().isEmpty()) {
```
pshapiro4broad: Is this required for a draft DAR?

src/test/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceTest.java
```diff
@@ -61,6 +61,24 @@
 @ExtendWith(MockitoExtension.class)
 class DataAccessRequestResourceTest {
 
+  private final AuthUser authUser = new AuthUser("test@test.com");
```
pshapiro4broad, line 64: Did you mean to move these fields? Maybe this is due to a IDE reformat? The google reformatter shouldn't change the order of fields.

src/test/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceTest.java
```diff
@@ -418,11 +419,27 @@ void testPostProgressReportUserNotAuthorized() throws IOException {
...
+
+  @Test
+  void testPostProgressReportCollabAndEthicsFilesWithoutLibraryCardsThrows() throws IOException {
+    when(userService.findUserByEmail(user.getEmail())).thenReturn(user);
+    DataAccessRequest parentDar = generateDataAccessRequest();
+    parentDar.getData().setCollaborationLetterLocation("collaborationLetterLocation");
+    parentDar.getData().setIrbDocumentLocation("irbDocumentLocation");
+    when(dataAccessRequestService.findByReferenceId(any())).thenReturn(parentDar);
+    Pair<InputStream, FormDataContentDisposition> collabFile = mockFormDataMultiPart("collab.txt");
+    Pair<InputStream, FormDataContentDisposition> ethicsFile = mockFormDataMultiPart("ethics.txt");
```
pshapiro4broad: This would be a nice place to use `var` IMO
```suggestion
    var collabFile = mockFormDataMultiPart("collab.txt");
    var ethicsFile = mockFormDataMultiPart("ethics.txt");
```

src/test/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceTest.java
```diff
@@ -815,7 +832,8 @@ void testCreateDataAccessRequestWithDAARestrictions() {
       fail("Initialization Exception: " + e.getMessage());
     }
 
-    try (Response response = resource.createDataAccessRequestWithDAARestrictions(authUser, info, "")) {
+    try (Response response = resource.createDataAccessRequestWithDAARestrictions(authUser, info,
+        "")) {
       assertEquals(HttpStatusCodes.STATUS_CODE_CREATED, response.getStatus());
     }
```
pshapiro4broad, line 838: It looks like many tests have the same test pattern. You could refactor the common code to a method that accepts the expected status code, to make future changes like this less work.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -85,59 +70,4 @@ public Response getDatasetByIdentifier(@Auth AuthUser authUser,
     }
   }
 
-  @POST
```
otchet-broad, line 88: This is the work @rushtong mentioned a new ticket for in standup.

src/main/java/org/broadinstitute/consent/http/resources/Resource.java
```diff
@@ -68,67 +69,22 @@ abstract public class Resource implements ConsentLogger {
       Map.entry("22021",
           ImmutablePair.of(Response.Status.BAD_REQUEST.getStatusCode(), "Invalid byte sequence"))
   );
-
-  protected Response createExceptionResponse(Exception e) {
```
otchet-broad, line 72: Reorganizing the file.  This moved.

src/test/java/org/broadinstitute/consent/http/resources/TDRResourceTest.java
```diff
@@ -120,113 +95,4 @@ void testGetDatasetByIdentifier404() {
     assertEquals(404, r.getStatus());
   }
 
-  // Created response when a new DAR draft is successful
```
otchet-broad, line 123: Tests also removed since the endpoints were removed.

src/test/java/org/broadinstitute/consent/http/service/DataAccessRequestServiceTest.java
```diff
@@ -195,6 +196,7 @@ void testGetUsersApprovedForDataset() {
   void testInsertDraftDataAccessRequest() {
     User user = new User();
     user.setUserId(1);
+    user.setLibraryCards(List.of(new LibraryCard()));
```
otchet-broad, line 200: It was expected that these tests broke because the user needs a Library Card now.

src/test/java/org/broadinstitute/consent/http/service/DataAccessRequestServiceTest.java
```diff
@@ -141,7 +142,7 @@ void testCreateDataAccessRequest_FailsIfNoLibraryCard() {
     user.setLibraryCards(List.of(new LibraryCard()));
     user.setLibraryCards(List.of());
     initService();
-    assertThrows(IllegalArgumentException.class, () -> {
+    assertThrows(NIHComplianceRuleException.class, () -> {
```
otchet-broad, line 146: New exception type being returned.

src/test/java/org/broadinstitute/consent/http/service/DataAccessRequestServiceTest.java
```diff
@@ -401,6 +408,23 @@ void testCreateDraftDarFromCanceledCollection() {
...
+    DataAccessRequestData data = new DataAccessRequestData();
+    data.setStatus(DarStatus.CANCELED.getValue());
+    dar.addDatasetId(1);
+    data.setReferenceId(UUID.randomUUID().toString());
+    dar.setData(data);
+    dar.setReferenceId(data.getReferenceId());
+    sourceCollection.addDar(dar);
+    initService();
+    assertThrows(NIHComplianceRuleException.class,
+        () -> service.createDraftDarFromCanceledCollection(user, sourceCollection));
```
otchet-broad, line 426: Ensure this fails when a user doesn't have a library card.

src/main/java/org/broadinstitute/consent/http/exceptions/NIHComplianceRuleException.java
```diff
@@ -0,0 +1,13 @@
...
+
+import com.google.api.client.http.HttpStatusCodes;
+import jakarta.ws.rs.ClientErrorException;
+
+public class NIHComplianceRuleException extends ClientErrorException {
+
+  public static final String MESSAGE = "NIH has received, reviewed, and processed your data access request. This request is denied. Please see Guide Notice [NOT-OD-25-083](https://grants.nih.gov/grants/guide/notice-files/NOT-OD-25-083.html) for more information.";
+
+  public NIHComplianceRuleException() {
+    super(MESSAGE, HttpStatusCodes.STATUS_CODE_UNPROCESSABLE_ENTITY);
```
otchet-broad, line 11: I used a 422 response here but can be convinced otherwise.
[https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/422](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/422)

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
rushtong: We've tried to move away from this more recently where we are not using streams, but this is a usage that is repeated throughout the codebase.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
rushtong: If you look at `createExceptionResponse(e)`, we try to be smart about translating various exceptions into logical HTTP response codes. This is a consistent pattern we use to trap unexpected exceptions from being returned as unclear responses.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
rushtong: This covers https://broadworkbench.atlassian.net/browse/DT-1523

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
rushtong: One thing we can do to simplify the check is to instantiate `User.libraryCards` as an empty list so it's never null. We've been doing that in other classes recently (see [Study](https://github.com/DataBiosphere/consent/blob/f0f02ea7a573021f8358fd1b4d8009c675f57974/src/main/java/org/broadinstitute/consent/http/models/Study.java#L17) for example)

rjohanek (commented):
Does this fix the bug in the UI (not displaying the error message) because now its returning and exception response instead of just throwing an exception?

src/main/java/org/broadinstitute/consent/http/exceptions/NIHComplianceRuleException.java
```diff
@@ -0,0 +1,13 @@
+package org.broadinstitute.consent.http.exceptions;
+
+import com.google.api.client.http.HttpStatusCodes;
+import jakarta.ws.rs.ClientErrorException;
+
+public class NIHComplianceRuleException extends ClientErrorException {
+
+  public static final String MESSAGE = "NIH has received, reviewed, and processed your data access request. This request is denied. Please see Guide Notice [NOT-OD-25-083](https://grants.nih.gov/grants/guide/notice-files/NOT-OD-25-083.html) for more information.";
```
rjohanek, line 8: do we have to display exactly and only this message? It seems like the old text "User must have a library card to create a DAR." seems much more helpful. Could we include both?

rushtong: There is a **must** condition in the CADR document about this message so I do think we need at least that.

rushtong (approved):
👍🏽

s-rubenstein (approved):

pshapiro4broad (approved):
Looks OK overall. If possible, I think it would be better to initialize `User.libraryCards` so callers don't need to do a null check.

otchet-broad:
@rjohanek - It does!  I'll just need to use the contents of the error message and format it as markdown.  That's part of DT-1519.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
otchet-broad: I'll confirm that we want things this way - but I'm pretty sure the answer was yes.  And that does the effective work of notifying the user in the Data Catalog that they cannot create a DAR - without having to complete the form and only be denied after.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
otchet-broad: Good catch.  Thank you!

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
rushtong: I do think we need to enforce a Library Card for draft DARs.

src/test/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceTest.java
otchet-broad: See comment about the code formatter.

src/main/java/org/broadinstitute/consent/http/resources/Resource.java
otchet-broad: This was a move by the IDE code formatter. Nothing was changed in this method.
I think we need to publish settings and fix the file that's currently checked into consent at ./idea/codeStyles/Project.xml. I'll make a ticket.

src/main/java/org/broadinstitute/consent/http/exceptions/NIHComplianceRuleException.java
otchet-broad: The UI ticket addresses some desired language on the "Apply for Access" screen that effectively blocks this form from appearing.  See: [https://github.com/DataBiosphere/duos-ui/pull/2828](https://github.com/DataBiosphere/duos-ui/pull/2828)

The other thing to note is that the DAR form already displays a warning at the top of the page that says the DAR cannot be submitted without a library card.  
image

The only time this particular message is displayed is when the user has:
1) Loaded a saved DAR
2) Suddenly regained a library card and reloaded the form.
3) Lost the gained library card in step 2.
4) Clicks submit on the DAR form so the request goes through the UI.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
otchet-broad: We have an unexpected number of tests that fail when I return an empty list.  I can spend time on that now or create a ticket.  Which does the team prefer?

otchet-broad: Got to the bottom of the issues.  User.getLibraryCards() will now always return a list.

src/main/java/org/broadinstitute/consent/http/resources/Resource.java
otchet-broad: Addressed the variable name.  Something doesn't like Map.of() and I didn't dive into why at this point.

otchet-broad:
Please see new note for a proposed minor change to the requirement in the ticket here: [https://broadworkbench.atlassian.net/browse/DT-1502](https://broadworkbench.atlassian.net/browse/DT-1502)

src/main/java/org/broadinstitute/consent/http/models/User.java
```diff
@@ -283,6 +283,9 @@ public void setLibraryCards(List<LibraryCard> cards) {
   }
 
   public List<LibraryCard> getLibraryCards() {
+    if (this.libraryCards == null) {
+      this.libraryCards = new ArrayList<>();
```
pshapiro4broad, line 287: Can this use a field initializer instead?

src/main/java/org/broadinstitute/consent/http/models/User.java
```diff
@@ -306,11 +309,11 @@ public void addProperty(UserProperty userProp) {
   }
 
   public void addLibraryCard(LibraryCard card) {
-    if (Objects.isNull(this.getLibraryCards())) {
-      this.setLibraryCards(new ArrayList<>());
+    if (this.libraryCards == null) {
+      this.libraryCards = new ArrayList<>();
```
pshapiro4broad, line 313: Not sure how consistent the consent code is, but I think using `this.` when it's not needed for disambiguation purposes makes the code harder to read
```suggestion
    if (libraryCards == null) {
     libraryCards = new ArrayList<>();
```

src/main/java/org/broadinstitute/consent/http/resources/Resource.java
pshapiro4broad: I would instead rather move consent to use spotless, and remove the IDE formatter settings from the repo. https://broadworkbench.atlassian.net/browse/DT-921

IMO reformatting is fine but allowing a reformat to reorder code like this is something we want to avoid, due to issues with code change attribution and PR diff size.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -294,6 +287,19 @@ public DataAccessRequest createDataAccessRequest(User user, DataAccessRequest da
     return findByReferenceId(referenceId);
   }
 
+  private void validateUserAndRequestObject(User user, DataAccessRequest dataAccessRequest) {
+    if (Objects.isNull(user) || Objects.isNull(dataAccessRequest) || Objects.isNull(
```
pshapiro4broad: While you're moving this code, I would also replace `isNull()` with `!= null`. And I don't see how this code can be called with null `user` or `dataAccessRequest`, so maybe some of these checks can be removed.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -294,6 +287,19 @@ public DataAccessRequest createDataAccessRequest(User user, DataAccessRequest da
...
+        dataAccessRequest.getReferenceId()) || Objects.isNull(dataAccessRequest.getData())) {
+      throw new IllegalArgumentException("User and DataAccessRequest are required");
+    }
+
+    if (user.getLibraryCards().isEmpty()) {
+      throw new NIHComplianceRuleException();
+    }
+
+    Date now = new Date();
```
pshapiro4broad: copy/paste error?
```suggestion
```

src/test/java/org/broadinstitute/consent/http/db/DarCollectionDAOTest.java
```diff
@@ -5,6 +5,7 @@
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.runners.model.MultipleFailureException.assertEmpty;
```
pshapiro4broad: unused import?
```suggestion
```

pshapiro4broad (approved):
Still looks ok, a few more comments

src/main/java/org/broadinstitute/consent/http/models/User.java
otchet-broad: The proposed change makes this use inconsistent with the pattern used throughout this class.  My opinion is that I agree with your assessment that it's easier to read the code without "this."  The open question is do I trade off the time to update the entire class, make the single change, or leave it as-is?

otchet-broad: The other thing in the back of my mind (and why I'm holding off on your other note about a test refactor) is that I'm wondering if the builder pattern would be better for some of these classes.  I'm still working through that in a mental exercise.

src/main/java/org/broadinstitute/consent/http/models/User.java
otchet-broad: There's a setter that can cause this.libraryCards to become null.

src/main/java/org/broadinstitute/consent/http/exceptions/LibraryCardRequiredException.java
```diff
@@ -0,0 +1,8 @@
+package org.broadinstitute.consent.http.exceptions;
+
+public class LibraryCardRequiredException extends UnprocessableEntityException{
+  public static final String MESSAGE = "Please register in DUOS with your institution's email and obtain a library card from your Signing Official in order to submit a data access request.";
```
otchet-broad, line 4: Updated "missing Library Card" error.

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -85,59 +70,4 @@ public Response getDatasetByIdentifier(@Auth AuthUser authUser,
...
-      @Auth AuthUser authUser,
-      @QueryParam("identifiers") String identifiers,
-      @QueryParam("projectTitle") String projectTitle) {
-    try {
-      // Ensure that the user is a registered DUOS and SAM user
-      User user = userService.findOrCreateUser(authUser);
-      if (Objects.isNull(identifiers) || identifiers.isBlank()) {
-        throw new BadRequestException("No dataset identifiers were provided");
-      } else {
-        DataAccessRequest newDar = tdrService.populateDraftDarStubFromDatasetIdentifiers(identifiers, projectTitle);
```
rushtong, line 104: This method service class is now only used in tests and can likely be removed

rushtong (approved):
Looks good, thank you. Address potential cleanup as you see fit 👍🏽

s-rubenstein (approved):

src/main/java/org/broadinstitute/consent/http/models/User.java
pshapiro4broad: > There's a setter that can cause this.libraryCards to become null.

There are no non-test use case where `libraryCards` is set to `null`. Instead of changing `User` to support tests in this way, it would be better to update the tests to not require `null`. But I can see that this might be a change for a future PR.

src/main/java/org/broadinstitute/consent/http/models/User.java
pshapiro4broad: > I'm wondering if the builder pattern would be better for some of these classes

My general philosophy is that all data classes like this should be immutable. If you need to mutate, you create a copy. Adding builders is one way to support immutable objects, it depends on the needs of the code. Often I find that the only need for builders is for tests, in which case moving the builder-type code into test classes is a good way to make that clear.

PR #2483, opened 2025-04-16
DT-1537: Ensure that study info is populated when indexing dataset terms

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1537

### Summary
Datasets, when queried in list context, do not have study information populated, so when we index a collection, we need to ensure that we have study info before publishing to the index.

This PR does some minor updates to the method signatures to clean up db lookups and simplify usages of the index methods.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
@@ -278,7 +278,12 @@ public class ElasticSearchService implements ConsentLogger {
   }
 
   public Response indexDatasets(List<Dataset> datasets, User user) throws IOException {
-    List<DatasetTerm> datasetTerms = datasets.stream().map(this::toDatasetTerm).toList();
+    // Datasets in list context may not have their study populated, so we need to ensure that is
+    // true before trying to index them in ES.
+    List<DatasetTerm> datasetTerms = datasets.stream()
+        .map(d -> datasetDAO.findDatasetById(d.getDatasetId()))
+        .map(this::toDatasetTerm)
+        .toList();
     return indexDatasetTerms(datasetTerms, user);
   }
 
--- a/src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
@@ -263,7 +263,7 @@ public class ElasticSearchService implements ConsentLogger {
    */
   public void synchronizeDatasetInESIndex(Dataset dataset, User user, boolean force) {
     if (force || dataset.getIndexedDate() != null) {
-      try (var response = indexDataset(dataset, user)) {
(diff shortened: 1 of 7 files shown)


src/test/java/org/broadinstitute/consent/http/service/ElasticSearchServiceTest.java
```diff
@@ -446,7 +449,7 @@ void testToDatasetTermNullStudyProps() {
   ArgumentCaptor<Request> request;
 
   @Test
-  void testIndexDatasets() throws IOException {
+  void testIndexDatasetTerms() throws IOException {
```
rushtong, line 452: This test was mis-named

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -278,7 +278,12 @@ public Response indexDataset(Dataset dataset, User user) throws IOException {
   }
 
   public Response indexDatasets(List<Dataset> datasets, User user) throws IOException {
-    List<DatasetTerm> datasetTerms = datasets.stream().map(this::toDatasetTerm).toList();
+    // Datasets in list context may not have their study populated, so we need to ensure that is
```
pshapiro4broad, line 281: If this code is only using the dataset IDs, I would change it to accept `List<Integer> datasetIds` instead, to make this clear. For example, `ElasticSearchService.indexStudy()` is calling this method after also calling `findDatasetsByIdList()`, resulting in duplicate dataset lookups.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -278,7 +278,12 @@ public Response indexDataset(Dataset dataset, User user) throws IOException {
...
+    // true before trying to index them in ES.
+    List<DatasetTerm> datasetTerms = datasets.stream()
+        .map(d -> datasetDAO.findDatasetById(d.getDatasetId()))
```
pshapiro4broad: If there are many datasets, isn't this inefficient? Why not use `DatasetDAO.findDatasetsByIdList()` instead?

pshapiro4broad (approved):
Looks OK, some possible improvements

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
rushtong: It is inefficient, but that method doesn't populate study information. Adding study info to `findDatasetsByIdList()` significantly degrades the performance of the query.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
rushtong: Yes, I think this will simplify some other usages as well 👍🏽

snf2ye (approved):

PR #2484, opened 2025-04-16
[DT-1487]Req. 10 - Enable Int. Collab. Approvals via DAR when LCA = Yes

rjohanek:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1487

### Summary
On DAR submission, add validation of Internal Collaborators. 
Internal Collaborators must:
1. Exist in DUOS
2. Have a Library Card (aka SO approval)
3. Have the same institution as the submitter
Returns error response if this is not the case

For approved DARs, adds the Internal Collaborators to the returned approved users list.

### Testing
Adds/Updates unit tests

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/DataAccessRequestData.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/DataAccessRequestData.java
@@ -557,11 +557,9 @@ public class DataAccessRequestData {
   }
 
   public List<Collaborator> getLabAndInternalCollaborators() {
-   List<Collaborator> empty = Collections.emptyList();
-    return Stream.of(Objects.requireNonNullElse(labCollaborators, empty),
-            Objects.requireNonNullElse(internalCollaborators, empty))
+    return Stream.of(getInternalCollaborators(), getLabCollaborators())
         .flatMap(List::stream)
-        .collect(Collectors.toList());
+        .toList();
   }
 
   public List<Collaborator> getExternalCollaborators() {
--- a/src/main/java/org/broadinstitute/consent/http/models/DataAccessRequestData.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/DataAccessRequestData.java
@@ -9,6 +9,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 @JsonInclude(Include.NON_NULL)
 public class DataAccessRequestData {
@@ -553,6 +554,14 @@ public class DataAccessRequestData {
   public void setInternalCollaborators(
       List<Collaborator> internalCollaborators) {
     this.internalCollaborators = internalCollaborators;
+  }
+
+  public List<Collaborator> getLabAndInternalCollaborators() {
+   List<Collaborator> empty = Collections.emptyList();
+    return Stream.of(Objects.requireNonNullElse(labCollaborators, empty),
+            Objects.requireNonNullElse(internalCollaborators, empty))
+        .flatMap(List::stream)
+        .collect(Collectors.toList());
   }
 
   public List<Collaborator> getExternalCollaborators() {
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -37,7 +37,6 @@ import org.broadinstitute.consent.http.enumeration.DarDocumentType;
 import org.broadinstitute.consent.http.enumeration.UserRoles;
 import org.broadinstitute.consent.http.exceptions.LibraryCardRequiredException;
 import org.broadinstitute.consent.http.exceptions.NIHComplianceRuleException;
-import org.broadinstitute.consent.http.exceptions.UnprocessableEntityException;
 import org.broadinstitute.consent.http.models.AuthUser;
 import org.broadinstitute.consent.http.models.DataAccessAgreement;
 import org.broadinstitute.consent.http.models.DataAccessRequest;
@@ -108,9 +107,6 @@ public class DataAccessRequestResource extends Resource {
       @Auth AuthUser authUser, @Context UriInfo info, String dar) {
     try {
       User user = findUserByEmail(authUser.getEmail());
-      if (Objects.isNull(user.getLibraryCards()) || user.getLibraryCards().isEmpty()) {
-        throw new IllegalArgumentException("User must have a library card to create a DAR.");
-      }
 
       DataAccessRequest payload = populateDarFromJsonString(user, dar);
       DataAccessRequest newDar = dataAccessRequestService.createDataAccessRequest(user, payload);
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -413,7 +413,9 @@ public class DataAccessRequestResource extends Resource {
     User user = userService.findUserByEmail(authUser.getEmail());
     DataAccessRequest parentDar = dataAccessRequestService.findByReferenceId(parentReferenceId);
     try {
-      checkAuthorizedUpdateUser(user, parentDar);
+      if (!user.getUserId().equals(parentDar.getUserId())) {
+        throw new ForbiddenException("User not authorized to update this Data Access Request");
+      }
     } catch (Exception e) {
       return createExceptionResponse(e);
     }
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -1,6 +1,7 @@
 package org.broadinstitute.consent.http.resources;
 
 import com.google.cloud.storage.BlobId;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import com.google.inject.Inject;
 import io.dropwizard.auth.Auth;
@@ -36,12 +37,14 @@ import org.broadinstitute.consent.http.cloudstore.GCSService;
 import org.broadinstitute.consent.http.enumeration.DarDocumentType;
 import org.broadinstitute.consent.http.enumeration.UserRoles;
 import org.broadinstitute.consent.http.models.AuthUser;
+import org.broadinstitute.consent.http.models.Collaborator;
 import org.broadinstitute.consent.http.models.DataAccessAgreement;
 import org.broadinstitute.consent.http.models.DataAccessRequest;
 import org.broadinstitute.consent.http.models.DataAccessRequestData;
 import org.broadinstitute.consent.http.models.DataUse;
 import org.broadinstitute.consent.http.models.Dataset;
 import org.broadinstitute.consent.http.models.Error;
+import org.broadinstitute.consent.http.models.LibraryCard;
 import org.broadinstitute.consent.http.models.User;
 import org.broadinstitute.consent.http.service.DaaService;
 import org.broadinstitute.consent.http.service.DataAccessRequestService;
@@ -110,6 +113,7 @@ public class DataAccessRequestResource extends Resource {
       }
 
       DataAccessRequest payload = populateDarFromJsonString(user, dar);
+      validateInternalCollaborators(payload, user);
       DataAccessRequest newDar = dataAccessRequestService.createDataAccessRequest(user, payload);
       Integer collectionId = newDar.getCollectionId();
       try {
@@ -123,6 +127,27 @@ public class DataAccessRequestResource extends Resource {
       return Response.created(uri).entity(newDar.convertToSimplifiedDar()).build();
     } catch (Exception e) {
       return createExceptionResponse(e);
+    }
+  }
+
+  @VisibleForTesting
+  public void validateInternalCollaborators(DataAccessRequest payload, User requestingUser) {
+    Integer institution = requestingUser.getInstitutionId();
+    List<Collaborator> internalCollaborators = payload.getData().getInternalCollaborators();
+    if (internalCollaborators != null && !internalCollaborators.isEmpty()) {
+      for (Collaborator collaborator : internalCollaborators) {
+        // throws not found exception
+        User collabUser = findUserByEmail(collaborator.getEmail());
+        if (!Objects.equals(collabUser.getInstitutionId(), institution)) {
+          throw new BadRequestException(
+              "Collaborator " + collaborator.getEmail() + " is not part of the same institution");
+        }
+        List<LibraryCard> libraryCards = collabUser.getLibraryCards();
+        if (libraryCards == null || libraryCards.isEmpty()) {
+          throw new BadRequestException(
+              "Collaborator " + collaborator.getEmail() + " does not have a library card.");
+        }
+      }
     }
   }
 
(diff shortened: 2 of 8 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -110,6 +113,7 @@ public Response createDataAccessRequest(
       }
 
       DataAccessRequest payload = populateDarFromJsonString(user, dar);
+      validateInternalCollaborators(payload, user);
```
rjohanek: the errors thrown in this method will be caught and returned as error responses with their respective response codes (404 and 400) in createExceptionResponse

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -126,6 +130,27 @@ public Response createDataAccessRequest(
     }
   }
 
+  @VisibleForTesting
+  public void validateInternalCollaborators(DataAccessRequest payload, User requestingUser) {
+    Integer institution = requestingUser.getInstitutionId();
+    List<Collaborator> internalCollaborators = payload.getData().getInternalCollaborators();
+    if (internalCollaborators != null && !internalCollaborators.isEmpty()) {
+      for (Collaborator collaborator : internalCollaborators) {
```
pshapiro4broad: Looking at `DataAccessRequestData.getInternalCollaborators()`, `internalCollaborators` will never be `null`. Also there's no need to check for an empty list since the `for` loop won't be executed if the list is empty.
```suggestion
    for (Collaborator collaborator : internalCollaborators) {
```

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -126,6 +130,27 @@ public Response createDataAccessRequest(
...
+        // throws not found exception
+        User collabUser = findUserByEmail(collaborator.getEmail());
+        if (!Objects.equals(collabUser.getInstitutionId(), institution)) {
+          throw new BadRequestException(
+              "Collaborator " + collaborator.getEmail() + " is not part of the same institution");
```
pshapiro4broad: Should the message also include both institutions, to help the user understand what the mismatch is?

fboulnois (approved):
👍

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
pshapiro4broad: I see two other code paths which end up calling `DataAccessRequestService.createDataAccessRequest()`. `createDataAccessRequest()` already does some validations, should these validations be done there as well, so that all uses of `createDataAccessReqeust()` are validated using these rules?

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -126,6 +130,27 @@ public Response createDataAccessRequest(
...
+        }
+        List<LibraryCard> libraryCards = collabUser.getLibraryCards();
+        if (libraryCards == null || libraryCards.isEmpty()) {
```
pshapiro4broad: I think `libraryCards` can't be `null` here.
```suggestion
        if (libraryCards.isEmpty()) {
```

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequestData.java
```diff
@@ -555,6 +556,14 @@ public void setInternalCollaborators(
     this.internalCollaborators = internalCollaborators;
   }
 
+  public List<Collaborator> getLabAndInternalCollaborators() {
```
pshapiro4broad, line 562: With this change, you can update `DarDecisionMetrics.setCountUniqueUser()` to use `getLabAndInternalCollaborators()` instead of both `getLabCollaborators()` and `getInternalCollaborators()`. 

Once that's done, you can remove `getLabCollaborators()` and `getInternalCollaborators()` since they're no longer used.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequestData.java
```diff
@@ -555,6 +556,14 @@ public void setInternalCollaborators(
...
+   List<Collaborator> empty = Collections.emptyList();
+    return Stream.of(Objects.requireNonNullElse(labCollaborators, empty),
+            Objects.requireNonNullElse(internalCollaborators, empty))
+        .flatMap(List::stream)
+        .collect(Collectors.toList());
```
pshapiro4broad: I would use `List.of()` instead of `Collections.emptyList()`. Also `List.of()` doesn't allocate an object so there's no need to create a local variable for it. And `.toList()` is a better way to write `.collect(Collectors.toList())`.
```suggestion
    return Stream.of(Objects.requireNonNullElse(labCollaborators, List.of()),
            Objects.requireNonNullElse(internalCollaborators, List.of())
        .flatMap(List::stream)
        .toList();
```

Also these fields should be fixed so they're never `null`, but that might be a change for a future PR.

src/test/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceTest.java
```diff
@@ -162,6 +165,99 @@ void testCreateDataAccessRequest() {
     assertEquals(Status.CREATED.getStatusCode(), response.getStatus());
   }
 
+  @NotNull
```
pshapiro4broad: Why did you add `@NotNull` to these methods? We don't use this information for compile time checks, and the IDE itself can already infer that the methods return non-null values.

src/test/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceTest.java
```diff
@@ -162,6 +165,99 @@ void testCreateDataAccessRequest() {
...
+  private User getRequestingUser() {
```
pshapiro4broad: Why is this `get` instead of `create`? I would use `create` for a method that creates an object.

src/test/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceTest.java
```diff
@@ -162,6 +165,99 @@ void testCreateDataAccessRequest() {
...
+    data.setInternalCollaborators(internalCollaborators);
+    DataAccessRequest dar = new DataAccessRequest();
+    dar.setData(data);
+    return dar;
+  }
+
+  @Test
+  void testValidateInternalCollaboratorsNone() {
+    User requestingUser = getRequestingUser();
+    DataAccessRequest dar = getDataAccessRequest(Collections.emptyList());
```
pshapiro4broad:
```suggestion
    DataAccessRequest dar = getDataAccessRequest(List.of());
```

src/test/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceTest.java
```diff
@@ -162,6 +165,99 @@ void testCreateDataAccessRequest() {
...
+    User requestingUser = getRequestingUser();
+    Collaborator invalidCollaborator = getCollaborator();
+    DataAccessRequest dar = getDataAccessRequest(List.of(invalidCollaborator));
+    when(userService.findUserByEmail(invalidCollaborator.getEmail())).thenThrow(
+        new NotFoundException("Unable to find User with the provided email: " + invalidCollaborator.getEmail()));
+
+    initResource();
+    NotFoundException exception = assertThrows(NotFoundException.class, () ->
+        resource.validateInternalCollaborators(dar, requestingUser));
+    assertEquals(exception.getMessage(), "Unable to find User with the provided email: " + invalidCollaborator.getEmail());
```
pshapiro4broad: If we want to validate that the exception we throw above is caught here, I would make that explicit in the test.
```
    NotFoundException thrown = new NotFoundException);
    when(userService.findUserByEmail(invalidCollaborator.getEmail())).thenThrow(thrown);
   ...
    NotFoundException caught = assertThrows(NotFoundException.class, () ->
        resource.validateInternalCollaborators(dar, requestingUser));
    assertEquals(caught, thrown);
```
The message itself is unimportant and doesn't affect the code behavior, so adding one adds unnecessary complexity to the test.

pshapiro4broad (approved):
Looks OK, a few possible code cleanups

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
rjohanek: I could definitely include the requesting user's institution but is it too revealing to include the collaborator's institution?

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
rjohanek: yeah that's a good idea

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
rjohanek: Why not? I see nothing on the User guaranteeing it can't be null

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequestData.java
rjohanek: I updated the usage in darDecisionMetrics but getInternalCollaborators is still used in the validation code

rushtong: As a side note here, `DarDecisionMetrics` and the underlying service code is in need of significant rework and is likely reporting incorrect information. I'm pretty sure we have some very old tickets around this.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequestData.java
rjohanek: I needed the type to be List<Collaborator> instead of List<?> which is why I created the variable, but if I just use the getters it resolves the type issue and the null case

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -108,9 +107,6 @@ public Response createDataAccessRequest(
       @Auth AuthUser authUser, @Context UriInfo info, String dar) {
     try {
       User user = findUserByEmail(authUser.getEmail());
-      if (user.getLibraryCards().isEmpty()) {
```
rjohanek, line 111: removing this check here because it is done in dataAccessRequestService.createDataAccessRequest(user, payload)

rjohanek: There is also a redundant check in createDataAccessRequestWithDAARestrictions and postProgressReport. I could remove those too. Does that sound right, @otchet-broad ?

otchet-broad: The change here looks good because this method is wrapped with a try/catch that handles the error thrown by dataAccessRequestService.createDataAccessRequest.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -138,9 +133,6 @@ public Response createDataAccessRequestWithDAARestrictions(
...
```
otchet-broad, line 141: This looks good as well.  It's wrapped, and also calling createDataAccessRequest.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
otchet-broad: @rjohanek I don't see a change for the postProgressReport in this PR.  Am I missing it?

rjohanek: @otchet-broad I just removed it!

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -421,7 +413,9 @@ public Response postProgressReport(
     User user = userService.findUserByEmail(authUser.getEmail());
     DataAccessRequest parentDar = dataAccessRequestService.findByReferenceId(parentReferenceId);
     try {
-      checkAuthorizedUpdateUser(user, parentDar);
+      if (!user.getUserId().equals(parentDar.getUserId())) {
+        throw new ForbiddenException("User not authorized to update this Data Access Request");
+      }
```
otchet-broad, line 418: Line 424 needs to be wrapped in the same try/catch logic so the correct response is returned to the client.  It would be the right time to refactor this to put the entire method in a try/catch that returns the correct response and remove the tightly wrapped try/catches that aren't needed.

otchet-broad: Not for committing exactly, but here's an attempt to be more explicit:
```suggestion
      try {
      User user = userService.findUserByEmail(authUser.getEmail());
      DataAccessRequest parentDar = dataAccessRequestService.findByReferenceId(parentReferenceId);
      checkAuthorizedUpdateUser(user, parentDar);
      DataAccessRequest payload = populateDarFromJsonString(user, dar);
      DataAccessRequest childDar = dataAccessRequestService.createDataAccessRequest(user, payload);

      for (Integer datasetId : childDar.getDatasetIds()) {
        Dataset dataset = datasetService.findDatasetById(datasetId);
        if (dataset == null) {
          throw new NotFoundException("Dataset " + datasetId + " not found");
        }
        DataUse dataUse = dataset.getDataUse();
        if (dataUse == null || dataUse.getCollaboratorRequired() == null
            || dataUse.getEthicsApprovalRequired() == null) {
          throw new BadRequestException("Dataset " + datasetId + " is missing data use(s)");
        }
        if (dataUse.getCollaboratorRequired()) {
          String parentCollabLocation = parentDar.getData().getCollaborationLetterLocation();
          if ((collabFileDetails == null || collabFileDetails.getSize() <= 0)
              && Strings.isNullOrEmpty(parentCollabLocation)) {
            throw new BadRequestException("Collaboration document is required");
          }
          childDar = updateDarWithDocumentContents(DarDocumentType.COLLABORATION, user, childDar,
              collabInputStream, collabFileDetails);
        }
        if (dataUse.getEthicsApprovalRequired()) {
          String parentEthicsLocation = parentDar.getData().getIrbDocumentLocation();
          if ((ethicsFileDetails == null || ethicsFileDetails.getSize() <= 0)
              && Strings.isNullOrEmpty(parentEthicsLocation)) {
            throw new BadRequestException("Ethics approval document is required");
          }
          childDar = updateDarWithDocumentContents(DarDocumentType.IRB, user, childDar,
              ethicsInputStream, ethicsFileDetails);
        }
      }
      return Response.ok(childDar.convertToSimplifiedDar()).build();
    } catch (Exception e) {
      return createExceptionResponse(e);
    }
```

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
pshapiro4broad: The code is recent so maybe your branch is behind, the current code in `User` looks like
```
  public List<LibraryCard> getLibraryCards() {
    if (this.libraryCards == null) {
      this.libraryCards = new ArrayList<>();
    }
    return this.libraryCards;
  }
```

otchet-broad (approved):
Nice work!

PR #2485, opened 2025-04-16
DT-1497: Add code to auto-set institution on user creation

pshapiro4broad:
### Addresses

Institution is no longer selected by user but is assigned based on a user's email address and the DUOS configured domain-to-institution mapping.

### Summary

Add code to auto-set institution on user creation based on the domain to institution configuration. Remove code paths where institution is changed after a user is created.

Many files were changed due to refactoring and API changes. Interesting changes are in `UserDAO` `InstitutionDomainMap`, `InstitutionService`, `UserService`.

### Behavior changes

When a user is created, their institution is assigned automatically based on email, if possible. If no institution is found, it will remain `null` until an admin user assigns it, or until an admin or signing official adds the user to a role (and, for a signing official, the signing official's institution also matches the user's email).

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
@@ -122,12 +122,14 @@ public interface UserDAO extends Transactional<UserDAO> {
       """)
   List<User> findUsersByEmailList(@BindList("emails") List<String> emails);
 
-  @SqlUpdate("INSERT INTO users (email, display_name, create_date) values (:email, :displayName, :createDate)")
+  @SqlUpdate("INSERT INTO users (email, display_name, institution_id, create_date) values (:email, :displayName, :institutionId, :createDate)")
   @GetGeneratedKeys
   Integer insertUser(@Bind("email") String email,
       @Bind("displayName") String displayName,
+      @Bind("institutionId") Integer institutionId,
       @Bind("createDate") Date createDate);
 
+  // Only used by tests; all users except Admin have an institution ID set on creation.
   @SqlUpdate("UPDATE users SET display_name=:displayName, institution_id=:institutionId WHERE user_id=:id")
   void updateUser(@Bind("displayName") String displayName,
       @Bind("id") Integer id,
--- a/src/main/java/org/broadinstitute/consent/http/enumeration/UserFields.java
+++ b/src/main/java/org/broadinstitute/consent/http/enumeration/UserFields.java
@@ -1,26 +1,20 @@
 package org.broadinstitute.consent.http.enumeration;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public enum UserFields {
-  ERA_EXPIRATION_DATE("eraExpiration", false),
-  ERA_STATUS("eraAuthorized", false),
-  SELECTED_SIGNING_OFFICIAL_ID("selectedSigningOfficialId", false),
-  SUGGESTED_SIGNING_OFFICIAL("suggestedSigningOfficial", false),
-  SUGGESTED_INSTITUTION("suggestedInstitution", false),
-  DAA_ACCEPTANCE("daaAcceptance", false);
+  ERA_EXPIRATION_DATE("eraExpiration"),
+  ERA_STATUS("eraAuthorized"),
+  SELECTED_SIGNING_OFFICIAL_ID("selectedSigningOfficialId"),
+  SUGGESTED_SIGNING_OFFICIAL("suggestedSigningOfficial"),
+  DAA_ACCEPTANCE("daaAcceptance");
 
-  public static final String LIBRARY_CARDS = "libraryCards";
-  public static final String LIBRARY_CARD_ENTRIES = "libraryCardEntries";
   private final String value;
-  private final Boolean required;
 
-  UserFields(String value, Boolean required) {
+  UserFields(String value) {
     this.value = value;
-    this.required = required;
   }
 
   public String getValue() {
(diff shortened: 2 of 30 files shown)


src/main/java/org/broadinstitute/consent/http/enumeration/UserFields.java
```diff
@@ -1,56 +1,27 @@
 package org.broadinstitute.consent.http.enumeration;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public enum UserFields {
-  ERA_EXPIRATION_DATE("eraExpiration", false),
```
pshapiro4broad, line 9: Cleanups along the way: removed `required` and unused methods.

src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
```diff
@@ -248,4 +241,40 @@ private DataAccessRequest createDataAccessRequest(Integer userId, Integer collec
     return dataAccessRequestDAO.findByReferenceId(referenceId);
   }
 
+  User createUserWithRoleInDac(Integer roleId, Integer dacId) {
```
pshapiro4broad, line 244: These methods were duplicated in subclasses; this change puts the shared code in the base class.

src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
```diff
@@ -122,12 +122,14 @@ WHERE LOWER(u.email) ILIKE ANY (array[<emails>])
...
 
-  @SqlUpdate("INSERT INTO users (email, display_name, create_date) values (:email, :displayName, :createDate)")
+  @SqlUpdate("INSERT INTO users (email, display_name, institution_id, create_date) values (:email, :displayName, :institutionId, :createDate)")
   @GetGeneratedKeys
   Integer insertUser(@Bind("email") String email,
       @Bind("displayName") String displayName,
+      @Bind("institutionId") Integer institutionId,
       @Bind("createDate") Date createDate);
 
+  // Only used by tests; all users except Admin have an institution ID set on creation.
```
pshapiro4broad, line 132: I could remove this API code completely changing tests to create two users: one to create the institute and another that uses the institute. Right now tests use one user for both purposes.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -163,33 +160,38 @@ public User updateUserFieldsById(UserUpdateFields userUpdateFields, Integer user
   }
 
   public void insertRoleAndInstitutionForUser(UserRole role, Integer institutionId,
-      Integer userId) {
+      User user) {
+    var userId = user.getUserId();
     try {
+      Institution institution = institutionService.findInstitutionForEmail(user.getEmail());
+      if (!institution.getId().equals(institutionId)) {
```
snf2ye: I'm probably just not understanding the order of operations here, but we expect for institution to be null if there isn't a matching institution, so shouldn't we handle that case here instead of throwing an NPE?

snf2ye (approved):

src/main/java/org/broadinstitute/consent/http/service/UserService.java
pshapiro4broad: This code path is only used for the admin or signing official operation where a role is being assigned to a user. So in practice there should always be an institution for the user's email at this point; if not, the user should be blocked from being updated with this role.

But I agree, failing with an NPE is not great. I'll check for a `null` institution and return a 400 type error.

src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
```diff
@@ -122,12 +122,14 @@ WHERE LOWER(u.email) ILIKE ANY (array[<emails>])
       """)
   List<User> findUsersByEmailList(@BindList("emails") List<String> emails);
 
-  @SqlUpdate("INSERT INTO users (email, display_name, create_date) values (:email, :displayName, :createDate)")
+  @SqlUpdate("INSERT INTO users (email, display_name, institution_id, create_date) values (:email, :displayName, :institutionId, :createDate)")
   @GetGeneratedKeys
   Integer insertUser(@Bind("email") String email,
       @Bind("displayName") String displayName,
+      @Bind("institutionId") Integer institutionId,
       @Bind("createDate") Date createDate);
```
fboulnois, line 130: Since in most case this refactor sets `institutionId` to be `null`, would it make sense to create a `insertUserWithInstitution` instead make `insertUser` not take an institution as the default? (or removing the bind and explicitly setting `institutionId` to `null` in the query for clarity?)

src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
```diff
@@ -86,6 +93,27 @@ public Institution findInstitutionById(Integer id) {
...
+          HttpStatusCodes.STATUS_CODE_SERVER_ERROR, e);
+    }
+  }
+
+  public Institution findInstitutionForEmail(String email) {
+    String name = getInstitutionDomainMap().getInstitutionForEmail(email);
+    if (name != null) {
+      var institutions = institutionDAO.findInstitutionsByName(name);
+      if (institutions.size() == 1) {
+        return institutions.get(0);
```
fboulnois, line 111: If we have multiple matches, should we throw an error instead? We've agreed that this case should never happen, so I think it would make sense to encode in our system.

src/main/java/org/broadinstitute/consent/http/models/InstitutionDomainMap.java
```diff
@@ -18,4 +18,13 @@ public void setInstitutionDomainMap(Map<String, Set<String>> institutionDomainMa
   public Set<String> getDomainsForInstitution(String institution) {
     return institutionDomainMap.get(institution);
   }
+
+  public String getInstitutionForEmail(String email) {
+    String domain = email.substring(email.indexOf('@') + 1);
+    return institutionDomainMap.entrySet().stream()
```
fboulnois, line 24: If this structure is inconvenient for searching, should we make an additional (reverse) mapping internally from `domain` -> `institution`?

fboulnois (approved):
A few comments:

src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
pshapiro4broad: All the cases that use `null` are in test code, and I'd rather not add test-only code to the deployment.

I should be able to refactor test uses so there's only one or two cases where `null` is passed in.

src/main/java/org/broadinstitute/consent/http/models/InstitutionDomainMap.java
pshapiro4broad: Since the structure isn't persisted across API calls there's no benefit in creating a reverse map, as the time to create it will match the linear search done here.

If we decide to change this to use an in-memory cache, I agree that creating a reverse map would be worthwhile.

src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
pshapiro4broad: I agree that it should be checked, but if it's checked here, that means we won't know about the problem until a user whose domain has this issue tries to create an account.

If we wanted to catch it earlier, we'd want to check for duplicates whenever the JSON data changes. Maybe we should have an upload/deploy process that checks the data before copying to the bucket? For now it could be a simple python script.

PR #2487, opened 2025-04-17
DT-1146 - Exclude expired DARS when retrieving approved DARs by Dataset Id

snf2ye:
https://broadworkbench.atlassian.net/browse/DT-1147

### Addresses
If a DAR was submitted more than a year ago, it is expired and therefore it should not be returned with the rest of the approved DARs.

### Summary
- Do not include DARS with a submission date of more than a year ago when querying for approved DARs

**Note**: This impacts the DARS returned in 3 spots:
 - Elastic Search Service, index Datasets
 - getDatasetMetricsData endpoint
 - getApprovedUsers endpoint

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
@@ -81,6 +81,7 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
           WHERE dar.draft = false
           AND final_access_vote.last_vote = TRUE
           AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)
+          AND dar.submission_date > now() - interval '1 year'
       """)
   List<DataAccessRequest> findApprovedDARsByDatasetId(@Bind("datasetId") Integer datasetId);
 
--- a/src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
@@ -349,16 +349,17 @@ class DataAccessRequestDAOTest extends DAOTestHelper {
 
   // local method to create a DAR
   protected DataAccessRequest createDAR(User user, Dataset dataset, String darCode, Timestamp submissionDate) {
+    var now = new Timestamp(new Date().getTime());
     Integer collectionId = darCollectionDAO.insertDarCollection(darCode, user.getUserId(),
-        submissionDate);
+        now);
     DataAccessRequest testDar = new DataAccessRequest();
     testDar.setCollectionId(collectionId);
     testDar.setReferenceId(UUID.randomUUID().toString());
     testDar.setUserId(user.getUserId());
-    testDar.setCreateDate(submissionDate);
-    testDar.setSortDate(submissionDate);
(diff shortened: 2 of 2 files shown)


src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
```diff
@@ -596,6 +601,39 @@ void testFindAllApprovedDataAccessRequestsByDatasetId_ApprovedThenDeniedCase() {
             .size());
   }
 
+
+  @ParameterizedTest
+  @ValueSource(longs = {200, 370})
+  void testFindAllApprovedDataAccessRequestsByDatasetId_ExpiredDARCase(long submissionDaysAgo) {
```
rjohanek, line 608: this looks good, but would it be clearer if instead of a parameterized test, you just create both of the dars and then just check that only the one you expect is returned?

rjohanek (approved):

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
snf2ye: I like the parameterized test in this case because we can clearly see that with all of the same setup, one returns a DAR and the other doesn't, with the only submission date changing.

rushtong: Can we add a test for when the submission date is null?

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -81,6 +81,7 @@ AND LOWER(v.type) = 'final') final_access_vote ON final_access_vote.reference_id
           WHERE dar.draft = false
           AND final_access_vote.last_vote = TRUE
           AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)
+          AND dar.submission_date > now() - interval '1 year'
```
rushtong, line 84: Love this!

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
snf2ye: Sure! What do we want the behavior to be if the submission date is null? Should it be included or excluded in these results?

rushtong: I'm pretty sure the null case would be a data problem, but if it is, it should be excluded from the results.

snf2ye:
@rushtong  - I want to highlight that this change will impact the DARs returned in a couple of other places. Is this an issue?

> Note: This impacts the DARS returned in 3 spots:
> 
> 1. Elastic Search Service, index Datasets
> 2. getDatasetMetricsData endpoint
> 3. getApprovedUsers endpoint

rushtong:
> @rushtong - I want to highlight that this change will impact the DARs returned in a couple of other places. Is this an issue?
> 
> > Note: This impacts the DARS returned in 3 spots:
> > 
> > 1. Elastic Search Service, index Datasets
> > 2. getDatasetMetricsData endpoint
> > 3. getApprovedUsers endpoint

That looks exactly right, thank you!

rushtong (approved):
👍🏽

PR #2488, opened 2025-04-17
DT-1518: Populate AuthUser with full Consent User object

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1518

### Summary
This PR adds a second principal to Dropwizard's `@Auth` annotation. We have the original `AuthUser` object which is essentially unchanged, and we now have a `DuosUser` option that is guaranteed to have a populated `User` object. This required some refactoring of the existing auth filters and follows the pattern in [Dropwizard's documentation for multiple principals](https://www.dropwizard.io/en/stable/manual/auth.html#multiple-principals-and-authenticators). 

A ton of common code necessary for both authorization and authentication was pulled into `AuthorizationHelper`

Additionally, as an example of implementation, `InstitutionResource` is refactored to use this new auth scheme.

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
```diff
@@ -67,42 +71,47 @@ private AuthUser buildAuthUserFromHeaders(Map<String, String> headers) {
...
    */
   private AuthUser getUserWithStatusInfo(AuthUser authUser) {
     if (authUser == null || authUser.getEmail() == null) {
       logWarn("AuthUser/email is null, cannot get user status info");
       return null;
     }
+    try {
+      authUser.setUser(userService.findUserByEmail(authUser.getEmail()));
+    } catch (Exception e) {
+      logException("Error finding Consent user: " + authUser.getEmail(), e);
```
pshapiro4broad: This will occur the first time a user logs into duos? So is it an exception or an expected code path?

If it's expected, I think we'd use `logInfo()` here instead of `logException()`, which uses ERROR logging.

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
```diff
@@ -67,42 +71,47 @@ private AuthUser buildAuthUserFromHeaders(Map<String, String> headers) {
...
+    }
     try {
       UserStatusInfo userStatusInfo = samService.getRegistrationInfo(authUser);
       if (Objects.nonNull(userStatusInfo)) {
         // safety check in case the call to generic user (i.e. Google) failed.
-        if (Objects.isNull(authUser.getEmail())) {
+        if (authUser.getEmail() == null) {
```
pshapiro4broad: If it's possible that the user email is `null` here, wouldn't we want to call `findUserByEmail()` after the Sam lookup?

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
```diff
@@ -67,42 +71,47 @@ private AuthUser buildAuthUserFromHeaders(Map<String, String> headers) {
...
-      Gson gson = new Gson();
       try {
         // Try to post the user to Sam if they have not registered previously
         UserStatus userStatus = samService.postRegistrationInfo(authUser);
-        if (Objects.nonNull(userStatus) && Objects.nonNull(userStatus.getUserInfo())) {
+        if ((userStatus != null) && (userStatus.getUserInfo() != null)) {
           authUser.setEmail(userStatus.getUserInfo().getUserEmail());
         } else {
+          Gson gson = GsonUtil.gsonBuilderWithAdapters().create();
           logWarn("Error posting to Sam, AuthUser not able to be registered: " + gson.toJson(authUser));
```
pshapiro4broad: Do we need to use `gson` for the purposes of logging `authuser`? Is this a logging message we will need to look at, and does it have information we wouldn't already have in the logs from the ingress proxy?

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
```diff
@@ -67,42 +71,47 @@ private AuthUser buildAuthUserFromHeaders(Map<String, String> headers) {
...
     } catch (NotFoundException e) {
-      Gson gson = new Gson();
       try {
         // Try to post the user to Sam if they have not registered previously
         UserStatus userStatus = samService.postRegistrationInfo(authUser);
-        if (Objects.nonNull(userStatus) && Objects.nonNull(userStatus.getUserInfo())) {
+        if ((userStatus != null) && (userStatus.getUserInfo() != null)) {
           authUser.setEmail(userStatus.getUserInfo().getUserEmail());
         } else {
+          Gson gson = GsonUtil.gsonBuilderWithAdapters().create();
```
pshapiro4broad: This is the same as 
```suggestion
          Gson gson = GsonUtil.buildGson();
```

src/main/java/org/broadinstitute/consent/http/models/AuthUser.java
```diff
@@ -20,7 +22,7 @@ public AuthUser(String email) {
   }
 
   public AuthUser deepCopy() {
```
pshapiro4broad: Why is this method needed? It looks like it's only used in `getUserWithStatusInfo()`, and in that case `authUser` is still being constructed so it should be OK to modify it rather than copying it.

fboulnois (approved):
👍

fboulnois (commented):
Do we have to be concerned about logging sensitive information?

src/main/java/org/broadinstitute/consent/http/models/AuthUser.java
```diff
@@ -69,4 +71,12 @@ public AuthUser setUserStatusInfo(UserStatusInfo userStatusInfo) {
     this.userStatusInfo = userStatusInfo;
     return this;
   }
+
+  public User getUser() {
```
pshapiro4broad: I was thinking that, in practice, a caller will know that either `User` must exist or not when this method is called. For example, `UserResource.updateSelf()` won't work unless the user exists in the database. So the goal is to `@Auth` inject an object that _must_ have the user provided. If the user isn't present, a `NotFoundException` should be thrown before its use.

One way to do this is to have two different `AuthUser` types. One type is used in the case where we don't know if the duos user account exists. The other is used when it must exist, and is checked on initialization. Otherwise the caller will still need to check for a non-`null` `user` every time.

pshapiro4broad (approved):
This is OK but I think more could be done to address code duplication.

Before merging up, I'd like to see at least one case where this new code is used so we can discuss how this change will affect the code.

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
rushtong: Gson, in this context, necessary for the current log method to work without changes. At present, I don't know if proxy logs have enough information to debug this issue.

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
rushtong: Yes, good point. I see that the comment here is pretty outdated as well - at one point in time, we queried Google for the user's information, but that isn't the case now.

rushtong:
> Do we have to be concerned about logging sensitive information?

We log email pretty consistently, and more recently, it's a required field, so I think we're safe with what we're logging in this class.

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
rushtong: I'll revisit this block. I think, in this situation, I prefer `logWarn()`. `logException()` triggers a sentry alert which is probably more alerting than we want.

src/main/java/org/broadinstitute/consent/http/models/AuthUser.java
rushtong: I'm going to explore this a little bit, thanks for the suggestion.

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
pshapiro4broad: What I'm asking here is, do we need to use `gson` to extract the information needed for logging? We have the object itself, can we instead log the important fields directly without first converting to JSON? Or is the JSON output needed in the logs for some other reason?

rushtong: We certainly could log this differently than we're doing now with selected information. I don't remember the history behind why we're logging the full object. The only reason for gson here would have been to do just that though, log the full object.

src/main/java/org/broadinstitute/consent/http/models/AuthUser.java
rushtong: This is now unused from refactoring and deleted.

src/main/java/org/broadinstitute/consent/http/ConsentApplication.java
```diff
@@ -249,17 +249,21 @@ public void run(ConsentConfiguration config, Environment env) {
...
-        .buildAuthFilter();
-    List<AuthFilter> filters = List.of(
-        defaultAuthFilter,
-        new OAuthCustomAuthFilter(authenticator, userRoleDAO));
+    // Requests annotated with @Auth AuthUser will be authenticated with this filter
+    final AuthFilter<String, AuthUser> primaryAuthFilter = new OAuthCustomAuthFilter<>(authenticator, userRoleDAO);
+    // Requests annotated with @Auth DUOSAuthUser will be authenticated with this filter and are guaranteed to have a populated User object
+    final AuthFilter<String, DUOSAuthUser> duosAuthUserFilter = new OAuthCustomAuthFilter<>(authenticator, userRoleDAO);
+    final PolymorphicAuthDynamicFeature<AuthUser> feature = new PolymorphicAuthDynamicFeature<>(
+        ImmutableMap.of(
```
pshapiro4broad: Instead of google APIs, I think we should use JDK APIs where possible. In this case the code can use `Map.of()` and `Set.of()` instead.
```suggestion
        Map.of(
```

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
```diff
@@ -64,49 +71,32 @@ private AuthUser buildAuthUserFromHeaders(Map<String, String> headers) {
...
         // Try to post the user to Sam if they have not registered previously
         UserStatus userStatus = samService.postRegistrationInfo(authUser);
-        if (Objects.nonNull(userStatus) && Objects.nonNull(userStatus.getUserInfo())) {
-          authUser.setEmail(userStatus.getUserInfo().getUserEmail());
-        } else {
-          logWarn("Error posting to Sam, AuthUser not able to be registered: " + gson.toJson(authUser));
-        }
+        // If we succeed, return a basic version of UserStatusInfo. Future API calls will
+        // return the full UserStatusInfo object.
+        return new UserStatusInfo()
```
pshapiro4broad: Does DUOS need `UserStatusInfo`? Could this code instead return `UserStatus`? The difference I see is that Info has `userSubjectId` and `adminEnabled` but I don't see these used inside consent.

src/main/java/org/broadinstitute/consent/http/models/DUOSAuthUser.java
```diff
@@ -0,0 +1,21 @@
...
+public class DUOSAuthUser extends AuthUser {
+
+  private final User user;
+
+  public DUOSAuthUser(AuthUser authUser, User user) {
+    super.setAuthToken(authUser.getAuthToken());
+    super.setEmail(authUser.getEmail());
+    super.setName(authUser.getName());
+    super.setAud(authUser.getAud());
+    super.setUserStatusInfo(authUser.getUserStatusInfo());
```
pshapiro4broad: Since these methods are not overloaded there's no need to use `super` here. But I would instead add a constructor to `AuthUser` so the fields aren't initialized using setters. That would let us make the fields `final` as the only other construction case is in `buildAuthUserFromHeaders()`, which could also pass all arguments.

src/main/java/org/broadinstitute/consent/http/models/DUOSAuthUser.java
```diff
@@ -0,0 +1,21 @@
+package org.broadinstitute.consent.http.models;
+
+
+public class DUOSAuthUser extends AuthUser {
```
pshapiro4broad: I think it's good that this is a subclass of `AuthUser`, to make migration easier. IMO calling it `DUOSUser` would also make sense, as a DUOS user is a kind of authenticated user, so `DUOSAuthUser` seems redundant.

Also note that the [google java style guidelines](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case) specify that this name should be `DuosAuthUser`. I realize that this isn't how it's been done before, but it's something to consider if we want to adopt those guidelines for DUOS as it's been done for other java terra services.

src/test/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticatorTest.java
```diff
@@ -56,10 +66,9 @@ void testAuthenticateGetUserInfoSuccess() {
     headerMap.put(ClaimsCache.OAUTH2_CLAIM_email, List.of("email"));
     headerMap.put(ClaimsCache.OAUTH2_CLAIM_name, List.of("name"));
     headerCache.loadCache(bearerToken, headerMap);
-    oAuthAuthenticator = new OAuthAuthenticator(samService);
+
     Optional<AuthUser> authUser = oAuthAuthenticator.authenticate(bearerToken);
-    assertTrue(authUser.isPresent());
-    assertNotNull(authUser.get().getEmail());
+    assertNotNull(authUser.orElseThrow().getEmail());
     assertNotNull(authUser.get().getAuthToken());
```
pshapiro4broad: To avoid calling `get()` again you could use `orElseThrow()` earlier
```
    AuthUser authUser = oAuthAuthenticator.authenticate(bearerToken).orElseThrow();
    assertNotNull(authUser.getEmail());
    assertNotNull(authUser.getAuthToken());
```

pshapiro4broad (approved):
Nice to see you got this working, and it's good to see an example of its use as well

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
rushtong: IIRC, the only usages of this are related to Terms of Service acceptance. It is very likely that we could revisit this.

src/main/java/org/broadinstitute/consent/http/authentication/DuosUserAuthorizer.java
```diff
@@ -0,0 +1,36 @@
...
+  public boolean authorize(DuosAuthUser user, String role, ContainerRequestContext context) {
+    boolean authorize = false;
+    if (StringUtils.isNotEmpty(role)) {
+      List<String> roles = userRoleDAO.findRoleNamesByUserEmail(user.getEmail());
+      List<String> existentRole = roles.stream()
+          .filter(r -> r.equalsIgnoreCase(role))
+          .collect(Collectors.toCollection(ArrayList::new));
+      if (CollectionUtils.isNotEmpty(existentRole)) {
+        authorize = true;
+      }
```
pshapiro4broad: Since you don't use the contents of `existentRole`, you can replace `filter(x).collect() .. isNotEmpty()` with `anyMatch(x)`

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
rushtong: Going to file this as a separate ticket to look into the right Sam entity to be looking at. The current implementation precedes quite a bit of Sam refactoring and all of the newer ToS revisions.
Ticket: https://broadworkbench.atlassian.net/browse/DT-1604

src/main/java/org/broadinstitute/consent/http/authentication/AuthenticatorHelper.java
rushtong, line 1: The bulk of this class was copied from `OAuthAuthenticator`

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
```diff
@@ -1,112 +1,41 @@
...
+        }
       }
-      return Optional.of(userWithStatus);
+      return Optional.of(authUser);
     }
     logException(new ServerErrorException("Error reading request headers", 500));
     return Optional.empty();
   }
-
-  private AuthUser buildAuthUserFromHeaders(Map<String, String> headers) {
```
rushtong, line 47: Moved to `AuthorizationHelper` so it can be reused by `DuosUserAuthenticator`

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
```diff
@@ -1,112 +1,41 @@
...
-        .setName(name);
-  }
-
-  /**
-   * Attempt to get the registration status of the current user and set the value on AuthUser
-   *
-   * @param authUser The AuthUser
-   * @return A cloned AuthUser with Sam registration status
-   */
-  private AuthUser getUserWithStatusInfo(AuthUser authUser) {
```
rushtong, line 72: Moved to `AuthorizationHelper` so it can be reused by `DuosUserAuthenticator`

src/main/java/org/broadinstitute/consent/http/authentication/AuthorizerHelper.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+public class AuthorizerHelper {
+
+  private final UserRoleDAO userRoleDAO;
+
+  protected AuthorizerHelper(UserRoleDAO userRoleDAO) {
+    this.userRoleDAO = userRoleDAO;
+  }
+
+  protected boolean authorize(AuthUser user, String role) {
```
rushtong: Copied from `UserAuthorizer` so it can be reused by `DuosUserAuthorizer`

src/main/java/org/broadinstitute/consent/http/models/DuosUser.java
```diff
@@ -0,0 +1,24 @@
...
+public class DuosUser extends AuthUser {
+
+  private final User user;
+
+  public DuosUser(AuthUser authUser, User user) {
+    super(authUser.getAuthToken(),
+        authUser.getEmail(),
+        authUser.getName(),
+        authUser.getAud(),
+        authUser.getUserStatusInfo());
```
pshapiro4broad: Another approach here would be to add a constructor for `AuthUser` that accepts `AuthUser`.

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
```diff
@@ -1,112 +1,41 @@
...
-        return Optional.of(user);
+      AuthUser authUser = buildAuthUserFromHeaders(headers);
+      if (authUser.getEmail() != null) {
+        authUser.setUserStatusInfo(getUserStatusInfo(authUser));
+        try {
+          User duosUser = userService.findUserByEmail(authUser.getEmail());
+          return Optional.of(new DuosUser(authUser, duosUser));
+        } catch (NotFoundException e) {
+          logWarn("User not found, authentication incomplete: %s".formatted(authUser.getEmail()));
+        }
```
pshapiro4broad: Is this needed now? If a `User` is required, you can use `DuosUser`. If it's not, you can use `AuthUser`. Until the code migrates to using `DuosUser`, we can continue to look up `User` manually on API calls.

src/main/java/org/broadinstitute/consent/http/authentication/AuthorizerHelper.java
```diff
@@ -0,0 +1,28 @@
...
+    boolean authorize = false;
+    if (StringUtils.isNotEmpty(role)) {
+      List<String> roles = userRoleDAO.findRoleNamesByUserEmail(user.getEmail());
+      return roles.stream().anyMatch(r -> r.equalsIgnoreCase(role));
+    }
+    return authorize;
```
pshapiro4broad: Since `role` is not declared `Nullable` in `Authorizer.getAuthorizationContext()` I don't think it can be `null` here.

Even if it were `null`, the code would still work the same and you could remove the `if`, since `String.equalsIgnoreCase()` accepts `null`.
```suggestion
    List<String> roles = userRoleDAO.findRoleNamesByUserEmail(user.getEmail());
    return roles.stream().anyMatch(r -> r.equalsIgnoreCase(role));
```

src/main/java/org/broadinstitute/consent/http/authentication/DuosUserAuthorizer.java
```diff
@@ -0,0 +1,18 @@
+package org.broadinstitute.consent.http.authentication;
+
+import io.dropwizard.auth.Authorizer;
+import jakarta.ws.rs.container.ContainerRequestContext;
+import org.broadinstitute.consent.http.db.UserRoleDAO;
+import org.broadinstitute.consent.http.models.DuosUser;
+
+public class DuosUserAuthorizer extends AuthorizerHelper implements Authorizer<DuosUser> {
```
pshapiro4broad: I think these subclasses can be avoided by making `AuthorizerHelper` a generic class. So where you're writing `new DuosUserAuthorizer()` you could write `new AuthorizerHelper<DuosUser>()`.

src/main/java/org/broadinstitute/consent/http/ConsentModule.java
```diff
@@ -200,7 +201,12 @@ OntologyService providesOntologyService() {
 
   @Provides
   OAuthAuthenticator providesOAuthAuthenticator() {
-    return new OAuthAuthenticator(providesSamService());
+    return new OAuthAuthenticator(providesSamService(), providesUserService());
```
pshapiro4broad: I was expecting that this code path would no longer need `UserService` now that we have `DuosUser`.

src/main/java/org/broadinstitute/consent/http/models/DuosUser.java
```diff
@@ -0,0 +1,24 @@
+package org.broadinstitute.consent.http.models;
+
+/**
+ * This class is used to represent an authenticated user that exists as a User in the Consent
+ * system. It extends the AuthUser class which is not guaranteed to have a user associated with it.
+ */
+public class DuosUser extends AuthUser {
```
pshapiro4broad, line 7: I wonder if it's a good idea to subclass here after all. I was thinking it would help with migration, where you could pass `DuosUser` instead of `AuthUser` without changing the code. But having it as a field could work too, and then we wouldn't need to copy the data into the base class, we could embed the `AuthUser` passed to the constructor.

src/main/java/org/broadinstitute/consent/http/resources/InstitutionResource.java
```diff
@@ -78,16 +72,15 @@ public Response getInstitution(@Auth AuthUser authUser, @PathParam("id") Integer
...
     try {
-      User user = userService.findUserByEmail(authUser.getEmail());
       Institution payload = GsonUtil.getInstance().fromJson(institution, Institution.class);
       List<Institution> conflicts = institutionService.findAllInstitutionsByName(payload.getName());
       if (!conflicts.isEmpty()) {
         throw new ConsentConflictException(
             "An institution exists with the name of '" + payload.getName() + "'");
       }
-      Institution newInstitution = institutionService.createInstitution(payload, user.getUserId());
+      Institution newInstitution = institutionService.createInstitution(payload, duosUser.getUser().getUserId());
```
pshapiro4broad, line 83: Maybe not worth doing now, but if `getUser().getUserId()` is a common pattern you could add `getUserId()` to `DuosUser`.

src/test/java/org/broadinstitute/consent/http/authentication/AuthenticatorTest.java
```diff
@@ -131,17 +141,45 @@ void testAuthenticateGetUserWithStatusInfoIncompleteClaims() throws Exception {
...
+  void testAuthenticateGetUserInfoWithDuosUser() {
+    headerMap.put(ClaimsCache.OAUTH2_CLAIM_access_token, List.of(bearerToken));
+    headerMap.put(ClaimsCache.OAUTH2_CLAIM_email, List.of("email"));
+    headerMap.put(ClaimsCache.OAUTH2_CLAIM_name, List.of("name"));
+    headerCache.loadCache(bearerToken, headerMap);
+    when(userService.findUserByEmail(headerMap.get(ClaimsCache.OAUTH2_CLAIM_email).get(0))).thenReturn(new User());
+
+    AuthUser authUser = oAuthAuthenticator.authenticate(bearerToken).orElseThrow();
+    assertInstanceOf(DuosUser.class, authUser);
+    assertNotNull(((DuosUser) authUser).getUser());
```
pshapiro4broad: if you kept the `user` passed to `thenReturn()` you could use `assertEquals()` here.

pshapiro4broad (approved):
Looks good overall, glad to see you got this working.

src/main/java/org/broadinstitute/consent/http/authentication/OAuthAuthenticator.java
rushtong: Good point - we don't need this here any longer

src/main/java/org/broadinstitute/consent/http/authentication/AuthorizerHelper.java
rushtong: Thanks, I'll add some tests for that.

src/main/java/org/broadinstitute/consent/http/authentication/DuosUserAuthorizer.java
rushtong: I like that idea. That might let me simplify the other abstract class too.

src/main/java/org/broadinstitute/consent/http/ConsentModule.java
rushtong: Once I remove the user lookup in `OAuthAuthenticator`, I can remove this as well.

src/main/java/org/broadinstitute/consent/http/authentication/DuosUserAuthorizer.java
rushtong: I'm refactoring so the two abstract classes are instead a single helper class that has all of the common functions.

src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
```diff
@@ -0,0 +1,104 @@
...
+
+  protected Cache<String, Map<String, String>> getCache() {
+    return claimsCache.cache;
+  }
+
+  protected AuthUser buildAuthUserFromHeaders(Map<String, String> headers) {
+    String aud = headers.get(ClaimsCache.OAUTH2_CLAIM_aud);
+    String token = headers.get(ClaimsCache.OAUTH2_CLAIM_access_token);
+    String email = headers.get(ClaimsCache.OAUTH2_CLAIM_email);
+    String name = headers.get(ClaimsCache.OAUTH2_CLAIM_name);
```
otchet-broad, line 43: Do we ever really use this value?  I'm wondering if we should get rid of `name` and instead use our User object's displayName property.

rushtong: We do use this. When a user is first signing in, we don't have a display name. We use this to do that initial population of the user's display name. Subsequently, users can update this in their profile.

src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
```diff
@@ -0,0 +1,104 @@
...
+    // Name is not a guaranteed header
+    if (name == null || name.equals("unknown")) {
```
otchet-broad, line 45: Given the name is not a guaranteed header, maybe we shouldn't even take the value if one is provided - because it doesn't seem like we rely on it.

src/main/java/org/broadinstitute/consent/http/authentication/UserAuthorizer.java
```diff
@@ -2,35 +2,18 @@
...
-      List<String> roles = userRoleDAO.findRoleNamesByUserEmail(user.getEmail());
-      List<String> existentRole = roles.stream()
-          .filter(r -> r.equalsIgnoreCase(role))
-          .collect(Collectors.toCollection(ArrayList::new));
-      if (CollectionUtils.isNotEmpty(existentRole)) {
-        authorize = true;
-      }
-    }
-    return authorize;
+    return authorizationHelper.authorize(user, role);
```
rushtong, line 17: Moved to `AuthorizationHelper` so it can be shared by `DuosUserAuthorizer`

src/main/java/org/broadinstitute/consent/http/models/AuthUser.java
```diff
@@ -15,13 +18,24 @@ public class AuthUser implements Principal {
...
   }
 
-  public AuthUser deepCopy() {
-    Gson gson = new Gson();
-    return gson.fromJson(gson.toJson(this), AuthUser.class);
+  public AuthUser(String authToken, String email, String name, String aud,
+      UserStatusInfo userStatusInfo) {
+    this.authToken = authToken;
+    this.email = email;
+    this.name = name;
```
otchet-broad, line 40: Same consideration might apply here about getting rid of the `name` here.

src/main/java/org/broadinstitute/consent/http/models/DuosUser.java
rushtong: My first approach was using a field. I tend to like this approach better as it opens us up for different kinds of users that might be an `AuthUser`. We don't have the use case now, but I could imagine a `ServiceAccountUser` that needs special treatment, like logging, for example.

src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
rushtong: Name is guaranteed for Google authentications and only not guaranteed for MS authentications. Google is the majority of our authentications and this does give us valuable info to improve the initial UX.

src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
rushtong: I should add, all of the methods in this class are essentially copied from the previous implementation so they can be re-used across the two Auth schemes. I'm hesitant to make too many changes here that might have unexpected and hard-to-detect regressions.

src/main/java/org/broadinstitute/consent/http/models/AuthUser.java
rushtong: Name is required to implement `Principal`

otchet-broad (approved):

PR #2489, opened 2025-04-18
DT-1497: Additional test cleanups from PR comments

pshapiro4broad:
### Addresses

Remove test-only API code from `UserDAO`, and remove many code duplications in database test code setup.

### Summary

----

- Verify all tests go green


src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
```diff
@@ -271,10 +247,13 @@ User createUserWithInstitution() {
...
-    userDAO.updateUser(name, userId, institutionId);
-    userRoleDAO.insertSingleUserRole(7, userId);
-    return userDAO.findUserById(userId);
+    User user = createUserWithRole(UserRoles.SIGNINGOFFICIAL.getRoleId(), institutionId);
+    return userDAO.findUserById(user.getUserId());
+  }
+
+  protected Institution getUserInstitution(User user) {
+    return institutionDAO.findInstitutionById(user.getInstitutionId());
   }
```
rushtong, line 258: In general, we've been trying to actively remove shared methods from this class. It has, at times, become an uber-service class and contributed to test confusion. In this case, all DAO tests have access to `institutionDAO` directly and can make this call.

rushtong (approved):
Cleanup looks pretty good, thank you 👍🏽

fboulnois (approved):
cleanups looks good 👍

src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
pshapiro4broad: > In general, we've been trying to actively remove shared methods from this class. It has, at times, become an uber-service class and contributed to test confusion.

What is the goal of removing shared methods from this class? If a shared method isn't here, are you saying we should instead duplicate code in our tests? Even in test code, duplication is a maintenance burden and differences can lead to inconsistencies across test classes.

>  In this case, all DAO tests have access to institutionDAO directly and can make this call.

The only reason I added this is because `createUserWithInstitution()` creates both an institution and a user and sometimes both are needed. `createUserWithInstitution()` has it but doesn't return it, maybe a better fix would be to make a method that returns both for those cases.

s-rubenstein (approved):

src/test/java/org/broadinstitute/consent/http/db/DAOTestHelper.java
rushtong: In the past, we've had many variations of `createUser` with and without a variety of things like roles, institution, required properties, missing properties, etc. That let to more sprawl with create entity X with Y and not Z, etc.

PR #2491, opened 2025-04-21
[DT-1520] DAR Expiration.

otchet-broad:
### Addresses
[https://broadworkbench.atlassian.net/browse/DT-1520](https://broadworkbench.atlassian.net/browse/DT-1520)
### Summary
Consent had two state machines for determining if a Data Access Request was in draft form or submitted for a Data Access Committee to consider for approval.  This PR aligns the code so that only one state machine exists.  It is now centered on submission_date.

When submission_date is non-null, the DAR has been submitted by a RESEARCHER to a DATA ACCESS COMMITTEE for consideration.  When the submission_date is null, it is considered a draft.  This was the behavior in the existing code paths except for some tests that leveraged methods to short circuit that requirement. 

The DataAccessRequest class now computes the fields `draft`, `expired`, and `expiredAt` fields when the submissionDate is set.  These three values are now returned to the UI for enrichment.

~~Of **critical importance** is the update to `DataAccessRequestDAO.findApprovedDARsByDatasetId` that now requires DARs to be submitted within the last year in order to be considered "approved."~~ <- This was already completed in [https://github.com/DataBiosphere/consent/pull/2487](https://github.com/DataBiosphere/consent/pull/2487)

Also added in `DataAccessRequestDAO` is `findSubmittedDarsByTimeRange`. This method is intended to be used by a to-be-built notification component that can alert users to expiring DARs.  It allows the caller to specify the date range of approved DARs so that notifications can be sent.  For example, give me the list of DARs that will expire 30 days from today.  Those DARs would have a submission_date during the time of 1745193600.000 to 1745279999.999 

To simplify our codebase, other test only methods that were impacting draft state were removed.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
@@ -86,7 +86,7 @@ public interface DarCollectionDAO extends Transactional<DarCollectionDAO> {
           "dd.dataset_id, " +
           "dar.id AS dar_id, dar.reference_id AS dar_reference_id, dar.collection_id AS dar_collection_id, "
           +
-          "dar.parent_id AS dar_parent_id, dar.draft AS dar_draft, dar.user_id AS dar_userId, " +
+          "dar.parent_id AS dar_parent_id, dar.user_id AS dar_userId, " +
           "dar.create_date AS dar_create_date, dar.sort_date AS dar_sort_date, dar.submission_date AS dar_submission_date, "
           +
           "dar.update_date AS dar_update_date, (dar.data #>> '{}')::jsonb AS data " +
@@ -127,10 +127,11 @@ public interface DarCollectionDAO extends Transactional<DarCollectionDAO> {
           UserProperty.QUERY_FIELDS_WITH_UP_PREFIX + QUERY_FIELD_SEPARATOR +
           "dar.id AS dar_id, dar.reference_id AS dar_reference_id, dar.collection_id AS dar_collection_id, "
           +
-          "dar.parent_id AS dar_parent_id, dar.draft AS dar_draft, dar.user_id AS dar_userId, " +
+          "dar.parent_id AS dar_parent_id, dar.user_id AS dar_userId, " +
           "dar.create_date AS dar_create_date, dar.sort_date AS dar_sort_date, dar.submission_date AS dar_submission_date, "
           +
-          "dar.update_date AS dar_update_date, (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data, dd.dataset_id " +
+          "dar.update_date AS dar_update_date, (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data, dd.dataset_id "
+          +
           "FROM dar_collection c " +
           "INNER JOIN users u ON c.create_user_id = u.user_id " +
           "LEFT JOIN user_property up ON u.user_id = up.user_id " +
@@ -165,7 +166,7 @@ public interface DarCollectionDAO extends Transactional<DarCollectionDAO> {
           + LibraryCard.QUERY_FIELDS_WITH_LC_PREFIX + QUERY_FIELD_SEPARATOR
           + "dd.dataset_id, "
           + "dar.id AS dar_id, dar.reference_id AS dar_reference_id, dar.collection_id AS dar_collection_id, "
-          + "dar.parent_id AS dar_parent_id, dar.draft AS dar_draft, dar.user_id AS dar_userId, "
+          + "dar.parent_id AS dar_parent_id, dar.user_id AS dar_userId, "
           + "dar.create_date AS dar_create_date, dar.sort_date AS dar_sort_date, dar.submission_date AS dar_submission_date, "
           + "dar.update_date AS dar_update_date, (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data, "
           + "e.election_id AS e_election_id, e.reference_id AS e_reference_id, e.status AS e_status, e.create_date AS e_create_date, "
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
@@ -121,26 +121,28 @@ public interface DarCollectionDAO extends Transactional<DarCollectionDAO> {
   @RegisterBeanMapper(value = UserProperty.class, prefix = "up")
   @UseRowReducer(DarCollectionReducer.class)
   @SqlQuery(
-      "SELECT c.*, " +
-          User.QUERY_FIELDS_WITH_U_PREFIX + QUERY_FIELD_SEPARATOR +
-          Institution.QUERY_FIELDS_WITH_I_PREFIX + QUERY_FIELD_SEPARATOR +
-          UserProperty.QUERY_FIELDS_WITH_UP_PREFIX + QUERY_FIELD_SEPARATOR +
-          "dar.id AS dar_id, dar.reference_id AS dar_reference_id, dar.collection_id AS dar_collection_id, "
-          +
-          "dar.parent_id AS dar_parent_id, dar.user_id AS dar_userId, " +
-          "dar.create_date AS dar_create_date, dar.sort_date AS dar_sort_date, dar.submission_date AS dar_submission_date, "
-          +
-          "dar.update_date AS dar_update_date, (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data, dd.dataset_id "
-          +
-          "FROM dar_collection c " +
-          "INNER JOIN users u ON c.create_user_id = u.user_id " +
-          "LEFT JOIN user_property up ON u.user_id = up.user_id " +
-          "LEFT JOIN institution i ON i.institution_id = u.institution_id " +
-          "INNER JOIN data_access_request dar ON c.collection_id = dar.collection_id " +
-          "LEFT JOIN dar_dataset dd on dd.reference_id = dar.reference_id " +
-          "WHERE c.collection_id = (SELECT collection_id FROM data_access_request WHERE reference_id = :referenceId) "
-          +
-          "AND (LOWER(data->>'status')!='archived' OR data->>'status' IS NULL) ")
+      """
+        SELECT c.*,
+        u.user_id as u_user_id, u.email as u_email, u.display_name as u_display_name, u.create_date
+        as u_create_date, u.email_preference as u_email_preference, u.institution_id
+        as u_institution_id, u.era_commons_id as u_era_commons_id, i.institution_id as i_id,
+        i.institution_name as i_name, i.it_director_name as i_it_director_name,  i.it_director_email
+        as i_it_director_email, i.create_date as i_create_date, i.update_date as i_update_date,
+        up.property_id AS up_property_id, up.user_id AS up_user_id, up.property_key
+        as up_property_key, up.property_value AS up_property_value,
+        dar.id AS dar_id, dar.reference_id AS dar_reference_id, dar.collection_id AS dar_collection_id,
+        dar.parent_id AS dar_parent_id, dar.user_id AS dar_userId,
+        dar.create_date AS dar_create_date, dar.sort_date AS dar_sort_date, dar.submission_date AS dar_submission_date,
+        dar.update_date AS dar_update_date, (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data, dd.dataset_id
+        FROM dar_collection c
+        INNER JOIN users u ON c.create_user_id = u.user_id
+        LEFT JOIN user_property up ON u.user_id = up.user_id
+        LEFT JOIN institution i ON i.institution_id = u.institution_id
+        INNER JOIN data_access_request dar ON c.collection_id = dar.collection_id
+        LEFT JOIN dar_dataset dd on dd.reference_id = dar.reference_id
+        WHERE c.collection_id = (SELECT collection_id FROM data_access_request WHERE reference_id = :referenceId)
+        AND (LOWER(data->>'status')!='archived' OR data->>'status' IS NULL)
+      """)
   DarCollection findDARCollectionByReferenceId(@Bind("referenceId") String referenceId);
 
   /**
@@ -200,9 +202,12 @@ public interface DarCollectionDAO extends Transactional<DarCollectionDAO> {
    *
    * @return Integer, ID of newly created DarCollection
    */
-  @SqlUpdate("INSERT INTO dar_collection " +
-      " (dar_code, create_user_id, create_date) " +
-      " VALUES (:darCode, :createUserId, :createDate)")
+  @SqlUpdate(
+      """
+        INSERT INTO dar_collection
+        (dar_code, create_user_id, create_date)
+        VALUES (:darCode, :createUserId, :createDate)
+      """)
   @GetGeneratedKeys
   Integer insertDarCollection(@Bind("darCode") String darCode,
       @Bind("createUserId") Integer createUserId,
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
@@ -60,26 +60,26 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
       (
           """
               SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, u.display_name as researcher_name,
-              i.institution_name, e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id as dd_datasetid,
-              (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name
+                i.institution_name, e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id as dd_datasetid,
+                (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name
               FROM dar_collection c
               INNER JOIN users u
-              ON u.user_id = c.create_user_id
+                ON u.user_id = c.create_user_id
               LEFT JOIN institution i
-              ON i.institution_id = u.institution_id
+                ON i.institution_id = u.institution_id
               INNER JOIN data_access_request dar
-              ON dar.collection_id = c.collection_id
+                ON dar.collection_id = c.collection_id
               LEFT JOIN (
-              SELECT election.*, MAX(election.election_id) OVER(PARTITION BY election.reference_id, election.dataset_id) AS latest
-              FROM election
-              WHERE LOWER(election.election_type) = 'dataaccess'
-              ) AS e
+                SELECT election.*, MAX(election.election_id) OVER(PARTITION BY election.reference_id, election.dataset_id) AS latest
+                FROM election
+                WHERE LOWER(election.election_type) = 'dataaccess'
+                ) AS e
               ON e.reference_id = dar.reference_id
               INNER JOIN dar_dataset dd
               ON dar.reference_id = dd.reference_id
               WHERE u.institution_id = :institutionId
-              AND (e.latest = e.election_id OR e.election_id IS NULL)
-              AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL )
+                AND (e.latest = e.election_id OR e.election_id IS NULL)
+                AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL )
           """
       )
   List<DarCollectionSummary> getDarCollectionSummariesForSO(
@@ -119,9 +119,9 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
       (
           """
               SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, u.display_name as researcher_name,
-              i.institution_name, e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id as dd_datasetid,
-              (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name,
-              (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'status' AS dar_status
+                i.institution_name, e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id as dd_datasetid,
+                (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name,
+                (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'status' AS dar_status
               FROM dar_collection c
               INNER JOIN users u
               ON u.user_id = c.create_user_id
@@ -130,17 +130,17 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
               INNER JOIN data_access_request dar
               ON dar.collection_id = c.collection_id
               LEFT JOIN (
-              SELECT election.*, MAX(election.election_id) OVER(PARTITION BY election.reference_id, election.dataset_id) AS latest
-              FROM election
-              WHERE LOWER(election.election_type) = 'dataaccess'
+                SELECT election.*, MAX(election.election_id) OVER(PARTITION BY election.reference_id, election.dataset_id) AS latest
+                FROM election
+                WHERE LOWER(election.election_type) = 'dataaccess'
               ) AS e
               ON e.reference_id = dar.reference_id
               INNER JOIN dar_dataset dd
               ON dar.reference_id = dd.reference_id
               WHERE c.create_user_id = :userId
-              AND (e.latest = e.election_id OR e.election_id IS NULL)
-              AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL )
-              AND (EXISTS (SELECT 1 FROM data_access_request WHERE (collection_id = c.collection_id and dar.submission_date is not null)))
+                AND (e.latest = e.election_id OR e.election_id IS NULL)
+                AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL )
+                AND (EXISTS (SELECT 1 FROM data_access_request WHERE (collection_id = c.collection_id and dar.submission_date is not null)))
         """
       )
   List<DarCollectionSummary> getDarCollectionSummariesForResearcher(
@@ -194,9 +194,9 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
       (
           """
               SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, u.display_name as researcher_name,
-              u.user_id as researcher_id, i.institution_name, i.institution_id, e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id as dd_datasetid,
-              (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name,
-              (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'status' AS dar_status
+                u.user_id as researcher_id, i.institution_name, i.institution_id, e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id as dd_datasetid,
+                (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name,
+                (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'status' AS dar_status
               FROM dar_collection c
               INNER JOIN users u
               ON u.user_id = c.create_user_id
@@ -205,16 +205,16 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
               INNER JOIN data_access_request dar
               ON dar.collection_id = c.collection_id
               LEFT JOIN (
-              SELECT election.*, MAX(election.election_id) OVER(PARTITION BY election.reference_id, election.dataset_id) AS latest
-              FROM election
-              WHERE LOWER(election.election_type) = 'dataaccess'
+                SELECT election.*, MAX(election.election_id) OVER(PARTITION BY election.reference_id, election.dataset_id) AS latest
+                FROM election
+                WHERE LOWER(election.election_type) = 'dataaccess'
               ) AS e
               ON e.reference_id = dar.reference_id
               INNER JOIN dar_dataset dd
               ON dar.reference_id = dd.reference_id
               WHERE c.collection_id = :collectionId
-              AND (e.latest = e.election_id OR e.election_id IS NULL)
-              AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL )
+                AND (e.latest = e.election_id OR e.election_id IS NULL)
+                AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL )
           """
       )
   DarCollectionSummary getDarCollectionSummaryByCollectionId(
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
@@ -62,7 +62,8 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
               +
               "i.institution_name, e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id as dd_datasetid, "
               +
-              "(regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name " +
+              "(regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name "
+              +
               "FROM dar_collection c " +
               "INNER JOIN users u " +
               "ON u.user_id = c.create_user_id " +
@@ -122,8 +123,10 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
               +
               "i.institution_name, e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id as dd_datasetid, "
               +
-              "(regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name, " +
-              "(regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'status' AS dar_status " +
+              "(regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name, "
+              +
+              "(regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'status' AS dar_status "
+              +
               "FROM dar_collection c " +
               "INNER JOIN users u " +
               "ON u.user_id = c.create_user_id " +
@@ -143,7 +146,7 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
               "WHERE c.create_user_id = :userId " +
               "AND (e.latest = e.election_id OR e.election_id IS NULL) " +
               "AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL ) " +
-              "AND (EXISTS (SELECT 1 FROM data_access_request WHERE (collection_id = c.collection_id and draft = false)))"
+              "AND (EXISTS (SELECT 1 FROM data_access_request WHERE (collection_id = c.collection_id and dar.submission_date is not null)))"
       )
   List<DarCollectionSummary> getDarCollectionSummariesForResearcher(
       @Bind("userId") Integer userId);
@@ -198,8 +201,10 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
               +
               "u.user_id as researcher_id, i.institution_name, i.institution_id, e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id as dd_datasetid, "
               +
-              "(regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name, " +
-              "(regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'status' AS dar_status " +
+              "(regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name, "
+              +
+              "(regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'status' AS dar_status "
+              +
               "FROM dar_collection c " +
               "INNER JOIN users u " +
               "ON u.user_id = c.create_user_id " +
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
@@ -58,31 +58,29 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
   @UseRowReducer(DarCollectionSummaryReducer.class)
   @SqlQuery
       (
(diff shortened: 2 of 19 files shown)


src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -35,69 +35,95 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
...
+                ) last_vote
+                FROM election e
+                INNER JOIN vote v ON e.election_id = v.electionid AND v.vote IS NOT NULL
+                WHERE e.dataset_id = :datasetId
+                AND LOWER(e.election_type) = 'dataaccess'
+                AND LOWER(v.type) = 'final') final_access_vote ON final_access_vote.reference_id = dar.reference_id
+              WHERE dar.submission_date BETWEEN SYMMETRIC now() - interval '1' year AND now()
+              AND final_access_vote.last_vote = TRUE
+              AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)
+              AND dar.submission_date > now() - interval '1 year'
```
otchet-broad: **Important change highlighted here** Approved DARs must have a submitted date within the last year, otherwise they're not approved. @rushtong

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -35,69 +35,95 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
...
+  @UseRowReducer(DataAccessRequestReducer.class)
+  @SqlQuery(
+      """
+              SELECT dar.id, dar.reference_id, dar.collection_id, dar.parent_id,
+                dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date,
+                (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data,
+                dd.dataset_id
+              FROM data_access_request dar
+              LEFT JOIN dar_dataset dd ON dd.reference_id = dar.reference_id
+              WHERE dar.submission_date BETWEEN SYMMETRIC to_timestamp(:begin) AND to_timestamp(:end)
```
otchet-broad: Now we can support time range queries for submitted DARs.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -58,30 +58,29 @@ List<DarCollectionSummary> getDarCollectionSummariesForDAC(
...
+              ON u.user_id = c.create_user_id
+              LEFT JOIN institution i
+              ON i.institution_id = u.institution_id
+              INNER JOIN data_access_request dar
+              ON dar.collection_id = c.collection_id
+              LEFT JOIN (
+              SELECT election.*, MAX(election.election_id) OVER(PARTITION BY election.reference_id, election.dataset_id) AS latest
+              FROM election
+              WHERE LOWER(election.election_type) = 'dataaccess'
+              ) AS e
```
rushtong: Minor nit: it would be nice to add a level of indentation to the inner select. Previous code didn't do that, but as a text block, it would be a lot more legible. Similar comment applies to other query reformats in this class.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
rushtong: In this query, there are now two restrictions on `submission_date`:
```
WHERE dar.submission_date BETWEEN SYMMETRIC now() - interval '1' year AND now()
```
and ... 
```
AND dar.submission_date > now() - interval '1 year'
```
Do they not both have a similar effect? I believe the second condition was originally added in https://github.com/DataBiosphere/consent/pull/2487

otchet-broad: They do.  I'm removing the one I added.

rushtong (approved):
Love the removal of `draft`. It would be interesting (not for this PR) to see if we could do something similar for the `status` field in data.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
```diff
@@ -121,25 +121,28 @@ List<DarCollection> findDARCollectionByCollectionIds(
...
-          "FROM dar_collection c " +
-          "INNER JOIN users u ON c.create_user_id = u.user_id " +
-          "LEFT JOIN user_property up ON u.user_id = up.user_id " +
-          "LEFT JOIN institution i ON i.institution_id = u.institution_id " +
-          "INNER JOIN data_access_request dar ON c.collection_id = dar.collection_id " +
-          "LEFT JOIN dar_dataset dd on dd.reference_id = dar.reference_id " +
-          "WHERE c.collection_id = (SELECT collection_id FROM data_access_request WHERE reference_id = :referenceId) "
-          +
-          "AND (LOWER(data->>'status')!='archived' OR data->>'status' IS NULL) ")
+      """
```
pshapiro4broad, line 124: It would've been nice to PR the text block change separately as it makes it hard to see the actual diff here.

Also note that changing to text block changes the string generated here. Originally it generated a string without line breaks, but as a text block there are now line breaks. This shouldn't be a problem but I've seen cases where it can mess up logging, if SQL statements are logged without removing line breaks first.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -35,69 +35,94 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
...
```
pshapiro4broad: Why is `SYMMETRIC` needed here? Is it possible that `begin` is after `end` instead of before it?

otchet-broad: Exactly.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -35,69 +35,94 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
...
+          """)
+  List<DataAccessRequest> findSubmittedDarsByTimeRange(@Bind("begin") double begin,
```
pshapiro4broad: Is there a way to write this so that it accepts a java object instead of a `double`? Maybe this is the simplest way but it seems awkward to have to convert into a FP number to send a time value to SQL.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
otchet-broad: It would have.  Unfortunately what this effort shows is that we added semgrep and it's only looking at changed files so we're finding the problems as we go.  You'll see failed CI/CD checks for runs before I had to reformat the SQL.

I'd like to discuss the idea of a ticket to have a wholesale fix for this as just a SQL formatting PR since it's a pain to switch formatting of unrelated code.  Suspecting that ticket to be back burnered, I went ahead and only changed the files that were necessary for this PR.

otchet-broad: @pshapiro4broad - I created [https://broadworkbench.atlassian.net/browse/DT-1551](https://broadworkbench.atlassian.net/browse/DT-1551) for discussion in today's standup parking lot.

s-rubenstein (commented):
Where possible, could you drop comments where the SQL text blocks are logically the same? It looks like a bunch of this is solely formatting changes but I am not sure which is functional and which is format

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
```diff
@@ -24,6 +25,9 @@
 @JsonInclude(Include.NON_NULL)
 public class DataAccessRequest {
 
+  public static final long EXPIRATION_DURATION_MILLIS = 365L * 24 * 60 * 60
+      * 1000; // 365 days/year 24 hours/day 60 minutes/hour 60 seconds/minute 1000 ms/second
```
pshapiro4broad: use `TimeUnit` to avoid math
```suggestion
  public static final long EXPIRATION_DURATION_MILLIS = TimeUnit.DAYS.toMillis(365L);
```

src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
```diff
@@ -121,25 +121,28 @@ List<DarCollection> findDARCollectionByCollectionIds(
...
+        SELECT c.*,
```
otchet-broad, line 125: Formatting.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
```diff
@@ -199,9 +202,12 @@ List<DarCollection> findDARCollectionByCollectionIds(
    *
    * @return Integer, ID of newly created DarCollection
    */
-  @SqlUpdate("INSERT INTO dar_collection " +
-      " (dar_code, create_user_id, create_date) " +
-      " VALUES (:darCode, :createUserId, :createDate)")
+  @SqlUpdate(
```
otchet-broad, line 205: Formatting

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -58,30 +58,29 @@ List<DarCollectionSummary> getDarCollectionSummariesForDAC(
...
-              "FROM election " +
-              "WHERE LOWER(election.election_type) = 'dataaccess'" +
-              ") AS e " +
-              "ON e.reference_id = dar.reference_id " +
-              "INNER JOIN dar_dataset dd " +
-              "ON dar.reference_id = dd.reference_id " +
-              "WHERE u.institution_id = :institutionId " +
-              "AND (e.latest = e.election_id OR e.election_id IS NULL) " +
-              "AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL ) "
+          """
```
otchet-broad, line 61: Formatting

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -118,32 +117,31 @@ WHERE LOWER(election.election_type) = 'dataaccess'
...
+                FROM election
+                WHERE LOWER(election.election_type) = 'dataaccess'
+              ) AS e
+              ON e.reference_id = dar.reference_id
+              INNER JOIN dar_dataset dd
+              ON dar.reference_id = dd.reference_id
+              WHERE c.create_user_id = :userId
+                AND (e.latest = e.election_id OR e.election_id IS NULL)
+                AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL )
+                AND (EXISTS (SELECT 1 FROM data_access_request WHERE (collection_id = c.collection_id and dar.submission_date is not null)))
```
otchet-broad, line 143: **switch from draft to submission date**

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -118,32 +117,31 @@ WHERE LOWER(election.election_type) = 'dataaccess'
...
-              "WHERE LOWER(election.election_type) = 'dataaccess'" +
-              ") AS e " +
-              "ON e.reference_id = dar.reference_id " +
-              "INNER JOIN dar_dataset dd " +
-              "ON dar.reference_id = dd.reference_id " +
-              "WHERE c.create_user_id = :userId " +
-              "AND (e.latest = e.election_id OR e.election_id IS NULL) " +
-              "AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL ) " +
-              "AND (EXISTS (SELECT 1 FROM data_access_request WHERE (collection_id = c.collection_id and draft = false)))"
+          """
```
otchet-broad, line 120: Formatting + see comment below.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -194,31 +192,30 @@ DarCollectionSummary getDarCollectionSummaryForDACByCollectionId(
...
-              "FROM election " +
-              "WHERE LOWER(election.election_type) = 'dataaccess'" +
-              ") AS e " +
-              "ON e.reference_id = dar.reference_id " +
-              "INNER JOIN dar_dataset dd " +
-              "ON dar.reference_id = dd.reference_id " +
-              "WHERE c.collection_id = :collectionId " +
-              "AND (e.latest = e.election_id OR e.election_id IS NULL) " +
-              "AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL )"
+          """
```
otchet-broad, line 195: Formatting

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -35,69 +36,94 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
...
-      "SELECT dd.dataset_id, dar.id, dar.reference_id, dar.collection_id, dar.parent_id, dar.draft, dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date, "
-          + "  (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data FROM data_access_request dar"
-          + "  LEFT JOIN dar_dataset dd on dd.reference_id = dar.reference_id "
-          + "  WHERE dar.draft != true "
-          + "  AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)")
+      """
+          SELECT dd.dataset_id, dar.id, dar.reference_id, dar.collection_id, dar.parent_id, dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date,
+            (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data FROM data_access_request dar
+          LEFT JOIN dar_dataset dd on dd.reference_id = dar.reference_id 
+          WHERE dar.submission_date is not null
```
otchet-broad, line 43: switching from draft to submission date

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -35,69 +36,94 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
...
-            WHERE e.dataset_id = :datasetId
-            AND LOWER(e.election_type) = 'dataaccess'
-            AND LOWER(v.type) = 'final') final_access_vote ON final_access_vote.reference_id = dar.reference_id
-          WHERE dar.draft = false
-          AND final_access_vote.last_vote = TRUE
-          AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)
-          AND dar.submission_date > now() - interval '1 year'
-      """)
+  @SqlQuery(
+      """
```
otchet-broad, line 63: remove dar.draft + formatting

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -35,69 +36,94 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
...
    */
   @UseRowReducer(DataAccessRequestReducer.class)
   @SqlQuery(
-      "SELECT dd.dataset_id, dar.id, dar.reference_id, dar.collection_id, dar.parent_id, dar.draft, dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date, "
-          + "  (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data FROM data_access_request dar"
-          + "  LEFT JOIN dar_dataset dd on dd.reference_id = dar.reference_id "
-          + "  WHERE dar.draft = true "
-          + "  AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL) "
-          + "  ORDER BY dar.update_date DESC")
+      """
```
otchet-broad, line 119: Remove dar.draft + formatting

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -107,13 +133,15 @@ AND dar.submission_date > now() - interval '1 year'
...
   @UseRowReducer(DataAccessRequestReducer.class)
   @SqlQuery(
-      "SELECT dd.dataset_id, dar.id, dar.reference_id, dar.collection_id, dar.parent_id, dar.draft, dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date, "
-          + "  (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data FROM data_access_request dar"
-          + "  LEFT JOIN dar_dataset dd on dd.reference_id = dar.reference_id "
-          + "  WHERE dar.draft = true "
-          + "  AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL) "
-          + "  AND dar.user_id = :userId "
-          + "  ORDER BY dar.sort_date DESC")
+      """
```
otchet-broad, line 136: remove dar.draft + Formatting.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -124,13 +152,15 @@ AND dar.submission_date > now() - interval '1 year'
...
   @UseRowReducer(DataAccessRequestReducer.class)
   @SqlQuery(
-      "SELECT dd.dataset_id, dar.id, dar.reference_id, dar.collection_id, dar.parent_id, dar.draft, dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date, "
-          + "  (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data FROM data_access_request dar"
-          + "  LEFT JOIN dar_dataset dd on dd.reference_id = dar.reference_id "
-          + "  WHERE dar.draft = false "
-          + "  AND dar.user_id = :userId "
-          + "  AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL) "
-          + "  ORDER BY dar.sort_date DESC")
+      """
```
otchet-broad, line 155: Remove dar.draft + formatting

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -175,10 +209,12 @@ AND dar.submission_date > now() - interval '1 year'
    */
   @RegisterArgumentFactory(JsonArgumentFactory.class)
   @SqlUpdate(
-      "UPDATE data_access_request "
-          + "SET data = to_jsonb(regexp_replace(:data, '\\\\u0000', '', 'g')), user_id = :userId, sort_date = :sortDate, "
-          + "submission_date = :submissionDate, update_date = :updateDate "
-          + "WHERE reference_id = :referenceId")
+      """
```
otchet-broad, line 212: formatting.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -192,50 +228,58 @@ void updateDataByReferenceId(
    *
    * @param referenceId String
    */
-  @SqlUpdate("DELETE FROM data_access_request WHERE reference_id = :referenceId")
+  @SqlUpdate(
```
otchet-broad, line 231: Formatting

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -192,50 +228,58 @@ void updateDataByReferenceId(
...
 
 
   @SqlUpdate("DELETE FROM data_access_request WHERE reference_id IN (<referenceIds>)")
   void deleteByReferenceIds(@BindList("referenceIds") List<String> referenceIds);
 
   @SqlUpdate(
-      "UPDATE data_access_request dar "
-          + "SET data=jsonb_set((regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb, '{status}', '\"Canceled\"') "
-          + "WHERE reference_id IN (<referenceIds>)")
+      """
```
otchet-broad, line 242: Formatting.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -192,50 +228,58 @@ void updateDataByReferenceId(
...
+      """)
   void cancelByReferenceIds(@BindList("referenceIds") List<String> referenceIds);
 
   /**
    * Delete all DataAccessRequests with the given collection id
    *
    * @param collectionId Integer
    */
-  @SqlUpdate("DELETE FROM data_access_request WHERE collection_id = :collectionId")
+  @SqlUpdate(
```
otchet-broad, line 254: Formatting.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -192,50 +228,58 @@ void updateDataByReferenceId(
...
+   * @param data        DataAccessRequestData DAR Properties
    */
   @RegisterArgumentFactory(JsonArgumentFactory.class)
   @SqlUpdate(
-      "INSERT INTO data_access_request "
-          + "(reference_id, user_id, create_date, sort_date, submission_date, update_date, data, draft) "
-          + "VALUES (:referenceId, :userId, :createDate, :sortDate, "
-          + ":submissionDate, :updateDate, to_jsonb(:data), true)")
+      """
+          INSERT INTO data_access_request
```
otchet-broad, line 273: remove draft.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
```diff
@@ -40,7 +44,13 @@ public class DataAccessRequest {
...
 
   @JsonProperty
-  public Boolean draft;
+  public Boolean draft = true;
+
+  @JsonProperty
+  public Boolean expired = false;
+
+  @JsonProperty
+  public long expiresAt = -1;
```
pshapiro4broad: I would prefer to not store computed data in a POJO. Could you instead annotate the getters for these values with `@JsonProperty`? Storing data here means that you need to recompute when their source data changes.

Or, the UI client could also compute these values, as that's the only use of them right now.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
```diff
@@ -302,6 +319,12 @@ private Map<String, Object> shallowCopy(DataAccessRequest dar) {
     if (Objects.nonNull(dar.getDraft())) {
       copy.put("draft", dar.getDraft());
     }
+    if (Objects.nonNull(dar.getExpired())) {
```
pshapiro4broad: These values can never be null so the null checks are unnecessary.

s-rubenstein (approved):

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -253,10 +297,12 @@ void insertDraftDataAccessRequest(
    */
   @RegisterArgumentFactory(JsonArgumentFactory.class)
   @SqlUpdate(
-      "INSERT INTO data_access_request "
-          + "(collection_id, reference_id, user_id, create_date, sort_date, submission_date, update_date, data) "
-          + "VALUES (:collectionId, :referenceId, :userId, :createDate, :sortDate, " +
-          ":submissionDate, :updateDate, to_jsonb(:data))")
+      """
```
otchet-broad, line 300: Formatting.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -267,44 +313,32 @@ void insertDataAccessRequest(
...
-      @Bind("draft") Boolean draft);
 
   @SqlUpdate(
-      "UPDATE data_access_request "
-          + "SET draft = false, collection_id = :collectionId "
-          + "WHERE reference_id = :referenceId")
-  void updateDraftForCollection(@Bind("collectionId") Integer collectionId,
+      """
+          UPDATE data_access_request
+            SET  submission_date = now(), collection_id = :collectionId
```
otchet-broad, line 320: remove the use of draft

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -35,69 +36,94 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
...
-          + "  (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data FROM data_access_request dar"
-          + "  LEFT JOIN dar_dataset dd on dd.reference_id = dar.reference_id "
-          + "  WHERE dar.draft != true "
-          + "  AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)")
+      """
+          SELECT dd.dataset_id, dar.id, dar.reference_id, dar.collection_id, dar.parent_id, dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date,
+            (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data FROM data_access_request dar
+          LEFT JOIN dar_dataset dd on dd.reference_id = dar.reference_id 
+          WHERE dar.submission_date is not null
+          AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)""")
```
pshapiro4broad, line 44: Order of `OR` seems odd here, wouldn't you want to check for `IS NULL` before calling `LOWER()`? Or maybe SQL is smart enough to reverse these two expressions.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -35,69 +36,94 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
...
+      """
+              SELECT dar.id, dar.reference_id, dar.collection_id, dar.parent_id,
+                dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date,
+                (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data,
+                dd.dataset_id
+              FROM data_access_request dar
+              LEFT JOIN dar_dataset dd ON dd.reference_id = dar.reference_id
+              WHERE dar.submission_date BETWEEN SYMMETRIC :begin AND :end
+          """)
+  List<DataAccessRequest> findSubmittedDarsByTimeRange(@Bind("begin") Timestamp begin,
```
pshapiro4broad, line 109: Why is this API being added? I don't see it used anywhere in this PR other than tests.

Before when I asked about `SYMMETRIC`, my question was really about how do we know that this needs to support out-of-order queries? Without an example of how it will be used in code, it's hard to tell.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -267,44 +313,32 @@ void insertDataAccessRequest(
...
+  void updateDraftToSubmittedForCollection(@Bind("collectionId") Integer collectionId,
       @Bind("referenceId") String referenceId);
 
   @RegisterRowMapper(DataAccessRequestDataMapper.class)
   @SqlQuery(
-      "SELECT (data #>> '{}')::jsonb AS data "
-          + "FROM data_access_request "
-          + "WHERE (LOWER(data->>'status') != 'archived' "
-          + "OR data->>'status' IS NULL)")
+      """
```
otchet-broad, line 328: formatting

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -267,44 +313,32 @@ void insertDataAccessRequest(
...
+            OR data->>'status' IS NULL)
+      """)
   List<DataAccessRequestData> findAllDataAccessRequestDatas();
 
   @SqlUpdate(
-      " UPDATE data_access_request"
-          + " SET data = jsonb_set ((data #>> '{}')::jsonb, '{status}', '\"Archived\"', true) "
-          + " WHERE reference_id IN (<referenceIds>)")
+      """
+         UPDATE data_access_request
```
otchet-broad, line 338: Formatting

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -315,24 +349,31 @@ void updateDraftForCollection(@Bind("collectionId") Integer collectionId,
    * @param datasetId   Integer
    */
   @SqlUpdate(
-      "INSERT INTO dar_dataset (reference_id, dataset_id) "
-          + "VALUES (:referenceId, :datasetId) "
-          + "ON CONFLICT DO NOTHING")
+      """
+          INSERT INTO dar_dataset (reference_id, dataset_id)
```
otchet-broad, line 353: Formatting.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -315,24 +349,31 @@ void updateDraftForCollection(@Bind("collectionId") Integer collectionId,
...
+      """)
   void insertDARDatasetRelation(@Bind("referenceId") String referenceId,
       @Bind("datasetId") Integer datasetId);
 
   @SqlBatch(
-      "INSERT INTO dar_dataset (reference_id, dataset_id) "
-          + "VALUES (:referenceId, :datasetId) "
-          + "ON CONFLICT DO NOTHING")
+      """
+          INSERT INTO dar_dataset (reference_id, dataset_id)
```
otchet-broad, line 362: Formatting

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -315,24 +349,31 @@ void updateDraftForCollection(@Bind("collectionId") Integer collectionId,
...
 
   /**
    * Delete rows which have the given reference id
    *
    * @param referenceId String
    */
-  @SqlUpdate("DELETE FROM dar_dataset WHERE reference_id = :referenceId")
+  @SqlUpdate(
+      """
+          DELETE FROM dar_dataset WHERE reference_id = :referenceId
```
otchet-broad, line 375: Formatting.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
pshapiro4broad: Thanks for these comments, they were very helpful!

src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
```diff
@@ -19,10 +19,11 @@ public class DarCollection {
   public static final String DAR_FILTER_QUERY_COLUMNS =
       "dar.id AS dar_id, dar.reference_id AS dar_reference_id, dar.collection_id AS dar_collection_id, "
           +
-          "dar.parent_id AS dar_parent_id, dar.draft AS dar_draft, dar.user_id AS dar_userId, " +
+          "dar.parent_id AS dar_parent_id, dar.user_id AS dar_userId, " +
```
otchet-broad, line 22: remove draft

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
```diff
@@ -157,6 +169,11 @@ public Timestamp getSubmissionDate() {
 
   public void setSubmissionDate(Timestamp submissionDate) {
     this.submissionDate = submissionDate;
+    draft = submissionDate == null;
+    expired = submissionDate != null
```
pshapiro4broad, line 172: Is there a unit test for this code?

src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
```diff
@@ -147,14 +148,14 @@ public void addDataset(Dataset dataset) {
...
 
-  public void setDatasets(Set<Dataset> datasets) {
-    this.datasets = datasets;
-  }
-
   public Set<Dataset> getDatasets() {
     return datasets;
   }
 
+  public void setDatasets(Set<Dataset> datasets) {
```
otchet-broad, line 155: Formatting.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
otchet-broad: The ticket includes an AC of "Mechanism to find expired DARs"
This is an example of a "Mechanism to find expired DARs"

src/main/java/org/broadinstitute/consent/http/models/DarCollectionSummary.java
```diff
@@ -134,6 +141,18 @@ public Timestamp getSubmissionDate() {
 
   public void setSubmissionDate(Timestamp submissionDate) {
     this.submissionDate = submissionDate;
+    if (submissionDate != null) {
+      this.expiresAt = Timestamp.from(Instant.ofEpochMilli(submissionDate.getTime() + DataAccessRequest.EXPIRATION_DURATION_MILLIS));
+      this.expired = this.expiresAt.before(Timestamp.from(Instant.now()));
```
pshapiro4broad, line 146: If you use `Instant` for `expiresAt` I think it's easier to understand what's going on here
```
    expiresAt = (submissionDate != null)
        ? submissionDate.toInstant().plus(365, ChronoUnit.DAYS) : Instant.MAX;
    expired = Instant.now().isAfter(expiresAt);
```

src/main/resources/assets/schemas/DataAccessRequest.yaml
```diff
@@ -22,6 +22,12 @@ properties:
   draft:
     type: boolean
     description: Draft status of this DAR
+  expired:
+    type: boolean
+    description: Expired status of this DAR
+  expiredAt:
+    type: number
+    description: When this DAR expires, in milliseconds since the epoch.  A value of -1 indicates this DAR does not have an expiration date because it is a draft.
```
pshapiro4broad: Another approach is to have these fields be `null` or missing unless `draft` is `false`. Or, instead of a sentinel value, use `Instant.MAX` which is a valid time offset and could be used in comparisons, vs `-1` which is invalid.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
otchet-broad: Two things:
1) I've kept the syntax as-is because it's widely used and didn't want to creep scope on this.
2) I'm wondering if there are other optimizations we could be doing here to avoid the data mutation that happens in the select.  
[https://broadworkbench.atlassian.net/browse/DT-1563](https://broadworkbench.atlassian.net/browse/DT-1563)

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
```diff
@@ -762,6 +779,40 @@ void testFindAllDataAccessRequestDatasArchived() {
...
+    darCollection.getDars().keySet().forEach((referenceId) -> {
+      dataAccessRequestDAO.updateDraftToSubmittedForCollection(darCollection.getDarCollectionId(),
+          referenceId);
+    });
+    Timestamp oneYearAgo = Timestamp.from(
+        Instant.ofEpochMilli(new Date().getTime() - DataAccessRequest.EXPIRATION_DURATION_MILLIS));
+    Timestamp oneMinuteInTheFuture = Timestamp.from(
+        Instant.ofEpochMilli(new Date().getTime() + (60 * 1000)));
+    List<DataAccessRequest> dars = dataAccessRequestDAO.findSubmittedDarsByTimeRange(oneYearAgo,
+        oneMinuteInTheFuture);
```
pshapiro4broad: I think `Instant` would be easier to read here
```suggestion
    Instant oneYearAgo = Instant.now().minus(365, ChronoUnit.DAYS);
    Instant oneMinuteInTheFuture = Instant.now().plus(1, ChronoUnit.MINUTES);
    List<DataAccessRequest> dars = dataAccessRequestDAO.findSubmittedDarsByTimeRange(Timestamp.from(oneYearAgo),
        Timestamp.from(oneMinuteInTheFuture));
```

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
```diff
@@ -762,6 +779,40 @@ void testFindAllDataAccessRequestDatasArchived() {
     assertEquals(1, returnedDAR.size());
   }
 
+  @Test
+  void testFindDARsByDateRange() throws InterruptedException {
+    DarCollection darCollection = createDarCollection();
+    darCollection.getDars().keySet().forEach((referenceId) -> {
+      dataAccessRequestDAO.updateDraftToSubmittedForCollection(darCollection.getDarCollectionId(),
+          referenceId);
+    });
```
pshapiro4broad: A one-line lambda doesn't need `{}`s
```suggestion
    darCollection.getDars().keySet().forEach((referenceId) ->
        dataAccessRequestDAO.updateDraftToSubmittedForCollection(darCollection.getDarCollectionId(), referenceId));
```

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
```diff
@@ -762,6 +779,40 @@ void testFindAllDataAccessRequestDatasArchived() {
...
+  void testFindDARsByDateRangeWithNoneInRange() throws InterruptedException {
+    DarCollection darCollection = createDarCollection();
+    darCollection.getDars().keySet().forEach((referenceId) -> {
+      dataAccessRequestDAO.updateDraftToSubmittedForCollection(darCollection.getDarCollectionId(),
+          referenceId);
+    });
+    Timestamp oneHourAgo = Timestamp.from(
+        Instant.ofEpochMilli(new Date().getTime() - (60 * 60 * 1000)));
+    Timestamp halfAnHourAgo = Timestamp.from(
+        Instant.ofEpochMilli(new Date().getTime() - (30 * 60 * 1000)));
```
pshapiro4broad:
```
    Instant oneHourAgo = Instant.now().minus(1, ChronoUnit.HOURS);
    Instant halfAnHourAgo = Instant.now().minus(30, ChronoUnit.MINUTES);
```

pshapiro4broad (approved):
Some possible changes; also I would like to avoid storing computed values if that's possible here. But looks OK overall.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
pshapiro4broad: If the only need is to find expired, could it be done with a query that has a single bound? Adding the service method that uses this API  (e.g. `DataAccessRequestService.findExpiredDars()`) would help me understand how this API works to meet the AC.

Is the idea that, instead of keeping track of which DARs we've notified users of, we instead find only those that expired in a specific time range? That could work but we'd need to be very precise about when the check is done, to avoid missing some DARs or notifying twice for others.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
otchet-broad: I went down this path and had to revert my changes.

https://github.com/google/gson/issues/232

We've configured [consent+Jersey ](https://github.com/DataBiosphere/consent/blob/acbe32328afec051cbc2f25f22aa0bc74572afdc/src/main/java/org/broadinstitute/consent/http/ConsentApplication.java#L192)to use Gson.  Gson doesn't (easily) allow us to implement this.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
otchet-broad: Exactly.  As discussed at standup this morning, we have a ticket to notify DAR submitters when their DAR will expire in 30 days (from today).  We'll know the starting time and ending time of the query range.  That's the use case I was specifically thinking about.  
Do others exist?  Sure.  
Are they in other tickets?  Yes.
Do I expect this to change?  Maybe.  It's just an example.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
pshapiro4broad: OK, I saw `JsonProperty` and thought that maybe jackson was being used here. Oh well. Seems unlikely to happen now as the ticket's been open for 10 years!

src/main/java/org/broadinstitute/consent/http/models/DarCollectionSummary.java
otchet-broad: Updated to use Timestamp and modify the code/tests to return null when a draft.

src/main/resources/assets/schemas/DataAccessRequest.yaml
otchet-broad: OBE since expiredAt won't return when it's null.  Docs updated.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
otchet-broad: Yes.  In DataAccessRequestDAOTest it's tested to confirm behavior.  Last few pushes refactored to make the tests stand out from others.

PR #2493, opened 2025-04-22
DT-1546: Verify that user's email matches institution during library card creation

pshapiro4broad:
### Summary

Verify that a user's institution (derived from their email) matches their library card's institution. This is verified on library card creation and when a library card is updated.

### Details

Since library cards can be created before a user's account is created in DUOS, there are a few code paths possible:

- When the user ID in the library card is `null`. In this case the user email must exist, but the user account may not exist. If the user account does exist, the user ID in the card is set to the correct value.
- When the user ID in the library card is set. In this case, the user must exist, and if the card specifies an email address, it must match (case insensitive) the user email in the library card. If the card doesn't specify an email address, its value is set using the user's email address.

Once this is complete, the library card contains the user's email address, and we can use it to match the derived institution against the library card's institution.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
@@ -8,7 +8,6 @@ import java.util.Date;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.stream.Collectors;
 import org.broadinstitute.consent.http.db.InstitutionDAO;
 import org.broadinstitute.consent.http.db.LibraryCardDAO;
 import org.broadinstitute.consent.http.db.UserDAO;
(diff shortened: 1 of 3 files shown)


snf2ye (approved):
Looks reasonable to me

src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
```diff
@@ -22,36 +21,38 @@ public class LibraryCardService {
...
-    Boolean isAdmin = checkIsAdmin(user);
+    boolean isAdmin = checkIsAdmin(user);
     //If user is not an admin, use user's institutionId rather than the value provided in the payload
     if (!isAdmin && !libraryCard.getInstitutionId().equals(user.getInstitutionId())) {
       throw new BadRequestException("Card payload not valid");
     }
     checkIfCardExists(libraryCard);
-    checkForValidInstitution(libraryCard.getInstitutionId());
-    LibraryCard processedCard = processUserOnNewLC(libraryCard);
+    processUserOnNewLC(libraryCard);
```
pshapiro4broad, line 45: Since this method mutates its argument, there was no need to have it return a value.

fboulnois (approved):
looks good, thank you 👍

PR #2496, opened 2025-04-23
[DT-1167]Make Progress Reports Submittable

rjohanek:
### Addresses
Ticket: https://broadworkbench.atlassian.net/browse/DT-1167
When testing the post progress report endpoint, there were a few issues. This PR addresses the following issues.
1. ParentId was not being recorded.
2. The CollectionId of the progress report was not the same as the parent DAR.
3. The progress report was being updated with documents following its submission. We don't want it to be modifiable after submission.
4. Any data sent in the progress report was being saved even if they were different from the parent DAR. There are some fields that should not be changeable/we want to ignore and copy from the parent DAR.
5. Progress Reports could be made from Draft DARs.
6. Progress Reports could be made without the required fields: progressReportSummary, intellectualPropertySummary.

### Summary
Changes and the issues they address:

New DAO method just for progress reports, including parent Id **(Issue 1)**

New populateProgressReportFromJsonString method that copies most information from the parent DAR (including CollectionId) and then overwrites the datasetIds and collaborators with the progress report information and adds the new progress report specific fields. **(Issue 2 and 4)**
- Question: What is the expected behavior for the document related fields? Outside of the populate method, in updateProgressReportWithDocuments the new documents will be uploaded if they are present and the document name/location fields are updated on the progress report to reflect the new documents. Just checking, does this sound like the behavior we want? Or would we rather just copy over the old documents if they exist?

Move the uploading of documents before the submission of the progress report. Pulled into a new method and refactored to share code - removing the call to updateByReferenceId for this case since the progress report does not exist yet **(Issue 3)**

Add validation to validateProgressReport that the parent DAR is not a draft, and that the required fields are present. **(Issue 5 and 6)**

### Testing
Unit testing 
Manual testing success case - now includes parentId, correct CollectionId, parentDAR information (ex. title, RUS, research purpose, etc), and new progress report information (collaborators, datasets, progressReportSummary, etc.)
See results of manual testing [here]( https://docs.google.com/document/d/1Ihws5ZHZfpzrvJhuKoXJIsEdpvzjuFGxFIrwpfdd-q8/edit?usp=sharing).

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -574,6 +570,56 @@ private DataAccessRequest populateDarFromJsonString(User user, String json) {
...
+    originalDataCopy.setLabCollaborators(newData.getLabCollaborators());
+
+    originalDataCopy.setProgressReportSummary(newData.getProgressReportSummary());
+    originalDataCopy.setIntellectualPropertySummary(newData.getIntellectualPropertySummary());
+    originalDataCopy.setPublications(newData.getPublications());
+    originalDataCopy.setPresentations(newData.getPresentations());
+    originalDataCopy.setDmi(newData.getDmi());
+    originalDataCopy.setResearchPlans(newData.getResearchPlans());
+    originalDataCopy.setCloseoutSupplement(newData.getCloseoutSupplement());
+
```
rjohanek, line 607: additionally, in updateProgressReportWithDocuments the collaboration letter/irb approval document fields maybe updated

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -589,6 +635,21 @@ private DataAccessRequest updateDarWithDocumentContents(
       DataAccessRequest dar,
       InputStream uploadInputStream,
       FormDataContentDisposition fileDetail) throws IOException {
+    uploadDocumentContents(type, dar, uploadInputStream, fileDetail);
+    return dataAccessRequestService.updateByReferenceId(user, dar);
```
rjohanek, line 653: refactoring shared code. uploadDocumentContents doesn't include the updateByReferenceId call so Progress Reports can use it prior to PR creation in the DB

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -533,15 +537,7 @@ private User findUserByEmail(String email) {
...
-    DataAccessRequestData data;
-    try {
-      data = DataAccessRequestData.fromString(json);
-    } catch (Exception e) {
-      throw new BadRequestException("Unable to parse DAR from JSON string");
-    }
-    if (Objects.isNull(data)) {
-      data = new DataAccessRequestData();
-    }
+    DataAccessRequestData data = populateDARData(json);
```
rjohanek, line 544: refactoring shared code

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -163,26 +164,10 @@ public List<DataAccessRequest> getDataAccessRequestsByUserRole(User user) {
...
-        dataAccessRequest.getReferenceId()) || Objects.isNull(dataAccessRequest.getData())) {
-      throw new IllegalArgumentException("User and DataAccessRequest are required");
-    }
-
-    if (user.getLibraryCards().isEmpty()) {
-      throw new NIHComplianceRuleException();
-    }
-
-    validateInternalCollaborators(dataAccessRequest, user);
+    validateDar(user, dataAccessRequest);
```
rjohanek, line 166: refactoring shared code

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -163,26 +164,10 @@ public List<DataAccessRequest> getDataAccessRequestsByUserRole(User user) {
...
 
     Date now = new Date();
-    long nowTime = now.getTime();
     DataAccessRequestData darData = dataAccessRequest.getData();
-    if (Objects.isNull(darData.getCreateDate())) {
-      darData.setCreateDate(nowTime);
-    }
-    darData.setSortDate(nowTime);
```
rjohanek, line 183: these are deprecated fields (not returned in the resulting DAR), just a cleanup

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -413,45 +414,48 @@ public Response postProgressReport(
     try {
       User user = userService.findUserByEmail(authUser.getEmail());
       DataAccessRequest parentDar = dataAccessRequestService.findByReferenceId(parentReferenceId);
-      if (!user.getUserId().equals(parentDar.getUserId())) {
-        throw new ForbiddenException("User not authorized to update this Data Access Request");
```
rjohanek, line 418: moved this validation to service code

src/test/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceTest.java
```diff
@@ -366,45 +375,19 @@ private Pair<InputStream, FormDataContentDisposition> mockFormDataMultiPart(Stri
   }
 
   @Test
-  void testPostProgressReportCollabAndEthicsFiles() throws IOException {
+  void testPostProgressReport() {
```
rjohanek, line 406: removed a bunch of postProgressReport tests and added a bunch of updateProgressReportWithDocuments in their place, now that the logic is pulled into that method

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -595,6 +639,21 @@ private DataAccessRequest updateDarWithDocumentContents(
...
+  /**
+   * Uploads the document contents to GCS and mutates the dar data object with the blobId.
+   *
+   * @param type            The type of document (IRB or Collaboration)
+   * @param dar             The Data Access Request
+   * @param uploadInputStream The input stream of the file to be uploaded
+   * @param fileDetail      The file details
+   * @throws IOException if an error occurs during upload
+   */
+  public void uploadDocumentContents(DarDocumentType type, DataAccessRequest dar, InputStream uploadInputStream,
```
otchet-broad, line 665: I think the use of this method in this way introduces the chance a child DAR could delete the IRB and Collaboration documents on the parent DAR.  Is that the behavior you expect here?

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -313,7 +313,44 @@ void insertDataAccessRequest(
...
+            :submissionDate, :updateDate, to_jsonb(:data))
+      """)
+  void insertProgressReport(
+      @Bind("parentId") Integer parentId,
+      @Bind("collectionId") Integer collectionId,
+      @Bind("referenceId") String referenceId,
+      @Bind("userId") Integer userId,
+      @Bind("createDate") Date createDate,
+      @Bind("sortDate") Date sortDate,
+      @Bind("submissionDate") Date submissionDate,
```
otchet-broad: It looks like all of the uses of insertProgressReport make use of the same date value for all of the date fields and it's always the current time.  Why not just insert the value of now() in the SQL for each of these?  Doing so will have the added benefit of causing someone in the future who wants to set submitted_date to null to have to contemplate the consequences of that decision.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -227,6 +212,74 @@ public DataAccessRequest createDataAccessRequest(User user, DataAccessRequest da
...
+          now,
+          now,
+          progressReport.getData());
+    syncDataAccessRequestDatasets(datasetIds, referenceId);
+    return findByReferenceId(referenceId);
+  }
+
+  public void validateProgressReport(User user, DataAccessRequest progressReport, DataAccessRequest parentDar) {
+    validateDar(user, progressReport);
+    if (parentDar.getDraft()) {
```
otchet-broad, line 240: Woohoo!  I like this in the service!

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -227,6 +212,74 @@ public DataAccessRequest createDataAccessRequest(User user, DataAccessRequest da
...
+      throw new BadRequestException("Progress report summary is required");
+    }
+    if (progressReport.getData().getIntellectualPropertySummary() == null ||
+        progressReport.getData().getIntellectualPropertySummary().isEmpty()) {
+      throw new BadRequestException("Intellectual Property Summary is required");
+    }
+  }
+
+  public void validateDar(User user, DataAccessRequest dar) {
+    if (Objects.isNull(user) || Objects.isNull(dar) || Objects.isNull(
```
otchet-broad, line 261: Nice refactor!

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
rushtong: This is a pattern we should be using across all inserts

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
rushtong: This is a scenario where we should be stopping the user before line 418 where files are uploaded.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -414,45 +414,48 @@ public Response postProgressReport(
...
-              collabInputStream, collabFileDetails);
+  public void updateProgressReportWithDocuments(InputStream collabInputStream,
+      FormDataContentDisposition collabFileDetails, InputStream ethicsInputStream,
+      FormDataContentDisposition ethicsFileDetails, DataAccessRequest childDar,
+      DataAccessRequest parentDar) throws IOException {
+    for (Integer datasetId : childDar.getDatasetIds()) {
+      Dataset dataset = datasetService.findDatasetById(datasetId);
+      if (dataset == null) {
+        throw new NotFoundException("Dataset " + datasetId + " not found");
+      }
```
rushtong, line 439: Probably scope creep here, but `datasetService.findDatasetById` really should throw the NFE. We do this a lot in the service layer now, not sure why this method doesn't do that. I think a follow-on ticket would be nice for this.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -575,6 +570,55 @@ private DataAccessRequest populateDarFromJsonString(User user, String json) {
...
+  /**
+   * Populate a new Data Access Request from the JSON string and the parent Data Access Request.
+   * Copies all the data from the parent dar, then overwrites the collaborators and datasets.
+   * Adds all progress report specific fields.
+   *
+   * @param json      The JSON string to populate the new Progress Report.
+   * @param parentDar The parent Data Access Request to copy data from.
+   * @return A new Progress Report populated with the provided JSON string and parent DAR data.
+   */
+  public DataAccessRequest populateProgressReportFromJsonString(String json, DataAccessRequest parentDar) {
```
rushtong, line 586: This feels more like a service layer method than a resource layer method, but it is building off of shared code in `populateDARData` (👍🏽) so I think it's fine here without doing a larger refactor.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
rushtong: I see this too - on line 585, we copy the parent dar's values for `irbDocumentLocation` and `collaborationLetterLocation` into the child. The end result here will be that the parent dar will contain a reference to a location that will have had it's underlying file store removed and the child will contain new values with references to new documents.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
rushtong: Nice cleanup.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -227,6 +212,74 @@ public DataAccessRequest createDataAccessRequest(User user, DataAccessRequest da
...
+          Integer.valueOf(progressReport.getParentId()),
+          progressReport.getCollectionId(),
+          referenceId,
+          user.getUserId(),
+          now,
+          now,
+          now,
+          now,
+          progressReport.getData());
+    syncDataAccessRequestDatasets(datasetIds, referenceId);
```
rushtong: Not for this PR - it's definitely scope creep. The insert and sync should really be in a single transaction-level method. This also applies to any call where we're modifying a DAR - draft updates, for example.

src/test/java/org/broadinstitute/consent/http/service/DataAccessRequestServiceTest.java
```diff
@@ -202,6 +206,46 @@ void testCreateDataAccessRequest_FailsIfNoLibraryCard() {
...
+    progressReport.setCollectionId(parentDar.getCollectionId());
+    parentDar.setSubmissionDate(Timestamp.from(Instant.now()));
+    User user = new User(1, "email@test.org", "Display Name", new Date());
+    user.setLibraryCards(List.of(new LibraryCard()));
+    parentDar.setUserId(user.getUserId());
+
+    doNothing().when(dataAccessRequestDAO)
+        .insertProgressReport(eq(parentDar.getId()), eq(progressReport.getCollectionId()), eq(progressReport.getReferenceId()), eq(user.getUserId()), any(), any(),
+            any(), any(), eq(progressReport.getData()));
+    doNothing().when(dataAccessRequestDAO).insertAllDarDatasets(argThat(new DarDatasetMatcher(progressReport)));
```
rushtong: The test runs fine without these statements since `dataAccessRequestDAO` is a mock

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
rjohanek: good catch, I'll move that back to the resource layer

src/test/java/org/broadinstitute/consent/http/service/DataAccessRequestServiceTest.java
rjohanek: okay I changed it to a verify

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
rjohanek: okay great point! Maybe in the populate method we shouldn't copy over those fields from the parent then.  Maybe just set them to null? Or copy over the progressReport's values?

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
pshapiro4broad: > Why not just insert the value of now() in the SQL for each of these?

I agree that the SQL should be setting any fields that we are seeing to `now()` in code, and they should be made non-null.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
otchet-broad: @rjohanek Good question on what to do here.
The thing crossing my mind at a high level is if we need to maintain a DAR exactly as it was submitted forever so that we can demonstrate what was approved at the time, if we need to do that, my suggestion is below:
1) If we have existing documents from a (parent) DAR, we copy the files (making new FileStorageObjects) to new locations with new blobIDs for them so they can live as independent artifacts for this DAR instance.
2) We accept new documents when posting a progress report and when we've got new documents, we only use those.

Might be a good follow-up question for our POs.

rjohanek: @otchet-broad Yes I think we do need to maintain the DAR exactly as it was submitted. If I understand correctly, if we just set the values to null in the populate method, then, we will never overwrite any parent values. The parent DAR will remain as it was submitted with its documents, and the progress report will have the new documents uploaded and set if there are any. I don't see why we would need to copy the parent documents?

otchet-broad: @rjohanek - great conversation.  I really appreciate your approach of incremental progress.   We discussed the updated code path and some follow-on tickets you would make for our product owners to consider:
1) Should collaboration agreements and IRB documents be copied over from the parent DAR to this new DAR?
2) When we delete any DAR in the DAR collection, should we delete all DARs? (e.g. a cascade delete)

IMHO, the answers to the questions above don't need to block completion of this ticket because the new tickets give us an opportunity to revisit the decisions made and make changes as needed.

otchet-broad (approved):
Great work with the refactor and logic encapsulation!  I appreciate all of the engagement and focus on getting this work through so our frontend work can proceed while our product owners answer some lingering questions.

rushtong (approved):
Thank you for the updates 👍🏽

PR #2498, opened 2025-04-24
[DT-1548] Validate user has eRA commons login set when submitting a DAR to a DAC.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-1548](https://broadworkbench.atlassian.net/browse/DT-1548)

### Summary

Ensures that a DAR submitted to a DAC is done so by a user with valid eRA account linked to their DUOS account.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
@@ -44,13 +44,14 @@ public class DataAccessRequestService implements ConsentLogger {
   private final MatchDAO matchDAO;
   private final VoteDAO voteDAO;
   private final UserDAO userDAO;
+  private final UserService userService;
   private final DataAccessRequestServiceDAO dataAccessRequestServiceDAO;
 
   private final DacService dacService;
 
   @Inject
   public DataAccessRequestService(CounterService counterService, DAOContainer container,
-      DacService dacService, DataAccessRequestServiceDAO dataAccessRequestServiceDAO) {
+      DacService dacService, DataAccessRequestServiceDAO dataAccessRequestServiceDAO, UserService userService) {
     this.counterService = counterService;
     this.dataAccessRequestDAO = container.getDataAccessRequestDAO();
     this.darCollectionDAO = container.getDarCollectionDAO();
@@ -60,6 +61,7 @@ public class DataAccessRequestService implements ConsentLogger {
     this.userDAO = container.getUserDAO();
     this.dacService = dacService;
     this.dataAccessRequestServiceDAO = dataAccessRequestServiceDAO;
+    this.userService = userService;
   }
 
   public List<DataAccessRequest> findAllDraftDataAccessRequests() {
@@ -171,6 +173,8 @@ public class DataAccessRequestService implements ConsentLogger {
     if (user.getLibraryCards().isEmpty()) {
       throw new NIHComplianceRuleException();
     }
+
+  userService.hasValidActiveERACredentials(user);
 
     validateInternalCollaborators(dataAccessRequest, user);
 
--- a/src/main/java/org/broadinstitute/consent/http/service/UserService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/UserService.java
@@ -10,6 +10,7 @@ import com.google.gson.JsonObject;
 import com.google.inject.Inject;
 import jakarta.ws.rs.BadRequestException;
 import jakarta.ws.rs.NotFoundException;
+import java.time.Instant;
 import java.util.Collections;
 import java.util.Date;
 import java.util.List;
@@ -32,6 +33,7 @@ import org.broadinstitute.consent.http.db.VoteDAO;
 import org.broadinstitute.consent.http.enumeration.UserFields;
 import org.broadinstitute.consent.http.enumeration.UserRoles;
 import org.broadinstitute.consent.http.exceptions.ConsentConflictException;
+import org.broadinstitute.consent.http.exceptions.LibraryCardRequiredException;
 import org.broadinstitute.consent.http.models.AuthUser;
 import org.broadinstitute.consent.http.models.DataAccessAgreement;
 import org.broadinstitute.consent.http.models.Institution;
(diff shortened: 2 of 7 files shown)


otchet-broad:
Soliciting feedback: I noticed that the User object isn't being enriched with UserProperty objects when we `UserService.findUserByEmail` or `UserService.findUserById`

I can refactor this implementation so that these two methods always include the populated  `List<UserProperty>` or leave it as is.  The nice part about the current implementation is that we don't need to make another trip to the database for these except where we need them.  The downside is that we need to bring the UserService where ever we need to make the call on this - and right now it's only one method in one service class (that doesn't have an issue with including the UserService).

rushtong:
> Soliciting feedback: I noticed that the User object isn't being enriched with UserProperty objects when we `UserService.findUserByEmail` or `UserService.findUserById`
> 
> I can refactor this implementation so that these two methods always include the populated `List<UserProperty>` or leave it as is. The nice part about the current implementation is that we don't need to make another trip to the database for these except where we need them. The downside is that we need to bring the UserService where ever we need to make the call on this - and right now it's only one method in one service class (that doesn't have an issue with including the UserService).

I'm happy with enriching the single user calls. I would want some additional timing metrics to know if adding the joins for lists of users degrades performance too much.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -172,6 +174,8 @@ public DataAccessRequest createDataAccessRequest(User user, DataAccessRequest da
       throw new NIHComplianceRuleException();
     }
 
+  userService.hasValidActiveERACredentials(user);
```
rushtong: nit: looks like indentation is off - could be the whole file needs reformatting

src/test/java/org/broadinstitute/consent/http/service/DataAccessRequestServiceTest.java
```diff
@@ -165,14 +165,31 @@ void testCreateDataAccessRequest_CreateWithSubmittedDar() {
...
     });
   }
 
+  @Test
+  void testCreateDataAccessRequestCreateWithoutERACommons() {
+    DataAccessRequest dar = generateDataAccessRequest();
+    dar.addDatasetIds(List.of(1, 2, 3));
+    dar.setCreateDate(new Timestamp(1000));
+    dar.setSortDate(new Timestamp(1000));
+    dar.setReferenceId("id");
```
rushtong: nit: I don't think the test relies on these values being set.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
otchet-broad: Off intentionally.  I accidentally landed some of these changes (in a non functional way) in an earlier PR.  I wanted to call attention to them here so they're reviewed.

otchet-broad:
> > Soliciting feedback: I noticed that the User object isn't being enriched with UserProperty objects when we `UserService.findUserByEmail` or `UserService.findUserById`
> > I can refactor this implementation so that these two methods always include the populated `List<UserProperty>` or leave it as is. The nice part about the current implementation is that we don't need to make another trip to the database for these except where we need them. The downside is that we need to bring the UserService where ever we need to make the call on this - and right now it's only one method in one service class (that doesn't have an issue with including the UserService).
> 
> I'm happy with enriching the single user calls. I would want some additional timing metrics to know if adding the joins for lists of users degrades performance too much.

It's been a while since I was in in this class.  What was interesting and surprising to me was the fact that we aren't joining users to some core items. **Library cards** are an independent lookup when getting a user by email - by far our most common access pattern.  

I guess my own philosophy on this is that every time I'm working with a User object, I expect it to be the full representation of that user, not a partial one. Lazily loaded, fine (if need be).  But in this case it wasn't clear that I'm being returned a partial object.  Are others surprised by that?

pshapiro4broad:
> I guess my own philosophy on this is that every time I'm working with a User object, I expect it to be the full representation of that user, not a partial one. Lazily loaded, fine (if need be). But in this case it wasn't clear that I'm being returned a partial object. Are others surprised by that?

I agree, in that having inconsistently initialized models can make for very confusing bugs and code. So I would prefer to never have partially initialized (or mutable!) models so it's always clear to the caller what the expected state is.

But, I'm not surprised of the state of things since it's much more common to see partial initializations like this than code that follows this rule.

I agree with @rushtong that adding this to single user lookups should be OK. It would be nice to think about if we want to better distinguish between empty and uninitialized state.

Regarding the current behavior, the `properties` field is on `User` but it looks like it's only initialized by `findAllDARCollections()`, `findDARCollectionByReferenceId()`, `findDARCollectionByCollectionId()` when fetching the `createUser` for `DarCollection`. The field is also mapped by `UserWithRolesReducer` but as far as I can tell, none of the queries that use this mapper join to `user_property`. So another approach here would be to move `User.properties` into `DarCollection` as a new field, like `createUserProperties`.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -467,32 +469,36 @@ public List<User> findUsersInJsonArray(String json, String arrayKey) {
...
-    List<UserProperty> eraExpirationProps = userProperties.stream().filter(
-            userProperty -> userProperty.getPropertyKey()
-                .equalsIgnoreCase(ERA_EXPIRATION_DATE.getValue()))
-        .toList();
-    if (eraStatusProps.size() == 1 && eraExpirationProps.size() == 1) {
-      if (!eraStatusProps.get(0).getPropertyValue().equalsIgnoreCase("true")) {
-        throw new BadRequestException("User does not have an Era Commons ID that is authorized.");
+    public void hasValidActiveERACredentials(User user) {
+      List<LibraryCard> cards = user.getLibraryCards();
+      List<UserProperty> userProperties = findAllUserProperties(user.getUserId());
```
pshapiro4broad: The more I look at this, the more I think I would move `properties` out of `User` rather than making it populated in more cases. Unless there are upcoming use cases where `properties` is needed outside of its use in `DarCollection.createUser`.

otchet-broad: @rushtong , @pshapiro4broad  - so I think we're in agreement.  I'll leave the PR as-is and make a ticket for us to consider something different later.
@pshapiro4broad - the one solid use case where I see the List used is when we build the object to serialize back to the client in findUserWithPropertiesByIdAsJsonObject.  Ironically, we never seem to populate the list in the user object.  The code is necessary in that method because it allows the DAR form in the UI to become aware that the user has a valid token.

otchet-broad: Ticket: [https://broadworkbench.atlassian.net/browse/DT-1571](https://broadworkbench.atlassian.net/browse/DT-1571)

rushtong: Yes, in agreement on enrichment.

Secondly, yes, the `findUserWithPropertiesByIdAsJsonObject` is the primary method to populate a user which is used quite a bit in the UI. In theory, that method could be refactored to make use of the enriched user. I think it might be odd to try and restrict user properties to the DarCollection - we need user info for admins, DAC members, SOs, etc., who would generally never create a DAR, but could be eRA Commons/RAS users.

s-rubenstein (approved):
LGTM, I agree with the follow up change to properties.

rjohanek (commented):
this looks good, sorry about the merge conflicts, now that createProgressReport exists, I think this logic needs to be added there too. There's the part where we add documents before the dar Service is called to create the PR, so maybe we need to check if there's a LC before that call too

fboulnois (approved):
looks good 👍

otchet-broad:
> this looks good, sorry about the merge conflicts, now that createProgressReport exists, I think this logic needs to be added there too. There's the part where we add documents before the dar Service is called to create the PR, so maybe we need to check if there's a LC before that call too

@rjohanek - Both createProgressReport and createDataAccessRequest call validateDar.  I've moved this code's validation of ERACredentials into validate just after library cards are checked. :-). Let me know what you think!

rjohanek (approved):
looks good thanks!!

PR #2502, opened 2025-04-28
[DT-99] Update DAR table to show latest in Dar Collection

rjohanek:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-99
We want to show the information from the latest DAR in the collection on the researcher console DAR page. We want to do this for all DAR tables in the UI which means updating all the DarCollectionSummaryDAO queries.

### Summary
Maintain the structure of the DarCollectionSummary by aggregating all referenceIds that share a collectionId and still include them in the summary, but update the information of the summary to only look at the the latest DAR by submission date instead.

### Testing
Manual testing shows the UI looks as expected and checked that the summary information represents the latest DAR
Added unit tests
I also tested performance and the new queries have a similar performance to the existing queries (~100ms on 600 Dar Collections)

### To Do
I only updated the researcher query, something similar needs to be done for the other queries. There is a ticket for the rest of this work: https://broadworkbench.atlassian.net/browse/DT-1598

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
@@ -142,11 +142,13 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
             SELECT *
             FROM data_access_request
             WHERE data_access_request.collection_id = c.collection_id
+            AND data_access_request.submission_date IS NOT NULL
             ORDER BY data_access_request.submission_date DESC
             LIMIT 1
         ) dar ON true
     INNER JOIN
         data_access_request dar_all ON dar_all.collection_id = c.collection_id
+        AND dar_all.submission_date IS NOT NULL
     LEFT JOIN
         LATERAL (
             SELECT
@@ -163,6 +165,7 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
         c.create_user_id = :userId
         AND (e.latest = e.election_id OR e.election_id IS NULL)
         AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)
+        AND (EXISTS (SELECT 1 FROM data_access_request WHERE (collection_id = c.collection_id AND dar.submission_date IS  NOT NULL)))
     GROUP BY
         c.collection_id, c.dar_code, dar.submission_date, dar.reference_id, u.display_name, i.institution_name,
         e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id, dar.data
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
@@ -93,11 +93,24 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
       SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, u.display_name as researcher_name,
         i.institution_name, e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id as dd_datasetid,
         (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name,
-        dac.name as dac_name
+        dac.name as dac_name, ARRAY_AGG(dar_all.reference_id) AS reference_ids
       FROM dar_collection c
       INNER JOIN users u ON u.user_id = c.create_user_id
       LEFT JOIN institution i ON i.institution_id = u.institution_id
-      INNER JOIN data_access_request dar ON dar.collection_id = c.collection_id
+      INNER JOIN
+              LATERAL (
+                  SELECT *
+                  FROM data_access_request
+                  WHERE data_access_request.collection_id = c.collection_id
+                  AND data_access_request.submission_date IS NOT NULL
+                  AND (LOWER(data_access_request.data->>'status') != 'archived' OR data_access_request.data->>'status' IS NULL)
+                  ORDER BY data_access_request.submission_date DESC
+                  LIMIT 1
+              ) dar ON true
+      INNER JOIN
+              data_access_request dar_all ON dar_all.collection_id = c.collection_id
+              AND dar_all.submission_date IS NOT NULL
+              AND (LOWER(dar_all.data->>'status') != 'archived' OR dar_all.data->>'status' IS NULL)
       LEFT JOIN (
         SELECT election.*, MAX(election.election_id) OVER(PARTITION BY election.reference_id, election.dataset_id) AS latest
         FROM election
@@ -107,7 +120,9 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
       LEFT JOIN dataset dataset on dataset.dataset_id = dd.dataset_id
       LEFT JOIN dac dac on dac.dac_id = dataset.dac_id
       WHERE (e.latest = e.election_id OR e.election_id IS NULL)
-        AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL )
+      GROUP BY
+              c.collection_id, c.dar_code, dar.submission_date, dar.reference_id, u.display_name, i.institution_name,
+              e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id, dar.data, dac.name
       """)
   List<DarCollectionSummary> getDarCollectionSummariesForAdmin();
 
@@ -151,16 +166,11 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
         data_access_request dar_all ON dar_all.collection_id = c.collection_id
         AND dar_all.submission_date IS NOT NULL
         AND (LOWER(dar_all.data->>'status') != 'archived' OR dar_all.data->>'status' IS NULL)
-    LEFT JOIN
-        LATERAL (
-            SELECT
-                election.*,
-                MAX(election.election_id) OVER (PARTITION BY election.reference_id, election.dataset_id) AS latest
-            FROM
-                election
-            WHERE
-                LOWER(election.election_type) = 'dataaccess'
-        ) e ON e.reference_id = dar.reference_id
+    LEFT JOIN (
+        SELECT election.*, MAX(election.election_id) OVER(PARTITION BY election.reference_id, election.dataset_id) AS latest
+        FROM election
+        WHERE LOWER(election.election_type) = 'dataaccess'
+    ) AS e ON e.reference_id = dar.reference_id
     INNER JOIN
         dar_dataset dd ON dar.reference_id = dd.reference_id
     WHERE
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
@@ -115,36 +115,59 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
   @RegisterBeanMapper(value = DarCollection.class)
   @RegisterBeanMapper(value = Election.class)
   @UseRowReducer(DarCollectionSummaryReducer.class)
-  @SqlQuery
-      (
-          """
-              SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, u.display_name as researcher_name,
-                i.institution_name, e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id as dd_datasetid,
-                (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name,
-                (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'status' AS dar_status
-              FROM dar_collection c
-              INNER JOIN users u
-              ON u.user_id = c.create_user_id
-              LEFT JOIN institution i
-              ON i.institution_id = u.institution_id
-              INNER JOIN data_access_request dar
-              ON dar.collection_id = c.collection_id
-              LEFT JOIN (
-                SELECT election.*, MAX(election.election_id) OVER(PARTITION BY election.reference_id, election.dataset_id) AS latest
-                FROM election
-                WHERE LOWER(election.election_type) = 'dataaccess'
-              ) AS e
-              ON e.reference_id = dar.reference_id
-              INNER JOIN dar_dataset dd
-              ON dar.reference_id = dd.reference_id
-              WHERE c.create_user_id = :userId
-                AND (e.latest = e.election_id OR e.election_id IS NULL)
-                AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL )
-                AND (EXISTS (SELECT 1 FROM data_access_request WHERE (collection_id = c.collection_id and dar.submission_date is not null)))
-        """
-      )
-  List<DarCollectionSummary> getDarCollectionSummariesForResearcher(
-      @Bind("userId") Integer userId);
+  @SqlQuery("""
+    SELECT
+        c.collection_id AS dar_collection_id,
+        c.dar_code,
+        dar.submission_date,
+        dar.reference_id AS dar_reference_id,
+        u.display_name AS researcher_name,
+        i.institution_name,
+        e.election_id,
+        e.status,
+        e.dataset_id,
+        e.reference_id AS election_reference_id,
+        dd.dataset_id AS dd_datasetid,
+        (regexp_replace(dar.data #>> '{}', '\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name,
+        (regexp_replace(dar.data #>> '{}', '\\u0000', '', 'g'))::jsonb ->> 'status' AS dar_status,
+        ARRAY_AGG(dar_all.reference_id)::text[] AS reference_ids
+    FROM
+        dar_collection c
+    INNER JOIN
+        users u ON u.user_id = c.create_user_id
+    LEFT JOIN
+        institution i ON i.institution_id = u.institution_id
+    INNER JOIN
+        LATERAL (
+            SELECT *
+            FROM data_access_request
+            WHERE data_access_request.collection_id = c.collection_id
+            ORDER BY data_access_request.submission_date DESC
+            LIMIT 1
+        ) dar ON true
+    INNER JOIN
+        data_access_request dar_all ON dar_all.collection_id = c.collection_id
+    LEFT JOIN
+        LATERAL (
+            SELECT
+                election.*,
+                MAX(election.election_id) OVER (PARTITION BY election.reference_id, election.dataset_id) AS latest
+            FROM
+                election
+            WHERE
+                LOWER(election.election_type) = 'dataaccess'
+        ) e ON e.reference_id = dar.reference_id
+    INNER JOIN
+        dar_dataset dd ON dar.reference_id = dd.reference_id
+    WHERE
+        c.create_user_id = :userId
+        AND (e.latest = e.election_id OR e.election_id IS NULL)
+        AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)
+    GROUP BY
+        c.collection_id, c.dar_code, dar.submission_date, dar.reference_id, u.display_name, i.institution_name,
+        e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id, dar.data
+""")
+  List<DarCollectionSummary> getDarCollectionSummariesForResearcher(@Bind("userId") Integer userId);
 
 
   @RegisterBeanMapper(value = DarCollectionSummary.class)
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
@@ -119,14 +119,14 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
               AND dar_all.submission_date IS NOT NULL
               AND (LOWER(dar_all.data->>'status') != 'archived' OR dar_all.data->>'status' IS NULL)
           LEFT JOIN (
-              SELECT DISTINCT ON (reference_id, dataset_id) *
-              FROM election
-              WHERE LOWER(election_type) = 'dataaccess'
-              ORDER BY reference_id, dataset_id, election_id DESC
-          ) e ON e.reference_id = dar.reference_id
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -115,36 +115,59 @@ WHERE LOWER(election.election_type) = 'dataaccess'
...
+        u.display_name AS researcher_name,
+        i.institution_name,
+        e.election_id,
+        e.status,
+        e.dataset_id,
+        e.reference_id AS election_reference_id,
+        dd.dataset_id AS dd_datasetid,
+        (regexp_replace(dar.data #>> '{}', '\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name,
+        (regexp_replace(dar.data #>> '{}', '\\u0000', '', 'g'))::jsonb ->> 'status' AS dar_status,
+        ARRAY_AGG(dar_all.reference_id)::text[] AS reference_ids
```
rjohanek, line 133: line 133 is a change

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -115,36 +115,59 @@ WHERE LOWER(election.election_type) = 'dataaccess'
...
+    LEFT JOIN
+        institution i ON i.institution_id = u.institution_id
+    INNER JOIN
+        LATERAL (
+            SELECT *
+            FROM data_access_request
+            WHERE data_access_request.collection_id = c.collection_id
+            ORDER BY data_access_request.submission_date DESC
+            LIMIT 1
+        ) dar ON true
```
rjohanek: lines 141-148 is a change

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -115,36 +115,59 @@ WHERE LOWER(election.election_type) = 'dataaccess'
...
+        ) e ON e.reference_id = dar.reference_id
+    INNER JOIN
+        dar_dataset dd ON dar.reference_id = dd.reference_id
+    WHERE
+        c.create_user_id = :userId
+        AND (e.latest = e.election_id OR e.election_id IS NULL)
+        AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)
+    GROUP BY
+        c.collection_id, c.dar_code, dar.submission_date, dar.reference_id, u.display_name, i.institution_name,
+        e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id, dar.data
```
rjohanek, line 186: group by is an addition

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -115,36 +115,59 @@ WHERE LOWER(election.election_type) = 'dataaccess'
...
+        u.display_name AS researcher_name,
+        i.institution_name,
+        e.election_id,
+        e.status,
+        e.dataset_id,
+        e.reference_id AS election_reference_id,
+        dd.dataset_id AS dd_datasetid,
+        (regexp_replace(dar.data #>> '{}', '\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name,
+        (regexp_replace(dar.data #>> '{}', '\\u0000', '', 'g'))::jsonb ->> 'status' AS dar_status,
+        ARRAY_AGG(dar_all.reference_id) AS reference_ids
```
rjohanek, line 156: line 156 is an addition

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -115,36 +115,59 @@ WHERE LOWER(election.election_type) = 'dataaccess'
...
+        LATERAL (
+            SELECT *
+            FROM data_access_request
+            WHERE data_access_request.collection_id = c.collection_id
+            ORDER BY data_access_request.submission_date DESC
+            LIMIT 1
+        ) dar ON true
+    INNER JOIN
+        data_access_request dar_all ON dar_all.collection_id = c.collection_id
+    LEFT JOIN
```
rjohanek: line 149-151 is an addition

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -90,61 +90,102 @@ List<DarCollectionSummary> getDarCollectionSummariesForSO(
...
+              dar.reference_id AS dar_reference_id,
+              u.display_name AS researcher_name,
+              i.institution_name,
+              e.election_id,
+              e.status,
+              e.dataset_id,
+              dd.dataset_id AS dd_datasetid,
+              (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name,
+              dac.name AS dac_name,
+              ARRAY_AGG(dar_all.reference_id) AS reference_ids
```
rjohanek, line 106: line 106 is an addition

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -90,61 +90,102 @@ List<DarCollectionSummary> getDarCollectionSummariesForSO(
...
+              SELECT DISTINCT ON (collection_id) *
+              FROM data_access_request
+              WHERE submission_date IS NOT NULL
+              AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL)
+              ORDER BY collection_id, submission_date DESC
+          ) dar ON dar.collection_id = c.collection_id
+          INNER JOIN data_access_request dar_all
+              ON dar_all.collection_id = c.collection_id
+              AND dar_all.submission_date IS NOT NULL
+              AND (LOWER(dar_all.data->>'status') != 'archived' OR dar_all.data->>'status' IS NULL)
```
rjohanek, line 120: lines 111-116 is a modification and lines 117-120 is an addition

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -90,61 +90,102 @@ List<DarCollectionSummary> getDarCollectionSummariesForSO(
...
+                  WHERE LOWER(election.election_type) = 'dataaccess'
+                ) AS e ON e.reference_id = dar.reference_id
+          INNER JOIN dar_dataset dd ON dar.reference_id = dd.reference_id
+          LEFT JOIN dataset ON dataset.dataset_id = dd.dataset_id
+          LEFT JOIN dac ON dac.dac_id = dataset.dac_id
+          WHERE (e.latest = e.election_id OR e.election_id IS NULL)
+          GROUP BY
+              c.collection_id, c.dar_code, dar.submission_date, dar.reference_id,
+              u.display_name, i.institution_name, e.election_id, e.status,
+              e.dataset_id, dd.dataset_id, dar.data, dac.name
```
rjohanek, line 133: lines 130-133 is an addition

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -90,61 +90,102 @@ List<DarCollectionSummary> getDarCollectionSummariesForSO(
...
+    LEFT JOIN (
+        SELECT election.*, MAX(election.election_id) OVER(PARTITION BY election.reference_id, election.dataset_id) AS latest
+        FROM election
+        WHERE LOWER(election.election_type) = 'dataaccess'
+    ) AS e ON e.reference_id = dar.reference_id
+    INNER JOIN
+        dar_dataset dd ON dar.reference_id = dd.reference_id
+    WHERE
+        c.create_user_id = :userId
+        AND (e.latest = e.election_id OR e.election_id IS NULL)
```
rjohanek, line 183: reduced where clause

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -90,61 +90,102 @@ List<DarCollectionSummary> getDarCollectionSummariesForSO(
...
+         SELECT DISTINCT ON (collection_id) *
+         FROM data_access_request
+         WHERE submission_date IS NOT NULL
+         AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL)
+         ORDER BY collection_id, submission_date DESC
+    ) dar ON dar.collection_id = c.collection_id
+    INNER JOIN
+        data_access_request dar_all ON dar_all.collection_id = c.collection_id
+        AND dar_all.submission_date IS NOT NULL
+        AND (LOWER(dar_all.data->>'status') != 'archived' OR dar_all.data->>'status' IS NULL)
```
rjohanek, line 173: lines 164-168 is modification, 170-173 is addition

src/main/java/org/broadinstitute/consent/http/db/mapper/DarCollectionSummaryReducer.java
```diff
@@ -41,7 +41,9 @@ public void accumulate(Map<Integer, DarCollectionSummary> map, RowView rowView)
       try {
         darReferenceId = rowView.getColumn("dar_reference_id", String.class);
         if (Objects.nonNull(darReferenceId)) {
-          summary.addReferenceId(darReferenceId);
+          if (!summary.getReferenceIds().contains(darReferenceId)) {
```
pshapiro4broad: Since `DarCollectionSummary.referenceIds` is a `Set`, this check shouldn't be necessary.

src/test/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAOTest.java
```diff
@@ -623,6 +624,107 @@ void testGetDarCollectionSummaryForAdmin_ArchivedCollection() {
...
+    dataAccessRequestDAO.insertDARDatasetRelation(archivedDar.getReferenceId(), dataset.getDatasetId());
+    // draft DAR
+    dataAccessRequestDAO.updateDataByReferenceId(draftDar.getReferenceId(), draftDar.userId, new Date(), null,
+        new Date(), draftDar.getData());
+    // archived DAR
+    dataAccessRequestDAO.archiveByReferenceIds(List.of(archivedDar.getReferenceId()));
+
+    List<DarCollectionSummary> summaries = switch (type) {
+      case "admin" -> darCollectionSummaryDAO.getDarCollectionSummariesForAdmin();
+      case "researcher" -> darCollectionSummaryDAO.getDarCollectionSummariesForResearcher(userId);
```
pshapiro4broad, line 654: Since the test doesn't create any collections using a different user, the result of `admin` and `researcher` is the same. It would be a more interesting test if two different users are used, to show that the `admin` can see all records and the `researcher` only gets records for that user id.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -90,61 +90,102 @@ List<DarCollectionSummary> getDarCollectionSummariesForSO(
...
+        (regexp_replace(dar.data #>> '{}', '\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name,
+        (regexp_replace(dar.data #>> '{}', '\\u0000', '', 'g'))::jsonb ->> 'status' AS dar_status,
+        ARRAY_AGG(dar_all.reference_id) AS reference_ids
+    FROM
+        dar_collection c
+    INNER JOIN
+        users u ON u.user_id = c.create_user_id
+    LEFT JOIN
+        institution i ON i.institution_id = u.institution_id
+    INNER JOIN (
```
pshapiro4broad, line 163: If some of these blocks are identical, could this query be built by combining the identical blocks? That would make it easier to see what's the same and what's different for each case. And it would make it easier to change all the related queries at once. For example, see `DarCollectionDAO.getCollectionAndDars`.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -90,61 +90,102 @@ List<DarCollectionSummary> getDarCollectionSummariesForSO(
...
+    INNER JOIN
+        dar_dataset dd ON dar.reference_id = dd.reference_id
+    WHERE
+        c.create_user_id = :userId
+        AND (e.latest = e.election_id OR e.election_id IS NULL)
+    GROUP BY
+        c.collection_id, c.dar_code, dar.submission_date, dar.reference_id, u.display_name, i.institution_name,
+        e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id, dar.data
+""")
+  List<DarCollectionSummary> getDarCollectionSummariesForResearcher(@Bind("userId") Integer userId);
```
pshapiro4broad, line 188: This method is called `getDarCollectionSummariesForResearcher()` but it doesn't seem to be specific to the researcher role. Maybe it should be called `getDarCollectionSummariesForUser()`?

src/test/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAOTest.java
```diff
@@ -623,6 +624,107 @@ void testGetDarCollectionSummaryForAdmin_ArchivedCollection() {
...
+
+    // Insert dataset relations for both DARs
+    // the older DAR has two datasets and the newer DAR has one
+    dataAccessRequestDAO.insertDARDatasetRelation(olderDar.getReferenceId(), dataset.getDatasetId());
+    dataAccessRequestDAO.insertDARDatasetRelation(olderDar.getReferenceId(), dataset1.getDatasetId());
+    dataAccessRequestDAO.insertDARDatasetRelation(newerDar.getReferenceId(), dataset.getDatasetId());
+
+    // Create an election for the older DAR
+    createElection(ElectionType.DATA_ACCESS.getValue(),
+        ElectionStatus.OPEN.getValue(),
```
pshapiro4broad: Maybe not a change for this PR, but `createElection()` should accept enum types not `String`
```suggestion
    createElection(ElectionType.DATA_ACCESS,
        ElectionStatus.OPEN,
```

src/test/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAOTest.java
```diff
@@ -623,6 +624,107 @@ void testGetDarCollectionSummaryForAdmin_ArchivedCollection() {
...
+    assertNotNull(summaries);
+    assertEquals(1, summaries.size());
+    DarCollectionSummary summary = summaries.get(0);
+    assertEquals(collectionId, summary.getDarCollectionId());
+
+    // Ensure the reference IDs include only the non-draft non-archived DAR
+    assertNotNull(summary.getReferenceIds());
+    assertEquals(1, summary.getReferenceIds().size());
+    assertTrue(summary.getReferenceIds().contains(olderDar.getReferenceId()));
+    assertFalse(summary.getReferenceIds().contains(draftDar.getReferenceId()));
```
pshapiro4broad: Since there's only one element you don't need to `assertFalse()`
```suggestion
    assertTrue(summary.getReferenceIds().contains(olderDar.getReferenceId()));
```

src/test/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAOTest.java
```diff
@@ -623,6 +624,107 @@ void testGetDarCollectionSummaryForAdmin_ArchivedCollection() {
...
+    assertNotNull(summaries);
+    assertEquals(1, summaries.size());
+    DarCollectionSummary summary = summaries.get(0);
+    assertEquals(collectionId, summary.getDarCollectionId());
+
+    // Ensure the reference IDs include both DARs
+    assertNotNull(summary.getReferenceIds());
+    assertEquals(2, summary.getReferenceIds().size());
+    assertTrue(summary.getReferenceIds().contains(olderDar.getReferenceId()));
+    assertTrue(summary.getReferenceIds().contains(newerDar.getReferenceId()));
```
pshapiro4broad: You could use `containsAll()` here
```suggestion
    assertTrue(summary.getReferenceIds().containsAll(List.of(olderDar.getReferenceId(), newerDar.getReferenceId())));
```

pshapiro4broad (approved):
Looks OK as far as I can tell. It would be nice to replace duplicate query fragments with a constant if possible.

src/test/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAOTest.java
rjohanek: I feel like that case (an admin seeing multiple users' collections) is covered in other tests, here I just wanted to show the new test cases of having multiple DARs in a collection

pshapiro4broad: I'm more interested in verifying that the `researcher` API only returns that researcher's data, not that admin can see everything. Since the admin API doesn't accept a user ID, it seems likely that it's working as expected, vs the researcher API, which we currently don't know if it's correctly filtering or not.

src/test/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAOTest.java
rjohanek: I'll make a ticket

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
rjohanek: I was talking to Elliot and thought that we couldn't do something like that because of SemGrep

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
rjohanek: It's true it could be used for any user, currently the only use case is for researchers but I'm open to changing the name

src/test/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAOTest.java
rjohanek: There are other tests for that case already, such as testGetDarCollectionSummaryForResearcher and testGetDarCollectionSummaryForResearcher_NoElectionsPresent

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
rushtong: Naming is a good point. In fact, this is the method hit when any user, admin, chair, member, or SO, goes to the researcher console to find their DARs, so the intention is that the results are eventually shown on a specific console page.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
rushtong: Semgrep complains about any kind of [string concatenation in queries](https://semgrep.dev/r?q=java.jboss.security.session_sqli.find-sql-string-concatenation)

rushtong: I suspect that we could use postgres functions to alleviate the pain here, but haven't looked into it enough to be confident it is a solution.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -90,61 +90,102 @@ List<DarCollectionSummary> getDarCollectionSummariesForSO(
...
+    FROM
+        dar_collection c
+    INNER JOIN
+        users u ON u.user_id = c.create_user_id
+    LEFT JOIN
+        institution i ON i.institution_id = u.institution_id
+    INNER JOIN (
+         SELECT DISTINCT ON (collection_id) *
+         FROM data_access_request
+         WHERE submission_date IS NOT NULL
```
rushtong, line 166: I believe this will filter out drafts from the result. For the researcher console, we need to show the user all of their DARs, both those submitted and those still in process.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -90,61 +90,102 @@ List<DarCollectionSummary> getDarCollectionSummariesForSO(
...
+         AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL)
+         ORDER BY collection_id, submission_date DESC
+    ) dar ON dar.collection_id = c.collection_id
+    INNER JOIN
+        data_access_request dar_all ON dar_all.collection_id = c.collection_id
+        AND dar_all.submission_date IS NOT NULL
```
rushtong, line 172: Similar filter issue here - researchers need to see all of their DARs.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
rjohanek: Drafts are added in at the service level, this is not a change to the functionality see daoCollectionService.getSummariesForRoleName

rushtong: Gotcha, was just testing this locally, you're right.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
rjohanek: this dao method is only called for researchers. each of those specified roles has their own dao method

src/test/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAOTest.java
```diff
@@ -623,6 +624,105 @@ void testGetDarCollectionSummaryForAdmin_ArchivedCollection() {
...
+
+    // Ensure the election from the older DAR is not included
+    assertTrue(summary.getElections().isEmpty());
+
+    // Ensure the summary represents the most recently submitted DAR
+    assertEquals(newerDar.getSubmissionDate(), summary.getSubmissionDate());
+    assertEquals(newerDar.getExpiresAt(), summary.getExpiresAt());
+    // should only be one dataset because the newer DAR has one dataset
+    assertEquals(1, summary.getDatasetIds().size());
+    assertTrue(summary.getDatasetIds().contains(dataset.getDatasetId()));
```
rushtong, line 723: Nice test condition

rushtong (approved):
This looks pretty reasonable given the complexity. I do think we could perhaps reduce that with a new db function approach we haven't explored yet, so in the meantime, moving forward with this makes the most sense 👍🏽

src/main/java/org/broadinstitute/consent/http/db/mapper/DarCollectionSummaryReducer.java
rjohanek: I made the set mutable, so now I can remove the check and it won't throw a 409

PR #2505, opened 2025-04-29
[DT-1573] Email notification for DAR expiration

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-1573](https://broadworkbench.atlassian.net/browse/DT-1573)

### Summary
This PR does two things.
Primary function: adds a resource to trigger DAR Expiring Soon and Expired notifications.
Secondary function: Since this will be invoked by a Resource, a new "Service Account" role has been added to the consent.  This role can now be added to Resources to limit endpoint to those users with a service account role.

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -59,28 +59,28 @@ List<DarCollectionSummary> getDarCollectionSummariesForDAC(
...
-                WHERE LOWER(election.election_type) = 'dataaccess'
-                ) AS e
-              ON e.reference_id = dar.reference_id
-              INNER JOIN dar_dataset dd
-              ON dar.reference_id = dd.reference_id
-              WHERE u.institution_id = :institutionId
-                AND (e.latest = e.election_id OR e.election_id IS NULL)
-                AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL )
-          """
+                  SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, u.display_name as researcher_name,
```
otchet-broad: Formatting.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -118,30 +118,30 @@ WHERE LOWER(election.election_type) = 'dataaccess'
...
-              ) AS e
-              ON e.reference_id = dar.reference_id
-              INNER JOIN dar_dataset dd
-              ON dar.reference_id = dd.reference_id
-              WHERE c.create_user_id = :userId
-                AND (e.latest = e.election_id OR e.election_id IS NULL)
-                AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL )
-                AND (EXISTS (SELECT 1 FROM data_access_request WHERE (collection_id = c.collection_id and dar.submission_date is not null)))
-        """
+                    SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, u.display_name as researcher_name,
```
otchet-broad: Formatting

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -193,29 +193,29 @@ DarCollectionSummary getDarCollectionSummaryForDACByCollectionId(
...
-                WHERE LOWER(election.election_type) = 'dataaccess'
-              ) AS e
-              ON e.reference_id = dar.reference_id
-              INNER JOIN dar_dataset dd
-              ON dar.reference_id = dd.reference_id
-              WHERE c.collection_id = :collectionId
-                AND (e.latest = e.election_id OR e.election_id IS NULL)
-                AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL )
-          """
+                  SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, u.display_name as researcher_name,
```
otchet-broad: Formatting

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -104,10 +105,14 @@ WHERE dar.submission_date > now() - interval '1 year'
                 dd.dataset_id
               FROM data_access_request dar
               LEFT JOIN dar_dataset dd ON dd.reference_id = dar.reference_id
-              WHERE dar.submission_date BETWEEN SYMMETRIC :begin AND :end
+              LEFT OUTER JOIN email_entity email ON email.entity_reference_id = dar.reference_id AND email.email_type = :emailType
```
otchet-broad, line 111: Adjusts the query to make it possible to find DARs that are expiring (not before a specific date) and notify them that they're either expiring soon or have expired.

src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
```diff
@@ -15,14 +15,16 @@ public enum UserRoles {
...
-  ITDIRECTOR(Resource.ITDIRECTOR, 9);
+  ITDIRECTOR(Resource.ITDIRECTOR, 9),
+  SERVICE_ACCOUNT(Resource.SERVICE_ACCOUNT, 10);
 
-  private final String roleName;
-  private final Integer roleId;
   private static final Set<UserRoles> NON_DAC_ROLES =
-      Set.of(ALUMNI, ADMIN, RESEARCHER, SIGNINGOFFICIAL, DATASUBMITTER, ITDIRECTOR);
+      Set.of(ALUMNI, ADMIN, RESEARCHER, SIGNINGOFFICIAL, DATASUBMITTER, ITDIRECTOR,
+          SERVICE_ACCOUNT);
```
otchet-broad, line 23: New role for SERVICE_ACCOUNT

src/main/java/org/broadinstitute/consent/http/enumeration/UserRoles.java
```diff
@@ -105,4 +103,12 @@ public static boolean isValidSoAdjustableRoleId(UserRoles role) {
     return SO_AUTHORIZED_ROLES_TO_ADJUST.contains(role);
   }
 
+  public String getRoleName() {
```
otchet-broad, line 106: Reorganizing the class.

src/main/java/org/broadinstitute/consent/http/resources/EmailNotifierResource.java
```diff
@@ -29,4 +30,17 @@ public Response sendReminderMessage(@PathParam("voteId") String voteId) {
       return createExceptionResponse(e);
     }
   }
+
+  @GET
+  @Path("/darExpirationNotices")
+  @RolesAllowed({SERVICE_ACCOUNT})
+  public Response sendDarExpirationNotices() {
+    try {
+      emailService.sendExpirationNotices();
```
otchet-broad, line 39: Simple, blocking call to service.

src/main/java/org/broadinstitute/consent/http/resources/EmailNotifierResource.java
```diff
@@ -29,4 +30,17 @@ public Response sendReminderMessage(@PathParam("voteId") String voteId) {
       return createExceptionResponse(e);
     }
   }
+
+  @GET
+  @Path("/darExpirationNotices")
+  @RolesAllowed({SERVICE_ACCOUNT})
```
otchet-broad, line 36: Limited to just SERVICE_ACCOUNT callers.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -37,9 +38,10 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
   @UseRowReducer(DataAccessRequestReducer.class)
   @SqlQuery(
       """
-          SELECT dd.dataset_id, dar.id, dar.reference_id, dar.collection_id, dar.parent_id, dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date,
+          SELECT collection.dar_code, dd.dataset_id, dar.id, dar.reference_id, dar.collection_id, dar.parent_id, dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date,
             (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data FROM data_access_request dar
           LEFT JOIN dar_dataset dd on dd.reference_id = dar.reference_id
+          LEFT JOIN dar_collection collection on collection.collection_id = dar.collection_id
```
otchet-broad, line 44: Joining to enrich with dar_code

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -64,8 +66,9 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
               SELECT dar.id, dar.reference_id, dar.collection_id, dar.parent_id,
                 dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date,
                 (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data,
-                dd.dataset_id
+                dd.dataset_id, collection.dar_code
               FROM data_access_request dar
+              LEFT JOIN dar_collection collection on collection.collection_id = dar.collection_id
```
otchet-broad, line 71: Joining to enrich with dar_code

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -101,13 +104,18 @@ WHERE dar.submission_date > now() - interval '1 year'
...
```
otchet-broad, line 109: Joining to enrich with dar_code

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -118,7 +126,8 @@ List<DataAccessRequest> findSubmittedDarsByTimeRange(@Bind("begin") Timestamp be
   @SqlQuery(
       """
               SELECT dd.dataset_id, dar.id, dar.reference_id, dar.collection_id, dar.parent_id, dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date,
-                (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data FROM data_access_request dar
+                (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data, collection.dar_code FROM data_access_request dar
+              LEFT JOIN dar_collection collection on collection.collection_id = dar.collection_id
```
otchet-broad, line 130: Joining to enrich with dar_code

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -135,7 +144,8 @@ List<DataAccessRequest> findSubmittedDarsByTimeRange(@Bind("begin") Timestamp be
...
```
otchet-broad, line 148: Joining to enrich with dar_code

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -173,7 +164,8 @@ List<DataAccessRequest> findSubmittedDarsByTimeRange(@Bind("begin") Timestamp be
...
```
otchet-broad, line 168: Joining to enrich with dar_code

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -190,7 +182,8 @@ List<DataAccessRequest> findSubmittedDarsByTimeRange(@Bind("begin") Timestamp be
   @SqlQuery(
       """
           SELECT dd.dataset_id, dar.id, dar.reference_id, dar.collection_id, dar.parent_id, dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date,
-            (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data FROM data_access_request dar
+            (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data, collection.dar_code FROM data_access_request dar
+          LEFT JOIN dar_collection collection on collection.collection_id = dar.collection_id
```
otchet-broad, line 186: Joining to enrich with dar_code

src/main/java/org/broadinstitute/consent/http/enumeration/EmailType.java
```diff
@@ -21,7 +21,8 @@ public enum EmailType {
   NEW_DAA_UPLOAD_SO(17, "new-daa-upload-signing-official.html"),
   DATASET_DENIED(18, "dataset-denied.html"),
   DATASET_APPROVED(19, "dataset-approved.html"),
-  DAR_EXPIRED(20, "dar-expired.html");
+  DAR_EXPIRED(20, "dar-expired.html"),
+  DAR_EXPIRING_SOON(21);
```
otchet-broad: Placeholder, awaiting @fboulnois  branch.

src/main/java/org/broadinstitute/consent/http/mail/message/DarExpiredMessage.java
```diff
@@ -7,10 +7,12 @@
 public class DarExpiredMessage extends MailMessage {
 
   private final String darCode;
+  private final String referenceId;
 
-  public DarExpiredMessage(User toUser, String darCode) {
+  public DarExpiredMessage(User toUser, String darCode, String referenceId) {
```
otchet-broad, line 12: Update to pass in referenceId since dar_code is reused by progress reports.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
```diff
@@ -42,6 +42,9 @@ public class DataAccessRequest {
   @JsonProperty
   public DataAccessRequestData data;
 
+  @JsonProperty
+  public String darCode;
```
otchet-broad, line 46: Field to hold the dar_code value.

src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
```diff
@@ -906,6 +906,16 @@ void insertUserRoleAndInstitution_roleOnly() {
     verify(userRoleDAO).insertSingleUserRole(role.getRoleId(), testUser.getUserId());
   }
 
+  @Test
+  void insertServiceAccountUserRole() {
```
otchet-broad, line 910: Verify the service account can be assigned.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
pshapiro4broad: Can you revert the format-only changes, and commit them in a separate PR? Or is there a reason why they need to be done now?

otchet-broad: I can revert this file.

src/main/java/org/broadinstitute/consent/http/resources/EmailNotifierResource.java
pshapiro4broad, line 39: There are many ways to think about this, but I had been thinking that this operation was a `DataAccessRequestResource` operation. The operation is "check for expired or expiring requests". Depending on the check results, some emails may be sent (and some data in the DAR may be updated); but IMO this is primarily a DAR operation, not an email one.

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
```diff
@@ -233,6 +241,71 @@ private void sendNewDARRequestEmail(
...
+    });
+  }
+
+  private void sendDARExpirationReminderNotices() {
+    EmailType emailType = EmailType.DAR_EXPIRATION_REMINDER;
+    String interval = "11 months";
+    String noEmailForUserToWarnFoundLogTemplate = "User %d (%s) not found for expiring warning.  DAR reference id: %s";
+    // Per value in ticket DT-1573
+    Timestamp minimumSubmittedDateForExpirations = Timestamp.from(Instant.ofEpochSecond(
+        LocalDate.of(2024, 9, 30).toEpochSecond(LocalTime.of(0, 0, 0, 0), ZoneOffset.UTC)));
```
pshapiro4broad: Can this be a named constant, so it can be shared with both code branches?

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
```diff
@@ -233,6 +241,71 @@ private void sendNewDARRequestEmail(
         user.getUserId());
   }
 
+  public void sendExpirationNotices() {
+    sendDARExpirationReminderNotices();
+    sendDARExpirationNotices();
+  }
+
+  private void sendDARExpirationNotices() {
```
pshapiro4broad, line 256: Instead of adding this to EmailService, I've been thinking that these should really be in their specific service classes. In this case, I would put the methods in `DataAccessRequestService`

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
```diff
@@ -233,6 +241,71 @@ private void sendNewDARRequestEmail(
...
+        String darCode = expiredDar.getDarCode();
+        String userName = user.getDisplayName();
+        if (user.getEmail() == null) {
+          // Do not throw here.  Log information about the DAR since this will continue
+          // to appear broken until manual intervention is taken to resolve the missing user
+          // email address
+          logWarn(String.format(noEmailForUserToWarnFoundLogTemplate,
+              expiredDar.getUserId(), userName, referenceId));
+        } else {
+          sendDarExpirationReminderMessage(user, darCode, user.getUserId(), referenceId);
```
pshapiro4broad: Except for `emailType`, `interval`, and the type of `MailMessage` created, these methods appear to be the same. If so, can they be refactored to avoid this duplication?

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
```diff
@@ -233,6 +241,71 @@ private void sendNewDARRequestEmail(
...
+        }
+      } catch (Exception e) {
+        logException(e);
```
pshapiro4broad, line 292: This code logs; do we expect to do anything with these logs? Since the API request is blocking, we could collect the errors, send them back in the response and log them in the cron service, where they might be more likely to be seen. The response could also include the number of emails sent of each type, etc.

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
```diff
@@ -233,6 +241,71 @@ private void sendNewDARRequestEmail(
...
+        LocalDate.of(2024, 9, 30).toEpochSecond(LocalTime.of(0, 0, 0, 0), ZoneOffset.UTC)));
+    List<DataAccessRequest> expiredDars = dataAccessRequestDAO.findAgedDARsByEmailTypeOlderThanInterval(
+        emailType.getTypeInt(), interval, minimumSubmittedDateForExpirations);
+    expiredDars.forEach(expiredDar -> {
+      try {
+        String referenceId = expiredDar.getReferenceId();
+        User user = userDAO.findUserById(expiredDar.getUserId());
+        String darCode = expiredDar.getDarCode();
+        String userName = user.getDisplayName();
+        if (user.getEmail() == null) {
```
pshapiro4broad, line 275: Under what circumstances can a user's email be `null`? `UserService.validateRequiredFields()` blocks user creation with an empty email and it's the only path I see to `UserDAO.insertUser()`.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
pshapiro4broad: I assumed that a DAR would have a dar_code field directly. Since it doesn't, is `darCode` needed for email sending? Or could `referenceId` be used instead?

src/test/java/org/broadinstitute/consent/http/authentication/UserAuthorizerTest.java
```diff
@@ -43,8 +43,14 @@ void testAuthorizeNotAuthorized() {
...
-      .thenReturn(List.of(UserRoles.CHAIRPERSON.getRoleName()));
+        .thenReturn(List.of(UserRoles.CHAIRPERSON.getRoleName()));
     assertTrue(authorizer.authorize(authorizedUser, Resource.CHAIRPERSON, context));
   }
 
+  @Test
+  void testAuthorizedServiceAccount() {
+    when(userRoleDAO.findRoleNamesByUserEmail(any())).thenReturn(
+        List.of(UserRoles.SERVICE_ACCOUNT.getRoleName()));
+    assertTrue(authorizer.authorize(authorizedUser, Resource.SERVICE_ACCOUNT, context));
```
pshapiro4broad, line 54: Instead of mocking `AuthUser`, why not create one? I think it's better to not mock POJOs if we can avoid it.
```suggestion
    AuthUser user = new AuthUser("email");
    when(userRoleDAO.findRoleNamesByUserEmail(user.getEmail()))
        .thenReturn(List.of(UserRoles.SERVICE_ACCOUNT.getRoleName()));
    assertTrue(authorizer.authorize(user, Resource.SERVICE_ACCOUNT, context));
```

src/test/java/org/broadinstitute/consent/http/mail/message/DarExpirationReminderMessageTest.java
```diff
@@ -33,11 +34,12 @@ String getElementTextById(Document document, String id) {
   void testGetDarExpiredTemplate() throws Exception {
     String userName = randomAlphabetic(10);
     String darCode = randomAlphabetic(10);
+    String darReferenceId = UUID.randomUUID().toString();
```
pshapiro4broad, line 37: Is there a reason why this needs to be a UUID?

src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
pshapiro4broad: Why test this role in particular? We don't have tests for the other roles, except `UserRoles.Researcher()` which is tested in `insertUserRoleAndInstitution_roleOnly()`.

If the intent is to test the database contents, as created by liquibase migrations (vs the behavior of `UserService`), then the test should loop over all the possible roles and test each.

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
```diff
@@ -129,9 +131,9 @@ void updateNonDraftToDraft() {
     List<DataAccessRequest> draftDars2 = dataAccessRequestDAO.findAllDraftDataAccessRequests();
     assertFalse(draftDars2.isEmpty());
     assertEquals(1, draftDars2.size());
+    assertNotNull(draftDars2.get(0).getDarCode());
```
pshapiro4broad: Do we know what the expected value of dar code is here, to use `assertEquals()` instead?

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
```diff
@@ -226,6 +229,9 @@ void testFindByReferenceIds() {
     assertNotNull(dars);
     assertFalse(dars.isEmpty());
     assertEquals(3, dars.size());
+    dars.forEach(dar->{
+      assertNotNull(dar.getDarCode());
+    });
```
pshapiro4broad: whitespace
```suggestion
    dars.forEach(dar -> {
      assertNotNull(dar.getDarCode());
    });
```

pshapiro4broad (commented):
Some comments so far, still looking at the test code.

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
otchet-broad: The value in darCode is the promoted way for users to think about these things.  The juice is definitely worth the squeeze on this one.

src/test/java/org/broadinstitute/consent/http/mail/message/DarExpirationReminderMessageTest.java
otchet-broad: It doesn't need to be a UUID.  It needs to be a string.  The code generating this value is using this same method to produce this string, so I'm using it here.

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
otchet-broad: I think we're going to need to do things with our logs, regardless of what is causing entries in them or which logger created them.
I like the idea of logging the metrics.  I think the program could benefit more from doing that overall.

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
otchet-broad: I had originally refactored this to do what you're suggesting only to move back to this because of different methods being used based on email type.  I could probably switch based on the value, but I'm not sure it will improve readability or reduce the lines of code in any meaningful way.

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
otchet-broad: It's still not a required parameter on the constructor.  If it's the case that no User object can be built in our future code, then shouldn't User be modified so it requires the attributes in all of the constructors?

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
otchet-broad: I think I like it with the refactor.  Let me know what you think.

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
otchet-broad: I think we might want to do this on a holistic level.  These aren't the only methods like this in the class like this.  I'd support a ticket to do that kind of refactor.

src/main/java/org/broadinstitute/consent/http/resources/EmailNotifierResource.java
otchet-broad: Funny, I think of this as "send emails to the people with DARs that are expired or expiring soon" and appreciate your perspective.  I could support a refactor ticket that moves both this method and the `sendReminderMessage` above it to their respective services to eliminate the need for this endpoint all together.

src/test/java/org/broadinstitute/consent/http/authentication/UserAuthorizerTest.java
otchet-broad: I reduced your suggestion SLOC by 33%. :-). No need to create the AuthUser at all.

src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
otchet-broad: It's to test that we can create a service account without an institution because we shouldn't have to assign the institution for this path to function in our code.  This applies the `least privilege` principle to the service account user.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -101,13 +104,18 @@ WHERE dar.submission_date > now() - interval '1 year'
...
-                dd.dataset_id
+                dd.dataset_id, collection.dar_code
               FROM data_access_request dar
+              LEFT JOIN dar_collection collection on collection.collection_id = dar.collection_id
               LEFT JOIN dar_dataset dd ON dd.reference_id = dar.reference_id
-              WHERE dar.submission_date BETWEEN SYMMETRIC :begin AND :end
+              LEFT OUTER JOIN email_entity email ON email.entity_reference_id = dar.reference_id AND email.email_type = :emailType
+              WHERE dar.submission_date >= :notBefore
+              AND (dar.submission_date < now() - :interval ::interval)
+              AND (email.email_type IS NULL)
```
rushtong, line 114: I'm confused by this restriction on null email type. How does this work if we're passing in an email type? The left outer join on line 111 along with this condition is counter-intuitive.

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
rushtong: It **should** always be true that email is non-null. Conceptually, it is the root identity of any DUOS user. If we're not enforcing that, then we should be.

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
rushtong: The `logException()` code path triggers a sentry event so we'll see all errors in slack.

src/main/resources/assets/api-docs.yaml
```diff
@@ -529,7 +529,21 @@ paths:
...
+        - Notifier
+      responses:
+        200:
+          description: Notification process started.
+        401:
+          description: Not authenticated.
+        403:
+          description: Not authorized.
+        500:
+          description: Server error.
```
rushtong: minor nit - we've been trying to reduce the size of this doc over time and using a `$ref` to point to the method descriptions.

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
```diff
@@ -41,6 +42,20 @@
 @ExtendWith(MockitoExtension.class)
 class DataAccessRequestDAOTest extends DAOTestHelper {
 
+  private static DataAccessRequestData createDataAccessRequestData(String darCode) {
+    DataAccessRequestData data = new DataAccessRequestData();
+    data.setProjectTitle("Project Title: " + RandomStringUtils.random(50, true, false));
```
rushtong, line 49: nit: we've been replacing these deprecated `RandomStringUtils` methods with updated versions in `AbstractTestHelper`. This is static method so we could make the helper methods static as well.

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
```diff
@@ -41,6 +42,20 @@
...
+    data.setDarCode(darCode);
```
rushtong, line 50: It would be nice to get away from this deprecated method. Commenting it out, tests pass for me locally.

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
rushtong: Looking at this test, it is very out of date. The method we're using here (`updateDataByReferenceId`) to try and turn a draft into a submitted DAR is not the right method. So, in this test scenario, we do not actually know what the real dar code should be. It's calculated when we create a Dar Collection, which this test method doesn't do.

rushtong: Update to previous comment - we actually are creating a Dar Collection in `createDataAccessRequestV3` so we should know what the dar code is, it should be `dar.getDarCode()`

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
```diff
@@ -826,34 +814,203 @@ void createProgressReport() {
...
+    createDataAccessRequest(collectionOneId, userOneId,
+        new Date(Instant.now().minus(360,
+            ChronoUnit.DAYS).toEpochMilli()));
+    createDataAccessRequest(collectionTwoId, userOneId,
+        new Date(Instant.now().minus(15,
+            ChronoUnit.DAYS).toEpochMilli()));
+
+    List<DataAccessRequest> dars = dataAccessRequestDAO.findAgedDARsByEmailTypeOlderThanInterval(
+        EmailType.DAR_EXPIRATION_REMINDER.getTypeInt(),
+        "11 months", Timestamp.from(Instant.now().minus(365, ChronoUnit.DAYS)));
```
rushtong: I wonder if there's a better way to codify the interval string. Maybe something like a custom Interval object ... `new DuosInterval(11, ChronoUnit.MONTHS.name).toString()`?

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
```diff
@@ -233,6 +245,54 @@ private void sendNewDARRequestEmail(
...
+          // Do not throw here.  Log information about the DAR since this will continue
+          // to appear broken until manual intervention is taken to resolve the missing user
+          // email address
+          logWarn(String.format(
```
rushtong: I think it might be nice to log this to Sentry as well so we can be notified via Slack. We don't have a warn level sentry logger, but I think this is an example where we would need that. Potential implementation:
```
  default void logSentryWarning(String message) {
    logWarn(message);
    SentryEvent event = new SentryEvent();
    event.setLevel(SentryLevel.WARNING);
    Message sentryMessage = new Message();
    sentryMessage.setMessage(message);
    event.setMessage(sentryMessage);
    Sentry.captureEvent(new SentryEvent());
  }
```

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
otchet-broad: I use the email type in the left outer join to enrich the result set with the messages we've sent matching the email type, then this code filters those rows because we've already sent that type of message.  The net effect is that we get a list of DARs we haven't sent the emailType notice to yet.  :-)

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
otchet-broad: We should make a ticket.  I've got a test that shows that we don't enforce this in code.  It will impact a number items across the codebase.

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
otchet-broad: Would making this a logError event satisfy that need just as well?

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
rushtong: > The net effect is that we get a list of DARs we haven't sent the emailType notice to yet. :-)

Thank you, that makes sense 👍🏽

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
rushtong: We don't have a `logError` currently, but if we did and it did something similar, then yes.

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
pshapiro4broad: While I agree that other methods could also be moved, I don't see why that means that these new methods need to be added here. While some methods here make DAO or other calls (like `sendReminderMessage()`) many others do not, such as `sendDataCustodianApprovalMessage()`, `sendDatasetSubmittedMessage()`, etc.

src/main/java/org/broadinstitute/consent/http/resources/EmailNotifierResource.java
pshapiro4broad: > Funny, I think of this as "send emails to the people with DARs that are expired or expiring soon" and appreciate your perspective.

Another thing to consider is that, in the future, we may decide that we want to do something other than sending an email when things are expired, or do additional work that's not directly email related.

> I could support a refactor ticket

If it will be refactored in the future, why not do it now before the code is initially added? I think we should avoid creating tech debt in a PR if possible.

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
pshapiro4broad: > We should make a ticket. I've got a test that shows that we don't enforce this in code. It will impact a number items across the codebase.

I'd be interested to see that test. The only method I see to create a user is `UserDAO.insertUser()`, and the only use of that is in `UserService.createUser()`, which is checking `User.email`.

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
pshapiro4broad: Or use `logException()` and create an exception here

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
```diff
@@ -233,6 +245,54 @@ private void sendNewDARRequestEmail(
...
+              "Email address for user %d (%s) not found for expiring warning.  DAR reference id: %s",
+              expiredDar.getUserId(), userName, referenceId));
+        } else {
+          switch (type) {
+            case DAR_EXPIRATION_REMINDER:
+              sendDarExpirationReminderMessage(user, darCode, user.getUserId(), referenceId);
+              break;
+            case DAR_EXPIRED:
+              sendDarExpiredMessage(user, darCode, user.getUserId(), referenceId);
+          }
```
pshapiro4broad, line 289: This is better but I would go one step further, and pass in the message constructor as an argument.
```
  interface MessageCreator {
    MailMessage createMessage(User user, String darCode, String referenceId);
  }

  private void sendDARExpirationNotices() {
    EmailType emailType = EmailType.DAR_EXPIRED;
    String interval = "1 year";
    sendDARMessageToList(emailType, interval, DarExpirationReminderMessage::new);
  }

  private void sendDARExpirationReminderNotices() {
    EmailType emailType = EmailType.DAR_EXPIRATION_REMINDER;
    String interval = "11 months";
    sendDARMessageToList(emailType, interval, DarExpiredMessage::new);
  }

  private void sendDARMessageToList(EmailType type, String interval, MessageCreator messageCreator) {
  [...]
            if (user.getEmail() == null) {
              // Do not throw here.  Log information about the DAR since this will continue
              // to appear broken until manual intervention is taken to resolve the missing user
              // email address
              logWarn(
                  String.format(
                      "Email address for user %d (%s) not found for expiring warning.  DAR reference id: %s",
                      expiredDar.getUserId(), userName, referenceId));
            } else {
              sendMessage(
                  messageCreator.createMessage(user, darCode, referenceId), user.getUserId());
            }
```
And then delete `sendDarExpiredMessage()` and `sendDarExpirationReminderMessage()`

src/test/java/org/broadinstitute/consent/http/authentication/UserAuthorizerTest.java
pshapiro4broad: The problem is that this doesn't validate the argument to `findRoleNamesByUserEmail()`. If the method being tested passes a different string to that method, this test doesn't detect that error. Creating an object with a value to verify allows us to make sure that this part of the code is working as expected.

Although I don't see what coverage this provides over `testAuthorizeAuthorized()`. If you want to test all roles, you could make this a parameterized test that tests all `UserRoles` members. But since there's no difference in how these values are handled, I don't think there's a benefit for that additional testing.

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
```diff
@@ -554,4 +557,112 @@ void sendDarExpirationReminderMessage() throws Exception {
...
+
+    when(userDAO.findUserById(user1.getUserId())).thenReturn(user1);
+    when(userDAO.findUserById(user2.getUserId())).thenReturn(user2);
+    List<DataAccessRequest> dars = List.of(dar1, dar2);
+    when(dataAccessRequestDAO.findAgedDARsByEmailTypeOlderThanInterval(any(), any(), any())).thenReturn(dars);
+    when(templateHelper.getTemplate(EmailType.DAR_EXPIRATION_REMINDER.templateName)).thenReturn(mock());
+    when(templateHelper.getTemplate(EmailType.DAR_EXPIRED.templateName)).thenReturn(mock());
+
+    assertDoesNotThrow(()->service.sendExpirationNotices());
+    verify(emailDAO).insert(eq(dar1.getReferenceId()),eq(null), eq(user1.getUserId()), eq(EmailType.DAR_EXPIRED.getTypeInt()), any(), any(), any(), any(), any());
```
pshapiro4broad: Can you reformat this new code using the google-java-formatter? I see a missing space here.

Also, you can use `isNull()` more descriptive way to write `eq(null)`

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
```diff
@@ -554,4 +557,112 @@ void sendDarExpirationReminderMessage() throws Exception {
...
+    verify(emailDAO).insert(eq(dar1.getReferenceId()),eq(null), eq(user1.getUserId()), eq(EmailType.DAR_EXPIRATION_REMINDER.getTypeInt()), any(), any(), any(), any(), any());
+    verify(emailDAO).insert(eq(dar2.getReferenceId()),eq(null), eq(user2.getUserId()), eq(EmailType.DAR_EXPIRED.getTypeInt()), any(), any(), any(), any(), any());
+    verify(emailDAO).insert(eq(dar2.getReferenceId()),eq(null), eq(user2.getUserId()), eq(EmailType.DAR_EXPIRATION_REMINDER.getTypeInt()), any(), any(), any(), any(), any());
+    verify(emailDAO, times(4)).insert(any(), any(), any(), any(), any(), any(), any(), any(), any());
```
pshapiro4broad: There is no need to do this if you've already verified the individual calls. If you want to ensure that only the verfified calls are made, you can use `verifyNoMoreInteractions()`
```suggestion
    verifyNoMoreInteractions(emailDAO);
```

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
```diff
@@ -554,4 +557,112 @@ void sendDarExpirationReminderMessage() throws Exception {
...
+    verify(emailDAO, times(2)).insert(any(), any(), any(), any(), any(), any(), any(), any(), any());
+    assertEquals(2, listAppender.list.size());
+  }
+
+  private DataAccessRequest getMockedDar(String darCode, String referenceId, User user) {
+    DataAccessRequest dar = mock(DataAccessRequest.class);
+    when(dar.getReferenceId()).thenReturn(referenceId);
+    when(dar.getDarCode()).thenReturn(darCode);
+    when(dar.getUserId()).thenReturn(user.getUserId());
+    return dar;
```
pshapiro4broad, line 707: Why not create an object here?
```suggestion
    DataAccessRequest dar = new DataAccessRequest();
    dar.setReferenceId(referenceId);
    dar.setDarCode(darCode);
    dar.setUserId(user.getUserId());
    return dar;
```

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
```diff
@@ -554,4 +557,112 @@ void sendDarExpirationReminderMessage() throws Exception {
...
+    DataAccessRequest dar2 = getMockedDar("DAR-12346", UUID.randomUUID().toString(), user2);
+
+    when(userDAO.findUserById(user1.getUserId())).thenReturn(user1);
+    when(userDAO.findUserById(user2.getUserId())).thenReturn(user2);
+    List<DataAccessRequest> dars = List.of(dar1, dar2);
+    when(dataAccessRequestDAO.findAgedDARsByEmailTypeOlderThanInterval(any(), any(), any())).thenReturn(dars);
+    when(templateHelper.getTemplate(EmailType.DAR_EXPIRATION_REMINDER.templateName)).thenReturn(mock());
+    when(templateHelper.getTemplate(EmailType.DAR_EXPIRED.templateName)).thenReturn(mock());
+
+    assertDoesNotThrow(()->service.sendExpirationNotices());
```
pshapiro4broad: What is the purpose of this assert here? Any thrown exception will cause the test to fail, so there's no need to use `assertDoesNotThrow()`.
```suggestion
    service.sendExpirationNotices();
```

pshapiro4broad (commented):
More comments, still looking

src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
pshapiro4broad: Why do we need to test service account separately from researcher? More specifically, how can the code we're testing be broken in a way that researcher will work and service account won't?

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
pshapiro4broad: The magic string `"11 months"` appears in many places in this PR. In case we change this interval, can this be replaced with a named constant? E.g., something like `DataAccessRequestService.EXPIRE_WARN_INTERVAL`.

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
```diff
@@ -826,34 +814,203 @@ void createProgressReport() {
...
-    List<DataAccessRequest> dars = dataAccessRequestDAO.findSubmittedDarsByTimeRange(oneYearAgo,
-        oneMinuteInTheFuture);
+  void testFindAgedDARsByEmailTypeOlderThanInterval() throws InterruptedException {
+    User userOne = createUserWithInstitution();
+    Integer userOneId = userOne.getUserId();
+    Integer collectionOneId = createDarCollection(userOneId);
+    Integer collectionTwoId = createDarCollection(userOneId);
+    createDataAccessRequest(collectionOneId, userOneId,
+        new Date(Instant.now().minus(360,
+            ChronoUnit.DAYS).toEpochMilli()));
```
pshapiro4broad, line 826: You can go from `Instant` to `Date` by using `Date.from()`
```suggestion
        Date.from(Instant.now().minus(360, ChronoUnit.DAYS)));
```

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
```diff
@@ -826,34 +814,203 @@ void createProgressReport() {
...
+        "11 months", Timestamp.from(Instant.now().minus(365, ChronoUnit.DAYS)));
     assertFalse(dars.isEmpty());
-    assertEquals(darCollection.getDars().size(), dars.size());
+    assertEquals(1, dars.size());
   }
 
   @Test
-  void testFindDARsByDateRangeWithNoneInRange() throws InterruptedException {
+  void testFindAgedDARsByEmailTypeOlderThanIntervalSkipsEntriesBeforeNotBefore()
+      throws InterruptedException {
```
pshapiro4broad, line 840: This declares `throws InterruptedException` but it doesn't call any APIs that can throw this exception.
```suggestion
  void testFindAgedDARsByEmailTypeOlderThanIntervalSkipsEntriesBeforeNotBefore() {
```

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
```diff
@@ -554,4 +557,112 @@ void sendDarExpirationReminderMessage() throws Exception {
...
+    user2.setDisplayName("Jane Doe");
+    user2.setEmail("jd@somewhereelse");
+
+    DataAccessRequest dar1 = getMockedDar("DAR-12345", UUID.randomUUID().toString(), user1);
+    DataAccessRequest dar2 = getMockedDar("DAR-12346", UUID.randomUUID().toString(), user2);
+
+    when(userDAO.findUserById(user1.getUserId())).thenReturn(user1);
+    when(userDAO.findUserById(user2.getUserId())).thenReturn(user2);
+    List<DataAccessRequest> dars = List.of(dar1, dar2);
+    when(dataAccessRequestDAO.findAgedDARsByEmailTypeOlderThanInterval(any(), any(), any())).thenReturn(dars);
```
pshapiro4broad, line 579: This `when()` (which is also a `verify`) could be more precise.
```suggestion
    when(dataAccessRequestDAO.findAgedDARsByEmailTypeOlderThanInterval(
            eq(EmailType.DAR_EXPIRATION_REMINDER.getTypeInt()), eq("11 months"), any()))
        .thenReturn(dars);
    when(dataAccessRequestDAO.findAgedDARsByEmailTypeOlderThanInterval(
            eq(EmailType.DAR_EXPIRED.getTypeInt()), eq("1 year"), any()))
        .thenReturn(dars);
```

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
```diff
@@ -554,4 +557,112 @@ void sendDarExpirationReminderMessage() throws Exception {
...
+    when(userDAO.findUserById(user1.getUserId())).thenReturn(user1);
+    when(userDAO.findUserById(user2.getUserId())).thenReturn(user2);
+
+    List<DataAccessRequest> dars = List.of(dar2, dar1);
+
+    when(dataAccessRequestDAO.findAgedDARsByEmailTypeOlderThanInterval(any(), any(), any())).thenReturn(dars);
+    when(templateHelper.getTemplate(EmailType.DAR_EXPIRATION_REMINDER.templateName)).thenReturn(mock());
+    when(templateHelper.getTemplate(EmailType.DAR_EXPIRED.templateName)).thenReturn(mock());
+
+    assertDoesNotThrow(()->service.sendExpirationNotices());
```
pshapiro4broad, line 659: Instead of using a log appender, you can use a mock logger and then you can use `verify()` as is used for other mocked APIs.
```suggestion
    Logger logger = mock();
    try (var logFactoryMock = mockStatic(LoggerFactory.class)) {
      logFactoryMock.when(() -> LoggerFactory.getLogger(EmailService.class)).thenReturn(logger);
      service.sendExpirationNotices();
    }
    verify(logger, times(2)).warn(any());
```

pshapiro4broad (approved):
My last batch of comments. Lots of little things but overall looks OK to me.

src/test/java/org/broadinstitute/consent/http/mail/message/DarExpirationReminderMessageTest.java
pshapiro4broad: If using a UUID vs anther random string doesn't affect the test results (or, more specifically, doesn't affect the code under test's contract) then there's no need to use one here. So I don't see a benefit to using a random UUID over `randomAlphabetic()` or a string constant.

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
otchet-broad: Had something like your suggestion and it worked for the test and the class, but broke tests in another class when running the full suite.  Settled on the approach used because the scope of this ticket is too big already.

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
otchet-broad: Good point!

otchet-broad: I'm making a static variable for each string for now.  Let me know if that satisfies this for now.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
fboulnois: Has the performance of any of these queries changed?

fboulnois (approved):
The suggestions from other colleagues seem reasonable.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
otchet-broad: Great question!  Not in a noticeable way at this point.  That being said, I think there are a number of queries that we may have to keep an eye on as we start using dar_collection in new ways.  That may cause us to want to consider moving dar_code into data_access_request and simplifying the relationships we have going on in all of this code.

rushtong (approved):
Thank you for the changes 👍🏽

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
pshapiro4broad: > but broke tests in another class when running the full suite

Interesting, I guess the static nature of the logger initialization means that it will only happen once across a test suite. Maybe using a spy() here would work, but that would be a bit more work and definitely not worth doing right now.

PR #2510, opened 2025-05-01
[DT-1601] Add a check that all datasets in a progress report must be approved

s-rubenstein:
### Summary

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
@@ -109,7 +109,7 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
               FROM data_access_request dar
               INNER JOIN dar_dataset dd ON dd.reference_id = dar.reference_id
               INNER JOIN (
-                SELECT DISTINCT e.reference_id, LAST_VALUE(v.vote)
+                SELECT DISTINCT e.reference_id, e.dataset_id, LAST_VALUE(v.vote)
                 OVER(
                   PARTITION BY e.reference_id
                     ORDER BY v.createdate
@@ -124,6 +124,7 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
               WHERE dar.submission_date > now() - interval '1 year'
               AND final_access_vote.last_vote = TRUE
               AND dar.reference_id = :darReferenceId
+              AND final_access_vote.dataset_id = dd.dataset_id
               AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)
           """)
   List<Integer> findApprovedDatasetsByDar(@Bind("darReferenceId") String darReferenceId);
--- a/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
@@ -3,6 +3,7 @@ package org.broadinstitute.consent.http.db;
 import java.sql.Timestamp;
 import java.util.Date;
 import java.util.List;
+import java.util.Set;
 import org.broadinstitute.consent.http.db.mapper.DataAccessRequestDataMapper;
 import org.broadinstitute.consent.http.db.mapper.DataAccessRequestMapper;
 import org.broadinstitute.consent.http.db.mapper.DataAccessRequestReducer;
@@ -103,28 +104,30 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
    */
   @SqlQuery(
       """
-              SELECT dd.dataset_id
-              FROM data_access_request dar
-              INNER JOIN dar_dataset dd ON dd.reference_id = dar.reference_id
-              INNER JOIN (
-                SELECT DISTINCT e.reference_id, e.dataset_id, LAST_VALUE(v.vote)
-                OVER(
-                  PARTITION BY e.dataset_id
-                    ORDER BY v.createdate
-                    RANGE BETWEEN
-                      UNBOUNDED PRECEDING AND
-                      UNBOUNDED FOLLOWING
-                ) last_vote
-                FROM election e
-                INNER JOIN vote v ON e.election_id = v.electionid AND v.vote IS NOT NULL
-                AND LOWER(e.election_type) = 'dataaccess'
-                AND LOWER(v.type) = 'final') final_access_vote ON final_access_vote.reference_id = dar.reference_id AND final_access_vote.dataset_id = dd.dataset_id
-              WHERE dar.submission_date > now() - interval '1 year'
-              AND final_access_vote.last_vote = TRUE
-              AND dar.reference_id = :darReferenceId
-              AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)
-          """)
-  List<Integer> findApprovedDatasetsByDar(@Bind("darReferenceId") String darReferenceId);
+      SELECT dd.dataset_id
+      FROM data_access_request dar
+      INNER JOIN dar_dataset dd ON dd.reference_id = dar.reference_id
+      INNER JOIN (
+        SELECT DISTINCT e.reference_id, e.dataset_id, LAST_VALUE(v.vote)
+        OVER(
+          PARTITION BY e.dataset_id
+            ORDER BY v.createdate
+            RANGE BETWEEN
+              UNBOUNDED PRECEDING AND
+              UNBOUNDED FOLLOWING
+        ) last_vote
+        FROM election e
+        INNER JOIN vote v ON e.election_id = v.electionid AND v.vote IS NOT NULL
+        AND LOWER(e.election_type) = 'dataaccess'
+        AND LOWER(v.type) = 'final') final_access_vote ON
+          final_access_vote.reference_id = dar.reference_id AND
+          final_access_vote.dataset_id = dd.dataset_id
+      WHERE dar.submission_date > now() - interval '1 year'
+      AND final_access_vote.last_vote = TRUE
+      AND dar.reference_id = :darReferenceId
+      AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)
+      """)
+  Set<Integer> findApprovedDatasetsByDar(@Bind("darReferenceId") String darReferenceId);
 
   /**
    * This query finds submitted DARs based on a date range.  This would be useful if we wanted to
--- a/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
@@ -6,11 +6,9 @@ import java.util.List;
 import org.broadinstitute.consent.http.db.mapper.DataAccessRequestDataMapper;
 import org.broadinstitute.consent.http.db.mapper.DataAccessRequestMapper;
 import org.broadinstitute.consent.http.db.mapper.DataAccessRequestReducer;
-import org.broadinstitute.consent.http.db.mapper.DatasetReducer;
 import org.broadinstitute.consent.http.models.DarDataset;
 import org.broadinstitute.consent.http.models.DataAccessRequest;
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -6,9 +6,11 @@
 import org.broadinstitute.consent.http.db.mapper.DataAccessRequestDataMapper;
 import org.broadinstitute.consent.http.db.mapper.DataAccessRequestMapper;
 import org.broadinstitute.consent.http.db.mapper.DataAccessRequestReducer;
+import org.broadinstitute.consent.http.db.mapper.DatasetReducer;
 import org.broadinstitute.consent.http.models.DarDataset;
 import org.broadinstitute.consent.http.models.DataAccessRequest;
 import org.broadinstitute.consent.http.models.DataAccessRequestData;
+import org.broadinstitute.consent.http.models.Dataset;
```
rushtong: These added imports are not used

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -87,6 +89,45 @@ WHERE dar.submission_date > now() - interval '1 year'
...
+   * votes. The `LAST_VALUE` function selects the last result in the partition, which would be
+   * `FALSE` in the example above. Outside the JOIN, we filter on groupings where the final vote
+   * value is `TRUE` so the denied election in the example would be filtered out.
+   *
+   * @param darReferenceId The DAR reference UUID
+   * @return List of approved Datasets for the DAR
+   */
+  @SqlQuery(
+      """
+              SELECT DISTINCT dd.dataset_id
```
rushtong: In theory, we shouldn't need distinct here since a dar can only reference a dataset once.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -87,6 +89,45 @@ WHERE dar.submission_date > now() - interval '1 year'
...
+                  PARTITION BY e.reference_id
+                    ORDER BY v.createdate
+                    RANGE BETWEEN
+                      UNBOUNDED PRECEDING AND
+                      UNBOUNDED FOLLOWING
+                ) last_vote
+                FROM election e
+                INNER JOIN vote v ON e.election_id = v.electionid AND v.vote IS NOT NULL
+                AND LOWER(e.election_type) = 'dataaccess'
+                AND LOWER(v.type) = 'final') final_access_vote ON final_access_vote.reference_id = dar.reference_id
```
rushtong: I _think_ we need an additional restriction on `e.dataset_id = dd.dataset_id` here.

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
```diff
@@ -601,6 +601,51 @@ void testFindAllApprovedDataAccessRequestsByDatasetId() {
...
+        now,
+        v2.getVoteId(),
+        false,
+        e2.getElectionId(),
+        now,
+        false);
+
+    assertEquals(2,
+        dataAccessRequestDAO.findApprovedDatasetsByDar(testDar1.getReferenceId())
+            .size());
```
rushtong: If we had one more test condition where there was a final vote for a dataset election that was false, we could be 100% sure we're not getting denied datasets back from the query.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -226,14 +226,18 @@ public DataAccessRequest createProgressReport(User user, DataAccessRequest progr
     validateProgressReport(user, progressReport, parentDar);
 
     String referenceId = progressReport.getReferenceId();
-    List<Integer> datasetIds = progressReport.getDatasetIds();
+    List<Integer> progressReportDatasetIds = progressReport.getDatasetIds();
+    List<Integer> darDatasetIds = dataAccessRequestDAO.findApprovedDatasetsByDar(parentDar.getReferenceId());
+    if(progressReportDatasetIds.stream().filter(datasetId -> !darDatasetIds.contains(datasetId)).toList().size() > 0) {
+      throw new BadRequestException("Progress report can only be created for approved datasets in the parent DAR");
+    }
```
rushtong: This is challenging for me to follow. I tried rewriting it a bit so I can wrap my brain around it:
```
    boolean validIds = new HashSet<>(darDatasetIds).containsAll(progressReportDatasetIds);
    if (!validIds) {
      throw new BadRequestException("Progress report can only be created for datasets in the parent DAR");
    }
```

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -87,6 +87,45 @@ WHERE dar.submission_date > now() - interval '1 year'
...
+                FROM election e
+                INNER JOIN vote v ON e.election_id = v.electionid AND v.vote IS NOT NULL
+                AND LOWER(e.election_type) = 'dataaccess'
+                AND LOWER(v.type) = 'final') final_access_vote ON final_access_vote.reference_id = dar.reference_id AND final_access_vote.dataset_id = dd.dataset_id
+              WHERE dar.submission_date > now() - interval '1 year'
+              AND final_access_vote.last_vote = TRUE
+              AND dar.reference_id = :darReferenceId
+              AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)
+          """)
+  List<Integer> findApprovedDatasetsByDar(@Bind("darReferenceId") String darReferenceId);
```
pshapiro4broad: If this is the only use of `findApprovedDatasetsByDar()`, you can declare it as returning a `Set` type and then you don't have to create one in the service code.
```suggestion
  Set<Integer> findApprovedDatasetsByDar(@Bind("darReferenceId") String darReferenceId);
```

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -87,6 +87,45 @@ WHERE dar.submission_date > now() - interval '1 year'
...
+   * votes. The `LAST_VALUE` function selects the last result in the partition, which would be
+   * `FALSE` in the example above. Outside the JOIN, we filter on groupings where the final vote
+   * value is `TRUE` so the denied election in the example would be filtered out.
+   *
+   * @param darReferenceId The DAR reference UUID
+   * @return List of approved Datasets for the DAR
+   */
+  @SqlQuery(
+      """
+              SELECT dd.dataset_id
```
pshapiro4broad: The google java formatter doesn't indent text blocks for us, so I don't think there's any rule about what the indent needs to be. I do think we should try to move the text to the left if possible, to make it easier to see the code without scrolling right. In TDR the convention is to align the text with the start of the `"""`.
```suggestion
  @SqlQuery(
      """
      SELECT dd.dataset_id
```

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -87,6 +87,45 @@ WHERE dar.submission_date > now() - interval '1 year'
...
+                  PARTITION BY e.dataset_id
+                    ORDER BY v.createdate
+                    RANGE BETWEEN
+                      UNBOUNDED PRECEDING AND
+                      UNBOUNDED FOLLOWING
+                ) last_vote
+                FROM election e
+                INNER JOIN vote v ON e.election_id = v.electionid AND v.vote IS NOT NULL
+                AND LOWER(e.election_type) = 'dataaccess'
+                AND LOWER(v.type) = 'final') final_access_vote ON final_access_vote.reference_id = dar.reference_id AND final_access_vote.dataset_id = dd.dataset_id
```
pshapiro4broad: I think the indent here isn't showing the `ON` grouping very well. Using the intellij formatter, this is how this query looks
```sql
SELECT dd.dataset_id
FROM data_access_request dar
         INNER JOIN dar_dataset dd ON dd.reference_id = dar.reference_id
         INNER JOIN (SELECT DISTINCT e.reference_id,
                                     e.dataset_id,
                                     LAST_VALUE(v.vote)
                                     OVER (
                                         PARTITION BY e.dataset_id
                                         ORDER BY v.createdate
                                         RANGE BETWEEN
                                             UNBOUNDED PRECEDING AND
                                             UNBOUNDED FOLLOWING
                                         ) last_vote
                     FROM election e
                              INNER JOIN vote v
                                         ON e.election_id = v.electionid AND v.vote IS NOT NULL
                                             AND LOWER(e.election_type) = 'dataaccess'
                                             AND LOWER(v.type) = 'final') final_access_vote
                    ON final_access_vote.reference_id = dar.reference_id AND
WHERE dar.submission_date > now() - interval '1 year'
  AND final_access_vote.last_vote = TRUE
  AND dar.reference_id = :darReferenceId
  AND (LOWER(dar.data ->> 'status') != 'archived' OR dar.data ->> 'status' IS NULL)
```

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -226,14 +227,19 @@ public DataAccessRequest createProgressReport(User user, DataAccessRequest progr
     validateProgressReport(user, progressReport, parentDar);
 
     String referenceId = progressReport.getReferenceId();
-    List<Integer> datasetIds = progressReport.getDatasetIds();
+    List<Integer> progressReportDatasetIds = progressReport.getDatasetIds();
+    List<Integer> darDatasetIds = dataAccessRequestDAO.findApprovedDatasetsByDar(parentDar.getReferenceId());
+    boolean allIdsValid = new HashSet<>(darDatasetIds).containsAll(progressReportDatasetIds);
+    if(!allIdsValid) {
```
pshapiro4broad: whitespace
```suggestion
    if (!allIdsValid) {
```

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
```diff
@@ -601,6 +601,73 @@ void testFindAllApprovedDataAccessRequestsByDatasetId() {
...
+        now,
+        v1.getVoteId(),
+        false,
+        e1.getElectionId(),
+        now,
+        false);
+
+    assertEquals(1,
+        dataAccessRequestDAO.findApprovedDatasetsByDar(testDar1.getReferenceId())
+            .size());
```
pshapiro4broad: Could this also assert the contents of the returned collection?

src/test/java/org/broadinstitute/consent/http/service/DataAccessRequestServiceTest.java
```diff
@@ -228,6 +229,7 @@ void createProgressReport() {
     user.setLibraryCards(List.of(new LibraryCard()));
     parentDar.setUserId(user.getUserId());
     when(dataAccessRequestDAO.findByReferenceId(progressReport.getReferenceId())).thenReturn(progressReport);
+    when(dataAccessRequestDAO.findApprovedDatasetsByDar(parentDar.getReferenceId())).thenReturn(List.of(APPROVED_PROGRESS_REPORT_DATASET_ID));
```
pshapiro4broad: This code already has the dataset ID so there's no need for a named constant.
```suggestion
    when(dataAccessRequestDAO.findApprovedDatasetsByDar(parentDar.getReferenceId()))
        .thenReturn(progressReport.getDatasetIds());
```

(or `Set.copyOf(progressReport.getDatasetIds())` if you change the DAO method return type)

This value is from `generateDataAccessRequest()` here:
```
    dar.addDatasetId(1);
```

pshapiro4broad (approved):
Looks OK as far as I can tell, a few minor comments

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
rushtong: Agreed, the other asserts could likely go away if we just asserted the contents were the same as dataset ids for dataset 1 and 2.

rushtong (approved):
Thank you for the changes 👍🏽

PR #2511, opened 2025-05-01
DT-1472: Populate eRA Commons Id on DAR submission

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1472

### Summary
This PR does the following:
* Add a new top level field on DAR for eRA Commons Id
  * This represents the id of the user at the time of DAR submission
* Back populate existing submitted DARs with the value of the create user
  * This isn't perfect as we can't know if this was the actual value at the time of submission, but it is the only value we have and is currently what we show to the DACs when they review a DAR.
* Update plumbing to save and return this value along with DAR records
* Insert this value from the user on DAR submission and on Progress Report submission
* Large number of minor updates to tests to address sonar warnings

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
@@ -4,7 +4,6 @@ import java.sql.Timestamp;
 import java.util.Date;
 import java.util.List;
 import java.util.Set;
-import org.broadinstitute.consent.http.db.mapper.DataAccessRequestDataMapper;
 import org.broadinstitute.consent.http.db.mapper.DataAccessRequestMapper;
 import org.broadinstitute.consent.http.db.mapper.DataAccessRequestReducer;
 import org.broadinstitute.consent.http.models.DarDataset;
@@ -66,8 +65,8 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
       """
               SELECT dar.id, dar.reference_id, dar.collection_id, dar.parent_id,
                 dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date,
-                (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data, dar.era_commons_id,
-                dd.dataset_id, collection.dar_code
+                (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data,
+                dd.dataset_id, collection.dar_code, dar.era_commons_id
               FROM data_access_request dar
               LEFT JOIN dar_collection collection on collection.collection_id = dar.collection_id
               INNER JOIN dar_dataset dd ON dd.reference_id = dar.reference_id AND dd.dataset_id = :datasetId
@@ -103,7 +102,7 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
    * value is `TRUE` so the denied election in the example would be filtered out.
    *
    * @param darReferenceId The DAR reference UUID
-   * @return List of approved Datasets for the DAR
+   * @return Set of approved Dataset Ids for the DAR
    */
   @SqlQuery(
       """
@@ -145,8 +144,8 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
       """
               SELECT dar.id, dar.reference_id, dar.collection_id, dar.parent_id,
                 dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date,
-                (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data, dar.era_commons_id,
-                dd.dataset_id, collection.dar_code
+                (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data,
+                dd.dataset_id, collection.dar_code, dar.era_commons_id
               FROM data_access_request dar
               LEFT JOIN dar_collection collection on collection.collection_id = dar.collection_id
               LEFT JOIN dar_dataset dd ON dd.reference_id = dar.reference_id
@@ -154,7 +153,7 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
               WHERE dar.submission_date >= :notBefore
               AND (dar.submission_date < now() - :interval ::interval)
               AND (email.email_type IS NULL)
-            
+          
           """)
   List<DataAccessRequest> findAgedDARsByEmailTypeOlderThanInterval(@Bind("emailType") Integer emailType,
       @Bind("interval") String interval, @Bind("notBefore") Timestamp notBefore);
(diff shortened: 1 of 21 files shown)


src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -351,16 +359,6 @@ void insertProgressReport(
...
 
-  @RegisterRowMapper(DataAccessRequestDataMapper.class)
-  @SqlQuery(
-      """
-          SELECT (data #>> '{}')::jsonb AS data
-          FROM data_access_request
-          WHERE (LOWER(data->>'status') != 'archived'
-            OR data->>'status' IS NULL)
-      """)
-  List<DataAccessRequestData> findAllDataAccessRequestDatas();
```
rushtong, line 397: This was only used in tests

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -475,7 +475,7 @@ public void hasValidActiveERACredentials(User user) {
     if (cards.isEmpty()) {
       throw new LibraryCardRequiredException();
     }
-    boolean hasEraCommonsId = cards.stream().anyMatch(c -> c.getEraCommonsId() != null);
+    boolean hasEraCommonsId = user.getEraCommonsId() != null;
```
rushtong: We need to look into why LCs have this information and if we can remove it. The real source of truth is the user, not the LC. Additionally, if we only allow a single LC to be created, look at what it would take to turn that into a single entity instead of a list. Ticket to come.

rushtong: This was fixed in https://github.com/DataBiosphere/consent/pull/2512

src/main/resources/changesets/changelog-consent-2025-05-01-add-era-commons-to-dar.xml
```diff
@@ -0,0 +1,18 @@
...
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+  <changeSet id="changelog-consent-2025-05-01-add-era-commons-to-dar" author="grushton">
+    <addColumn tableName="data_access_request">
+      <column name="era_commons_id" type="varchar(255)"/>
+    </addColumn>
+    <sql>
+      UPDATE data_access_request dar
+      SET era_commons_id = users.era_commons_id FROM (select user_id, era_commons_id FROM users) AS users
+      WHERE dar.user_id = users.user_id
+      AND dar.submission_date IS NOT NULL;
```
pshapiro4broad, line 12: It would be nice if we could make this not null. We know that it won't be null in the future, correct?

rushtong: On dev, at least, I know there are null values. It is also a valid use case that a user can submit a dar, then remove their era commons link. It would be nice to enforce this more but I think we need to review the workflows in place.

pshapiro4broad: > then remove their era commons link

Do you mean remove it from the `User` or from the dar? If they remove it from `User` the dar isn't affected (in the future, once this code is in place) so the dar should still be non-null.

For the historical null value cases, we could put a placeholder value like `UNKNOWN`.

rushtong: > Do you mean remove it from the User or from the dar?

A user can unlink their account which removes it from their user, not anything else. 

> dar should still be non-null

Draft DARs will **not** have this value populated until submission so it has to be nullable.

> placeholder value like `UNKNOWN`

We could do that. It's not required, but I'm also not sure what that gets us. The UI will know either way if the value is defined or `UNKNOWN`

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -4,10 +4,8 @@
 import java.util.Date;
 import java.util.List;
 import java.util.Set;
-import org.broadinstitute.consent.http.db.mapper.DataAccessRequestDataMapper;
 import org.broadinstitute.consent.http.db.mapper.DataAccessRequestMapper;
 import org.broadinstitute.consent.http.db.mapper.DataAccessRequestReducer;
-import org.broadinstitute.consent.http.enumeration.EmailType;
```
otchet-broad, line 10: Thank you!  I'm looking forward to spotless.

rushtong: Me too 👍🏽

src/main/resources/changesets/changelog-consent-2025-05-01-add-era-commons-to-dar.xml
otchet-broad: @rushtong  - your responses make sense to me given the state machine around the presence of an eRA Commons Id and the state machine around DataAccessRequest eRA Commons Id in this.  One thought about adding the "UNKNOWN" value that @pshapiro4broad raises is with this value set, we could consider adding eraCommonsId to our state machine check on DARs to know they're draft.

otchet-broad: If we add test coverage to that contract, we'll know when we've got code that is attempting to do something our expected state machine doesn't support.

src/test/java/org/broadinstitute/consent/http/service/dao/DataAccessRequestServiceDAOTest.java
```diff
@@ -152,7 +152,7 @@ private DataAccessRequest createDataAccessRequest(Integer userId, Integer collec
       String darCode) {
     DataAccessRequestData data = new DataAccessRequestData();
     data.setProjectTitle("Project Title: " + RandomStringUtils.random(50, true, false));
-    data.setDarCode(darCode);
+//    data.setDarCode(darCode);
```
otchet-broad: Commented code.

It did make me wonder if we ever take this value from user input and use it for something.

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
```diff
@@ -845,36 +842,12 @@ void testFindByReferenceIdsArchived() {
     assertTrue(returnedDAR.isEmpty());
   }
 
-  // findAllDataAccessRequestDatas should exclude archived DARs
-  @Test
```
otchet-broad, line 849: Thank you!

otchet-broad (approved):

src/test/java/org/broadinstitute/consent/http/service/dao/DataAccessRequestServiceDAOTest.java
rushtong: My fixes to these test classes keep getting overwritten from merges - I need to review this class again.

src/main/resources/changesets/changelog-consent-2025-05-01-add-era-commons-to-dar.xml
rushtong: > One thought about adding the "UNKNOWN" value that @pshapiro4broad raises is with this value set, we could consider adding eraCommonsId to our state machine check on DARs to know they're draft.

I have not looked at prod data, but there is the possibility of existing, submitted, DARs to have no value (or `UNKNOWN` if we did take this approach). In that case, it would not really tell you anything more about the DAR that we don't already know. It would only be helpful for future DARs. This is definitely true in dev, but the likelihood in prod is admittedly low.

src/test/java/org/broadinstitute/consent/http/service/dao/DataAccessRequestServiceDAOTest.java
rushtong: Setting `darCode` on the data object is how we used to store the source of truth for this value. When we implemented `DarCollection`, we moved it there so that was the new source of truth. Tests, however, lagged behind.

fboulnois (approved):
In the future, consider putting the test cleanups and other changes in a separate PR, it will make the core PR relating to the feature easier to review (with less files). Otherwise, lgtm 👍

PR #2514, opened 2025-05-05
DT-1597: Update Unreviewed to Submitted for DAR Collection Summaries

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-1597

## Summary
Update the `Unreviewed` status to `Submitted`

### Updated View
Screenshot 2025-05-05 at 9 36 08 AM

Compared to the old view:
Screenshot 2025-05-05 at 9 38 06 AM

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/enumeration/DarCollectionStatus.java
+++ b/src/main/java/org/broadinstitute/consent/http/enumeration/DarCollectionStatus.java
@@ -1,7 +1,7 @@
 package org.broadinstitute.consent.http.enumeration;
 
 public enum DarCollectionStatus {
-  UNREVIEWED("Unreviewed"),
+  SUBMITTED("Submitted"),
   IN_PROCESS("In Process"),
   COMPLETE("Complete"),
   DRAFT("Draft"),
--- a/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
@@ -95,7 +95,7 @@ public class DarCollectionService implements ConsentLogger {
     //if there are some elections open, status is in process
     //if all elections are closed or canceled and electionCount == datasetCount, status is complete
     if (electionCount.equals(0)) {
-      summary.setStatus(DarCollectionStatus.UNREVIEWED.getValue());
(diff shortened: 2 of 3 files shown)


snf2ye (approved):

src/main/java/org/broadinstitute/consent/http/enumeration/DarCollectionStatus.java
```diff
@@ -1,7 +1,7 @@
 package org.broadinstitute.consent.http.enumeration;
 
 public enum DarCollectionStatus {
-  UNREVIEWED("Unreviewed"),
+  SUBMITTED("Submitted"),
```
pshapiro4broad, line 4: This seems fine, but we could also have the `value` diverge from the enum name. if the value is the UI version of this state.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -95,7 +95,7 @@ private void determineCollectionStatus(DarCollectionSummary summary,
     //if there are some elections open, status is in process
     //if all elections are closed or canceled and electionCount == datasetCount, status is complete
     if (electionCount.equals(0)) {
-      summary.setStatus(DarCollectionStatus.UNREVIEWED.getValue());
+      summary.setStatus(DarCollectionStatus.SUBMITTED.getValue());
```
pshapiro4broad, line 98: If all the uses of this enum are as the string constant, does it need to be an enum? Either `DarCollectionSummary` should be type `DarCollectionStatus` instead of `String`, or these enums could be replaced with `static final String` constants.

pshapiro4broad (approved):
I don't want to hold up this PR, but this approach means that UI text is driven by the back end, which doesn't seem great to me.

I think it would be better if the backend generated an enum value (specified in the openapi yaml) and the front end converted that enum value into the text representation. That way we could keep the backend logic the same and decide in the UI how we want that state to be displayed.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rushtong: These certainly could be strings instead. There is future work to consolidate the state machine logic around status and actions, so we can probably tackle this there.

PR #2515, opened 2025-05-05
[DT-99] Update db-connect.sh to use gcloud secrets

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-99

### Summary

Follow-up to Rae's work to allow the `db-connect.sh` script to connect using `gcloud` instead of `vault`.

----

- Verify all tests go green

### DIFF
--- a/scripts/db-connect.sh
+++ b/scripts/db-connect.sh
@@ -75,12 +75,11 @@ case $1 in
     * ) error "ENV must be one of dev, alpha, or staging";;
 esac
 
-VAULT_PATH="secret/dsde/firecloud/$1/consent/secrets/postgres/app_sql_user"
-VAULT_JSON=$(vault read -format=json "$VAULT_PATH")
+PG_JSON=$(gcloud secrets versions access latest --secret=consent-postgres-creds)
 
-INSTANCE=$(echo "$VAULT_JSON" | jq -r .data.instance_name)
-USERNAME=$(echo "$VAULT_JSON" | jq -r .data.username)
-PASSWORD=$(echo "$VAULT_JSON" | jq -r .data.password)
+INSTANCE=$(echo "$PG_JSON" | jq -r .instance_name)
+USERNAME=$(echo "$PG_JSON" | jq -r .username)
+PASSWORD=$(echo "$PG_JSON" | jq -r .password)
 
 INSTANCE="broad-dsde-$1:us-central1:$INSTANCE"
 


rushtong (approved):
👍🏽

otchet-broad (approved):

scripts/db-connect.sh
```diff
@@ -75,12 +75,11 @@ case $1 in
     * ) error "ENV must be one of dev, alpha, or staging";;
 esac
 
-VAULT_PATH="secret/dsde/firecloud/$1/consent/secrets/postgres/app_sql_user"
-VAULT_JSON=$(vault read -format=json "$VAULT_PATH")
+PG_JSON=$(gcloud secrets versions access latest --secret=consent-postgres-creds)
```
pshapiro4broad, line 78: I think you need to map env (`$1`) to a google project and pass `--project <>`.

scripts/db-connect.sh
```diff
@@ -75,12 +75,11 @@ case $1 in
...
 
-INSTANCE=$(echo "$VAULT_JSON" | jq -r .data.instance_name)
-USERNAME=$(echo "$VAULT_JSON" | jq -r .data.username)
-PASSWORD=$(echo "$VAULT_JSON" | jq -r .data.password)
+INSTANCE=$(echo "$PG_JSON" | jq -r .instance_name)
+USERNAME=$(echo "$PG_JSON" | jq -r .username)
+PASSWORD=$(echo "$PG_JSON" | jq -r .password)
```
pshapiro4broad, line 82: May as well change this to use redirection over `echo` as part of this. It works the same but avoids creating a subprocess for the `echo`
```suggestion
INSTANCE=$(jq -r .instance_name < "$PG_JSON")
USERNAME=$(jq -r .username < "$PG_JSON")
PASSWORD=$(jq -r .password < "$PG_JSON")
```

fboulnois: `jq` tries to treat the redirected json as a file, so this fails.

scripts/db-connect.sh
fboulnois: This path is the same in all three environments.

pshapiro4broad: The problem isn't the path, the problem is that the current default google project may not be correct for the env that the user is requesting. In order to give the user the correct password and instance name, the command needs a `--project` argument. For example, see how this is done in the TDR script https://github.com/DataBiosphere/jade-data-repo/blob/develop/ops/db-connect.sh#L129

scripts/db-connect.sh
pshapiro4broad: Sorry, I meant using `<<<`
```
INSTANCE=$(jq -r .instance_name <<< "$PG_JSON")
USERNAME=$(jq -r .username <<< "$PG_JSON")
PASSWORD=$(jq -r .password <<< "$PG_JSON")
```

PR #2516, opened 2025-05-05
[Dt-1598]Update DAR table to show latest in Dar Collection (Part 2)

rjohanek:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1598
This continues work from this PR: https://github.com/DataBiosphere/consent/pull/2502
We want to show the information from the latest DAR in the collection. We want to do this for all DAR tables in the UI which means updating all the DarCollectionSummaryDAO queries.

### Summary
Maintain the structure of the DarCollectionSummary by aggregating all referenceIds that share a collectionId and still include them in the summary, but update the information of the summary to only look at the the latest DAR by submission date instead. After this PR all DarCollectionSummaryDAO queries will have been updated.
Updates the following queries:
getDarCollectionSummariesForDAC
getDarCollectionSummariesForSO
getDarCollectionSummaryForDACByCollectionId
getDarCollectionSummaryByCollectionId

### Testing
Added unit tests for DAR collections with multiple DARs
Added unit tests for the DAC queries that include elections and votes

----

- Verify all tests go green


src/test/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAOTest.java
```diff
@@ -636,22 +647,22 @@ void testGetDarCollectionSummaryDraftAndArchivedNotIncluded(String type) {
 
     // Create two DataAccessRequests in the same collection
     DataAccessRequest olderDar = createDataAccessRequest(collectionId, userId);
-    DataAccessRequest draftDar = createDataAccessRequest(collectionId, userId);
```
rjohanek, line 638: removed draft because you can't have a draft progress report and you shouldn't/won't be able to have two original DARs in a collection

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -25,14 +25,24 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
       SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, u.display_name as researcher_name,
         i.institution_name, e.election_id, e.status, e.dataset_id, e.reference_id, v.voteid as v_vote_id, dd.dataset_id as dd_datasetid,
         v.user_id as v_user_id, v.vote as v_vote, v.electionid as v_election_id, v.createdate as v_create_date, v.updatedate as v_update_date, v.type as v_type,
-        (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name
+        (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name,
+        ARRAY_AGG(dar_all.reference_id) AS reference_ids
```
rushtong, line 29: Where is this populated on the `DarCollectionSummary`?

src/test/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAOTest.java
```diff
@@ -723,6 +738,68 @@ void testGetDarCollectionSummaryTwoDataAccessRequests(String type) {
     assertTrue(summary.getDatasetIds().contains(dataset.getDatasetId()));
   }
 
+  @ParameterizedTest
+  @ValueSource(strings= {"DatasetIds", "CollectionId"})
+  void testGetDarCollectionSummaryTwoDataAccessRequestsForDACsWithVotes(String type) {
```
rushtong, line 766: I'm finding this test challenging to follow. I think it would be easier to read if some common setup functions were broken out into separate helper methods. For example, this is how I read what this test is doing:
1. create approved data access request with election, votes, and datasets
2. create approved PR from approved DAR (i.e. with new election, new votes, and DAR's dataset(s))
3. test `darCollectionSummaryDAO.getDarCollectionSummariesForDAC` in its own method
4. test `darCollectionSummaryDAO.getDarCollectionSummaryForDACByCollectionId` in its own method

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -25,14 +25,24 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
...
       FROM dar_collection c
       INNER JOIN users u
         ON u.user_id = c.create_user_id
       LEFT JOIN institution i
         ON i.institution_id = u.institution_id
-      INNER JOIN data_access_request dar
-        ON dar.collection_id = c.collection_id
+      INNER JOIN (
```
fboulnois, line 35: I am finding these SQL queries hard to follow, they're very complex. I think we discussed refactoring these DAR-related queries to simplify them into smaller queries. Is that correct?

fboulnois (approved):
👍

rushtong (approved):
I have similar concerns about query complexity - I think we need to deal with this sooner rather than later.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
otchet-broad: **Thought that is beyond the scope of this pull request and mine that's also waiting to merge.**
I'm actually wondering if we can get rid of the entire dar_collection table/relationship and find groupings based on the darCode where we need them.  What I'm starting to realize is that we no longer really want to operate on the collection level.  Instead we really want to operate on the DAR level.  For example, elections really need to be tied to the DAR (we already use the DAR reference ID to do that), but current code loops over all DARs in the collection.  That behavior is no longer correct as our recent decision to allow users to exclude datasets in progress reports is much more impactful than I realized.  It might make sense to put together a design document with a plan for this so we can trace the use cases we're relying on, map them to the data_access_request table and see what we need to migrate over (darCode at a minimum).

src/test/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAOTest.java
otchet-broad: @raejohanek  - Let me know if you think it would be helpful for me to land my work first so you can leverage some of the new helpers that I've made there.  Otherwise, I'll hold off merging mine to deconflict the merge until after you're set.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
rjohanek: this is populated in summary.referenceIds in line 19 of the reducer, thanks for bringing this up bc now we can remove where dar_reference_ids are explicitly added in the reducer

rjohanek:
> I have similar concerns about query complexity - I think we need to deal with this sooner rather than later.

Okay I just filed a ticket for this work: https://broadworkbench.atlassian.net/browse/DT-1638

PR #2517, opened 2025-05-06
[DT-1619] Enrich DataAccessRequest and DarCollectionSummary information with progress report information.

otchet-broad:
### Addresses
[https://broadworkbench.atlassian.net/browse/DT-1619](https://broadworkbench.atlassian.net/browse/DT-1619)

### Summary

- Consolidates logic for how we determine if a DAR is a progress report.
~~- Makes the DAR tree available in DarCollectionSummary so the UI has deterministic information it needs to jump to the correct DAR.~~

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
@@ -22,8 +22,8 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
   @RegisterBeanMapper(value = Election.class)
   @UseRowReducer(DarCollectionSummaryReducer.class)
   @SqlQuery("""
-      SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, u.display_name as researcher_name,
-        i.institution_name, e.election_id, e.status, e.dataset_id, e.reference_id, v.voteid as v_vote_id, dd.dataset_id as dd_datasetid,
+      SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, dar.parent_id as dar_parent_id,
+       u.display_name as researcher_name, i.institution_name, e.election_id, e.status, e.dataset_id, e.reference_id, v.voteid as v_vote_id, dd.dataset_id as dd_datasetid,
         v.user_id as v_user_id, v.vote as v_vote, v.electionid as v_election_id, v.createdate as v_create_date, v.updatedate as v_update_date, v.type as v_type,
         (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name
       FROM dar_collection c
@@ -59,8 +59,8 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
   @SqlQuery
       (
           """
-              SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, u.display_name as researcher_name,
-                i.institution_name, e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id as dd_datasetid,
+              SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, dar.parent_id as dar_parent_id,
+                u.display_name as researcher_name, i.institution_name, e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id as dd_datasetid,
                 (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name
               FROM dar_collection c
               INNER JOIN users u
@@ -95,6 +95,7 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
               c.dar_code,
               dar.submission_date,
               dar.reference_id AS dar_reference_id,
+              dar.parent_id AS dar_parent_id,
               u.display_name AS researcher_name,
               i.institution_name,
               e.election_id,
@@ -128,7 +129,7 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
           LEFT JOIN dac ON dac.dac_id = dataset.dac_id
           WHERE (e.latest = e.election_id OR e.election_id IS NULL)
           GROUP BY
-              c.collection_id, c.dar_code, dar.submission_date, dar.reference_id,
+              c.collection_id, c.dar_code, dar.submission_date, dar.reference_id, dar.parent_id,
               u.display_name, i.institution_name, e.election_id, e.status,
               e.dataset_id, dd.dataset_id, dar.data, dac.name
       """)
@@ -144,6 +145,7 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
         c.dar_code,
         dar.submission_date,
         dar.reference_id AS dar_reference_id,
+        dar.parent_id AS dar_parent_id,
         u.display_name AS researcher_name,
         i.institution_name,
         e.election_id,
@@ -153,7 +155,8 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
         dd.dataset_id AS dd_datasetid,
         (regexp_replace(dar.data #>> '{}', '\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name,
         (regexp_replace(dar.data #>> '{}', '\\u0000', '', 'g'))::jsonb ->> 'status' AS dar_status,
-        ARRAY_AGG(dar_all.reference_id) AS reference_ids
+        dar_all.reference_id as dar_all_reference_id,
+        dar_all.parent_id as dar_all_parent_id
     FROM
         dar_collection c
     INNER JOIN
@@ -182,7 +185,7 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
         c.create_user_id = :userId
         AND (e.latest = e.election_id OR e.election_id IS NULL)
     GROUP BY
-        c.collection_id, c.dar_code, dar.submission_date, dar.reference_id, u.display_name, i.institution_name,
+        c.collection_id, c.dar_code, dar.submission_date, dar.reference_id, dar.parent_id, dar_all.reference_id, dar_all.parent_id, u.display_name, i.institution_name,
         e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id, dar.data
 """)
   List<DarCollectionSummary> getDarCollectionSummariesForResearcher(@Bind("userId") Integer userId);
@@ -234,7 +237,7 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
   @SqlQuery
       (
           """
-              SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, u.display_name as researcher_name,
+              SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, dar.parent_id as dar_parent_id, u.display_name as researcher_name,
                 u.user_id as researcher_id, i.institution_name, i.institution_id, e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id as dd_datasetid,
                 (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name,
                 (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'status' AS dar_status
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
@@ -155,8 +155,7 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
         dd.dataset_id AS dd_datasetid,
         (regexp_replace(dar.data #>> '{}', '\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name,
         (regexp_replace(dar.data #>> '{}', '\\u0000', '', 'g'))::jsonb ->> 'status' AS dar_status,
-        dar_all.reference_id as dar_all_reference_id,
-        dar_all.parent_id as dar_all_parent_id
+        ARRAY_AGG(dar_all.reference_id) AS reference_ids
     FROM
         dar_collection c
     INNER JOIN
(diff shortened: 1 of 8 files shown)


src/test/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAOTest.java
```diff
@@ -963,4 +964,76 @@ void testGetDarCollectionSummaryByCollectionId_ArchivedCollection() {
...
+
+    assertEquals(summary.getDarCollectionId(), dar.getCollectionId());
+    //create child progress report
+    String progressReportReferenceId = createProgressReportFromDAR(dar);
+    validateSummaryObjectWithParent(userId, progressReportReferenceId, dar);
+    //create grandchild progress report
+    DataAccessRequest progressReportDar = dataAccessRequestDAO.findByReferenceId(progressReportReferenceId);
+    String progressReportReferenceId2 = createProgressReportFromDAR(progressReportDar);
+    validateSummaryObjectWithParent(userId, progressReportReferenceId2, progressReportDar);
+    //create a second grandchild, peer to the first
```
otchet-broad: Yes, our model currently supports this!
Closing this path will mean we can likely shift to building a LinkedList or some other order preserving structure to simplify to the "chaining" model we spoke about.  I do think we need to understand our PO's perspective on if they think we'll need to support branching before making the change away from this.

rushtong: Agreed that we need to confirm the chained approach with product or go back to the tech design phase.

src/main/java/org/broadinstitute/consent/http/models/DarCollectionSummary.java
```diff
@@ -52,6 +52,12 @@ public class DarCollectionSummary {
   @JsonProperty
   private int datasetCount;
 
+  @JsonProperty
+  private Map<String, Set<String>> parentToReferenceIds;
```
otchet-broad: Left the key as <String> knowing that we're moving back to using the UUID String value here.

src/main/java/org/broadinstitute/consent/http/db/mapper/DarCollectionSummaryReducer.java
```diff
@@ -44,6 +49,28 @@ public void accumulate(Map<Integer, DarCollectionSummary> map, RowView rowView)
           summary.addReferenceId(darReferenceId);
         }
 
+        if (hasColumn(rowView, "dar_all_reference_id", String.class)) {
+          aggReferenceId = rowView.getColumn("dar_all_reference_id", String.class);
+          if (Objects.nonNull(aggReferenceId)) {
+            summary.addReferenceId(aggReferenceId);
+            if (hasColumn(rowView, "dar_all_parent_id", String.class)) {
+              aggParentId = rowView.getColumn("dar_all_parent_id", String.class);
```
otchet-broad: ~~Pulling in as String to be consistent with our other implementation.  Knowing we're going back the the UUID referenceId made me do this now, but if we want this released before we're going to do that work, I would change to Integer so the contents work as expected on the front end.~~

src/main/java/org/broadinstitute/consent/http/db/mapper/DarCollectionSummaryReducer.java
```diff
@@ -44,6 +49,28 @@ public void accumulate(Map<Integer, DarCollectionSummary> map, RowView rowView)
...
+              if (Objects.nonNull(aggParentId)) {
+                summary.addParentToReferenceId(aggParentId, aggReferenceId);
+              } else {
+                summary.addParentToReferenceId(null, aggReferenceId);
```
otchet-broad: I should probably switch the key to this entry to the id of the DAR for now, updating it to the referenceId when we have it.
From a graph theory perspective, switching to the referenceId for keys makes this a much cleaner implementation.  That will fall in naturally when we move parent_id to hold the referenceId.  I don't think the alternative of using the data_access_request.id field makes sense for the Set values because the front-end code and our APIs already have access patterns based on the referenceId

src/main/java/org/broadinstitute/consent/http/models/DarCollectionSummary.java
otchet-broad: I actually changed my mind and made this an integer, since that's where we are right now.  This should be adjusted when we make the change.

src/main/java/org/broadinstitute/consent/http/db/mapper/DarCollectionSummaryReducer.java
otchet-broad: Reverted building out this implementation based on conversation in ticket.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -22,8 +22,8 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
   @RegisterBeanMapper(value = Election.class)
   @UseRowReducer(DarCollectionSummaryReducer.class)
   @SqlQuery("""
-      SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, u.display_name as researcher_name,
-        i.institution_name, e.election_id, e.status, e.dataset_id, e.reference_id, v.voteid as v_vote_id, dd.dataset_id as dd_datasetid,
+      SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, dar.parent_id as dar_parent_id,
```
otchet-broad, line 25: Add new column and adjust query line for length.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -59,8 +59,8 @@ List<DarCollectionSummary> getDarCollectionSummariesForDAC(
   @SqlQuery
       (
           """
-              SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, u.display_name as researcher_name,
-                i.institution_name, e.election_id, e.status, e.dataset_id, e.reference_id, dd.dataset_id as dd_datasetid,
+              SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id as dar_reference_id, dar.parent_id as dar_parent_id,
```
otchet-broad: Add new column and adjust query line for length.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -194,7 +196,8 @@ WHERE LOWER(election.election_type) = 'dataaccess'
   @RegisterBeanMapper(value = Election.class)
   @UseRowReducer(DarCollectionSummaryReducer.class)
   @SqlQuery("""
-      SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, u.display_name as researcher_name, u.user_id as researcher_id,
+      SELECT c.collection_id as dar_collection_id, c.dar_code, dar.submission_date, dar.reference_id AS dar_reference_id,
```
otchet-broad, line 236: Add new columns and adjust query line for length.

src/test/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAOTest.java
otchet-broad: Removed this part of the test since we're still deciding if we support it or not (outside the scope of this ticket).

rushtong (approved):
Looks reasonable to me 👍🏽

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
```diff
@@ -378,4 +385,11 @@ private Map<String, Object> shallowCopy(DataAccessRequest dar) {
     }
     return copy;
   }
+
+  // Simple state machine for determining if the Data Access Request
+  // is a progress report.
+  private void updateProgressReportState() {
+    progressReport = !getDraft() && (getParentId() != null);
+  }
```
snf2ye, line 393: What if there is a DAR that is a draft and has a parent id?
I wonder if checking both of these fields will leave dars orphaned as neither a progress report nor a dar.

otchet-broad: Drafts cannot have progress reports.  It's a business rule enforced by src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java:251 when a progress report is submitted to the backend service.

snf2ye (approved):
Change LGTM. I'm still a little unclear on how these fields would be used, but I still think it is a positive change

src/main/java/org/broadinstitute/consent/http/db/mapper/DarCollectionSummaryReducer.java
```diff
@@ -44,6 +45,11 @@ public void accumulate(Map<Integer, DarCollectionSummary> map, RowView rowView)
           summary.addReferenceId(darReferenceId);
         }
 
+        darParentId = rowView.getColumn("dar_parent_id", Integer.class);
+        if (Objects.nonNull(darParentId)){
+          summary.addParentToReferenceId(darParentId, darReferenceId);
+        }
```
pshapiro4broad: `RowMapperHelper` has `hasOptionalColumn()` which avoids the need for a local variable and removes the need to specify the field name in two places.
```suggestion
        hasOptionalColumn(rowView, "dar_parent_id", Integer.class)
            .ifPresent(darParentId -> summary.addParentToReferenceId(darParentId, darReferenceId));
```

Given that the common use of `hasOptionalColumn()` is to call a setter, an even more streamlined version of this would accept `Consumer` instead of returning `Optional`.

```
  default <T> void withColumnDo(RowView rowView, String columnName, Class<T> clazz, Consumer<T> consumer) {
    try {
      T value = rowView.getColumn(columnName, clazz);
      if (value != null) {
        consumer.accept(value);
      }
    } catch (Exception e) {
      logDebug("RowView does not contain column %s".formatted(columnName));
    }
  }
```

```
        withColumnDo(rowView, "dar_parent_id", Integer.class, 
            darParentId -> summary.addParentToReferenceId(darParentId, darReferenceId));
```

src/main/java/org/broadinstitute/consent/http/models/DarCollectionSummary.java
```diff
@@ -251,6 +267,14 @@ public void addDacName(String dacName) {
     }
   }
 
+  public boolean getProgressReport() {
+    return progressReport;
+  }
+
+  private void updateProgressReportStatus() {
```
pshapiro4broad, line 274: Computed values that require an explicit update like this are hard to keep working correctly. Since gson still doesn't support getter-based properties, I can think of two approaches

- only generate these objects at the point before serialization, and have a generic "update computed fields" method which handles all cases. Or, change our code to use a builder pattern, and generate computed field values when the object is built
- switch to using Jackson or some other JSON library for the cases where we want to do this

otchet-broad: I'd like to see if we can return to using Jackson but given current time constraints, I don't think that's possible in the context of this PR.

otchet-broad:
@raejohanek - Thanks again for your help mobbing on the tests after the merge!  This is ready for your eyes!

src/test/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAOTest.java
```diff
@@ -1050,4 +1056,126 @@ void testGetDarCollectionSummaryByCollectionId_ArchivedCollection() {
...
+    Dac dac = createDac();
+    User user = createUserWithInstitution();
+    User userChair = createUserWithRoleInDac(UserRoles.CHAIRPERSON.getRoleId(), dac.getDacId());
+    Integer userId = user.getUserId();
+    Integer chairId = userChair.getUserId();
+    Integer dacId = dac.getDacId();
+    Dataset dataset = createDatasetWithDac(userId, dacId);
+
+    DarCollectionSummary summary = createDarWithVotes(userId, chairId, dataset.getDatasetId());
+    return Pair.of(Pair.of(userId, chairId), summary);
```
rushtong: This must have been fun. I don't see the need for the nested Pairs though. I might have considered a custom record class to make this a bit easier to work with:
```
public record Setup(int userId, int chairId, DarCollectionSummary summary)
```

rushtong (approved):
New tests look good, thanks for the changes. Feel free to address comment as you see fit 👍🏽

src/main/java/org/broadinstitute/consent/http/models/DarCollectionSummary.java
```diff
@@ -87,6 +93,15 @@ public void addReferenceId(String id) {
     this.referenceIds.add(id);
   }
 
+  public void addParentToReferenceId(Integer parentId, String referenceId) {
+    parentToReferenceIds.computeIfAbsent(parentId, k -> new HashSet<>()).add(referenceId);
```
rjohanek: this is a map from parent databaseId to child reference Id right? not parent databaseId to parent referenceId? Maybe a comment or renaming parameter to childReferenceId could clarify

src/main/java/org/broadinstitute/consent/http/models/DarCollectionSummary.java
```diff
@@ -251,6 +267,14 @@ public void addDacName(String dacName) {
     }
   }
 
+  public boolean getProgressReport() {
+    return progressReport;
+  }
+
+  private void updateProgressReportStatus() {
+    progressReport = !getParentToReferenceIds().isEmpty() && Objects.nonNull(getSubmissionDate());
```
rjohanek, line 275: why do we need to check that the submission date is nonNull?

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
```diff
@@ -378,4 +385,11 @@ private Map<String, Object> shallowCopy(DataAccessRequest dar) {
     }
     return copy;
   }
+
+  // Simple state machine for determining if the Data Access Request
+  // is a progress report.
+  private void updateProgressReportState() {
+    progressReport = !getDraft() && (getParentId() != null);
```
rjohanek, line 392: why do we need to check that it isn't a draft? I know progress reports don't support draft status so they won't be drafts, but why do we need to check that here? Is checking parentId not enough?

src/main/java/org/broadinstitute/consent/http/models/DarCollectionSummary.java
otchet-broad: We have this rule that valid progress reports can't be cancelled and are (presently) immutable.  This is to ensure the rule is codified throughout the code.

rjohanek (approved):
looks good, I like the new tests

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
otchet-broad: Your knowledge of `progress reports don't support draft status` isn't codified anywhere.  It's just our unwritten rule.  This codifies that so future developers are aware.

src/main/java/org/broadinstitute/consent/http/models/DarCollectionSummary.java
otchet-broad: Good point on clarifying.  I'll rename.
I also suspect we're going to change this in the very near future to be something closer to referenceId (parent) to set<referernceId> children.

pshapiro4broad (approved):

PR #2520, opened 2025-05-09
[DT-1570]Add CREATE_PROGRESS_REPORT Action

rjohanek:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1570

### Summary
Adds a new action CREATE_PROGRESS_REPORT
Adds the action to a DarCollectionSummary for Researchers if there at least one DAR in the collection with at least one approved Dataset

### Bug fix
This also includes a fix to findApprovedDatasetsByDar - this method is used to determine if a progress report can be made. It was filtering out expired Datasets, but a DAR can still be made in the case that all access is expired. So, I removed that line in the query and added a test. I think this bug might have existed because another similar method, findApprovedDARsByDatasetId, does need to filter out expired DARs, because it is used to determine access.

### Testing
Adds new unit test and expands existing unit test

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
@@ -102,47 +102,6 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
    * `FALSE` in the example above. Outside the JOIN, we filter on groupings where the final vote
    * value is `TRUE` so the denied election in the example would be filtered out.
    *
-   * @param darReferenceId The DAR reference UUID
-   * @return Set of approved Dataset Ids for the DAR
-   */
-  @SqlQuery(
-      """
-      SELECT dd.dataset_id
-      FROM data_access_request dar
-      INNER JOIN dar_dataset dd ON dd.reference_id = dar.reference_id
-      INNER JOIN (
-        SELECT DISTINCT e.reference_id, e.dataset_id, LAST_VALUE(v.vote)
-        OVER(
-          PARTITION BY e.dataset_id
-            ORDER BY v.createdate
-            RANGE BETWEEN
-              UNBOUNDED PRECEDING AND
-              UNBOUNDED FOLLOWING
-        ) last_vote
-        FROM election e
-        INNER JOIN vote v ON e.election_id = v.electionid AND v.vote IS NOT NULL
-        AND LOWER(e.election_type) = 'dataaccess'
-        AND LOWER(v.type) = 'final') final_access_vote ON
-          final_access_vote.reference_id = dar.reference_id AND
-          final_access_vote.dataset_id = dd.dataset_id
-      WHERE final_access_vote.last_vote = TRUE
-      AND dar.reference_id = :darReferenceId
-      AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)
-      """)
-  Set<Integer> findDatasetApprovalsByDar(@Bind("darReferenceId") String darReferenceId);
-
-  /**
-   * This query finds dataset ids on dar-dataset combinations where the most recent vote is true.
-   * This includes datasets that are a part of expired DARs, UNLIKE findApprovedDARsByDatasetId.
-   * The query accomplishes this by creating a view that is a grouping
-   * of election reference ids and LAST vote in the group of final votes for all data access
-   * elections. We need to group them due to the case of multiple elections on a dar-dataset
-   * request. Election 1 may have been denied. Election 2 may have been approved. Election 3 may
-   * have been denied again. When we partition over the election reference id, we'll get all final
-   * votes. The `LAST_VALUE` function selects the last result in the partition, which would be
-   * `FALSE` in the example above. Outside the JOIN, we filter on groupings where the final vote
-   * value is `TRUE` so the denied election in the example would be filtered out.
-   *
    * @param darReferenceIds The DARs reference UUIDs
    * @return Set of approved Dataset Ids for the list of DARs
    */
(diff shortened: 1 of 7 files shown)


src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -159,6 +159,12 @@ private void processDarCollectionSummariesForResearcher(List<DarCollectionSummar
       int electionCount = elections.values().size();
       elections.values().forEach(election -> updateStatusCount(statusCount, election.getStatus()));
       s.addAction(DarCollectionActions.REVIEW.getValue());
+      // if any DARs in the collection have approved datasets, include create progress report action
+      if (s.getReferenceIds().stream()
+          .map(dataAccessRequestDAO::findApprovedDatasetsByDar)
```
rushtong: This will work for now, but it is a looped DAO call inside of another loop of summaries and will become problematic over time when a user has lots of DARs with many datasets. It would be more efficient to query for the datasets before the loop. This might require an additional query that does the same work as `dataAccessRequestDAO::findApprovedDatasetsByDar` but instead for a list of reference ids.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -124,13 +125,53 @@ AND LOWER(e.election_type) = 'dataaccess'
         AND LOWER(v.type) = 'final') final_access_vote ON
           final_access_vote.reference_id = dar.reference_id AND
           final_access_vote.dataset_id = dd.dataset_id
-      WHERE dar.submission_date > now() - interval '1 year'
-      AND final_access_vote.last_vote = TRUE
+      WHERE final_access_vote.last_vote = TRUE
```
rjohanek: bug fix

rushtong: I'm not sure this is a bug fix. The name is `findApprovedDatasetsByDar` and expired DAR-Dataset combinations are, by definition, NOT approved. In the PR description, it looks like this method was being mis-used for something it wasn't meant for. I think a real fix would be to create a method that works for the specific case in question. If the only use for this method is for progress reports, then this method should be renamed since it is not doing what it says it is now.

rjohanek: this method was created for the purpose of checking if a PR can be created. that is its only use. I can rename it! But, I'm not sure what would be clearer, findApprovedAndExpiredDatasetsByDar?

snf2ye: This was the only change in my previous PR - https://github.com/DataBiosphere/consent/pull/2487/files
So, unless another change is made, we would now include expired DARS for the getApprovedUsers endpoint and users would not longer get their access revoked via TDR.

rushtong: I think the distinction we're trying to make here is datasets with approvals vs approved datasets. The phrase, "approved datasets", has specific meaning due to existing APIs and queries. Dataset approvals is actually a different thing and that's what I think we're looking for here. Maybe `findDatasetApprovalsByDARReferenceId`?

rjohanek: @snf2ye that change that you made is for the findApprovedDarsByDatasetId method. That is an important change that remains intact. This bug fix is for a different method findApprovedDatasetsByDar which is not used for the getApprovedUsers endpoint. It is used in the postProgressReport endpoint. Specifically, it is used to determine if a parent dar has any approvals on any datasets, then we can create a progress report. Thus we must include expired dars in this case. This fixes the bug in this method while leaving all logic for the getApprovedUsers endpoint untouched.

snf2ye: Ah, that makes sense. I forgot that this line of code was added in another place.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -124,12 +125,52 @@ AND LOWER(e.election_type) = 'dataaccess'
...
+              UNBOUNDED FOLLOWING
+        ) last_vote
+        FROM election e
+        INNER JOIN vote v ON e.election_id = v.electionid AND v.vote IS NOT NULL
+        AND LOWER(e.election_type) = 'dataaccess'
+        AND LOWER(v.type) = 'final') final_access_vote ON
+          final_access_vote.reference_id = dar.reference_id AND
+          final_access_vote.dataset_id = dd.dataset_id
+      WHERE final_access_vote.last_vote = TRUE
+        AND dar.reference_id IN (<darReferenceIds>)
```
snf2ye, line 129: The main difference between findDatasetApprovalsByDar and findDatasetApprovalsByDars is that you can pass in multiple dar reference ids here, right? And it looks like this is the only line that is different.  

Is there an opportunity to share the code for this query between these two methods? 
I actually think this line would be valid even in the case that there is only a single reference id. Do we even need two queries? We could just wrap the singleton case in a list.

rushtong: > Do we even need two queries? We could just wrap the singleton case in a list.

I think this is a good idea. 

The issue we face with re-using string blocks in queries is that semgrep considers that an anti-pattern prone to sql injection problems. There might be ways around that with sql functions, but we haven't dived into that yet. Another potential solution would be to use multiple queries that are more legible.

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
```diff
@@ -675,13 +681,136 @@ void testFindAllApprovedDatasetsByDar() {
...
+        e4.getElectionId(),
+        now,
+        false);
+
+    Set<Integer> approvedDatasetIds = dataAccessRequestDAO.findDatasetApprovalsByDars(
+        List.of(testDar1.getReferenceId(), testDar2.getReferenceId()));
+    assertEquals(2, approvedDatasetIds.size());
+    assertTrue(approvedDatasetIds.containsAll(List.of(dataset1.getDatasetId(), dataset4.getDatasetId())));
+  }
+  @Test
```
rushtong: minor nit: formatting

rushtong (approved):
Thanks for the updates, address comments as you see fit 👍🏽

snf2ye (approved):

PR #2522, opened 2025-05-09
Minor updates to db-connect script

pshapiro4broad:
### Addresses

- add `--project` argument to `gcloud secrets` operation, so the correct google project is always used
- add support for `prod`, remove `alpha` since we don't have an `alpha` deployment anymore

### Testing

Manually tested locally against `dev`, `staging`, and `prod`. Note that `prod` database connection requires authenticating using a suitable user account and being on the Broad VPN.

----

- Verify all tests go green

### DIFF
--- a/scripts/db-connect.sh
+++ b/scripts/db-connect.sh
@@ -6,6 +6,19 @@
 # USAGE: ./db-connect.sh ENV
 #   ENV must be one of dev, staging, or prod
 #
+# NOTE: This does not support bee deployments. You can connect to a bee deployment database
+# by running these commands, where <pshapiro-test> is the name of your bee:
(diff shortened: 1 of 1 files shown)


rushtong (approved):
👍🏽  Looks good, thank you. I think doing this for Bees might be an interesting task, but probably beyond the intended scope of this PR.

pshapiro4broad:
> @rushtong I think doing this for Bees might be an interesting task

I was just doing this recently so I added a comment on how to do this. It's a different approach than the script takes, so I didn't add support for it, but someone could do that later on if they like.

s-rubenstein (approved):

PR #2523, opened 2025-05-09
DT-1564, Progress Report DAC flow

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-1564](https://broadworkbench.atlassian.net/browse/DT-1564)

### Summary

- Notify the DAC chair on progress report submission
- Notify the DAC members when the chair opens an election on progress reports
- Notify the Researcher when the DAC approves a progress report
- Keep the API and message frequency the same

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
```diff
@@ -144,6 +148,16 @@ public void addDar(DataAccessRequest dar) {
...
   }
 
+  public DataAccessRequest getMostRecentDar() {
+    List<DataAccessRequest> dars = new java.util.ArrayList<>(
+        this.getDars().values().stream().toList());
+    dars.sort(Comparator.comparing(DataAccessRequest::getSubmissionDate).reversed());
+    if (dars.isEmpty()) {
+      return null;
+    }
+    return dars.get(0);
```
pshapiro4broad: You can pass a comparator to `Stream.sorted()`. Also since you want the first element or null, you can get the last element or null, without reversing. Or, you can get the min or max value directly from a stream with a comparator and avoid sorting it.
```suggestion
    return dars.values().stream()
        .max(Comparator.comparing(DataAccessRequest::getSubmissionDate))
        .orElse(null);
```

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
```diff
@@ -203,17 +207,25 @@ public void sendNewDARCollectionMessage(Integer collectionId)
           sendList.put(dac.getName(), matchingDatasetsForDac);
         }
       }
-      sendNewDARRequestEmail(user, sendList, researcherName, collection.getDarCode());
+      // If the dar is not a progress report, use the DAR template else use the PR template.
+      if (dar.getProgressReport()) {
```
pshapiro4broad: We now have both `DarCollectionSummary.getProgressReport()` and `DataAccessRequest.getProgressReport()`. 

Could we add `getProgressReport()` to `DarCollection` as `getMostRecentDar().getProgressReport()`? Then we can use `DarCollection.getProgressReport()` in `DarCollectionSummary()` instead of having it compute it separately.

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
```diff
@@ -180,17 +181,20 @@ public void sendNewDARCollectionMessage(Integer collectionId)
               + collectionId);
       return;
     }
-    List<User> distinctUsers = getDistinctAdminAndChairUsersForCollection(collection);
+    // Do this, but only for a single DAR
+    DataAccessRequest dar = collection.getMostRecentDar();
```
pshapiro4broad: This approach seems fine, but I think this should be a method in `DarCollectionService`, not in `EmailService`, since it has specific knowledge about the dar collection business logic.

otchet-broad: Perspective appreciated.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -449,6 +449,13 @@ public Response postProgressReport(
           ethicsFileDetails, payload, parentDar);
       DataAccessRequest progressReport = dataAccessRequestService.createProgressReport(user,
           payload, parentDar);
+      try {
+        // This service method knows how to distinguish between a DAR and a Progress Report.
+        emailService.sendNewDARCollectionMessage(parentDar.getCollectionId());
```
snf2ye: Unlikely scenario, but just to test my understanding: 
Since we're only passing along a collection ID, if two progress reports are created in quick succession, we could end up never sending an email about the first progress report and two emails about the second one. This is because we're always just getting the latest DAR in `sendNewDARCollectionMessage`.
- progress report 1 created
- progress report 2 created
- Email send about progress report 2
- Email send about progress report 2

otchet-broad: Correct, although this really just points out that there are a number of edge cases in this general logic.  You could have canceled the election before the email was actually sent and then you would be sending a message about a canceled election.

otchet-broad: ... and in your example, other than date sent, the emails will be identical since they both reference the same dar code (e.g. DAR-123).

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
otchet-broad: I'm not sure I'm following the thought here.  I don't see where DarCollectionSummary currently uses DarCollection today.  Maybe we can walk through your idea.

In a similar vain though, I'm wondering if we can eliminate DarCollection.  As you'll see in my refactor it's clear that we're moving into spaces where DARs can be grouped together by an ID (collection ID or darCode).  I'm not sure we really need the DarCollection object anymore.

src/main/java/org/broadinstitute/consent/http/db/mapper/DataAccessRequestMapper.java
```diff
@@ -25,7 +25,7 @@ public DataAccessRequest map(ResultSet resultSet, StatementContext statementCont
     if (hasColumn(resultSet, "dar_code")) {
       dar.setDarCode(resultSet.getString("dar_code"));
     }
-    dar.setParentId(resultSet.getString("parent_id"));
+    dar.setParentId(resultSet.getInt("parent_id"));
```
otchet-broad, line 28: We were coercing the `bigint` value in the database into a `String`.  I expect we'll go back to `String` in a different pull request, but this change makes our consent and duos-ui code clear on what the value is today.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -619,7 +630,7 @@ public DataAccessRequest populateProgressReportFromJsonString(String json,
 
     String referenceId = UUID.randomUUID().toString();
     newDar.setReferenceId(referenceId);
-    newDar.setParentId(parentDar.getId().toString());
+    newDar.setParentId(parentDar.getId());
```
otchet-broad, line 616: @raejohanek , @rushtong - I think this is a point of confusion.  The DAR body takes the parentId, the endpoint takes the referenceId of the parent DAR.  When we convert parentId, we might want to be more explicit with this and call it parentReferenceId.  This should help disambiguate from the numeric `id` property we're using now.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -664,19 +678,25 @@ public DarCollection cancelDarCollectionElectionsAsChair(DarCollection collectio
...
-      List<String> createdElectionReferenceIds = collectionServiceDAO.createElectionsForDarCollection(
-          user, collection);
+      DataAccessRequest dar = collection.getMostRecentDar();
+      List<String> createdElectionReferenceIds = collectionServiceDAO.createElectionsForDarByUser(
+          user, dar);
       if (createdElectionReferenceIds.isEmpty()) {
         var e = new IllegalStateException(
-            "No elections were created for DAR Collection: %s".formatted(
-                collection.getDarCode()));
+            "No elections were created for DAR Collection: %s %s".formatted(
```
otchet-broad, line 681: The collection is ambiguous.  Adding the reference ID makes it clearer about what was happening at the time this error was encountered.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -363,4 +383,70 @@ public Collection<DataAccessRequest> getApprovedDARsForDataset(Dataset dataset)
     return dataAccessRequestDAO.findApprovedDARsByDatasetId(dataset.getDatasetId());
   }
 
+  public void sendExpirationNotices() {
```
otchet-broad, line 385: Moving this code from the EmailService to the DataAccessRequestService to keep business logic pertaining to DARs out of the Email Service.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -700,4 +720,116 @@ private void cancelElectionsForReferenceIds(List<String> referenceIds) {
       }
     });
   }
+
+  private void sendDarNewCollectionElectionMessage(List<User> users, String darCode)
```
otchet-broad: Moving DarCollection business logic from the EmailService to the DarCollectionService.

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
```diff
@@ -52,64 +30,34 @@
...
-  private final DatasetDAO datasetDAO;
-  private final DacDAO dacDAO;
   private final FreeMarkerTemplateHelper templateHelper;
   private final SendGridAPI sendGridAPI;
   private final String fromAccount;
   private final String serverUrl;
 
   @Inject
   public EmailService(
-      DarCollectionDAO collectionDAO,
```
otchet-broad, line 93: All of the DAOs needed for business logic can now be removed from the EmailService.

src/main/java/org/broadinstitute/consent/http/service/VoteService.java
```diff
@@ -289,30 +283,25 @@ public void sendDatasetApprovalNotifications(List<Vote> votes, User user) {
       logException("Error indexing datasets for approved DARs: " + e.getMessage(), e);
     }
 
-    // For each dar collection, email the researcher summarizing the approved datasets in that collection
-    collections.forEach(c -> {
+    // For each dar, email the researcher summarizing the approved datasets in that dar
```
otchet-broad, line 286: Work on DARs, not collection objects because we don't want to take all of the actions contemplated by the original code on all of the DARs in the collection.  We only want to work on the DARs from the elections that were tied to the votes provided as input to this method.

src/main/java/org/broadinstitute/consent/http/service/dao/DarCollectionServiceDAO.java
```diff
@@ -37,17 +37,17 @@ public DarCollectionServiceDAO(DatasetDAO datasetDAO, ElectionDAO electionDAO, J
...
    * initiating a new set of elections. Any DAR elections in open state should be ignored.
    *
-   * @param user       The User initiating new elections for a collection
-   * @param collection The DarCollection
+   * @param user       The User initiating new elections for a data access request
+   * @param dataAccessRequest The DataAccessRequest
    * @return List of reference ids for which a DAR election was created
    */
-  public List<String> createElectionsForDarCollection(User user, DarCollection collection)
+  public List<String> createElectionsForDarByUser(User user, DataAccessRequest dataAccessRequest)
```
otchet-broad: We now perform this action on a specific DAR, not the entire collection because we should not be operating on the prior data access requests that were part of the collection at this point.

src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
```diff
@@ -1271,4 +1335,32 @@ private Election createMockElection() {
     return election;
   }
 
+  private Dataset createDataset(Integer dacId) {
```
otchet-broad, line 1356: This was moved from EmailServiceTest

src/main/java/org/broadinstitute/consent/http/mail/message/NewProgressReportCaseMessage.java
```diff
@@ -0,0 +1,41 @@
...
+
+  public NewProgressReportCaseMessage(User toUser, String referenceId, String type) {
+    super(toUser, EmailType.NEW_PROGRESS_REPORT_CASE);
+    this.referenceId = referenceId;
+    this.type = type;
+  }
+
+  @Override
+  public String createSubject() {
+    if (type.equals("Data Use Limitations")) {
```
pshapiro4broad: Maybe it's time to make a constant for this magic string? I now see it in three places, and it appears to come from `Election.getElectionType()`, but I don't see how the original value is set. Also it looks like its usage in `NewCaseMessage` may be bogus...

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
rushtong: > I'm wondering if we can eliminate DarCollection

I think we need to move in this direction ... will file a ticket.

rushtong: https://broadworkbench.atlassian.net/browse/DT-1665

src/main/java/org/broadinstitute/consent/http/enumeration/EmailType.java
```diff
@@ -22,7 +22,10 @@ public enum EmailType {
   DATASET_DENIED(18, "dataset-denied.html"),
   DATASET_APPROVED(19, "dataset-approved.html"),
   DAR_EXPIRED(20, "dar-expired.html"),
-  DAR_EXPIRATION_REMINDER(21, "dar-expiration-reminder.html");
+  DAR_EXPIRATION_REMINDER(21, "dar-expiration-reminder.html"),
+  NEW_PROGRESS_REPORT(22, "new-progress-report-request.html"),
```
rushtong: I would be confused seeing this enum without context - can we rename it to more closely match the function?
```suggestion
  NEW_PROGRESS_REPORT_REQUEST(22, "new-progress-report-request.html"),
```

src/main/java/org/broadinstitute/consent/http/mail/message/NewProgressReportCaseMessage.java
```diff
@@ -0,0 +1,41 @@
...
+      return String.format(NEWCASE_DUL, referenceId);
+    } else {
```
rushtong: This should be a defunct case. We don't make DUL elections any more which is where I believe this string comes from. I would prefer that this message handle just the `Data Access Request` case

src/main/resources/freemarker/new-progress-report-case.html
```diff
@@ -0,0 +1,37 @@
...
+            </td>
+        </tr>
+        <tr>
+            <td id="userName" style="padding: 15px 15px 0px; font-family: 'Montserrat', sans-serif; font-size: 22px; color: #1F3B50; text-align: left; line-height: 25px; display: block; font-weight: 500;">
+                Hello ${userName},
+            </td>
+        </tr>
+        <tr>
+            <td style="padding: 15px 15px 0px; font-family: 'Montserrat', sans-serif; font-size: 16px; color: #1F3B50; text-align: justify; line-height: 25px;">
+                ${electionType} Review case id ${entityName}, has been created. Please click on the following link to log your vote on it:
```
rushtong: I realize that this is essentially a clone of the DAR version. I don't think `${electionType}` needs to be injected here, I think it will always be a Data Access type since that's all that we have currently. There is the RP election type, but we don't send separate emails for that type, nor should we.

src/test/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceTest.java
```diff
@@ -250,10 +241,7 @@ void testUpdateByReferenceIdForbidden() {
     try {
       when(userService.findUserByEmail(any())).thenReturn(invalidUser);
       when(dataAccessRequestService.findByReferenceId(any())).thenReturn(dar);
-      resource =
-          new DataAccessRequestResource(daaService,
-              dataAccessRequestService, emailService, gcsService, userService, datasetService,
-              matchService);
+      initResource();
```
rushtong: It might be possible to move all of these `initResource();` calls into a `@BeforeEach` method and simplify this more.

rushtong (approved):
Thank you for the large refactor, this looks pretty good. Address comments inline as you see fit 👍🏽

src/main/java/org/broadinstitute/consent/http/mail/message/NewProgressReportCaseMessage.java
rushtong: See https://broadworkbench.atlassian.net/browse/DT-1666 for follow-on work for this problem.

src/test/java/org/broadinstitute/consent/http/resources/DataAccessRequestResourceTest.java
otchet-broad: Good call!

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -363,4 +381,72 @@ public Collection<DataAccessRequest> getApprovedDARsForDataset(Dataset dataset)
...
+        User user = userDAO.findUserById(expiredDar.getUserId());
+        String darCode = expiredDar.getDarCode();
+        String userName = user.getDisplayName();
+        if (user.getEmail() == null) {
+          // Do not throw here.  Log information about the DAR since this will continue
+          // to appear broken until manual intervention is taken to resolve the missing user
+          // email address
+          logException(new Exception(String.format(
+              "Email address for user %d (%s) not found for expiring warning.  DAR reference id: %s",
+              expiredDar.getUserId(), userName, referenceId)));
```
pshapiro4broad: Since this is inside a try/catch that logs the exception, you could throw here and it would be caught and logged. That way you don't need to use `else`
```suggestion
          throw new Exception(String.format(
              "Email address for user %d (%s) not found for expiring warning.  DAR reference id: %s",
              expiredDar.getUserId(), userName, referenceId));
```

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -363,4 +381,72 @@ public Collection<DataAccessRequest> getApprovedDARsForDataset(Dataset dataset)
...
+        } else {
+          switch (type) {
+            case DAR_EXPIRATION_REMINDER:
```
pshapiro4broad: Up to you, but it bothers met that we need to re-derive which caller we have from `type`, as the caller is known above. If you wanted to, you could pass the `sendDarExpirationReminderMessage()` method as a parameter to this method. Or, if `EmailService.sendMessage()` was `public`, you could pass the message constructor. This also lets us avoid handling a `default` which can be annoying.
```
  interface MessageCreator {
    MailMessage createMessage(User user, String darCode, String referenceId);
  }

  private void sendDARExpirationNotices() {
    EmailType emailType = EmailType.DAR_EXPIRED;
    sendDARMessageToList(emailType, EXPIRE_NOTICE_INTERVAL, DarExpiredMessage::new);
  }

  private void sendDARExpirationReminderNotices() {
    EmailType emailType = EmailType.DAR_EXPIRATION_REMINDER;
    sendDARMessageToList(emailType, EXPIRE_WARN_INTERVAL, DarExpirationReminderMessage::new);
  }

  private void sendDARMessageToList(EmailType type, String interval, MessageCreator messageCreator) {
[...]
        } else {
          emailService.sendMessage(messageCreator.createMessage(user, darCode, referenceId), user.getUserId());
        }
```

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -363,4 +381,72 @@ public Collection<DataAccessRequest> getApprovedDARsForDataset(Dataset dataset)
...
+    });
+  }
+
+  public void sendReminderMessage(Integer voteId) throws IOException, TemplateException {
+    Vote vote = voteDAO.findVoteById(voteId);
+    Election election = electionDAO.findElectionWithFinalVoteById(vote.getElectionId());
+    DarCollection collection = darCollectionDAO.findDARCollectionByReferenceId(
+        election.getReferenceId());
+    User user = findUserById(vote.getUserId());
+    String voteUrl = serverUrl + "dar_collection/%d".formatted(collection.getDarCollectionId());
```
pshapiro4broad, line 440: Since this change is a refactoring, you don't need to do this now, but I think the need to use `serverUrl` here is a sign that the templating for this email isn't quite correct. The email service code already has `serverUrl` and passes it to `MailMessage.createModel()`. If we want to build a path relative to `serverUrl` in the email, we can do it by using two substitutions:
```
                <a href="${serverUrl}${votePath}" style="text-decoration: none; font-family: 'Montserrat', sans-serif; color: #00609F; font-size: 20px; font-weight: bold;">
                    Click here to log your vote
                </a>
```
```
    String votePath = "dar_collection/%d".formatted(collection.getDarCollectionId());
```

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -363,4 +381,72 @@ public Collection<DataAccessRequest> getApprovedDARsForDataset(Dataset dataset)
...
+    emailService.sendReminderMessage(user, vote, collection.getDarCode(), election.getElectionType(), voteUrl);
+    voteDAO.updateVoteReminderFlag(voteId, true);
```
pshapiro4broad, line 442: I wonder if we need a separate db operation for this, vs using the email table to determine if the reminder has been sent or not. Hmm, looking at this more, we are updating `Vote.isReminderSent` but don't seem to be looking at it, so if it was intended to prevent multiple reminders from being sent, it doesn't appear to be doing that.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -363,4 +381,72 @@ public Collection<DataAccessRequest> getApprovedDARsForDataset(Dataset dataset)
...
+  }
+
+  private User findUserById(Integer id) throws IllegalArgumentException {
+    User user = userDAO.findUserById(id);
```
pshapiro4broad, line 446: Again, this is a code refactoring, but there are many calls to `userDAO.findUserById()` in consent which do a similar check. Also I see that in `sendDARMessageToList()` we call `findUserById()` and not check for `null`. Probably worth auditing the code for consistency here.

Although if `userID` is a foreign key in `Vote`, this should never fail, correct?

src/main/java/org/broadinstitute/consent/http/service/EmailService.java
pshapiro4broad: Thank you!

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -132,10 +136,10 @@ public Response createDataAccessRequest(
       DataAccessRequest newDar = dataAccessRequestService.createDataAccessRequest(user, payload);
       Integer collectionId = newDar.getCollectionId();
```
pshapiro4broad: Maybe I'm missing something, but I would expect this block and the one in `createDataAccessRequestWithDAARestrictions` to be replaced with
```
      sendNewDarCollectionMessage(newDar.getCollectionId());
```

pshapiro4broad: Also it looks like these two methods could be merged into one, since the only differences are `enforceDAARestrictions()` and `logDARSubmission()`, and it's possible we want `logDARSubmission()` for both cases.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -449,12 +453,27 @@ public Response postProgressReport(
...
     }
   }
 
+  private void sendNewDarCollectionMessage(Integer collectionId) {
+    try {
+      // This service method knows how to distinguish between a DAR and a Progress Report.
+      darCollectionService.sendNewDARCollectionMessage(collectionId);
+    } catch (Exception e) {
+      // non-fatal exception
+      logCaughtEmailException(collectionId, e);
```
pshapiro4broad, line 444: In the expiration notification we catch and log exceptions on a per-user basis, instead of at the top level. Making this per-user as well would probably be better.

pshapiro4broad (commented):
Comments so far, about halfway through

src/main/java/org/broadinstitute/consent/http/mail/message/NewProgressReportCaseMessage.java
rushtong: It is bogus, see https://broadworkbench.atlassian.net/browse/DT-1666

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
otchet-broad: I thought about this as well.  For now, we're keeping the notification frequency on these consistent with the current behavior in the DAR flow.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -694,4 +715,87 @@ private void cancelElectionsForReferenceIds(List<String> referenceIds) {
...
+
+  public void sendNewDARCollectionMessage(Integer collectionId)
+      throws IOException, TemplateException {
+    DarCollection collection = darCollectionDAO.findDARCollectionByCollectionId(collectionId);
+    if (collection == null) {
+      logWarn(
+          "Sending new DAR Collection message: Could not find collection for specified collection id: "
+              + collectionId);
+      return;
+    }
```
pshapiro4broad, line 728: I don't see how `collection` can be `null` at this point. This seems like a server error if it does occur.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -694,4 +715,87 @@ private void cancelElectionsForReferenceIds(List<String> referenceIds) {
       }
     });
   }
+
+
+  public void sendNewDARCollectionMessage(Integer collectionId)
```
pshapiro4broad, line 720: Does it still make sense to call this from `DataAccessRequestResource`? Or should it be called from `DataAccessRequestService`, from `createDataAccessRequest()` and `createProgressReport()`?

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -694,4 +715,87 @@ private void cancelElectionsForReferenceIds(List<String> referenceIds) {
...
+    }
+    String researcherName = researcher == null ? "Unknown" : researcher.getDisplayName();
+    // Only do this for the DAR... dacDAO.findDacsForDatasetIds(dar.getDatasetIds())
+    Collection<Dac> dacsInDAR = dacDAO.findDacsForDatasetIds(dar.getDatasetIds());
+    // Use only the datasets from the dar
+    List<Integer> datasetIds = dar.getDatasetIds();
+    List<Dataset> datasetsInDAR =
+        datasetIds.isEmpty() ? List.of() : datasetDAO.findDatasetsByIdList(datasetIds);
+
+    Map<String, List<String>> sendList = new HashMap<>();
```
pshapiro4broad, line 746: I don't think this is correct. `sendList` will be shared across all calls to `emailService` in this loop, instead of having a different map for each call.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -694,4 +715,87 @@ private void cancelElectionsForReferenceIds(List<String> referenceIds) {
...
+    for (User user : distinctUsers) {
+      List<Dac> matchingDacsForUser = getMatchingDacs(user, dacsInDAR);
+      for (Dac dac : matchingDacsForUser) {
+        List<String> matchingDatasetsForDac = getMatchingDatasets(dac, datasetsInDAR);
+        if (matchingDatasetsForDac != null) {
+          sendList.put(dac.getName(), matchingDatasetsForDac);
+        }
+      }
```
pshapiro4broad, line 754: If you do want to generate a new map each time, I think this would work
```
      Map<String, List<String>> sendList =
          getMatchingDacs(user, dacsInDAR).stream()
              .collect(Collectors.toMap(Dac::getName, dac -> getMatchingDatasets(dac, datasetsInDAR)));
```

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -694,4 +715,87 @@ private void cancelElectionsForReferenceIds(List<String> referenceIds) {
...
+        .filter(u -> Boolean.TRUE.equals(u.getEmailPreference()))
+        .distinct()
+        .toList();
+  }
+
+  private List<Dac> getMatchingDacs(User user, Collection<Dac> dacsInDAR) {
+    List<Integer> dacIDs = user.getRoles().stream()
+        .filter(ur -> ur.getDacId() != null)
+        .map(UserRole::getDacId)
+        .toList();
```
pshapiro4broad: Slightly less code to call `getDacId` first
```suggestion
        .map(UserRole::getDacId)
        .filter(Objects::nonNull)
        .toList();
```

src/main/java/org/broadinstitute/consent/http/service/VoteService.java
```diff
@@ -322,13 +311,18 @@ public void sendDatasetApprovalNotifications(List<Vote> votes, User user) {
...
-              translation);
+          if(dar.getProgressReport()) {
+            emailService.sendResearcherProgressReportApproved(darCode, researcherId, datasetMailDTOs,
+                translation);
+          } else {
+            emailService.sendResearcherDarApproved(darCode, researcherId, datasetMailDTOs,
+                translation);
+          }
         } catch (Exception e) {
           logException("Error sending researcher dar approved email: " + e.getMessage(), e);
```
pshapiro4broad, line 318: If the common operation for failed email operations is to log, it would be better to do it in the email code instead of in each method that calls the email service.

src/main/java/org/broadinstitute/consent/http/service/VoteService.java
```diff
@@ -322,13 +311,18 @@ public void sendDatasetApprovalNotifications(List<Vote> votes, User user) {
         String translation = String.join(";", dataUseTranslations);
```
pshapiro4broad: This `String.join()` be done directly from the stream. Also, there's no need for an intermediate `toList()` step
```
        String translation = approvedDatasetsInDar.stream()
            .map(dataset -> useRestrictionConverter.translateDataUse(dataset.getDataUse(), DataUseTranslationType.DATASET))
            .distinct()
            .collect(Collectors.joining(";"));
```

src/main/java/org/broadinstitute/consent/http/service/dao/DarCollectionServiceDAO.java
pshapiro4broad: Why not call this parameter `dar`? That's what the code used before for this concept.
```suggestion
  public List<String> createElectionsForDarByUser(User user, DataAccessRequest dar)
```

src/main/java/org/broadinstitute/consent/http/service/dao/DarCollectionServiceDAO.java
```diff
@@ -73,11 +73,14 @@ public List<String> createElectionsForDarCollection(User user, DarCollection col
           //    4. Create an RP Election
           //    5. Create member votes for rp election
           //        5a. Chair Vote for chair
-          collection.getDars().values().forEach(dar -> {
-            dar.getDatasetIds().forEach(datasetId -> {
+
+          // Only take actions on the most recent DAR/Progress report in the collection
+          // This means a chair cannot reopen an election in any other DAR in the collection besides
+          // the most recently submitted progress report.
+          dataAccessRequest.getDatasetIds().forEach(datasetId -> {
```
pshapiro4broad: In general I like forms like `forEach()` but in cases like this, it doesn't seem to add much and causes the code to be indented further than a `for` loop, so I would use `for` instead here
```suggestion
          for (var datasetId: dataAccessRequest.getDatasetIds()) {
```

src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
```diff
@@ -306,17 +318,47 @@ void testCreateElectionsForDarCollection() throws Exception {
...
-    when(darCollectionServiceDAO.createElectionsForDarCollection(any(), any())).thenReturn(
+    when(darCollectionServiceDAO.createElectionsForDarByUser(any(), any())).thenReturn(
         List.of("electionId"));
     when(voteDAO.findVoteUsersByElectionReferenceIdList(any())).thenReturn(List.of(new User()));
 
     service.createElectionsForDarCollection(user, collection);
-    verify(darCollectionServiceDAO).createElectionsForDarCollection(any(), any());
+    verify(darCollectionServiceDAO, times(1)).createElectionsForDarByUser(any(), eq(dar));
+    verify(voteDAO, times(1)).findVoteUsersByElectionReferenceIdList(any());
+    verify(darCollectionDAO, times(1)).findDARCollectionByCollectionId(any());
```
pshapiro4broad: `times(1)` is redundant as this is the default for `verify()`
```suggestion
    verify(darCollectionServiceDAO).createElectionsForDarByUser(any(), eq(dar));
    verify(voteDAO).findVoteUsersByElectionReferenceIdList(any());
    verify(darCollectionDAO).findDARCollectionByCollectionId(any());
```

Also, these calls are already verified below. And, `createElectionsForDarByUser()` and `findVoteUsersByElectionReferenceIdList()` do not need to be verified at all because they are mocked, and this test uses strict stubbing.

src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
```diff
@@ -306,17 +318,47 @@ void testCreateElectionsForDarCollection() throws Exception {
...
+
+    when(darCollectionServiceDAO.createElectionsForDarByUser(any(), any())).thenReturn(
+        List.of("electionId"));
+    when(voteDAO.findVoteUsersByElectionReferenceIdList(any())).thenReturn(List.of(new User()));
+
+    service.createElectionsForDarCollection(user, collection);
+    verify(darCollectionServiceDAO, times(1)).createElectionsForDarByUser(any(), eq(progressReport));
+    verify(voteDAO, times(1)).findVoteUsersByElectionReferenceIdList(any());
+    verify(darCollectionDAO, times(1)).findDARCollectionByCollectionId(any());
+    verify(emailService, times(1)).sendProgressReportNewCollectionElectionMessage(any(), any());
```
pshapiro4broad: This also has redundant `times(1)` calls. I know that our existing tests rely heavily on `any()`, but I think it's worth moving away from that for new tests
```suggestion
    User voteUser = new User();
    String electionId = "electionId";

    when(darCollectionServiceDAO.createElectionsForDarByUser(user, progressReport))
        .thenReturn(List.of(electionId));
    when(voteDAO.findVoteUsersByElectionReferenceIdList(List.of(electionId)))
        .thenReturn(List.of(voteUser));

    service.createElectionsForDarCollection(user, collection);
    verify(darCollectionDAO).findDARCollectionByCollectionId(collection.getDarCollectionId());
    verify(emailService)
        .sendProgressReportNewCollectionElectionMessage(List.of(voteUser), collection.getDarCode());
  }
```

src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
```diff
@@ -1230,6 +1272,30 @@ private DarCollectionSummary createDarCollectionSummaryWithElections() {
...
+    collection.setDarCollectionId(1);
+    collection.setDatasets(Set.of(d1, d2));
+    DataAccessRequest dar = new DataAccessRequest();
+    dar.setReferenceId(UUID.randomUUID().toString());
+    dar.setSubmissionDate(Timestamp.from(Instant.now()));
+    dar.setDatasetIds(List.of(d1.getDatasetId(), d2.getDatasetId()));
+    collection.addDar(dar);
+
+    assertDoesNotThrow(() -> service.sendNewDARCollectionMessage(collection.getDarCollectionId()));
+  }
```
pshapiro4broad: This test provides coverage -- code is executed -- but verifies very little about the code behavior. This kind of test results in false coverage data and is worse than not having a test at all IMO. Here's a version that verifies (some) behavior
```
  void testSendNewDARCollectionMessage() throws Exception {
    Dac dac = new Dac();
    dac.setDacId(1);
    User chairperson = createUserWithRole(UserRoles.CHAIRPERSON, dac.getDacId());
    dac.setChairpersons(List.of(chairperson));
    dac.setName("DAC-01");
    User researcher = createUserWithRole(UserRoles.RESEARCHER, null);

    Dataset d1 = createDataset(dac.getDacId());
    Dataset d2 = createDataset(dac.getDacId());

    DarCollection collection = new DarCollection();
    collection.setDarCode("01");
    collection.setDarCollectionId(1);
    collection.setDatasets(Set.of(d1, d2));
    collection.setCreateUserId(researcher.getUserId());
    dar.setReferenceId(UUID.randomUUID().toString());
    dar.setSubmissionDate(Timestamp.from(Instant.now()));
    dar.setDatasetIds(List.of(d1.getDatasetId(), d2.getDatasetId()));
    collection.addDar(dar);

    when(darCollectionDAO.findDARCollectionByCollectionId(collection.getDarCollectionId()))
        .thenReturn(collection);
    when(userDAO.findUserById(researcher.getUserId())).thenReturn(researcher);
    when(userDAO.findUsersForDatasetsByRole(dar.getDatasetIds(),
        Collections.singletonList(UserRoles.CHAIRPERSON.getRoleName()))).thenReturn(Set.of(chairperson));
    when(dacDAO.findDacsForDatasetIds(dar.getDatasetIds())).thenReturn(Set.of(dac));
    when(datasetDAO.findDatasetsByIdList(dar.getDatasetIds())).thenReturn(List.of(d1, d2));
    service.sendNewDARCollectionMessage(collection.getDarCollectionId());
    verify(emailService)
        .sendNewDARRequestEmail(
            chairperson,
            Map.of(dac.getName(), List.of(d1.getDatasetIdentifier(), d2.getDatasetIdentifier())),
            researcher.getDisplayName(),
            collection.getDarCode());
  }
```

src/test/java/org/broadinstitute/consent/http/service/DataAccessRequestServiceTest.java
```diff
@@ -789,4 +803,126 @@ public boolean matches(String argument) {
...
+    collection.setDarCode("DAR-12345");
+    when(darCollectionDAO.findDARCollectionByReferenceId(any())).thenReturn(collection);
+
+    User user = new User();
+    user.setDisplayName(RandomStringUtils.randomAlphanumeric(10));
+    user.setEmail(RandomStringUtils.randomAlphanumeric(10));
+    when(userDAO.findUserById(any())).thenReturn(user);
+
+    initService();
+    assertDoesNotThrow(() -> service.sendReminderMessage(vote.getVoteId()));
```
pshapiro4broad: If these are also copied over tests, I guess that's OK, just note that these are also not really testing our code's behavior. For this test, to improve it, add to initService():
```
    config.getServicesConfiguration().setLocalURL("local_url/");
```

And then add here:
```
    verify(emailService)
        .sendReminderMessage(
            user,
            vote,
            collection.getDarCode(),
            election.getElectionType(),
            "local_url/dar_collection/" + collection.getDarCollectionId());
    verify(voteDAO).updateVoteReminderFlag(vote.getVoteId(), true);
```

src/test/java/org/broadinstitute/consent/http/service/VoteServiceTest.java
```diff
@@ -521,27 +578,20 @@ void testNotifyResearchersOfDarApproval_2Dars_1Collection() throws Exception {
...
 
     when(electionDAO.findElectionsByIds(any())).thenReturn(List.of(e1, e2));
     when(dataAccessRequestDAO.findByReferenceIds(any())).thenReturn(List.of(dar1, dar2));
-    when(darCollectionDAO.findDARCollectionByCollectionIds(any())).thenReturn(List.of(c));
     when(datasetDAO.findDatasetsByIdList(any())).thenReturn(List.of(d1, d2));
     when(userDAO.findUserById(any())).thenReturn(researcher);
 
     initService();
     service.sendDatasetApprovalNotifications(List.of(v1, v2), researcher);
     // Since we have 1 collection with different DAR/Datasets, we should be sending 1 email
```
pshapiro4broad, line 594: Comment is out of date

src/test/java/org/broadinstitute/consent/http/service/VoteServiceTest.java
```diff
@@ -460,7 +464,60 @@ void testUpdateRationaleByVoteIds_NonDataAccessElection() {
...
+
+    when(electionDAO.findElectionsByIds(any())).thenReturn(List.of(e1));
+    when(dataAccessRequestDAO.findByReferenceIds(any())).thenReturn(List.of(dar1));
+    when(datasetDAO.findDatasetsByIdList(any())).thenReturn(List.of(d1));
+    when(userDAO.findUserById(any())).thenReturn(researcher);
+
+    initService();
+    service.sendDatasetApprovalNotifications(List.of(v1), researcher);
+    // Since we have 1 collection with different DAR/Datasets, we should be sending 1 email
+    verify(emailService, times(1)).sendResearcherProgressReportApproved(any(), any(), anyList(), any());
```
pshapiro4broad:
```suggestion
    verify(emailService).sendResearcherProgressReportApproved(any(), any(), anyList(), any());
```

src/test/java/org/broadinstitute/consent/http/service/dao/DarCollectionServiceDAOTest.java
```diff
@@ -151,22 +149,18 @@ void testCreateElectionsForDarCollectionWithMultipleDatasetsForAdmin() throws Ex
...
    */
   @Test
-  void testCreateElectionsForDarCollectionChair() throws Exception {
+  void testCreateElectionsForDarByUserChair() throws Exception {
     DarCollection collection = setUpDarCollectionWithDacDataset();
-    Optional<DataAccessRequest> dar = collection.getDars().values().stream().findFirst();
-    assertTrue(dar.isPresent());
-    Integer datasetId = dar.get().getDatasetIds().get(0);
+    DataAccessRequest dar = collection.getDars().values().stream().findFirst().orElse(null);
+    assertNotNull(dar);
```
pshapiro4broad: You can skip the assert and by using `orElseThrow()`
```suggestion
    DataAccessRequest dar = collection.getDars().values().stream().findFirst().orElseThrow();
```

src/test/java/org/broadinstitute/consent/http/service/dao/DarCollectionServiceDAOTest.java
```diff
@@ -263,10 +253,60 @@ void testCreateElectionsForDarCollectionWithMultipleDatasetsForChair() throws Ex
...
 
-    // Make sure we did not create elections for the DAR/Dataset that the chair does not have access to.
-    List<Election> nonCreatedElections =
-        electionDAO.findLastElectionsByReferenceIds(List.of(dar2.getReferenceId()));
-    assertTrue(nonCreatedElections.isEmpty());
+    assertEquals(8, createdVotes.size());  // 1 dataset X 2 elections/dataset X 4 votes/election each = 8 votes.
+
+    // Find the dac chairperson for the second Dataset in the DAR.
+    // refresh the collection
+    collection = darCollectionDAO.findDARCollectionByCollectionId(collection.getDarCollectionId());
```
pshapiro4broad: `collection` is unused here

src/test/java/org/broadinstitute/consent/http/service/dao/DarCollectionServiceDAOTest.java
```diff
@@ -372,50 +408,59 @@ void testCreateElectionsForDarCollectionAfterCancelingEarlierElectionsAsChair()
...
+    dataAccessRequestDAO.insertProgressReport(
+        dar.getId(),
+        dar.getCollectionId(),
+        referenceId,
+        dar.getUserId(),
+        dar.getData());
+    DataAccessRequest progressReport = dataAccessRequestDAO.findByReferenceId(referenceId);
+    dar.getDatasetIds().forEach(datasetId -> {
+      dataAccessRequestDAO.insertDARDatasetRelation(referenceId, datasetId);
+    });
```
pshapiro4broad: No `{}` needed for a one line lambda
```suggestion
    dar.getDatasetIds().forEach(datasetId ->
      dataAccessRequestDAO.insertDARDatasetRelation(referenceId, datasetId));
```

pshapiro4broad (approved):
Lots of changes, lots of comments, but looks OK overall.

It worries me a bit to see so many unit tests that are providing "coverage" but aren't asserting code behavior. If we don't want to fix this now, we may want to remove the tests or otherwise disable them so we don't get a false sense of our test coverage.

src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
```diff
@@ -127,6 +128,12 @@ public void addDar(DataAccessRequest dar) {
     dars.putIfAbsent(dar.getReferenceId(), dar);
   }
 
+  public DataAccessRequest getMostRecentDar() {
```
rjohanek, line 131: I don't know if this is relevant/helpful but my recent PR adds a latestReferenceId field to the DarCollectionSummary https://github.com/DataBiosphere/consent/pull/2528

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -162,10 +166,10 @@ public Response createDataAccessRequestWithDAARestrictions(
       DataAccessRequest newDar = dataAccessRequestService.createDataAccessRequest(user, payload);
       Integer collectionId = newDar.getCollectionId();
       try {
-        emailService.sendNewDARCollectionMessage(collectionId);
+        darCollectionService.sendNewDARCollectionMessage(collectionId);
       } catch (Exception e) {
         // non-fatal exception
-        logException("Exception sending email for collection id: " + collectionId, e);
+        logCaughtEmailException(collectionId, e);
```
rjohanek: why not replace this try catch with a call to  this class's sendNewDarCollectionMessage

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -658,19 +673,25 @@ private DarCollection cancelDarCollectionElectionsAsChair(DarCollection collecti
...
         throw e;
       }
       try {
         List<User> voteUsers = voteDAO.findVoteUsersByElectionReferenceIdList(
             createdElectionReferenceIds);
-        emailService.sendDarNewCollectionElectionMessage(voteUsers, collection);
+        if (dar.getProgressReport()) {
+          emailService.sendProgressReportNewCollectionElectionMessage(voteUsers, collection.getDarCode());
+        } else {
+          emailService.sendDarNewCollectionElectionMessage(voteUsers, collection.getDarCode());
```
rjohanek, line 692: why can't sendNewDARCollectionMessage be used here?

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
pshapiro4broad: Is there code that uses `isReminderSent()`? I don't see any code that reads this field, only code that updates it. If it is unused, I don't see how updating it or not will affect the current behavior.

otchet-broad: There's an interface on the frontend that expects a value there, but nothing appears to read it there either.

[https://broadworkbench.atlassian.net/browse/DT-1672](https://broadworkbench.atlassian.net/browse/DT-1672)

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
otchet-broad: Correct on both the foreign key and the never fail.  It's also clear from UserService that the deleteUserByEmail method remove votes as well.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
otchet-broad: Good catch.  Thank you.  I started the refactor, I just didn't finish the refactor.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
otchet-broad: I think we should tackle this when we tackle getting rid of DarCollection.  I've made a note in the ticket.[https://broadworkbench.atlassian.net/browse/DT-1665](https://broadworkbench.atlassian.net/browse/DT-1665)

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
otchet-broad: It looks like the original code was doing this as well.  See EmailService (left side of the diff ~ 197)

```
    Map<String, List<String>> sendList = new HashMap<>();
    for (User user : distinctUsers) {
      List<Dac> matchingDacsForUser = getMatchingDacs(user, dacsInDAR);
      for (Dac dac : matchingDacsForUser) {
        List<String> matchingDatasetsForDac = getMatchingDatasets(dac, datasetsInDAR);
        if (matchingDatasetsForDac != null) {
          sendList.put(dac.getName(), matchingDatasetsForDac);
        }
      }
      sendNewDARRequestEmail(user, sendList, researcherName, collection.getDarCode());
    }
```

I'm going to make a new ticket for this because I'm not sure what the correct behavior is.  As-is keeps the current behavior.  We can define the new one in the [ticket](https://broadworkbench.atlassian.net/browse/DT-1673).

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
otchet-broad: I've added this as a possible implementation to the ticket referenced in your previous comment.

src/main/java/org/broadinstitute/consent/http/service/VoteService.java
otchet-broad: [https://broadworkbench.atlassian.net/browse/DT-1674](https://broadworkbench.atlassian.net/browse/DT-1674)

src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
otchet-broad: This was a result of a merge with your recent PR.   Bottom line is that we were both adding coverage.  Now consolidated.

src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
otchet-broad: Thanks for the code.  I've replaced the existing method with this.  FWIW, this too was a migrated test from EmailServiceTest

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
otchet-broad: It was supposed to be and now is.  :-)

Total case of **squirrel!**

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
otchet-broad: This one is about sending notices to DAC members that the need to vote (either on a progress report ~690) or on a traditional DAR (~692) **when a chair opens an election.**  The sendNewDarCollectionMessage goes to the chairperson notifying them that the can open an election.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
otchet-broad: Agree that we need to think about this more.  I'd also like to visit deep linking from emails because we seem to be inconsistent about using them.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
otchet-broad: I suspect we're going to revisit this so I'm going to leave as-is for now.

src/main/java/org/broadinstitute/consent/http/service/dao/DarCollectionServiceDAO.java
otchet-broad: With the variable renamed to `dar` I think I prefer the forEach.

rushtong (approved):
Thank you for the updates 👍🏽

PR #2524, opened 2025-05-09
DT-1616: Disallow Cancel for submitted progress reports

pshapiro4broad:
### Addresses https://broadworkbench.atlassian.net/browse/DT-1616

Progress reports share logic with data access requests, but some operations that are valid for DARs are not valid for progress reports. This removes the Cancel action for submitted progress reports.

### Summary

Prevent the Cancel operation, and don't add `Cancel` to the list of valid actions for a Data collection that's a submitted progress report.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
@@ -7,7 +7,6 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.broadinstitute.consent.http.util.gson.GsonUtil;
@@ -48,7 +47,7 @@ public class DarCollection {
   private Integer updateUserId;
 
   @JsonProperty
-  private Map<String, DataAccessRequest> dars;
+  private final Map<String, DataAccessRequest> dars = new HashMap<>();
 
   @JsonProperty
   private Set<Dataset> datasets;
@@ -121,27 +120,11 @@ public class DarCollection {
   }
 
   public Map<String, DataAccessRequest> getDars() {
-    if (Objects.isNull(dars)) {
-      return new HashMap<>();
-    }
     return dars;
   }
 
-  public void setDars(Map<String, DataAccessRequest> dars) {
-    this.dars = dars;
-  }
-
   public void addDar(DataAccessRequest dar) {
-    if (Objects.isNull(dars)) {
-      this.setDars(new HashMap<>());
-    }
-    if (Objects.nonNull(dar)) {
-      String referenceId = dar.getReferenceId();
-      DataAccessRequest savedDar = dars.get(referenceId);
-      if (Objects.isNull(savedDar)) {
-        dars.put(referenceId, dar);
-      }
-    }
+    dars.putIfAbsent(dar.getReferenceId(), dar);
   }
 
   public void addDataset(Dataset dataset) {
--- a/src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
@@ -190,7 +190,7 @@ public class DarCollectionResource extends Resource {
         actingRole = validateUserHasRoleName(user, roleName);
       }
 
-      DarCollection cancelledCollection = darCollectionService.cancelDarCollectionElections(user, collection, actingRole);
+      DarCollection cancelledCollection = darCollectionService.cancelDarCollectionByRole(user, collection, actingRole);
       ComplianceLogger.logDARCancellation(user, cancelledCollection.getDatasets().stream().toList(),
               (ContainerRequest) request, Response.Status.OK.getStatusCode());
       return Response.ok().entity(cancelledCollection).build();
--- a/src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
@@ -8,7 +8,6 @@ import jakarta.annotation.security.RolesAllowed;
 import jakarta.ws.rs.BadRequestException;
 import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.DELETE;
-import jakarta.ws.rs.ForbiddenException;
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.NotFoundException;
 import jakarta.ws.rs.POST;
@@ -20,7 +19,6 @@ import jakarta.ws.rs.QueryParam;
 import jakarta.ws.rs.core.Context;
 import jakarta.ws.rs.core.Request;
 import jakarta.ws.rs.core.Response;
-import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import org.broadinstitute.consent.http.enumeration.DarStatus;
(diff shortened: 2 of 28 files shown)


src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -560,16 +556,20 @@ public List<DarCollection> addDatasetsToCollections(List<DarCollection> collecti
...
-        .collect(Collectors.toList());
-
-    if (referenceIds.isEmpty()) {
+    if (dars.isEmpty()) {
       logWarn("DAR Collection ID: [%s] does not have any associated DAR ids".formatted(
           collection.getDarCollectionId()));
       return collection;
     }
 
+    DarCollectionSummary summary = darCollectionSummaryDAO
```
pshapiro4broad, line 565: This change doesn't affect behavior. With this ordering change, if the `dars.isEmpty()` check fails, we don't make an unnecessary DAO call.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -304,17 +300,17 @@ public List<DarCollectionSummary> getSummariesForRole(User user, UserRoles role)
         processDarCollectionSummariesForMember(summaries, userId);
         break;
       case RESEARCHER:
-        summaries = darCollectionSummaryDAO.getDarCollectionSummariesForResearcher(userId);
-        processDarCollectionSummariesForResearcher(summaries);
+        var darSummaries = darCollectionSummaryDAO.getDarCollectionSummariesForResearcher(userId);
+        processDarCollectionSummariesForResearcher(darSummaries);
         List<DataAccessRequest> drafts = dataAccessRequestDAO.findAllDraftsByUserId(userId);
-        summaries.addAll(drafts
```
pshapiro4broad, line 315: This change removes the code's reliance on modifying the result of the DAO call and creates a new collection instead.

rjohanek: okay, just wondering what is the benefit of this?

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -548,6 +544,41 @@ public List<DarCollection> addDatasetsToCollections(List<DarCollection> collecti
...
+   * @param user       The User initiating the cancel
+   * @param collection The DarCollection
+   * @param role       The role of the user, must be one of ADMIN, CHAIRPERSON, or RESEARCHER
+   * @return The DarCollection whose elections have been canceled
+   */
+  public DarCollection cancelDarCollectionElections(User user, DarCollection collection, UserRoles role) {
+    DarCollectionSummary summary = darCollectionSummaryDAO
+        .getDarCollectionSummaryByCollectionId(collection.getDarCollectionId());
+    if (summary.getProgressReport()) {
+      throw new BadRequestException("Cannot cancel a progress report");
```
rjohanek: I thought the purpose of this ticket was to make the DAR not cancellable if it is a progress report? This seems like you are making the election not cancellable if it is on a progress report? Is that what we really want?

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
pshapiro4broad: The DAO call is a JDBI library, not code we wrote. Right now it produces mutable lists, but I think it would be better to avoid making assumptions about how it will work in the future.

In general, if we have an API that returns a collection, I think it's better to assume that the collection will be immutable, so we can freely use `Stream.toList()` and not worry about downstream usage.

pshapiro4broad: But if our code is depending on mutability of the DAO API results throughout, and that's a known feature of JDBI, I can change this back and make the test code supply mutable results to match the API behavior.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
pshapiro4broad: I don't fully understand the consent operations flow so I may have gotten this wrong, let's talk about it.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -556,20 +558,22 @@ public List<DarCollection> addDatasetsToCollections(List<DarCollection> collecti
...
-    Collection<DataAccessRequest> dars = collection.getDars().values();
-    List<String> referenceIds = dars.stream()
-        .map(DataAccessRequest::getReferenceId)
-        .collect(Collectors.toList());
-
-    if (referenceIds.isEmpty()) {
-      logWarn("DAR Collection ID: [%s] does not have any associated DAR ids".formatted(
-          collection.getDarCollectionId()));
-      return collection;
+  private DarCollection cancelDarCollectionAsResearcher(DarCollection collection, User user) {
```
rjohanek, line 570: should this be named cancelDarCollectionElectionsAsResearcher

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rjohanek: I think the point of this work was to avoid getting a PR in draft state. I don't think this PR is capturing that yet. I think the draft DAR endpoint needs to be updated. I think the work you've done for elections is valid, but I don't think it was the intended goal of this ticket.

pshapiro4broad: I see what you mean, I'll update the PR.

src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
```diff
@@ -121,27 +120,11 @@ public void setUpdateUserId(Integer updateUserId) {
...
-      this.setDars(new HashMap<>());
-    }
-    if (Objects.nonNull(dar)) {
-      String referenceId = dar.getReferenceId();
-      DataAccessRequest savedDar = dars.get(referenceId);
-      if (Objects.isNull(savedDar)) {
-        dars.put(referenceId, dar);
-      }
-    }
+    dars.putIfAbsent(dar.getReferenceId(), dar);
```
otchet-broad, line 127: Why not just put?  Couldn't the value of `dar` have changed?

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rjohanek: nevermind, it isn't cancelling elections it is cancelling all the DARs in the collection

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
```diff
@@ -192,16 +190,7 @@ public Response cancelDarCollectionByCollectionId(
...
-          switch (actingRole) {
-            case ADMIN -> darCollectionService.cancelDarCollectionElectionsAsAdmin(collection);
-            case CHAIRPERSON ->
-                darCollectionService.cancelDarCollectionElectionsAsChair(collection, user);
-            default -> {
-              validateUserIsCreator(user, collection);
-              yield darCollectionService.cancelDarCollectionAsResearcher(collection);
-            }
-          };
+      DarCollection cancelledCollection = darCollectionService.cancelDarCollectionElections(user, collection, actingRole);
```
otchet-broad: Does cancelDarCollectionElections now handle the case of the RESEARCHER turning a non-progress report DAR back into a draft?  If so, the name feels off here.

otchet-broad: I see that it does.  Maybe name it: cancelDarCollectionByRole?

otchet-broad (commented):
Overall, looks good.

Just a thought on naming and a question of put vs putIfAbsent from me.  This is what I expected.

Thank you!

src/main/java/org/broadinstitute/consent/http/models/DarCollection.java
pshapiro4broad: > Why not just put? Couldn't the value of dar have changed?

The short answer is, this is what it was doing before, so this change preserves the current behavior.

Looking at its usage, it's only used while processing the result of a SQL operation, in `DarCollectionReducer`. In that case, I wouldn't expect to see it used more then once for the same dar. So `put()` _should_ be safe.

otchet-broad (approved):
Thanks!

rjohanek (approved):
this makes sense to me thank you

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
pshapiro4broad: I've updated the name and the method comment.

PR #2526, opened 2025-05-14
DT-1660: Remove LC institution and era commons values

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1660

### Summary
This is a cleanup PR to remove the redundant institution and era commons values stored as part of the library card. In all cases, these values should be derived from the user it is associated to.

There is one functional change in [LibraryCardService](https://github.com/DataBiosphere/consent/pull/2526/files#diff-b6e15bc5490b0ca477a66cc3c347f0ffc1df7f803d95c75306d52848ba5a602dR42) where I modified the institutional verification check to be skipped if the user is an administrator. This ensures that SOs can only create a LC for a user in their institution. There are new tests that cover this behavior.

There will need to be a follow up PR to remove the institution from the admin LC view.

I tried to update tests as little as possible to keep the diff as as reasonable as possible - there are a number of things that should be done to all of the test classes that would be too much for this PR. 

### Testing
This branch has been tested with a local UI. The SO view does not allow for an institution selection when creating LCs, so there are no changes needed there. The Admin view does, but still works regardless of the institution selected - it is essentially ignored when a LC is created. 

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
```diff
@@ -37,18 +37,14 @@ public LibraryCardService(LibraryCardDAO libraryCardDAO, InstitutionDAO institut
...
-    //If user is not an admin, use user's institutionId rather than the value provided in the payload
-    if (!isAdmin && !libraryCard.getInstitutionId().equals(user.getInstitutionId())) {
-      throw new BadRequestException("Card payload not valid");
-    }
     checkIfCardExists(libraryCard);
     processUserOnNewLC(libraryCard);
-    checkForValidInstitution(libraryCard.getInstitutionId(), libraryCard.getUserEmail());
+    if (!isAdmin) {
+      checkForValidInstitution(user.getInstitutionId(), libraryCard.getUserEmail());
+    }
```
rushtong, line 44: This is the one functional change in this PR. This ensures that SOs have to have the same institution as the user for whom the LC is being created.

src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
```diff
@@ -5,46 +5,40 @@
...
+  @SqlUpdate("""
+      UPDATE library_card SET
+            id = :libraryCardId, 
+            user_id = :userId, 
+            user_name = :userName, 
+            user_email = :userEmail, 
+            update_user_id = :updateUserId, 
+            update_date = :updateDate 
+            WHERE id = :libraryCardId
+      """)
```
rushtong, line 39: Change required to make semgrep happy.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -115,19 +115,18 @@ public Response createLibraryCardDaaRelation(
...
-          .filter(card -> card.getInstitutionId() == authedUser.getInstitutionId().intValue())
-          .findFirst();
-      if (matchingCard.isEmpty()) {
-        LibraryCard createdLc = libraryCardService.createLibraryCardForSigningOfficial(user, authedUser);
-        matchingCard = Optional.of(createdLc);
+      LibraryCard workingLibraryCard;
+      if (libraryCards.isEmpty()) {
+        workingLibraryCard = libraryCardService.createLibraryCardForSigningOfficial(user, authedUser);
+      } else {
+        workingLibraryCard = libraryCards.get(0);
```
rushtong, line 122: A future PR will restrict users to having a single library card. In the meantime, this should preserve the existing behavior.

src/test/java/org/broadinstitute/consent/http/service/LibraryCardServiceTest.java
```diff
@@ -53,32 +54,77 @@ void initService() {
...
     assertEquals(createdCard, service.createLibraryCard(payload, adminUser));
-    verify(libraryCardDAO).insertLibraryCard(eq(user.getUserId()), eq(institution.getId()),
-        eq(payload.getEraCommonsId()), eq(payload.getUserName()), eq(user.getEmail()),
+    verify(libraryCardDAO).insertLibraryCard(eq(user.getUserId()),
+        eq(payload.getUserName()), eq(user.getEmail()),
         eq(payload.getCreateUserId()), any());
   }
 
+  @Test
+  // Test SO LC create with userId and email success case
```
rushtong, line 81: New test cases here to ensure the new functionality is exercised. just prior to this, there is an existing test case for Admins.

src/main/java/org/broadinstitute/consent/http/models/LibraryCard.java
```diff
@@ -56,22 +48,6 @@ public void setUserId(Integer userId) {
     this.userId = userId;
   }
 
-  public Integer getInstitutionId() {
```
pshapiro4broad, line 59: Are there any UI changes needed to remove references to these properties?

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
pshapiro4broad: Since we're using JDK 21 we can use `getFirst()` and `getLast()` which work on ordered collections.
```suggestion
        workingLibraryCard = libraryCards.getFirst();
```

Similar to `size() == 0` vs `isEmpty()` I think `getFirst()` is easier to read than `get(0)`.

src/main/java/org/broadinstitute/consent/http/service/LibraryCardService.java
pshapiro4broad: I don't think this is a functional change. Before the code validated that the SO and LC had the same institution, then validated that the LC and the user had the same institution. So it was also verifying that the SO and user had the same institution.

src/main/resources/assets/schemas/LibraryCard.yaml
```diff
@@ -6,14 +6,6 @@ properties:
   userId:
     type: integer
     description: ID of associated user
-  institution:
```
pshapiro4broad, line 9: Are there corresponding UI model changes needed?

pshapiro4broad (approved):
Looks good

src/main/java/org/broadinstitute/consent/http/models/LibraryCard.java
rushtong: Yes, there will need to be UI changes. I'll follow up with a ticket on that but the UI functions even with the mis-matched model.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
rushtong: This doesn't compile for me locally even though it is JDK 21 - not sure why that is. I'd like to tackle that separately.

src/main/resources/assets/schemas/LibraryCard.yaml
rushtong: Yes, will follow up with a new ticket

fboulnois (approved):
cleanup looks good 👍

PR #2527, opened 2025-05-15
[DT-1624]Consent: Update data use acknowledgement on progress report submit

rjohanek:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1624
We are going to add a section to include Data Use Acknowledgements in the Progress Report Application, so we need to save these new values on the backend.

### Summary
Set the Data Use Acknowledgement fields to the new value from the PR instead of copying from the parent.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
@@ -18,6 +18,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.UUID;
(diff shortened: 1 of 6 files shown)


rushtong (approved):
👍🏽

pshapiro4broad (approved):
This looks OK although I wonder why methods like `populateProgressReportFromJsonString()` aren't on `DataAccessRequest` instead of `DataAccessRequestResource`. They are pure data operations that don't need any injected services.

For example, `populateProgressReportFromJsonString(String, DataAccessRequest)` could be a static method `DataAccessRequest.fromJsonWithParent()`.

PR #2528, opened 2025-05-15
[DT-1663]Update CreateProgressReport Action to only be included if the Latest DAR in the collection has an approved dataset

rjohanek:
### Addresses

The previous code included the action if any DAR in the collection had approved datasets, now that we are assuming chaining of PRs (and not branching) this needs to be changed to only include the action if the latest DAR in the collection has approved datasets. The work that this modifies was originally done here: https://github.com/DataBiosphere/consent/pull/2520

### Summary

Adds a new field on the summary that stores the latest DAR referenceId. This field was already being selected in the query, so only an update to the reducer was required to set the field.
Update DAO test to check for latest reference Id
In the collectionService, instead of checking for approvals of all DARs in the collection, only check for approvals for the latest DAR in the collection.
TO DO: update collectionService test

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
@@ -126,10 +126,10 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
           final_access_vote.reference_id = dar.reference_id AND
           final_access_vote.dataset_id = dd.dataset_id
       WHERE final_access_vote.last_vote = TRUE
-        AND dar.reference_id IN (<darReferenceIds>)
+        AND dar.reference_id = :darReferenceId
         AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)
       """)
-  Set<Integer> findDatasetApprovalsByDars(@BindList("darReferenceIds") List<String> darReferenceIds);
+  Set<Integer> findDatasetApprovalsByDar(@Bind("darReferenceId") String darReferenceIds);
 
   /**
    * This query finds submitted DARs based on a date range.  This would be useful if we wanted to
--- a/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
@@ -102,7 +102,7 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
    * `FALSE` in the example above. Outside the JOIN, we filter on groupings where the final vote
    * value is `TRUE` so the denied election in the example would be filtered out.
    *
-   * @param darReferenceIds The DARs reference UUIDs
+   * @param darReferenceId The DARs reference UUID
    * @return Set of approved Dataset Ids for the list of DARs
    */
   @SqlQuery(
@@ -129,7 +129,7 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
         AND dar.reference_id = :darReferenceId
         AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)
       """)
-  Set<Integer> findDatasetApprovalsByDar(@Bind("darReferenceId") String darReferenceIds);
+  Set<Integer> findDatasetApprovalsByDar(@Bind("darReferenceId") String darReferenceId);
 
   /**
    * This query finds submitted DARs based on a date range.  This would be useful if we wanted to
--- a/src/main/java/org/broadinstitute/consent/http/db/mapper/DarCollectionSummaryReducer.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/mapper/DarCollectionSummaryReducer.java
@@ -39,11 +39,11 @@ public class DarCollectionSummaryReducer implements
       }
 
       try {
-        darReferenceId = rowView.getColumn("dar_reference_id", String.class);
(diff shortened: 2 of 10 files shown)


snf2ye (approved):

fboulnois (approved):
👍

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -151,8 +151,9 @@ private void processDarCollectionSummariesForResearcher(List<DarCollectionSummar
       int electionCount = elections.size();
       elections.values().forEach(election -> updateStatusCount(statusCount, election.getStatus()));
       s.addAction(DarCollectionActions.REVIEW);
-      //if any DARs in the collection have approved datasets, include create progress report action
-      Set<Integer> datasetIds = dataAccessRequestDAO.findDatasetApprovalsByDars(List.copyOf(s.getReferenceIds()));
+      //if the latest DAR in the collection has at least one approved dataset,
+      //include the create progress report action
+      Set<Integer> datasetIds = dataAccessRequestDAO.findDatasetApprovalsByDars(List.of(s.getLatestReferenceId()));
```
pshapiro4broad: With this change, I think that both callers to this function only pass a single reference ID. The other caller is in `DataAccessReqeustService.createProgressReport()`. If this is correct, the DAO method should be updated so it no longer accepts a `List`.

src/main/java/org/broadinstitute/consent/http/db/mapper/DarCollectionSummaryReducer.java
```diff
@@ -40,6 +40,9 @@ public void accumulate(Map<Integer, DarCollectionSummary> map, RowView rowView)
 
       try {
         darReferenceId = rowView.getColumn("dar_reference_id", String.class);
+        if (Objects.nonNull(darReferenceId)) {
+          summary.setLatestReferenceId(darReferenceId);
```
pshapiro4broad, line 44: Is this correct? This code will result in the last DAR with a reference ID in the query response being stored in the `latestReferenceId` field. But the queries that use this reducer are using say `ORDER BY collection_id, submission_date DESC`. Doesn't that result in the most recent DAR being sorted first, so that this will return the oldest reference ID?

pshapiro4broad (commented):
I don't think I understand how the sort order of the query affects the reducer's handling of `darReferenceId`.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rjohanek: yeah that's true

src/main/java/org/broadinstitute/consent/http/db/mapper/DarCollectionSummaryReducer.java
rjohanek: I do not think the sort order is relevant here. the reference Id is only being selected from the join on the dar table that only selects the latest dar (aliased dar in the query).

rjohanek: maybe the aliasing could be clearer here

rushtong: This does look confusing to me. We're selecting something with a name that is different from what it is named on the constructed object. Can we specifically select for `latest_dar_reference_id` in the query to make this clear?

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -151,8 +151,9 @@ private void processDarCollectionSummariesForResearcher(List<DarCollectionSummar
       int electionCount = elections.size();
       elections.values().forEach(election -> updateStatusCount(statusCount, election.getStatus()));
       s.addAction(DarCollectionActions.REVIEW);
-      //if any DARs in the collection have approved datasets, include create progress report action
-      Set<Integer> datasetIds = dataAccessRequestDAO.findDatasetApprovalsByDars(List.copyOf(s.getReferenceIds()));
+      //if the latest DAR in the collection has at least one approved dataset,
+      //include the create progress report action
+      Set<Integer> datasetIds = dataAccessRequestDAO.findDatasetApprovalsByDar(s.getLatestReferenceId());
```
rushtong, line 171: I see that this DAO call is already in prod code, but having this in a loop will definitely impact performance. We should investigate a way to collect this information before we hit the loop.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -126,10 +126,10 @@ AND LOWER(v.type) = 'final') final_access_vote ON
           final_access_vote.reference_id = dar.reference_id AND
           final_access_vote.dataset_id = dd.dataset_id
       WHERE final_access_vote.last_vote = TRUE
-        AND dar.reference_id IN (<darReferenceIds>)
+        AND dar.reference_id = :darReferenceId
         AND (LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)
       """)
-  Set<Integer> findDatasetApprovalsByDars(@BindList("darReferenceIds") List<String> darReferenceIds);
+  Set<Integer> findDatasetApprovalsByDar(@Bind("darReferenceId") String darReferenceIds);
```
pshapiro4broad: typo
```suggestion
  Set<Integer> findDatasetApprovalsByDar(@Bind("darReferenceId") String darReferenceId);
```

rushtong (approved):
Pending existing PR feedback, 👍🏽

PR #2529, opened 2025-05-15
[DT-1606] Enforce institutional parity

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-1606](https://broadworkbench.atlassian.net/browse/DT-1606)

### Summary

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
@@ -6,10 +6,12 @@ import jakarta.ws.rs.BadRequestException;
 import jakarta.ws.rs.NotAcceptableException;
 import jakarta.ws.rs.NotFoundException;
 import java.sql.SQLException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Date;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import org.apache.commons.validator.routines.EmailValidator;
@@ -30,6 +32,7 @@ import org.broadinstitute.consent.http.models.DataAccessRequest;
 import org.broadinstitute.consent.http.models.DataAccessRequestData;
 import org.broadinstitute.consent.http.models.Dataset;
 import org.broadinstitute.consent.http.models.Election;
+import org.broadinstitute.consent.http.models.Institution;
 import org.broadinstitute.consent.http.models.LibraryCard;
 import org.broadinstitute.consent.http.models.User;
 import org.broadinstitute.consent.http.service.dao.DataAccessRequestServiceDAO;
@@ -44,6 +47,7 @@ public class DataAccessRequestService implements ConsentLogger {
   private final DataAccessRequestDAO dataAccessRequestDAO;
   private final DarCollectionDAO darCollectionDAO;
   private final ElectionDAO electionDAO;
+  private final InstitutionService institutionService;
   private final MatchDAO matchDAO;
   private final VoteDAO voteDAO;
   private final UserDAO userDAO;
@@ -54,7 +58,8 @@ public class DataAccessRequestService implements ConsentLogger {
 
   @Inject
   public DataAccessRequestService(CounterService counterService, DAOContainer container,
-      DacService dacService, DataAccessRequestServiceDAO dataAccessRequestServiceDAO, UserService userService) {
+      DacService dacService, DataAccessRequestServiceDAO dataAccessRequestServiceDAO,
+      UserService userService, InstitutionService institutionService ) {
     this.counterService = counterService;
     this.dataAccessRequestDAO = container.getDataAccessRequestDAO();
     this.darCollectionDAO = container.getDarCollectionDAO();
@@ -65,6 +70,7 @@ public class DataAccessRequestService implements ConsentLogger {
     this.dacService = dacService;
     this.dataAccessRequestServiceDAO = dataAccessRequestServiceDAO;
     this.userService = userService;
+    this.institutionService = institutionService;
   }
 
   public List<DataAccessRequest> findAllDraftDataAccessRequests() {
@@ -281,6 +287,7 @@ public class DataAccessRequestService implements ConsentLogger {
 
     validateInternalCollaborators(dar, user);
     validateNoKeyPersonnelDuplicates(dar.getData());
+    validatePersonnelInSameInstitution(user, dar.getData());
   }
 
   @VisibleForTesting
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -279,25 +285,20 @@ public void validateDar(User user, DataAccessRequest dar) {
...
-    validateInternalCollaborators(dar, user);
+    validateInternalCollaborators(dar);
     validateNoKeyPersonnelDuplicates(dar.getData());
+    validatePersonnelInSameInstitution(user, dar.getData());
   }
 
   @VisibleForTesting
-  public void validateInternalCollaborators(DataAccessRequest payload, User requestingUser) {
-    Integer institution = requestingUser.getInstitutionId();
+  public void validateInternalCollaborators(DataAccessRequest payload) {
```
pshapiro4broad, line 311: Visible for testing methods should use `protected` not `public`, or default, as the test should be in the same package as the code under test.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -359,6 +360,92 @@ public void validateNoKeyPersonnelDuplicates(DataAccessRequestData darData) {
...
+    List<String> collaboratorsEmails =
+        darData.getInternalCollaborators().stream().map(Collaborator::getEmail).toList();
+    List<String> labStaffEmails =
+        darData.getLabCollaborators().stream().map(Collaborator::getEmail).toList();
+
+    List<String> invalidMembers = new ArrayList<>();
+
+    if (emailDoesNotMatchInstitution(submitterInstitution, piEmail)) {
+      invalidMembers.add("Principal Investigator: " + piEmail);
+    }
```
pshapiro4broad: To avoid code duplication, I would make this method accept the type of item being checked and the list of errors to append to
```suggestion
    verifyEmailMatchesInstitution(submitterInstitution, piEmail, "Principal Investigator", invalidMembers);
```

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -359,6 +360,92 @@ public void validateNoKeyPersonnelDuplicates(DataAccessRequestData darData) {
...
+    if (emailDoesNotMatchInstitution(submitterInstitution, itEmail)) {
+      invalidMembers.add("IT Director: " + itEmail);
+    }
+
+    getErrorSummary(
+            collaboratorsEmails,
+            submitterInstitution,
+            "Internal Collaborator member: ",
+            "Internal Collaborator members: ")
+        .ifPresent(invalidMembers::add);
```
pshapiro4broad: To me, this doesn't seem like a good use of `Optional`. The method is `private`, so I don't think there's a problem with passing in an array to be mutated, and every caller to the method is doing the same thing with the result.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -359,6 +360,92 @@ public void validateNoKeyPersonnelDuplicates(DataAccessRequestData darData) {
...
+    msg.append(String.join(", ", errors));
+    return msg.toString();
+  }
+
+  private boolean emailDoesNotMatchInstitution(Institution institution, String email) {
+    Institution foundInstitution = institutionService.findInstitutionForEmail(email);
+    if (foundInstitution == null || institution == null) {
+      return true;
+    }
+    return !institution.getId().equals(foundInstitution.getId());
```
pshapiro4broad: This is the same as comparing `Institution`
```suggestion
    return !institution.equals(foundInstitution);
```

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -359,6 +360,92 @@ public void validateNoKeyPersonnelDuplicates(DataAccessRequestData darData) {
...
+          "All listed personnel must share the same institutional affiliation.  The following list of roles and members must have email addresses associated with your institution: "
+              + String.join(", ", invalidMembers));
+    }
+  }
+
+  private Optional<String> getErrorSummary(
+      List<String> emails,
+      Institution institution,
+      String categorySingular,
+      String categoryPlural) {
```
pshapiro4broad: Personally, I wouldn't bother with singular vs plural messaging, that seems not worth the extra effort to handle the case where it happens to be singular. Although if you wanted to combine all the checks into a single method, and we always add "s" for a plural, maybe that would be worth it.

You can have one validation method:
```
  private void validateInstitutionEmails(
      List<String> emails, Institution institution, String category, List<String> errors) {
    List<String> invalid = emails.stream()
        .filter(email -> emailDoesNotMatchInstitution(institution, email))
        .toList();
    if (!invalid.isEmpty()) {
      String prefix = category;
      if (invalid.size() != 1) {
        prefix += "s";
      }
      errors.add(prefix + ": " + String.join(", ", errors));
    }
  }
```

And then use it for both the list and non-list cases:
```
    validateInstitutionEmails(List.of(itEmail), submitterInstitution, "IT Director", invalidMembers);
    validateInstitutionEmails(
        collaboratorsEmails, submitterInstitution, "Internal Collaborator member", invalidMembers);
```

Or, if you don't want to handle plurals
```
  private void validateInstitutionEmails(
      List<String> emails, Institution institution, String category, List<String> errors) {
    String invalid = emails.stream()
        .filter(email -> emailDoesNotMatchInstitution(institution, email))
        .collect(Collectors.joining(","));
    if (!invalid.isEmpty()) {
      errors.add(category + ": " + invalid);
    }
  }
```

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -359,6 +360,92 @@ public void validateNoKeyPersonnelDuplicates(DataAccessRequestData darData) {
...
+  private List<String> findEmailAddressesNotInInstitution(
+      List<String> emailAddresses, Institution institution) {
+    ArrayList<String> errors = new ArrayList<>();
+    emailAddresses.forEach(
+        collaborator -> {
+          if (emailDoesNotMatchInstitution(institution, collaborator)) {
+            errors.add(collaborator);
+          }
+        });
+    return errors;
```
pshapiro4broad, line 440: Since you're using a `List<String>` to produce another `List<String>`, using a streaming operation is usually a better way to express this.

src/test/java/org/broadinstitute/consent/http/service/DataAccessRequestServiceTest.java
```diff
@@ -119,17 +121,16 @@ private void initService() {
...
   @Test
   void testCreateDataAccessRequest_Update() {
     DataAccessRequest dar = generateDataAccessRequest();
     dar.setCollectionId(null);
     dar.addDatasetIds(List.of(1, 2, 3));
-    User user = new User(1, "email@test.org", "Display Name", new Date());
-    user.setLibraryCards(List.of(new LibraryCard()));
-    user.setEraCommonsId("eraCommonsId");
+    User user = createUserWithPrerequisites();
+    when(institutionService.findInstitutionForEmail(any())).thenReturn(user.getInstitution());
```
pshapiro4broad, line 151: Maybe this should be done in `generateDataAccessRequest()`? It looks like every caller to that method will need this mock, and this test user can be added as a field.

src/test/java/org/broadinstitute/consent/http/service/DataAccessRequestServiceTest.java
```diff
@@ -782,11 +750,199 @@ void testValidateNoKeyPersonnelDuplicatesSO() {
...
+    labStaffMember2.setEmail(badEmailAddress);
+    data.setLabCollaborators(List.of(labStaffMember, labStaffMember2));
+    when(institutionService.findInstitutionForEmail(goodEmailAddress)).thenReturn(goodInstitution);
+    when(institutionService.findInstitutionForEmail(badEmailAddress)).thenReturn(null);
+    initService();
+    assertThrows(IllegalArgumentException.class, () -> service.validatePersonnelInSameInstitution(user, data));
+  }
+
+  @Test
+  void testValidatePersonnelInSameInstitution_NoCollaborators() {
```
pshapiro4broad, line 963: This code is testing each case separately but since the.code tests all cases at the same time, you could combine and write a test that checks for all invalid cases at once. Or, have one setup that creates the data will all the different cases, and have separate tests that assert each case in the error message.

src/test/java/org/broadinstitute/consent/http/service/DataAccessRequestServiceTest.java
```diff
@@ -782,11 +750,199 @@ void testValidateNoKeyPersonnelDuplicatesSO() {
...
+    data.setInternalCollaborators(List.of(collaborator));
+    Collaborator labStaffMember = new Collaborator();
+    labStaffMember.setEmail(badEmailAddress);
+    data.setLabCollaborators(List.of(labStaffMember));
+
+    Institution badInstitution = new Institution();
+    badInstitution.setId(2);
+    when(institutionService.findInstitutionForEmail(badEmailAddress)).thenReturn(badInstitution);
+    initService();
+    assertThrows(IllegalArgumentException.class, () -> service.validatePersonnelInSameInstitution(user, data));
```
pshapiro4broad: The code is very careful to format an error message with all the users and the field names but that behavior is not asserted here. `assertThrows()` returns the thrown exception, so this should also assert that the error message is as expected.

pshapiro4broad (approved):
I think there's a more succinct way to write the validation, but the code does the validation as described.

The tests aren't validating the error message contents, and could be written in a shorter way to avoid code duplication as all test cases can be done at once.

rjohanek (approved):
looks good to me!

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
otchet-broad: DT-1564 corrects setting @VisibleForTesting methods to protected across the project.  I'm updating the method I added to also correct this.

fboulnois (approved):
looks good 👍

PR #2530, opened 2025-05-16
DT-1627: Prevent the creation of "sibling" progress reports

pshapiro4broad:
https://broadworkbench.atlassian.net/browse/DT-1627

### Addresses

Until we add support for "branching" progress reports, we disallow creating more than one progress report for a single data access request.

### Summary

To prevent multiple progress reports for a single data access request, make the parent ID for a data access request a unique key. Trying to insert another progress report results in a JDBI exception, which is caught and converted into a 400 client error (HTTP BAD_REQUEST).

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
@@ -34,6 +34,7 @@ import org.broadinstitute.consent.http.models.LibraryCard;
 import org.broadinstitute.consent.http.models.User;
 import org.broadinstitute.consent.http.service.dao.DataAccessRequestServiceDAO;
 import org.broadinstitute.consent.http.util.ConsentLogger;
+import org.jdbi.v3.core.JdbiException;
 import org.jdbi.v3.core.statement.UnableToExecuteStatementException;
 
 public class DataAccessRequestService implements ConsentLogger {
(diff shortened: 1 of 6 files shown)


src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -235,12 +236,17 @@ public DataAccessRequest createProgressReport(User user, DataAccessRequest progr
...
+    try {
+      dataAccessRequestDAO.insertProgressReport(
           Integer.valueOf(progressReport.getParentId()),
           progressReport.getCollectionId(),
           referenceId,
           user.getUserId(),
           progressReport.getData());
+    } catch (JdbiException e) {
+      throw new BadRequestException(
+          "Unable to create progress report for Data Access Request " + parentDar.getReferenceId());
```
pshapiro4broad, line 271: I thought about adding `e` to this exception, but it includes the complete text of the database operation, which I don't think we want to expose in our API. The failure is non-specific but our UI should prevent this from occurring. We can add better error reporting in the future it it's an issue.

fboulnois (approved):
Looks good 👍

rushtong (approved):
👍🏽

src/test/java/org/broadinstitute/consent/http/service/DataAccessRequestServiceTest.java
```diff
@@ -255,7 +251,30 @@ void createProgressReportFailsIfNonApprovedDatasets() {
...
+    progressReport.setParentId(parentDar.getId().toString());
+    progressReport.setCollectionId(parentDar.getCollectionId());
+    parentDar.setSubmissionDate(Timestamp.from(Instant.now()));
+    User user = new User(1, "email@test.org", "Display Name", new Date());
+    user.setLibraryCards(List.of(new LibraryCard()));
+    user.setEraCommonsId("eraCommonsId");
+    parentDar.setUserId(user.getUserId());
+    when(dataAccessRequestDAO.findDatasetApprovalsByDars(List.of(parentDar.getReferenceId()))).thenReturn(
+        Set.copyOf(progressReport.getDatasetIds()));
+    doThrow(new UnableToExecuteStatementException("Test exception"))
```
s-rubenstein, line 270: When this is returned to the user, what its it coerced to?

s-rubenstein (approved):

PR #2532, opened 2025-05-19
DT-1680: Fix deletion of LCs with no active DUOS user.

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1680

### Summary
Library Cards can be created for users that do not exist in the system yet, but the delete endpoint was explicitly expecting a user to exist. This bug was introduced with [this PR](https://github.com/DataBiosphere/consent/pull/2526)

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/LibraryCardResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/LibraryCardResource.java
@@ -127,7 +127,7 @@ public class LibraryCardResource extends Resource {
     }
     try {
       // If user is not an admin and SO institutionID doesn't match the user's throw an exception
-      if (lcUser != null && !checkIsAdmin(user) && !lcUser.getInstitutionId().equals(user.getInstitutionId())) {
+      if (lcUser != null && !checkIsAdmin(user) && !lcUser.getInstitution().equals(user.getInstitution())) {
         throw new ForbiddenException("You are not authorized to delete this library card");
       }
       libraryCardService.deleteLibraryCardById(id);
--- a/src/main/java/org/broadinstitute/consent/http/resources/LibraryCardResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/LibraryCardResource.java
@@ -8,6 +8,7 @@ import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.DELETE;
 import jakarta.ws.rs.ForbiddenException;
 import jakarta.ws.rs.GET;
+import jakarta.ws.rs.NotFoundException;
 import jakarta.ws.rs.POST;
 import jakarta.ws.rs.PUT;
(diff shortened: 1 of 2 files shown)


src/test/java/org/broadinstitute/consent/http/resources/LibraryCardResourceTest.java
```diff
@@ -256,6 +256,19 @@ void deleteLibraryCardThrowsNotFoundException() {
...
+  void deleteLibraryCardUserNotFound() {
+    LibraryCard card = mockLibraryCardSetup();
+    card.setUserId(null);
+    when(userService.findUserByEmail(user.getEmail())).thenReturn(user);
+    when(userService.findUserById(null)).thenThrow(new NotFoundException());
+    when(libraryCardService.findLibraryCardById(anyInt())).thenReturn(card);
+    initResource();
+
+    Response response = resource.deleteLibraryCard(authUser, 1);
+    assertEquals(HttpStatusCodes.STATUS_CODE_NO_CONTENT, response.getStatus());
```
rjohanek: this looks good, do we want to also verify that libraryCardService.deleteLibraryCardById(id) was called?

rjohanek (approved):

src/test/java/org/broadinstitute/consent/http/resources/LibraryCardResourceTest.java
rushtong: Good idea, will add that

src/main/java/org/broadinstitute/consent/http/resources/LibraryCardResource.java
```diff
@@ -118,10 +119,15 @@ public Response updateLibraryCard(@Auth AuthUser authUser, @PathParam("id") Inte
...
+    User lcUser = null;
+    try {
+      lcUser = userService.findUserById(card.getUserId());
+    } catch (NotFoundException nfe) {
+      // LC User can be null - do not need to error here
+    }
     try {
       // If user is not an admin and SO institutionID doesn't match the user's throw an exception
-      if (!checkIsAdmin(user) && !lcUser.getInstitutionId().equals(user.getInstitutionId())) {
+      if (lcUser != null && !checkIsAdmin(user) && !lcUser.getInstitutionId().equals(user.getInstitutionId())) {
```
otchet-broad: In other pull requests, we've started to use the Institution object's built-in equals comparison so that we can hold all of the business logic there.  It largely does the same thing as you're doing here right now.

```suggestion
      if (lcUser != null && !checkIsAdmin(user) && !lcUser.getInstitution().equals(user.getInstitution())) {
```

otchet-broad (approved):
Good catch!

pshapiro4broad (approved):

fboulnois (approved):
👍

PR #2535, opened 2025-05-21
DT-1607 Implement truth table as check when we resolve user in request.

otchet-broad:
### Addresses
[https://broadworkbench.atlassian.net/browse/DT-1607](https://broadworkbench.atlassian.net/browse/DT-1607)
### Summary

At the point where we extract an email address from an Oauth claim, perform validation logic to adjust the user object based on a truth table of rules defined in the ticket.

The general flavor of the rules is that when the institution derived from the email address of the requestor doesn't match our expectations for library cards or the values stored in our database,  fix our database to align with these rules.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
+++ b/src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
@@ -49,8 +49,7 @@ public class AuthorizationHelper implements ConsentLogger {
       logWarn(String.format(
           "Reading oauth2 claim headers: email is null, auth user is incomplete. Aud: %s Name: %s",
           aud, name));
-    }
-    if (email != null) {
+    } else {
       User user = userService.findUserByEmail(email);
       if (user != null) {
         userService.enforceInstitutionAndLibraryCardTruthTable(user);
--- a/src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
+++ b/src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
@@ -50,9 +50,11 @@ public class AuthorizationHelper implements ConsentLogger {
           "Reading oauth2 claim headers: email is null, auth user is incomplete. Aud: %s Name: %s",
           aud, name));
     } else {
-      User user = userService.findUserByEmail(email);
-      if (user != null) {
+      try {
+        User user = userService.findUserByEmail(email);
         userService.enforceInstitutionAndLibraryCardTruthTable(user);
+      } catch (NotFoundException nfe) {
+        // nothing to do.  new user.
       }
     }
     return new AuthUser(token, email, name, aud);
--- a/src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
+++ b/src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
@@ -99,7 +99,6 @@ public class AuthorizationHelper implements ConsentLogger {
     boolean authorize = false;
     try {
       User user = userService.findUserByEmail(authUser.getEmail());
-      user = userService.enforceInstitutionAndLibraryCardTruthTable(user);
       return user.getRoles().stream().anyMatch(r -> r.getName().equalsIgnoreCase(role));
     } catch (NotFoundException e) {
       logWarn("User not found, authorization incomplete: %s".formatted(authUser.getEmail()));
--- a/src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
+++ b/src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
@@ -50,6 +50,12 @@ public class AuthorizationHelper implements ConsentLogger {
           "Reading oauth2 claim headers: email is null, auth user is incomplete. Aud: %s Name: %s",
           aud, name));
     }
+    if (email != null) {
+      User user = userService.findUserByEmail(email);
+      if (user != null) {
+        userService.enforceInstitutionAndLibraryCardTruthTable(user);
+      }
+    }
     return new AuthUser(token, email, name, aud);
   }
 
@@ -94,6 +100,7 @@ public class AuthorizationHelper implements ConsentLogger {
     boolean authorize = false;
     try {
       User user = userService.findUserByEmail(authUser.getEmail());
+      user = userService.enforceInstitutionAndLibraryCardTruthTable(user);
       return user.getRoles().stream().anyMatch(r -> r.getName().equalsIgnoreCase(role));
     } catch (NotFoundException e) {
       logWarn("User not found, authorization incomplete: %s".formatted(authUser.getEmail()));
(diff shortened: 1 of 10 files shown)


src/main/java/org/broadinstitute/consent/http/service/DaaService.java
```diff
@@ -166,8 +166,8 @@ public void sendDaaRequestEmails(User user, Integer daaId) throws Exception {
         DataAccessAgreement daa = findById(daaId);
         String daaName = daa.getFile().getFileName();
         User toUser = new User();
-        toUser.setEmail(signingOfficial.email);
-        toUser.setDisplayName(signingOfficial.displayName);
+        toUser.setEmail(signingOfficial.getEmail());
```
otchet-broad, line 169: The `SimplifiedUser` changes in this PR are just some cleanup I did along the way.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -122,15 +123,15 @@ public User updateUserFieldsById(UserUpdateFields userUpdateFields, Integer user
       // Handle Roles
       if (Objects.nonNull(userUpdateFields.getUserRoleIds())) {
         List<Integer> currentRoleIds = userRoleDAO.findRolesByUserId(userId).stream()
-            .map(UserRole::getRoleId).collect(Collectors.toList());
+            .map(UserRole::getRoleId).toList();
```
otchet-broad, line 125: The toList changes make the value returned immutable so that we preserve the database being the source of truth.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -454,12 +454,127 @@ public void hasValidActiveERACredentials(User user) {
...
+  public User enforceInstitutionAndLibraryCardTruthTable(User user) {
+    Institution institutionFromEmail = institutionService.findInstitutionForEmail(user.getEmail());
+    Institution institutionFromDatabase = null;
+    try {
+      institutionFromDatabase = institutionService.findInstitutionById(user.getInstitutionId());
+    } catch (NotFoundException nfe) {
+      // do nothing.
+    }
+
+    boolean hasInstitutionMatchingEmailDomain =
```
fboulnois, line 474: A link to the JIRA truth table would be helpful to evaluate the cases.

fboulnois (approved):
I verified that the code matches the truth table, however I did not fully validate that the truth table logic was correct.

src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
```diff
@@ -49,6 +49,11 @@ protected AuthUser buildAuthUserFromHeaders(Map<String, String> headers) {
       logWarn(String.format(
           "Reading oauth2 claim headers: email is null, auth user is incomplete. Aud: %s Name: %s",
           aud, name));
+    } else {
+      User user = userService.findUserByEmail(email);
```
rushtong: This will throw an NFE which will be problematic for authenticated identities that do not yet exist in the system

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -454,12 +454,127 @@ public void hasValidActiveERACredentials(User user) {
...
   public static class SimplifiedUser {
 
-    public Integer userId;
-    public String displayName;
-    public String email;
-    public Integer institutionId;
+    private Integer userId;
+    private String displayName;
+    private String email;
+    private Integer institutionId;
```
rushtong, line 586: Looking back on this class, it really should be a record.

src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
rushtong: More thoughts on this code path. Another usage of this method is in `DuosUserAuthenticator` where we also make the same call to `userService.findUserByEmail(email)` so we'd be making the call twice for `@Auth DuosUser` usages. When resources use `@Auth AuthUser`, we usually follow up with a call to `userService.findUserByEmail()`. I wonder if calling the enforcement there would be more efficient.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
otchet-broad: I was thinking the same thing when I made the last getter but didn't know if that was too much of a refactor right now.  I may take a stab at it when I handle the NFE.

rushtong: Agreed, definitely not a priority

src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
otchet-broad: I had started down that path and pulled the changes later in the day yesterday.  When we migrate to all of our endpoints using DuosUserAuthenticator, we can move this call but right now it appeared that our `@PermitAll` annotations do not exercise either of those paths.

rushtong: Yes, that is true - forgot about that one 👍🏽

rushtong (approved):
Looks good to me thank you 👍🏽

src/main/java/org/broadinstitute/consent/http/service/UserService.java
otchet-broad: Looks like this will be more involved because of changes we'll need to make with the constructor that maps a user.  Not a high bar, but not worth the squeeze at the moment.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -454,12 +454,135 @@ public void hasValidActiveERACredentials(User user) {
...
+
+  @VisibleForTesting
+  protected boolean hasLibraryCard(User user) {
+    return !user.getLibraryCards().isEmpty();
+  }
+
+  @VisibleForTesting
+  protected boolean hasMatchingInstitutionInDatabase(
+      Institution institutionFromEmail, Institution institutionFromDatabase) {
+    if (institutionFromEmail == null || institutionFromDatabase == null) {
```
otchet-broad, line 575: If either the institution from the map lookup or the institution found in the database for the user are null, this check returns false.  This covers case 11.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
pshapiro4broad: If you're not going to convert this to a record now, I wouldn't bother with any cleanups in this PR, as any changes you make will end up being changed again when the record migration is done.

Having a custom constructor for a record is easy to do. Also, if you modify the code for `SimplifiedUser` slightly, you can enable the intellij refactoring operation to convert it to a `record` which takes care of getters etc automatically for you.

pshapiro4broad: Actually, looking at the code, I don't see why we have `SimplifiedUser` at all. It looks like every case we start with `User`, then build a `SimplifiedUser`, and then later convert that back into a `User`.

PR #2536, opened 2025-05-22
DT-1662: Ensure single Library Card per User

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-1662

## Summary
This PR enforces that a user can only have a single library card. There are a significant number of test changes to support this requirement. For backwards compatibility with the UI, we're populating the User object with a `libraryCard` and a list of `libraryCards`. When the UI is updated to match this, we can come back and remove the deprecated field.

### Notes
There will need to be a follow-on PR in the UI to deal with the user model describing an array of library cards. See https://github.com/DataBiosphere/duos-ui/blob/ee9e70640759c33bed9e6b9f8b5393f4692dea8b/src/types/model.ts#L51 for where we need to start there.

See these tickets for that work:
* https://broadworkbench.atlassian.net/browse/DT-1695
* https://broadworkbench.atlassian.net/browse/DT-1696

There are deletes for duplicate `user_id` and `user_email` rows. In Prod, there are no records affected. In Dev and Staging, there are 2 and 1 record respectively that will be removed.

### Testing
Tested the db update with a recent db download from
* ✅ Dev
* ✅ Staging
* ✅ Prod

Tested the UI Admin/SO Library Card table functionality
* ✅  Dev
* ✅ Staging
* ✅ Prod

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
@@ -27,16 +27,38 @@ public interface UserDAO extends Transactional<UserDAO> {
   @RegisterBeanMapper(value = User.class, prefix = "u")
   @RegisterBeanMapper(value = UserRole.class)
   @RegisterBeanMapper(value = Institution.class, prefix = "i")
+  @RegisterBeanMapper(value = LibraryCard.class, prefix = "lc")
   @UseRowReducer(UserWithRolesReducer.class)
-  @SqlQuery("SELECT "
-      + User.QUERY_FIELDS_WITH_U_PREFIX + QUERY_FIELD_SEPARATOR
-      + Institution.QUERY_FIELDS_WITH_I_PREFIX + QUERY_FIELD_SEPARATOR
-      + "     ur.user_role_id, ur.user_id, ur.role_id, ur.dac_id, r.name  "
-      + " FROM users u "
-      + " LEFT JOIN user_role ur ON ur.user_id = u.user_id "
-      + " LEFT JOIN roles r ON r.role_id = ur.role_id "
-      + " LEFT JOIN institution i ON u.institution_id = i.institution_id"
-      + " WHERE u.user_id = :userId")
+  @SqlQuery("""
+            SELECT
+          u.user_id as u_user_id,
+          u.email as u_email,
+          u.display_name as u_display_name,
+          u.create_date as u_create_date,
+          u.email_preference as u_email_preference,
+          u.institution_id as u_institution_id,
+          u.era_commons_id as u_era_commons_id,
+          i.institution_id as i_id,
+          i.institution_name as i_name,
+          i.it_director_name as i_it_director_name,
+          i.it_director_email as i_it_director_email,
+          i.create_date as i_create_date,
+          i.update_date as i_update_date,
+          ur.user_role_id as ur_user_role_id, ur.user_id as ur_user_id,
+          ur.role_id as ur_role_id, ur.dac_id as ur_dac_id, r.name as ur_name,
+          lc.id AS lc_id, lc.user_id AS lc_user_id,
+          lc.user_name AS lc_user_name, lc.user_email AS lc_user_email,
+          lc.create_user_id AS lc_create_user_id, lc.create_date AS lc_create_date,
+          lc.update_user_id AS lc_update_user_id,
+          ld.daa_id as lc_daa_id
+      FROM users u
+      LEFT JOIN user_role ur ON ur.user_id = u.user_id
+      LEFT JOIN roles r ON r.role_id = ur.role_id
+      LEFT JOIN institution i ON u.institution_id = i.institution_id
+      LEFT JOIN library_card lc ON lc.user_id = u.user_id
+      LEFT JOIN lc_daa ld ON lc.id = ld.lc_id
+      WHERE u.user_id = :userId
+      """)
   User findUserById(@Bind("userId") Integer userId);
 
   @RegisterBeanMapper(value = User.class, prefix = "u")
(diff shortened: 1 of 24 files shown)


src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
```diff
@@ -175,19 +218,6 @@ Integer insertUser(@Bind("email") String email,
...
-  @UseRowReducer(UserWithRolesReducer.class)
-  @SqlQuery("SELECT "
-      + User.QUERY_FIELDS_WITH_U_PREFIX + QUERY_FIELD_SEPARATOR
-      + "     ur.user_role_id, ur.user_id, ur.role_id, ur.dac_id, r.name "
-      + " FROM users u "
-      + " LEFT JOIN user_role ur ON ur.user_id = u.user_id "
-      + " LEFT JOIN roles r ON r.role_id = ur.role_id "
-      + " WHERE LOWER(u.email) = LOWER(:email) "
-      + " AND r.role_id = :roleId")
-  User findUserByEmailAndRoleId(@Bind("email") String email, @Bind("roleId") Integer roleId);
```
rushtong, line 190: Only used in tests

src/main/java/org/broadinstitute/consent/http/db/mapper/UserWithRolesReducer.java
```diff
@@ -56,34 +48,56 @@ public void accumulate(Map<Integer, User> map, RowView rowView) {
...
         user.addProperty(p);
       }
     } catch (MappingException e) {
-      // Ignore any attempt to map a column that doesn't exist
+      logDebug("Error adding User Property to User: %s".formatted(e.getMessage()));
+    }
+  }
+
+  // Some queries look for `user_role_id` while those that use a prefix look for `u_user_role_id`
+  private UserRole mapUserRoleFromRowView(RowView rowView, Integer userId) {
```
rushtong, line 78: In theory, I don't think this method should be necessary. However, I found that the mapping wasn't working for cases where the prefix was `ur_` and couldn't get that to work like I think it should.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
```diff
@@ -186,10 +181,9 @@ public Response deleteDaaForUser(
       @PathParam("daaId") Integer daaId,
       @PathParam("userId") Integer userId) {
     try {
-      User user = userService.findUserByEmail(authUser.getEmail());
-      List<LibraryCard> libraryCards = libraryCardService.findLibraryCardsByUserId(userId);
```
rushtong, line 190: This is a bug. When we remove DAAs from a library card, we need to remove them from the user passed into the method via `userId`. This code will remove a DAA from the `AuthUser`'s library card.

src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
```diff
@@ -28,16 +28,38 @@ public interface UserDAO extends Transactional<UserDAO> {
...
+          ur.user_role_id as ur_user_role_id, ur.user_id as ur_user_id,
+          ur.role_id as ur_role_id, ur.dac_id as ur_dac_id, r.name as ur_name,
+          lc.id AS lc_id, lc.user_id AS lc_user_id,
+          lc.user_name AS lc_user_name, lc.user_email AS lc_user_email,
+          lc.create_user_id AS lc_create_user_id, lc.create_date AS lc_create_date,
+          lc.update_user_id AS lc_update_user_id,
+          ld.daa_id as lc_daa_id
+      FROM users u
+      LEFT JOIN user_role ur ON ur.user_id = u.user_id
+      LEFT JOIN roles r ON r.role_id = ur.role_id
```
fboulnois, line 57: Should we pull this common section into something separate referenced by this method and the other one?

src/test/java/org/broadinstitute/consent/http/db/UserDAOTest.java
```diff
@@ -191,7 +207,7 @@ void testFindUsersWithLCsAndInstitution() {
 
     User user2 = createUserWithInstitution();
     int lcId2 = libraryCardDAO.insertLibraryCard(user2.getUserId(),
-        user.getDisplayName(), user.getEmail(), user.getUserId(), new Date());
+        user2.getDisplayName(), user2.getEmail(), user2.getUserId(), new Date());
```
fboulnois, line 210: Good catch.

fboulnois (approved):
Looks good, a suggestion:

src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
otchet-broad: @fboulnois if we do that, won't we get complaints from SemGrep?  Or are you suggesting we also change that configuration at this time?

otchet-broad: If I had to guess, it's why this class is being refactored in this way.

src/main/java/org/broadinstitute/consent/http/db/mapper/UserWithRolesReducer.java
```diff
@@ -56,34 +48,56 @@ public void accumulate(Map<Integer, User> map, RowView rowView) {
         }
       }
     } catch (MappingException e) {
-      //Ignore institution mapping errors, possible for new users to not have an institution
+      logDebug("Error adding Institution to User: %s".formatted(e.getMessage()));
```
otchet-broad, line 51: Thanks for adding the case of the debug logs.  I find them helpful for local development.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
otchet-broad: Good catch!

otchet-broad: Should this method also be checking that the SO is from the same institution as the user?

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
otchet-broad, line 1: @rushtong - Do I understand correctly that everyone can read every DAA?

src/test/java/org/broadinstitute/consent/http/db/LibraryCardDAOTest.java
```diff
@@ -32,28 +30,51 @@ void testInsertLibraryCard() {
...
+      libraryCardDAO.insertLibraryCard(user1.getUserId(), user1.getDisplayName(), user1.getEmail(),
+          user1.getUserId(), new Date());
+      fail("Should have thrown a UNIQUE_VIOLATION exception");
     } catch (Exception e) {
-      assertEquals(PSQLState.FOREIGN_KEY_VIOLATION.getState(),
+      assertEquals(PSQLState.UNIQUE_VIOLATION.getState(),
+          ((PSQLException) e.getCause()).getSQLState());
+    }
+    User user2 = createUser();
+    // Test Unique on library_card.user_email - note that we're using the same email as user1
```
otchet-broad, line 71: Nice!

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
rushtong: @otchet-broad - yes, all DAAs are intentionally public.

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
rushtong: This is currently not enabled in any environment so I want to tackle any explicit functional changes to DAA work as a distinct task. We will need to re-evaluate the entire feature before we do any further work here.

src/main/java/org/broadinstitute/consent/http/models/User.java
```diff
@@ -101,7 +101,7 @@ public User(String json) {
     // Nor do we need to retrieve the full institution object from user-provided data.
     JsonObject filteredUserJsonObject = filterFields(
         userJsonObject,
-        Arrays.asList("createDate", "institution", "libraryCards"));
+        Arrays.asList("createDate", "institution", "libraryCard"));
```
otchet-broad, line 104: Do we also have to support `libraryCards` until they're completely deprecated?

src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
rushtong: > If I had to guess, it's why this class is being refactored in this way.

This is the primary reason why this is refactored, yes. The string concatenation approach has been problematic to maintain and triggers semgrep issues. I would like to see if we can come up with a better way of re-using large query blocks in some other way, but for now, this is the least bad option as I see it.

src/main/java/org/broadinstitute/consent/http/models/User.java
rushtong: I don't think we do. This method constructs a `User` object. Since `User` doesn't have a `libraryCards` field so if a user-provided string does contain `libraryCards`, it will be ignored on the next line in `gson.fromJson()`

otchet-broad (approved):

src/main/java/org/broadinstitute/consent/http/resources/DaaResource.java
otchet-broad: Made a ticket to remind ourselves: https://broadworkbench.atlassian.net/browse/DT-1704

PR #2537, opened 2025-05-23
[DT-1690] Remove update library card endpoint

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1690

### Summary

Removes the update library card endpoint. There is a bunch of other update library card functionality that we cannot remove because when we create a user, they may already have library cards associated with their email which we then associate to the user, which is also considered a "library card update".

----

- Verify all tests go green


snf2ye (approved):

rushtong (approved):
Thank you 👍🏽

PR #2540, opened 2025-05-27
Replace SimplifiedUser with User

pshapiro4broad:
### Addresses

It would be simpler to not have `SimplifiedUser` since it is a subset of `User`. It's true that it's unclear in our code when we have a completely populated object and an incompletely populated one, and having classes like `SimplifiedUser` can make this more apparent. But this practice isn't followed elsewhere in the code for other DAOs, and having two different objects means that we need to convert between them.

If we decide that we want two different types, another approach would be to make `SimplifiedUser` a base class of `User`, so conversion wouldn't be necessary.

### Summary

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/InstitutionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/InstitutionDAO.java
@@ -6,7 +6,6 @@ import org.broadinstitute.consent.http.db.mapper.InstitutionMapper;
 import org.broadinstitute.consent.http.db.mapper.InstitutionWithUsersReducer;
 import org.broadinstitute.consent.http.models.Institution;
 import org.broadinstitute.consent.http.models.User;
-import org.broadinstitute.consent.http.service.UserService.SimplifiedUser;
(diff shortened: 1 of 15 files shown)


rushtong:
`SimplifiedUser` is a concept designed for the yet-to-be-released feature of Data Access Agreements. The idea was to expose the minimum amount of information to SOs so that we're not leaking information we shouldn't be. It's been a while, but [here is a doc](https://docs.google.com/document/d/181oiyqC3SQC20mFUY_wsDo9EDnOOaEXPuq2TZD-zNbQ) describing DAAs.

pshapiro4broad:
>  @rushtong `SimplifiedUser` is a concept designed for the yet-to-be-released feature of Data Access Agreements. The idea was to expose the minimum amount of information to SOs so that we're not leaking information we shouldn't be.

OK, that makes sense. In that case I'd push SimplifiedUser to the outer part of the API, e.g. use `User` at the DAO and service layer and convert to `SimplifiedUser` when creating the API response.

PR #2541, opened 2025-05-27
[DT-1607] Update users so that library cards and institutions match expected values.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-1607](https://broadworkbench.atlassian.net/browse/DT-1607)

### Summary

This pull request is based on updated requirements added to the ticket noted above.  Rules were simplified in the ticket to be as follows:

1) Assign institution to user based on email domain mapping (assign null if no match)

2) If there is no institution matching an email domain, then remove the library card if it exists

3) Else, check if the SO and researcher belong to the same institution using the email/domain map.  If there's a mismatch, remove library card.  Do nothing if there is a match.

Because this code reads and potentially updates the database at each request, care is taken to minimize this activity.

Minor refactoring of the SimplifiedUser object takes place in this pull request to address minor Sonar feedback related to code that's been in place for some time now.  A deeper refactoring of this that either removes SimplifiedUser or makes additional enhancements to it is beyond the scope of this effort.

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
```diff
@@ -49,6 +49,13 @@ protected AuthUser buildAuthUserFromHeaders(Map<String, String> headers) {
       logWarn(String.format(
           "Reading oauth2 claim headers: email is null, auth user is incomplete. Aud: %s Name: %s",
           aud, name));
+    } else {
+      try {
```
otchet-broad, line 53: If we've got an email address for the user, look them up and run the enforcement logic.

Needs to be here because of PermitAll in our code.  

Future refactoring might include replacing DuosUser by adding an Optional of the User to AuthUser so that it can leverage the obtained User object to be used in existing endpoints.

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
```diff
@@ -166,8 +166,8 @@ public void sendDaaRequestEmails(User user, Integer daaId) throws Exception {
         DataAccessAgreement daa = findById(daaId);
         String daaName = daa.getFile().getFileName();
         User toUser = new User();
-        toUser.setEmail(signingOfficial.email);
-        toUser.setDisplayName(signingOfficial.displayName);
+        toUser.setEmail(signingOfficial.getEmail());
```
otchet-broad, line 169: Getters and Setters for SimplifiedUser have been added.

@pshapiro4broad - I think there's more conversation needed on your draft pull request about partial objects and if we should be using them in consent.  In the mean time, I'm cleaning up some of the issues Sonar has with the current implementation of SimplifiedUser.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -122,15 +122,15 @@ public User updateUserFieldsById(UserUpdateFields userUpdateFields, Integer user
       // Handle Roles
       if (Objects.nonNull(userUpdateFields.getUserRoleIds())) {
         List<Integer> currentRoleIds = userRoleDAO.findRolesByUserId(userId).stream()
-            .map(UserRole::getRoleId).collect(Collectors.toList());
+            .map(UserRole::getRoleId).toList();
```
otchet-broad, line 126: Makes this list immutable.

Changes from collect(Collectors.toList()) to toList() in this pull request are for collections that should not be modified.

src/test/java/org/broadinstitute/consent/http/resources/UserResourceTest.java
```diff
@@ -55,7 +54,7 @@
 import org.postgresql.util.PSQLState;
 
 @ExtendWith(MockitoExtension.class)
-class UserResourceTest {
+class UserResourceTest extends AbstractTestHelper {
```
otchet-broad, line 57: Extending AbstractTestHelper allows us to migrate from deprecated Random methods.

src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
```diff
@@ -849,13 +856,257 @@ void testFindUsersInJsonArrayInvalidKey() {
...
+  }
+
   private User generateUserWithoutInstitution() {
     User u = generateUser();
     u.setInstitutionId(null);
     return u;
   }
 
-  private User generateUser() {
+  private static User generateUser() {
```
otchet-broad, line 1084: Made static so it can be leveraged in a static way for Parameterized Test.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -454,12 +454,135 @@ public void hasValidActiveERACredentials(User user) {
...
+        dropLibraryCardForUser(user);
+        modifiedUser = true;
+      }
+    }
+    return modifiedUser;
+  }
+
+  @VisibleForTesting
+  protected boolean handleUserWithoutInstitutionInMap(User user, Institution institutionFromDatabase) {
+    if (hasLibraryCard(user)) {
```
otchet-broad, line 520: Need for this is reinforced by [https://broadworkbench.atlassian.net/browse/DT-1669](https://broadworkbench.atlassian.net/browse/DT-1669)

src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
rushtong: `Optional<User>` on `AuthUser` was my original approach that we declined in favor of the current approach. Looking forward to revisiting this.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -454,12 +454,135 @@ public void hasValidActiveERACredentials(User user) {
...
+  }
+
+  private void dropLibraryCardForUser(User user) {
+    libraryCardDAO.deleteAllLibraryCardsByUser(user.getUserId());
+  }
+
+  @VisibleForTesting
+  protected boolean hasInstitutionMatchingEmailDomain(Institution institution) {
+    return institution != null;
+  }
```
rushtong: Naming here is confusing. I would assume this would mean that the method tells you if an institution and an email are a match or not.

src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
```diff
@@ -49,6 +49,13 @@ protected AuthUser buildAuthUserFromHeaders(Map<String, String> headers) {
...
+        User user = userService.findUserByEmail(email);
+        userService.enforceInstitutionAndLibraryCardRules(user);
```
rushtong: Another approach would be to fully push the user lookup into `enforceInstitutionAndLibraryCardRules` and only pass in the email string as an argument.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -454,12 +454,135 @@ public void hasValidActiveERACredentials(User user) {
...
+    }
+  }
+
+  @VisibleForTesting
+  protected boolean handleUserWithInstitutionInMap(User user, Institution institutionFromEmail, Institution institutionFromDatabase) {
+    boolean modifiedUser = false;
+    if (!hasMatchingInstitutionInDatabase(institutionFromEmail, institutionFromDatabase)) {
+      assignInstitutionToUser(user, institutionFromEmail.getId());
+      // Set these so we don't have to look the user up again in the database for the next bit of
+      // work with the library card.
```
rushtong: I'm a little hesitant about mutability here. This method assumes that all future usages will be in conjunction with `the next bit of work with the library card.`

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -454,12 +454,135 @@ public void hasValidActiveERACredentials(User user) {
...
 
+  /**
+   * Compliance method defined in
+   * <a href="https://broadworkbench.atlassian.net/browse/DT-1607">this ticket</a> that implements
+   * a set of rules in order to ensure Library Card and Institution matching rules are
+   * adhered to when authorizing users of the system.
+   * @param user the User being evaluated
+   * @return user with the Institution and Library Card rules applied
+   */
+  public User enforceInstitutionAndLibraryCardRules(User user) {
```
rushtong: One concern I have about the number of potential db updates is that they are not all encapsulated in a single transaction. I think this is a good example of the need to make a series of updates with a rollback. Maybe this can be moved to `UserServiceDAO` instead?

src/main/java/org/broadinstitute/consent/http/service/UserService.java
otchet-broad: Would your preference be to set the institution at 500 and reconstitute the user from the database at 501 with something like user = findUserById(user.getId())?

src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
pshapiro4broad: I had the same thought. This is the only use of this method, and this code doesn't need to know about checking for unknown users.

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
pshapiro4broad: OK. I think a better fix here would be to make the fields `final` so a getter isn't needed, but I don't know how to configure JDBI to use a non-bean API for object construction.

src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
```diff
@@ -81,13 +80,13 @@ public void deleteInstitutionById(Integer id) {
     institutionDAO.deleteInstitutionById(id);
   }
 
-  public Institution findInstitutionById(Integer id) {
+  public Institution findInstitutionById(Integer id) throws NotFoundException {
```
pshapiro4broad, line 83: Why did you add this? In general, I don't think it's a good idea to declare unchecked exceptions, unless they're directly thrown in the class, since if at some point this function no longer throws the exception, there's no way for the compiler or an IDE to determine this.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -454,12 +454,135 @@ public void hasValidActiveERACredentials(User user) {
     }
   }
 
+  /**
+   * Compliance method defined in
+   * <a href="https://broadworkbench.atlassian.net/browse/DT-1607">this ticket</a> that implements
```
pshapiro4broad: I don't think it's a good idea to put links to JIRA in the source code. Instead, we require that all PRs have JIRA tickets and maintiners can easily see the blame for any code to find out the ticket that the code was added for. Putting it in the code results in duplicate tracking and adds more work for a developer.
```suggestion
   * Compliance method that implements
```

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -454,12 +454,135 @@ public void hasValidActiveERACredentials(User user) {
...
+    Institution institutionFromEmail = institutionService.findInstitutionForEmail(user.getEmail());
+    Institution institutionFromDatabase = null;
+    try {
+      institutionFromDatabase = institutionService.findInstitutionById(user.getInstitutionId());
+    } catch (NotFoundException nfe) {
+      // do nothing.
+    }
+
+    boolean hasInstitutionMatchingEmailDomain =
+        hasInstitutionMatchingEmailDomain(institutionFromEmail);
```
pshapiro4broad: To me, this boolean obfuscates the code. Looking at the flow below, it's important that we assert that `institutionFromDatabase != null` before using it in the first branch of the `if`. This is what `hasInstitutionMatchingEmailDomain()` is doing, but there's no way for a reader to know that without looking at the method.

Also, calling a method that checks for `null` `hasInstitutionMatchingEmailDomain()` doesn't make sense to me. What does checking for null have to do with an institution matching an email domain? If anything, it should be `isInstitutionNotNull()`.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -454,12 +454,135 @@ public void hasValidActiveERACredentials(User user) {
...
+   * @return user with the Institution and Library Card rules applied
+   */
+  public User enforceInstitutionAndLibraryCardRules(User user) {
+    Institution institutionFromEmail = institutionService.findInstitutionForEmail(user.getEmail());
+    Institution institutionFromDatabase = null;
+    try {
+      institutionFromDatabase = institutionService.findInstitutionById(user.getInstitutionId());
+    } catch (NotFoundException nfe) {
+      // do nothing.
+    }
```
pshapiro4broad: Why do we need `institutionFromDatabase` here? Its only use is to get `institutionId`, which we already have, and to set `institution` in `User`, which is unused since any changes to `User` result in re-reading it from the database below.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -454,12 +454,135 @@ public void hasValidActiveERACredentials(User user) {
...
+   * @param user the User being evaluated
+   * @return user with the Institution and Library Card rules applied
+   */
+  public User enforceInstitutionAndLibraryCardRules(User user) {
+    Institution institutionFromEmail = institutionService.findInstitutionForEmail(user.getEmail());
+    Institution institutionFromDatabase = null;
+    try {
+      institutionFromDatabase = institutionService.findInstitutionById(user.getInstitutionId());
+    } catch (NotFoundException nfe) {
+      // do nothing.
```
pshapiro4broad: Under what circumstances can this occur? How does `user` have an `institutionId` which isn't in the database? This seems like a `500` issue if it can occur.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -454,12 +454,135 @@ public void hasValidActiveERACredentials(User user) {
...
+      return user;
+    }
+  }
+
+  @VisibleForTesting
+  protected boolean handleUserWithInstitutionInMap(User user, Institution institutionFromEmail, Institution institutionFromDatabase) {
+    boolean modifiedUser = false;
+    if (!hasMatchingInstitutionInDatabase(institutionFromEmail, institutionFromDatabase)) {
+      assignInstitutionToUser(user, institutionFromEmail.getId());
+      // Set these so we don't have to look the user up again in the database for the next bit of
```
pshapiro4broad: I don't think the code works this way. This method returns `true` so we end up calling `findUserByEmail()` on line 490 above.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -454,12 +454,135 @@ public void hasValidActiveERACredentials(User user) {
...
+
+    if (modifiedUser) {
+      return findUserByEmail(user.getEmail());
+    } else {
+      return user;
+    }
+  }
+
+  @VisibleForTesting
+  protected boolean handleUserWithInstitutionInMap(User user, Institution institutionFromEmail, Institution institutionFromDatabase) {
```
pshapiro4broad: This code is refactored but I don't find the method names helpful to understand what each method is doing. Given how short the methods are, I would inline them unless we can articulate what each is doing for a user and institution.

pshapiro4broad (commented):
A few comments so far

src/main/java/org/broadinstitute/consent/http/service/UserService.java
otchet-broad: Sonar complained about the complexity when it was inline.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
otchet-broad: Expected.  The next bit of code this is referring to is the next meaningful line of code in this function.

I'm expecting that the `enforce` function always return the service's view of the object.  Maybe I can clarify that in some way.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -454,12 +454,135 @@ public void hasValidActiveERACredentials(User user) {
...
+    return modifiedUser;
+  }
+
+  @VisibleForTesting
+  protected boolean handleUserWithoutInstitutionInMap(User user, Institution institutionFromDatabase) {
+    if (hasLibraryCard(user)) {
+      dropLCAndInstitutionForUser(user);
+      return true;
+    } else {
+      if (institutionFromDatabase != null) {
```
pshapiro4broad: This check for `!= null` seems redundant to me. You can remove it, and then remove the `assignInstitution()` from `dropLCAndInstitutionForUser()`, and always call it here. If it happens that it's already `null`, the `UPDATE` will have no effect. In fact, why not make the entire method unconditional? The work to check to see if an update is needed doesn't seem worth the extra logic here.
```
  protected boolean handleUserWithoutInstitutionInMap(User user) {
    userDAO.updateInstitutionId(user.getUserId(), null);
    libraryCardDAO.deleteAllLibraryCardsByUser(user.getUserId());
    return true;
  }
```

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -454,12 +454,135 @@ public void hasValidActiveERACredentials(User user) {
...
+        assignInstitutionToUser(user, null);
+        return true;
+      }
+    }
+    return false;
+  }
+
+  private void assignInstitutionToUser(User user, Integer institutionId) {
+    userDAO.updateInstitutionId(user.getUserId(), institutionId);
```
pshapiro4broad: I'm not a big fan of one-line methods, is there a need for this? I could maybe see a need for `removeUserInstitution()` which always sets it to `null`.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -454,12 +454,135 @@ public void hasValidActiveERACredentials(User user) {
...
+      return findUserByEmail(user.getEmail());
+    } else {
+      return user;
+    }
+  }
+
+  @VisibleForTesting
+  protected boolean handleUserWithInstitutionInMap(User user, Institution institutionFromEmail, Institution institutionFromDatabase) {
+    boolean modifiedUser = false;
+    if (!hasMatchingInstitutionInDatabase(institutionFromEmail, institutionFromDatabase)) {
```
pshapiro4broad: I don't think a method call is needed here. You can replace this with `equals()`
```suggestion
    if (!institutionFromEmail.equals(institutionFromDatabase)) {
```

src/test/java/org/broadinstitute/consent/http/service/DaaServiceTest.java
```diff
@@ -213,8 +213,8 @@ void testSendDaaRequestEmailsWithSigningOfficials() throws Exception {
     when(user.getInstitutionId()).thenReturn(1);
 
     SimplifiedUser signingOfficial = mock(SimplifiedUser.class);
```
pshapiro4broad, line 215: IMO this is not a good use of `mock()`, since `SimplifiedUser` is a POJO
```suggestion
    SimplifiedUser signingOfficial = new SimplifiedUser();
```

src/test/java/org/broadinstitute/consent/http/service/DaaServiceTest.java
```diff
@@ -213,8 +213,8 @@ void testSendDaaRequestEmailsWithSigningOfficials() throws Exception {
...
-    signingOfficial.displayName = "Official Name";
-    signingOfficial.email = "official@example.com";
+    signingOfficial.setDisplayName("Official Name");
```
pshapiro4broad, line 216: These statements are no-ops since `signingOfficial` is a mock. You can verify this by deleting this code and seeing that the tests still pass.

This is a symptom of using `any()` to `verify()` APIs in this test. If the test was matching on actual field data, then it wouldn't pass.

pshapiro4broad (commented):
A few more comments

src/main/java/org/broadinstitute/consent/http/service/UserService.java
pshapiro4broad: By enforce function do you mean `handleUserWithInstitutionInMap()`? Since the User will be refetched immediately after this returns, what is the requirement to modify the User that's passed in?

src/main/java/org/broadinstitute/consent/http/service/UserService.java
pshapiro4broad: Given that there's a foreign key constraint on `user.institution_id`, I think this code can be replaced with a check later on like
```
      if (!institutionFromEmail.getId().equals(user.getInstitutionId())) {
```
in the case where `institutionFromEmail != null`.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -454,12 +454,135 @@ public void hasValidActiveERACredentials(User user) {
...
+    boolean modifiedUser = false;
+    if (hasLibraryCard(user)) {
+      try {
+        User lcIssuer = findUserById(user.getLibraryCards().get(0).getCreateUserId());
+        Institution lcIssuerInstitution = institutionService.findInstitutionForEmail(lcIssuer.getEmail());
+        if (!user.getInstitution().equals(lcIssuerInstitution)) {
+          dropLibraryCardForUser(user);
+          modifiedUser = true;
+        }
+      } catch (NotFoundException nfe) {
```
pshapiro4broad, line 511: It's unclear here but the only lookup that can fail here is inside `institutionService.findInstitutionForEmail()` (vs `findUserById()`, which can be moved outside the `try`).

Given that the `null` return case for `findInstitutionForEmail()` is the same as the `catch (NotFoundException)` case, I would instead change `findInstitutionForEmail()` to not throw but return `null` instead (which it already does in the where no domain mapping exists). That would let you simplify this code and remove the try/catch.

src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
```diff
@@ -849,13 +856,257 @@ void testFindUsersInJsonArrayInvalidKey() {
...
+          Arguments.of(null, testUser, libraryCards1, institution2, true),
+          Arguments.of(null, testUser, libraryCards1, null, true),
+          Arguments.of(null, testUser, null, institution2, true),
+          Arguments.of(null, testUser, null, null, false)
+      );
+    }
+  }
+
+  @ParameterizedTest
+  @ArgumentsSource(InstitutionAndLibraryCardVariationsProvider.class)
```
pshapiro4broad: Instead of `ArgumentsProvider` you can use `MethodSource` which is less code
```suggestion
  static Stream<Arguments> testEnforceInstitutionAndLibraryCardVariations() {
    User testUser = generateUser();
    Institution institution1 = new Institution();
    institution1.setId(1);
    Institution institution2 = new Institution();
    institution2.setId(2);
    List<LibraryCard> libraryCards1 = List.of(new LibraryCard());
    LibraryCard libraryCard2 = new LibraryCard();
    libraryCard2.setCreateUserId(1);
    return Stream.of(
        Arguments.of(institution1, testUser, libraryCards1, institution1, true),
        Arguments.of(institution1, testUser, List.of(libraryCard2), institution1, true),
        Arguments.of(institution1, testUser, libraryCards1, institution2, true),
        Arguments.of(institution1, testUser, libraryCards1, null, true),
        Arguments.of(institution1, testUser, null, institution1, false),
        Arguments.of(institution1, testUser, null, institution2, true),
        Arguments.of(institution1, testUser, null, null, true),
        Arguments.of(null, testUser, libraryCards1, institution2, true),
        Arguments.of(null, testUser, libraryCards1, null, true),
        Arguments.of(null, testUser, null, institution2, true),
        Arguments.of(null, testUser, null, null, false)
    );
  }

  @ParameterizedTest
  @MethodSource
```

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -454,12 +454,135 @@ public void hasValidActiveERACredentials(User user) {
...
+    }
+
+    if (handleLibraryCardForUser(user)) {
+      modifiedUser = true;
+    }
+
+    return modifiedUser;
+  }
+
+  @VisibleForTesting
```
pshapiro4broad, line 501: I think most of these could be inlined, but even if they're not, I would avoid testing them individually unless it's very difficult to do so. In general a unit test should work at the API level of a class, so the external class behavior is tested. Testing non-API methods like `handleLibraryCardForUser()` can lead to otherwise unnecessary test changes if the internal code is refactored in the future.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
fboulnois: I agree with Phil that this method is hard to read, here's how I might implement it instead in pseudocode:

```
// could be null, that's ok
Institution userInstitutionFromEmail = institutionService.findInstitutionForEmail(user.getEmail());

assignInstitutionToUser(user, userInstitutionFromEmail.getId());

if (userInstitutionFromEmail == null && hasLibraryCard(user)) {
  dropLibraryCardForUser(user);
  return false;
}

Institution soInstitutionFromEmail = institutionService.findInstitutionForEmail(so.getEmail());
if (userInstitutionFromEmail.getId() != soInstitutionFromEmail.getId()) && hasLibraryCard(user)) {
  dropLibraryCardForUser(user);
  return false;
}

// else do nothing
return true;
```

fboulnois (approved):
Testing seems reasonable, otherwise a suggestion:

src/main/java/org/broadinstitute/consent/http/service/UserService.java
rushtong: That might be better, yes, but it also leads to a good number of re-fetching user data. If all of the logic were wrapped in a single transactional layer, we could avoid the multiple fetches and re-fetches.

src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
otchet-broad: Directly thrown in this class.  See `isInstitutionNull` called on line 85 throws a NotFoundException ~line 139.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
otchet-broad: This was specifically requested in the last PR to reference the truth table the code was built around (before the ticket criteria was changed).  @fboulnois  - mind if I remove the link?

src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
```diff
@@ -849,13 +856,257 @@ void testFindUsersInJsonArrayInvalidKey() {
...
+  void testEnforceInstitutionAndLibraryCardVariations(Institution institutionFromMap, User testUser, List<LibraryCard> cards, Institution institutionFromDatabase, boolean expectsUserMod) {
+    testUser.setLibraryCards(cards);
+    User alteredUser = new User();
+    alteredUser.setEmail(testUser.getEmail());
+    when(institutionService.findInstitutionForEmail(testUser.getEmail())).thenReturn(institutionFromMap);
+    when(institutionService.findInstitutionById(testUser.getInstitutionId())).thenReturn(institutionFromDatabase);
+    if (expectsUserMod) {
+      when(userDAO.findUserByEmail(testUser.getEmail())).thenReturn(alteredUser);
+      validateAlteredUserIsReturned(testUser, service.enforceInstitutionAndLibraryCardRules(testUser));
```
pshapiro4broad: Since the last line in `enforceInstitutionAndLibraryCardRules()` is `return findUserByEmail(user.getEmail())`, if the user is modified, this should instead use `assertEquals()`.
```suggestion
      assertEquals(testUser, service.enforceInstitutionAndLibraryCardRules(testUser));
```

Similarly the `else` case is `assertEquals(testUser, ...)`

src/main/java/org/broadinstitute/consent/http/service/UserService.java
pshapiro4broad: After editing the code a bit, I believe this is equivalent. It works differently than the original so the tests don't pass though.
```
  public User enforceInstitutionAndLibraryCardRules(User user) {
    Institution institutionFromEmail = institutionService.findInstitutionForEmail(user.getEmail());

    boolean modifiedUser = false;

    if (institutionFromEmail != null) {
      if (!institutionFromEmail.getId().equals(user.getInstitutionId())) {
        userDAO.updateInstitutionId(user.getUserId(), institutionFromEmail.getId());
        modifiedUser = true;
      }
      LibraryCard card = user.getLibraryCards().stream().findFirst().orElse(null);
      if (card != null) {
        User lcIssuer = findUserById(card.getCreateUserId());
        Institution lcIssuerInstitution =
            institutionService.findInstitutionForEmail(lcIssuer.getEmail());
        if (!user.getInstitution().equals(lcIssuerInstitution)) {
          libraryCardDAO.deleteAllLibraryCardsByUser(user.getUserId());
          modifiedUser = true;
        }
      }
    } else {
      userDAO.updateInstitutionId(user.getUserId(), null);
      libraryCardDAO.deleteAllLibraryCardsByUser(user.getUserId());
      modifiedUser = true;
    }

    if (modifiedUser) {
      user = findUserByEmail(user.getEmail());
    }
    return user;
  }
```

src/main/java/org/broadinstitute/consent/http/service/UserService.java
pshapiro4broad: Yes, that would be easier to read IMO. You could also push `hasLibraryCard()` into `dropLibraryCard()` since both callers are doing that, and use `return dropLibraryCardForUser(user);`

src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
otchet-broad: The User equals comparison only considers the userId property.  It would always be true.

The modification to the user we're expecting are properties that are outside of that equivalence rule.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
otchet-broad: Do we care about the number of database calls here?  If not, I can go back to an earlier implementation that always sets the institution, even when it's not needed.
I had something close to this codified but then refactored to eliminate updating the database if it wasn't needed.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
otchet-broad: Can we always expect institutionFromEmail to not be null?  I thought we could have visitors to DUOS who were not users and did not have institutions, but went through OAuth.  @rushtong Am I remembering that correctly?

src/main/java/org/broadinstitute/consent/http/service/UserService.java
otchet-broad: See my question to @fboulnois about reducing database calls here.  This was optimization to reduce the number of operations needed.  If we don't care about that, then I'm happy to remove it.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
otchet-broad: By `enforce` I meant `enforceInstitutionAndLibraryCardRules`. All that said, I've removed mutating the value by passing the needed value to the function outside of the user object.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
otchet-broad: Eliminated the need to do this.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
rushtong: `institutionFromEmail` can indeed be null. Imagine a user from a new institution that we don't know about. That user signs in with `user@new.institution.biz` and we know nothing about it. That user would have successfully auth-ed and has a real identity.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
otchet-broad: Refactored to both limit database interactions and wrap updates to multiple entities in a single transaction.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
otchet-broad: Although there's a FK, institution_id can still be set to null.  That's how a user can have an institutionId that isn't in the database.

I think the modification I made is a bit cleaner on this.  Let me know what you think.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
otchet-broad: I intentionally waited for Greg's Library Card changes to land and refactored a bit.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
fboulnois: My thought is to go for the most straightforward approach that satisfies the requirements until we know that it is a performance bottleneck, at which point we can look at alternate solutions.

I would hope that an update on an indexed table should be pretty fast.

rushtong (approved):
Thank you for the changes, this definitely looks better 👍🏽

src/main/java/org/broadinstitute/consent/http/service/UserService.java
rushtong: I agree with starting with the straightforward approach. However, this code path will be hit for every API call which also makes calls to Sam for user information. I generally lean towards the cautious side with functionality that is hit this frequently. The more we can simplify db interactions, the better.

PR #2542, opened 2025-05-30
[DT-1756] Update validation rules to change how progress reports are validated.

otchet-broad:
### Addresses
[https://broadworkbench.atlassian.net/browse/DT-1756](https://broadworkbench.atlassian.net/browse/DT-1756)

### Summary

This updates the progress report validation code so that users can submit progress reports on DARs where the IT/SO/PI are not validated.  This will allow existing DARs to have progress reports created for them.

Logic is added into the Progress Report flow to ensure that Internal Collaborators and Lab staff are from the same institution.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
@@ -16,7 +16,6 @@ import java.time.LocalTime;
 import java.time.ZoneOffset;
 import java.util.Collection;
 import java.util.Date;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
@@ -294,14 +293,14 @@ roles and members must have email addresses associated with your institution:\s"
           + String.join(", ", errorSummary));
     }
 
-    if (Boolean.TRUE.equals(parentDar.getDraft())) {
+    if (parentDar.getDraft()) {
       throw new BadRequestException(
           "Cannot create a progress report for a draft Data Access Request");
     }
     if (progressReport.getDatasetIds() == null || progressReport.getDatasetIds().isEmpty() ) {
       throw new BadRequestException("At least one dataset is required");
     }
-    if (!new HashSet<>(parentDar.getDatasetIds()).containsAll(progressReport.getDatasetIds())) {
+    if (!Set.copyOf(parentDar.getDatasetIds()).containsAll(progressReport.getDatasetIds())) {
       throw new BadRequestException("Progress report can only be created for datasets in the parent DAR");
     }
     if (progressReport.getData().getProgressReportSummary() == null ||
@@ -355,7 +354,7 @@ roles and members must have email addresses associated with your institution:\s"
    * @return The updated DataAccessRequest
    */
   public DataAccessRequest updateByReferenceId(User user, DataAccessRequest dar) {
-    if (Boolean.FALSE.equals(dar.getDraft())) {
+    if (!dar.getDraft()) {
       throw new SubmittedDARCannotBeEditedException();
     }
     try {
--- a/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
@@ -60,7 +60,9 @@ public class DataAccessRequestService implements ConsentLogger {
   private static final String MEMBER = "member";
   private static final String MEMBERS = MEMBER + "s: ";
   public static final String ALL_LISTED_PERSONNEL_MUST_SHARE_THE_SAME_INSTITUTION =
-      "All listed personnel must share the same institutional affiliation.  The following list of roles and members must have email addresses associated with your institution: ";
+      """
+All listed personnel must share the same institutional affiliation.  The following list of \
+roles and members must have email addresses associated with your institution:\s""";
   private static final String INTERNAL_COLLABORATOR = "Internal Collaborator";
   private static final String LAB_STAFF = "Lab staff";
   private final CounterService counterService;
@@ -318,7 +320,7 @@ public class DataAccessRequestService implements ConsentLogger {
       throw new NIHComplianceRuleException();
     }
 
-    userService.hasValidActiveERACredentials(user);
+    userService.validateActiveERACredentials(user);
 
     validateInternalCollaborators(dar);
   }
--- a/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
@@ -16,6 +16,7 @@ import java.time.LocalTime;
 import java.time.ZoneOffset;
 import java.util.Collection;
 import java.util.Date;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
@@ -43,7 +44,6 @@ import org.broadinstitute.consent.http.models.DataAccessRequestData;
 import org.broadinstitute.consent.http.models.Dataset;
 import org.broadinstitute.consent.http.models.Election;
 import org.broadinstitute.consent.http.models.Institution;
-import org.broadinstitute.consent.http.models.LibraryCard;
 import org.broadinstitute.consent.http.models.User;
 import org.broadinstitute.consent.http.models.Vote;
 import org.broadinstitute.consent.http.service.dao.DataAccessRequestServiceDAO;
@@ -57,6 +57,12 @@ public class DataAccessRequestService implements ConsentLogger {
   protected static final Timestamp MINIMUM_SUBMITTED_DATE_FOR_DAR_EXPIRATIONS = Timestamp.from(
       Instant.ofEpochSecond(
           LocalDate.of(2024, 9, 30).toEpochSecond(LocalTime.of(0, 0, 0, 0), ZoneOffset.UTC)));
+  private static final String MEMBER = "member";
+  private static final String MEMBERS = MEMBER + "s: ";
+  public static final String ALL_LISTED_PERSONNEL_MUST_SHARE_THE_SAME_INSTITUTION =
+      "All listed personnel must share the same institutional affiliation.  The following list of roles and members must have email addresses associated with your institution: ";
+  private static final String INTERNAL_COLLABORATOR = "Internal Collaborator";
+  private static final String LAB_STAFF = "Lab staff";
   private final CounterService counterService;
   private final DataAccessRequestDAO dataAccessRequestDAO;
   private final DarCollectionDAO darCollectionDAO;
@@ -128,7 +134,7 @@ public class DataAccessRequestService implements ConsentLogger {
     return dar;
   }
 
-  //NOTE: rewrite method into new servicedao method on another ticket
+  //NOTE: rewrite method into new service DAO method on another ticket
   public DataAccessRequest insertDraftDataAccessRequest(User user, DataAccessRequest dar) {
     if (Objects.isNull(user) || Objects.isNull(dar) || Objects.isNull(
         dar.getReferenceId()) || Objects.isNull(dar.getData())) {
@@ -184,9 +190,9 @@ public class DataAccessRequestService implements ConsentLogger {
   /**
    * Generate a DataAccessRequest from the provided DAR. The provided DAR may or may not exist in
    * draft form, so it covers both cases of converting an existing draft to submitted and creating a
-   * brand new DAR from scratch.
+   * brand-new DAR from scratch.
    *
-   * @param user              The create User
+   * @param user              The creating User
    * @param dataAccessRequest DataAccessRequest with populated DAR data
    * @return The created DAR.
    */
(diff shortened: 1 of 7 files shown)


src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -57,6 +57,12 @@ public class DataAccessRequestService implements ConsentLogger {
   protected static final Timestamp MINIMUM_SUBMITTED_DATE_FOR_DAR_EXPIRATIONS = Timestamp.from(
       Instant.ofEpochSecond(
           LocalDate.of(2024, 9, 30).toEpochSecond(LocalTime.of(0, 0, 0, 0), ZoneOffset.UTC)));
+  private static final String MEMBER = "member";
+  private static final String MEMBERS = MEMBER + "s: ";
+  public static final String ALL_LISTED_PERSONNEL_MUST_SHARE_THE_SAME_INSTITUTION =
+      "All listed personnel must share the same institutional affiliation.  The following list of roles and members must have email addresses associated with your institution: ";
```
rushtong: minor legibility question - can we use a text block to make this a little easier to scan in code?

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -304,6 +321,10 @@ public void validateDar(User user, DataAccessRequest dar) {
     userService.hasValidActiveERACredentials(user);
```
rushtong: Not your code, but I was a little thrown by the name of this `void` method. It throws if the user doesn't have valid credentials. Barring renaming the method to something like `validateActiveERACredentials` (definitely outside the scope of this PR), maybe we can add a comment here for clarity.

src/test/java/org/broadinstitute/consent/http/service/DataAccessRequestServiceTest.java
```diff
@@ -375,10 +368,52 @@ void validateProgressReport() {
...
+    progressReportData.setInternalCollaborators(Collections.singletonList(collaborator1));
+    Collaborator collaborator2 = new Collaborator();
+    collaborator2.setEmail("2" + user.getEmail());
+    progressReportData.setLabCollaborators(Collections.singletonList(collaborator2));
+    DataAccessRequest parentDar = generateDataAccessRequest();
+    parentDar.setSubmissionDate(Timestamp.from(Instant.now()));
+    parentDar.setDatasetIds(List.of(1, 2, 3));
+    parentDar.setUserId(user.getUserId());
+    when(userDAO.findUserByEmail(collaborator1.getEmail())).thenReturn(user);
+    when(institutionService.findInstitutionForEmail(any())).thenReturn(null);
```
rushtong, line 438: I found this test challenging to trace since the collaborators are all set up with the same email domains and, in theory, **should** all match the same institution as the user. The key to the test passing is that there is a deeply nested check on the institution so this is why the test succeeds. It would be clearer to me if we had real collaborators with real emails in different orgs than the user created on line 397.

rushtong (approved):
Looks reasonable to me, see comments inline 👍🏽

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
otchet-broad: Easy peasy to rename.  Done.

rushtong (approved):
Thanks for the updates 👍🏽

s-rubenstein (approved):

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -274,15 +282,26 @@ public DataAccessRequest createProgressReport(User user, DataAccessRequest progr
...
+
+    if (Boolean.TRUE.equals(parentDar.getDraft())) {
       throw new BadRequestException(
           "Cannot create a progress report for a draft Data Access Request");
     }
     if (progressReport.getDatasetIds() == null || progressReport.getDatasetIds().isEmpty() ) {
       throw new BadRequestException("At least one dataset is required");
     }
-    if (!parentDar.getDatasetIds().containsAll(progressReport.getDatasetIds())) {
+    if (!new HashSet<>(parentDar.getDatasetIds()).containsAll(progressReport.getDatasetIds())) {
```
pshapiro4broad: You can use `Set.copyOf()` here instead of `new HashSet<>()`
```suggestion
    if (!Set.copyOf(parentDar.getDatasetIds()).containsAll(progressReport.getDatasetIds())) {
```

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -274,15 +282,26 @@ public DataAccessRequest createProgressReport(User user, DataAccessRequest progr
...
+        INTERNAL_COLLABORATOR + " " + MEMBER + ": ", INTERNAL_COLLABORATOR + " " + MEMBERS, errorSummary);
+    getErrorSummary(progressReport.getData().getLabCollaborators().stream().map(Collaborator::getEmail).toList(), user.getInstitution(),
+        LAB_STAFF + " " + MEMBER + ": ", LAB_STAFF + " " + MEMBERS, errorSummary);
+
+    if (!errorSummary.isEmpty()) {
+      throw new BadRequestException( ALL_LISTED_PERSONNEL_MUST_SHARE_THE_SAME_INSTITUTION
+          + String.join(", ", errorSummary));
+    }
+
+    if (Boolean.TRUE.equals(parentDar.getDraft())) {
```
pshapiro4broad: This means that if `parentDar.getDraft()` is `null`, it's not a draft. Is that correct? Is there a case where this field is `null`?

Another way to express this is to use `Objects.requireNonNullElse()` which may make the behavior more apparent.
```suggestion
    if (Objects.requireNonNullElse(parentDar.getDraft(), false)) {
```

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -332,7 +355,7 @@ public void validateInternalCollaborators(DataAccessRequest payload) {
    * @return The updated DataAccessRequest
    */
   public DataAccessRequest updateByReferenceId(User user, DataAccessRequest dar) {
-    if (!dar.getDraft()) {
+    if (Boolean.FALSE.equals(dar.getDraft())) {
```
pshapiro4broad: This check is different than the other one, in that it maps `null` to `true`, not `false`. If this field can be `null`, it should work consistently. If it can't be `null`, we should make the getter return `boolean` instead of `Boolean`.

src/test/java/org/broadinstitute/consent/http/service/DataAccessRequestServiceTest.java
```diff
@@ -375,10 +368,52 @@ void validateProgressReport() {
...
+    collaborator2.setEmail("eve@yetanotherdomain.org");
+    progressReportData.setLabCollaborators(Collections.singletonList(collaborator2));
+    DataAccessRequest parentDar = generateDataAccessRequest();
+    parentDar.setSubmissionDate(Timestamp.from(Instant.now()));
+    parentDar.setDatasetIds(List.of(1, 2, 3));
+    parentDar.setUserId(user.getUserId());
+    when(userDAO.findUserByEmail(collaborator1.getEmail())).thenReturn(user);
+    when(institutionService.findInstitutionForEmail(any())).thenReturn(null);
+    BadRequestException badRequestException = assertThrows(BadRequestException.class, () -> service.validateProgressReport(user, progressReport, parentDar));
+    assertTrue(badRequestException.getMessage().contains("All listed personnel must share the same institutional affiliation.  The following list of roles and members must have email addresses associated with your institution: Internal Collaborator member: alice@1otherdomain.org, Lab staff member: eve@yetanotherdomain.org"));
```
pshapiro4broad: This is a great case for using `assertThat()` over `assertTrue()`. If you use `assertThat()`, it will print out the strings so you can see why the `contains()` is failing.
```suggestion
    assertThat(badRequestException.getMessage(), containsString("All listed personnel must share the same institutional affiliation.  The following list of roles and members must have email addresses associated with your institution: Internal Collaborator member: alice@1otherdomain.org, Lab staff member: eve@yetanotherdomain.org"));
```

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
otchet-broad: Good point.  I did check DataAccessRequest and we currently establish a value to true on instance creation and only allow it to be mutated to true and false again using internally controlled values that never set this to null.  Additionally, there's no public setter on this value.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
otchet-broad: I agree and think this will make things simpler going forward.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -291,7 +309,7 @@ public void validateProgressReport(User user, DataAccessRequest progressReport,
     }
   }
 
-  public void validateDar(User user, DataAccessRequest dar) {
+  private void validateCommonDarAndProgressReportElements(User user, DataAccessRequest dar) {
     if (Objects.isNull(user) || Objects.isNull(dar) || Objects.isNull(
```
rjohanek, line 304: couldn't validateInternalCollaborators be called from validateCommonDarAndProgressReportElements

rjohanek: and validateInternalCollaborators should call payload.getData().getLabAndInternalCollaborators() instead of payload.getData().getInternalCollaborators(). I think thats a bug lab collaborators not getting validated maybe. It seems like now we have validation for internal collaborators in two places: validatePersonnelInSameInstitution, and validateInternalCollaborators

otchet-broad: @rjohanek - Check out the refactor.  I've left the calls to both lab and internal collaborators separate so that we can give better error messaging about which elements need to be fixed.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -43,20 +43,28 @@
...
 import org.broadinstitute.consent.http.models.Election;
 import org.broadinstitute.consent.http.models.Institution;
-import org.broadinstitute.consent.http.models.LibraryCard;
 import org.broadinstitute.consent.http.models.User;
 import org.broadinstitute.consent.http.models.Vote;
 import org.broadinstitute.consent.http.service.dao.DataAccessRequestServiceDAO;
 import org.broadinstitute.consent.http.util.ConsentLogger;
 import org.jdbi.v3.core.JdbiException;
 import org.jdbi.v3.core.statement.UnableToExecuteStatementException;
+import org.jetbrains.annotations.NotNull;
```
rushtong: Is this right?

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -43,20 +43,28 @@
...
   public static final String EXPIRE_NOTICE_INTERVAL = "1 year";
   protected static final Timestamp MINIMUM_SUBMITTED_DATE_FOR_DAR_EXPIRATIONS = Timestamp.from(
       Instant.ofEpochSecond(
           LocalDate.of(2024, 9, 30).toEpochSecond(LocalTime.of(0, 0, 0, 0), ZoneOffset.UTC)));
+  private static final String MEMBER = "member";
+  private static final String MEMBERS = MEMBER + "s: ";
+  public static final String ALL_LISTED_PERSONNEL_MUST_SHARE_THE_SAME_INSTITUTION =
+      """
+All listed personnel must share the same institutional affiliation and have a library card.  The following list of \
+roles and members must have email addresses associated with your institution or library cards issued:\s""";
```
rushtong: nit: indentation. These can be indented to the standard and tests still pass.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -274,15 +282,17 @@ public DataAccessRequest createProgressReport(User user, DataAccessRequest progr
...
+
     if (parentDar.getDraft()) {
       throw new BadRequestException(
           "Cannot create a progress report for a draft Data Access Request");
     }
     if (progressReport.getDatasetIds() == null || progressReport.getDatasetIds().isEmpty() ) {
       throw new BadRequestException("At least one dataset is required");
     }
-    if (!parentDar.getDatasetIds().containsAll(progressReport.getDatasetIds())) {
+    if (!Set.copyOf(parentDar.getDatasetIds()).containsAll(progressReport.getDatasetIds())) {
```
rushtong, line 294: Nice change to set context

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
otchet-broad: Gah.  Introduced by method extraction helper in IntelliJ.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -301,29 +311,37 @@ public void validateDar(User user, DataAccessRequest dar) {
...
+  public void validateInternalCollaborators(User user, DataAccessRequest progressReport) {
+    List<String> errorSummary = getCollaboratorAndLibraryCardErrors(user, progressReport.getData());
+
+    if (!errorSummary.isEmpty()) {
+      throw new BadRequestException( ALL_LISTED_PERSONNEL_MUST_SHARE_THE_SAME_INSTITUTION
+          + String.join(", ", errorSummary));
     }
   }
 
+  @NotNull
```
rushtong: I would favor standard java if possible: `javax.annotation.@Nonnull`

rushtong (approved):
Looks good, thanks for the updates 👍🏽

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
otchet-broad: What I really like is [https://projectlombok.org/features/NonNull](https://projectlombok.org/features/NonNull). Embrace the NPE.

rjohanek (approved):
looks good thanks for the changes

PR #2548, opened 2025-06-04
DT-1788: Ensure user never has the same institution as institutionFromEmail

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1788

### Summary
Minor test update to address flakiness. Without this change, the test fails when the test user's randomly generated institution id is the same as the manually set institution from email id.

----

- Verify all tests go green

### DIFF
--- a/src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
@@ -903,7 +903,7 @@ class UserServiceTest extends AbstractTestHelper {
   void handleUserWithInstitutionInMap_DifferentInDatabaseWithLibraryCard() {
     User testUser = generateUser();
     // Ensure that testUser has an institution id that will never be the same as the institution from email
-    testUser.setInstitutionId(-1);
+    testUser.setInstitutionId(3);
     User signingOfficial = generateUser();
     LibraryCard lc = new LibraryCard();
     lc.setCreateUserId(signingOfficial.getUserId());
--- a/src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
@@ -902,6 +902,8 @@ class UserServiceTest extends AbstractTestHelper {
   @Test
   void handleUserWithInstitutionInMap_DifferentInDatabaseWithLibraryCard() {
     User testUser = generateUser();
+    // Ensure that testUser has an institution id that will never be the same as the institution from email
+    testUser.setInstitutionId(-1);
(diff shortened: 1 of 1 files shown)


fboulnois (approved):
👍

src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
```diff
@@ -902,6 +902,8 @@ void handleUserWithInstitutionInMap_DifferentInDatabase() {
   @Test
   void handleUserWithInstitutionInMap_DifferentInDatabaseWithLibraryCard() {
     User testUser = generateUser();
+    // Ensure that testUser has an institution id that will never be the same as the institution from email
+    testUser.setInstitutionId(-1);
```
fboulnois: This seems like kind of a hacky fix, should we instead set this to a specific value and the institution id to a specific value so that they can never match?

rushtong: > should we instead set this to a specific value and the institution id to a specific value so that they can never match?

I think that's exactly what I'm doing here. We're setting the institution id for the test user to a specific value, and then later on in the test setup, we're setting the institution ids on line 912 and 915 to specific values.

fboulnois: `institutionId` with a value of `-1` doesn't really make sense in our system, though.

src/test/java/org/broadinstitute/consent/http/service/UserServiceTest.java
```diff
@@ -902,6 +902,8 @@ void handleUserWithInstitutionInMap_DifferentInDatabase() {
   @Test
   void handleUserWithInstitutionInMap_DifferentInDatabaseWithLibraryCard() {
     User testUser = generateUser();
+    // Ensure that testUser has an institution id that will never be the same as the institution from email
+    testUser.setInstitutionId(3);
     User signingOfficial = generateUser();
```
rjohanek, line 907: I'm confused about what is going on in this test. The institution is also set on line 913. So, I think this is just overwritten?

rushtong: The expectation of the test is that the user's database institution does NOT match the institution derived from their email address. To make that condition happen, they have to be different IDs. Even though `testUser.institution` is set on line 913, the `testUser.institutionId` is NOT overwritten. In `generateUser`, test user is given a random institution id between 1 and 100. When that value is the same as  `institutionFromEmail` (or 1), this test will fail.

rushtong:
Closing in favor of https://github.com/DataBiosphere/consent/pull/2552

PR #2551, opened 2025-06-04
DT-1706: Update the swagger docs for the NIH endpoints

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1706

### Summary
* Update the swagger docs so that they work for the current NIH/eRA Commons account linking
* Minor refactoring of the resource class and tests to bring them up to standards.
* Removed unused `datasetPermissions` field

----

- Verify all tests go green


src/main/resources/assets/paths/nih.yaml
```diff
@@ -0,0 +1,29 @@
+delete:
+  summary: Delete eRA Commons Account
+  description: Removes only eRA Commons Account related information from the logued user, requesting this action
```
rjohanek: is logued a typo

rjohanek (approved):
this looks good, does this fix the broken swagger for this API that we noticed the other day?

src/main/resources/assets/paths/nih.yaml
rushtong: Yes, and I copied the same typo from the original version in `api-docs.yaml`

rushtong:
> this looks good, does this fix the broken swagger for this API that we noticed the other day?

@rjohanek - Yes, it does fix that by moving the post body to `requestBody` in `nih.yaml`

src/main/resources/assets/paths/nih.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - User
+    - eRA Commons
+    - Nih
+  responses:
+    200:
+      description: Returns an empty response if the operation was successful
+post:
+  summary: Store eRA Commons information
+  description: Save user's eRA Commons account
+  requestBody:
```
rushtong, line 14: This is one part of the fix. Previously, this was a `parameters` which was valid for previous swagger versions, but not this one.

snf2ye (approved):

PR #2553, opened 2025-06-05
DT-1659, DT-1675: Filter approved users to those with library cards

rushtong:
### Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DT-1675
* Separate UI work will be necessary to address the suspension aspects of DT-1675.
* Separate backend work will be necessary to address the notification aspects of DT-1675.

Completely addresses https://broadworkbench.atlassian.net/browse/DT-1659

### Summary
This PR adds additional backend filtering to ensure that there is a library card for every approved user we provide as the approved users for a dataset.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
@@ -138,7 +138,7 @@ public interface LibraryCardDAO extends Transactional<LibraryCardDAO> {
 
   /**
    * Finds library cards by user emails.
-   * @param emails A list of email addresses
+   * @param emails A list of LOWER cased email addresses
    * @return List of LibraryCard objects associated with the provided emails.
    */
   @RegisterBeanMapper(value = LibraryCard.class)
@@ -146,7 +146,7 @@ public interface LibraryCardDAO extends Transactional<LibraryCardDAO> {
   @SqlQuery("""
       SELECT *
       FROM library_card
-      WHERE user_email ILIKE ANY (ARRAY[<emails>])
+      WHERE LOWER(user_email) in (<emails>)
       """)
   List<LibraryCard> findByUserEmails(
       @BindList(value = "emails", onEmpty = EmptyHandling.NULL_STRING) List<String> emails);
--- a/src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
@@ -139,7 +139,7 @@ public interface LibraryCardDAO extends Transactional<LibraryCardDAO> {
 
   /**
    * Finds library cards by user emails.
-   * @param emails A list of LOWER cased email addresses
+   * @param emails A list of email addresses
    * @return List of LibraryCard objects associated with the provided emails.
    */
   @RegisterBeanMapper(value = LibraryCard.class)
@@ -147,7 +147,7 @@ public interface LibraryCardDAO extends Transactional<LibraryCardDAO> {
   @SqlQuery("""
       SELECT *
       FROM library_card
-      WHERE LOWER(user_email) in (<emails>)
+      WHERE LOWER(user_email) = ANY(ARRAY(SELECT LOWER(UNNEST(ARRAY[<emails>]))))
       """)
   List<LibraryCard> findByUserEmails(
       @BindList(value = "emails", onEmpty = EmptyHandling.NULL_STRING) List<String> emails);
--- a/src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
@@ -154,7 +154,7 @@ public interface LibraryCardDAO extends Transactional<LibraryCardDAO> {
   /**
    * Finds library cards by user ids.
    * @param userIds A list of user IDs
-   * @return List of LibraryCard objects associated with the provided emails.
+   * @return List of LibraryCard objects associated with the provided ids.
    */
   @RegisterBeanMapper(value = LibraryCard.class)
   @UseRowReducer(LibraryCardReducer.class)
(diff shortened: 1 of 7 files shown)


src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
```diff
@@ -133,4 +135,19 @@ INSERT INTO lc_daa (lc_id, daa_id)
...
+   * Finds library cards by user emails.
+   * @param emails A list of email addresses
+   * @return List of LibraryCard objects associated with the provided emails.
+   */
+  @RegisterBeanMapper(value = LibraryCard.class)
+  @UseRowReducer(LibraryCardReducer.class)
+  @SqlQuery("""
+      SELECT *
+      FROM library_card
+      WHERE user_email ILIKE ANY (ARRAY[<emails>])
```
fboulnois: Hrm, this tripped me up because we're not actually using a pattern, we're using the `ILIKE` to effectively cast all of the elements to lowercase to do the comparison. I'm also wondering what happens if a user manages to get a `%` or `_` through to this point, it could find something that shouldn't match, for example `noa_@example.com` would match `noah@example.com` and `noam@example.com`.

fboulnois (commented):
See below:

src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
rushtong: My original approach was to lower case`user_email` and then ensure all passed in email values were lower cased, but that seemed clunky to me - but I can go back to that. I'll add some additional testing for regexes in emails.

rushtong: Ok, this approach doesn't work - I''l go back to the original approach. These addresses are all returned when we query for any single one of them:
* `user_@domain.com`
* `user%@domain.com`
* `user*@domain.com`

snf2ye (approved):

snf2ye:
Are there already tickets to cover the rest of [DT-1675](https://broadworkbench.atlassian.net/browse/DT-1675)? 

[DT-1675]: https://broadworkbench.atlassian.net/browse/DT-1675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

rushtong:
> Are there already tickets to cover the rest of [DT-1675](https://broadworkbench.atlassian.net/browse/DT-1675)?

@snf2ye - not that I'm aware of. That ticket hasn't been refined yet which is when we would normally split it out.

[DT-1675]: https://broadworkbench.atlassian.net/browse/DT-1675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java
```diff
@@ -133,4 +135,19 @@ INSERT INTO lc_daa (lc_id, daa_id)
...
+   * Finds library cards by user emails.
+   * @param emails A list of LOWER cased email addresses
+   * @return List of LibraryCard objects associated with the provided emails.
+   */
+  @RegisterBeanMapper(value = LibraryCard.class)
+  @UseRowReducer(LibraryCardReducer.class)
+  @SqlQuery("""
+      SELECT *
+      FROM library_card
+      WHERE LOWER(user_email) in (<emails>)
```
fboulnois:
```suggestion
      WHERE lower(user_email) = any(array(SELECT lower(unnest(array[<emails>]))));
```

Ugly, but does the entire lowercase in the query instead of half in the query and half in the code.

fboulnois (approved):
One suggestion:

rushtong:
@fboulnois @snf2ye - I found a second case where we look for approved users in the Elastic Search indexing. I've added a filter in that check as well to handle library cards - please have a second look at your convenience, thank you.

src/main/java/org/broadinstitute/consent/http/service/TDRService.java
```diff
@@ -74,9 +73,15 @@ public ApprovedUsers getApprovedUsersForDataset(AuthUser authUser, Dataset datas
...
-    List<ApprovedUser> approvedUsers = Stream.of(labCollaborators, userEmails)
+    // Filter to users where that have a library card
+    List<String> allEmails = Stream.of(labCollaborators, userEmails)
         .flatMap(List::stream)
         .distinct()
+        .toList();
+
+    List<ApprovedUser> approvedUsers = libraryCardDAO.findByUserEmails(allEmails)
+        .stream()
+        .map(LibraryCard::getUserEmail)
```
otchet-broad, line 84: Looks good.  

Took me a few to walk the path of the library card email address, and I understand you're using that address because we have library cards issued to non-users who might be granted access.

The thing I'm left thinking about is if we should be checking the library card email address to see if it's got a valid institution in the institution map.  For non-users with library cards, we'll never perform revalidation of the library card to institutional mapping because those individuals never likely log into DUOS.

rushtong: > Took me a few to walk the path of the library card email address, and I understand you're using that address because we have library cards issued to non-users who might be granted access.

Exactly, there is the case for collaborators to be added as approved users. Since we don't store user ids for collaborators, but we do have emails, we can collate that with the library card emails.

> The thing I'm left thinking about is if we should be checking the library card email address to see if it's got a valid institution in the institution map. For non-users with library cards, we'll never perform revalidation of the library card to institutional mapping because those individuals never likely log into DUOS.

Yes, this is true - I think we need more work on locking down collaborators, but it's outside the scope of this task. We can quickly remove a collaborator's access by deleting their LC which is something we **cannot** do in the current code.

otchet-broad (approved):

PR #2156, opened 2023-10-02
[DUOS-2682][risk=no] Unicode fix for DAR Data queries

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DUOS-2682

### Summary
The unicode `\u0000` is unsupported when selecting as a string in Postgres.
This PR replaces it with the empty string when trying to deserialize it into JSON.
Tested against copy of production db with existing invalid data.

----
Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

PR #2555, opened 2025-06-06
DT-1792: Restrict researcher actions on summaries with closeouts

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-1792

I suggest hiding whitespace due to some of the sql formatting updates

## Summary
* Add closeout to the summary queries
* Prevent the `create progress report` action for researchers when a closeout is present
* Minor query updates to include a missing DAR status query field
* Minor query updates to standardize unicode filters, e.g. `\\\\u0000`

## Testing
Tested locally with DAR-672
1. Click the update button
2. Fill out a progress report with a closeout
3. Navigate back to the researcher console DAR tab
4. Note that DAR-672 no longer has an action for "Update"

### Screens
Screenshot 2025-06-06 at 1 34 35 PM

Screenshot 2025-06-06 at 1 35 31 PM

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -172,60 +232,6 @@ WHERE LOWER(election.election_type) = 'dataaccess'
...
-        latest_dar.parent_id AS latest_dar_parent_id,
-        u.display_name AS researcher_name,
-        i.institution_name,
-        e.election_id,
-        e.status,
-        e.dataset_id,
-        e.reference_id AS election_reference_id,
-        dd.dataset_id AS dd_datasetid,
-        (regexp_replace(latest_dar.data #>> '{}', '\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name,
-        (regexp_replace(latest_dar.data #>> '{}', '\\u0000', '', 'g'))::jsonb ->> 'status' AS dar_status,
```
rushtong, line 194: It looks like some `\\\\` were replaced with `\\`. The 4 backslashes were added in https://github.com/DataBiosphere/consent/pull/2156 so this PR fixes these unintentional changes.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -169,9 +165,11 @@ private void processDarCollectionSummariesForResearcher(List<DarCollectionSummar
       //if the latest DAR in the collection has at least one approved dataset,
       //include the create progress report action
       Set<Integer> datasetIds = dataAccessRequestDAO.findDatasetApprovalsByDar(s.getLatestReferenceId());
-      if (!datasetIds.isEmpty()) {
-        s.addAction(DarCollectionActions.CREATE_PROGRESS_REPORT);
-      }
+      // Can only create a progress report if there are approved datasets and no closeout supplement
+      if (!datasetIds.isEmpty() && s.getCloseoutSupplement() == null) {
+          s.addAction(DarCollectionActions.CREATE_PROGRESS_REPORT);
+        }
```
rushtong, line 171: This is the actual fix.

fboulnois (approved):
👍

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
otchet-broad: I'm wondering if this code should also consider the case where closeoutSupplement is not null but the reason list is empty.  Thoughts?

rushtong: I'm not sure we want that. If we did have that case, then the UI would allow for a child closeout to be created from parent closeout. The back end may restrict that, but the UI would be confused.

otchet-broad: The front end has logic to ensure that one "reason" is currently included.  You'll notice that in my PR for DT-1785, that I added similar logic into that as well.

otchet-broad: From our conversation at standup, @rushtong is going to make a ticket for`consent` input validation so we can keep the current code as-is.

otchet-broad (approved):
Thanks!

PR #2556, opened 2025-06-06
[DT-1785] Memorialize the SO approval of a closeout progress report.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-1785](https://broadworkbench.atlassian.net/browse/DT-1785)

### Summary

Adds new Resource endpoint and swagger documentation to enable SO to record approvals on a Closeout Progress Report
Adds business logic to enforce closeout approval business rules needed by the flow
Adds Compliance logging for the action
Modifies the database schema for data_access_request to store the SO and when they approved the closeout
Updates existing queries and objects needed to return the information on DARs, Dar Collections, Dar Collection Summaries
Adds test coverage for the feature.

Does not alter the metrics query at this time because that query seems like it might have specific requirements that aren't currently captured.

image

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
@@ -38,8 +38,10 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
   @UseRowReducer(DataAccessRequestReducer.class)
   @SqlQuery(
       """
-          SELECT collection.dar_code, dd.dataset_id, dar.id, dar.reference_id, dar.collection_id, dar.parent_id, dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date,
-            (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data, dar.era_commons_id
+          SELECT collection.dar_code, dd.dataset_id, dar.id, dar.reference_id, dar.collection_id, 
+            dar.parent_id, dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date,
+            (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data, dar.era_commons_id,
+            dar.closeout_so_approval_timestamp, dar.closeout_approving_so_id
           FROM data_access_request dar
           LEFT JOIN dar_dataset dd on dd.reference_id = dar.reference_id
           LEFT JOIN dar_collection collection on collection.collection_id = dar.collection_id
@@ -67,7 +69,8 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
               SELECT dar.id, dar.reference_id, dar.collection_id, dar.parent_id,
                 dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date,
                 (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data,
-                dd.dataset_id, collection.dar_code, dar.era_commons_id
+                dd.dataset_id, collection.dar_code, dar.era_commons_id,
+                dar.closeout_so_approval_timestamp, dar.closeout_approving_so_id
               FROM data_access_request dar
               LEFT JOIN dar_collection collection on collection.collection_id = dar.collection_id
               INNER JOIN dar_dataset dd ON dd.reference_id = dar.reference_id AND dd.dataset_id = :datasetId
@@ -146,7 +149,8 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
               SELECT dar.id, dar.reference_id, dar.collection_id, dar.parent_id,
                 dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date,
                 (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data,
-                dd.dataset_id, collection.dar_code, dar.era_commons_id
+                dd.dataset_id, collection.dar_code, dar.era_commons_id,
+                dar.closeout_so_approval_timestamp, dar.closeout_approving_so_id
               FROM data_access_request dar
               LEFT JOIN dar_collection collection on collection.collection_id = dar.collection_id
               LEFT JOIN dar_dataset dd ON dd.reference_id = dar.reference_id
@@ -167,8 +171,10 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
   @UseRowReducer(DataAccessRequestReducer.class)
   @SqlQuery(
       """
-              SELECT dd.dataset_id, dar.id, dar.reference_id, dar.collection_id, dar.parent_id, dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date,
-                (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data, collection.dar_code, dar.era_commons_id
+              SELECT dd.dataset_id, dar.id, dar.reference_id, dar.collection_id, dar.parent_id,
+              dar.user_id, dar.create_date, dar.sort_date, dar.submission_date, dar.update_date,
+              (regexp_replace(dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb AS data, collection.dar_code, 
+              dar.era_commons_id, dar.closeout_so_approval_timestamp, dar.closeout_approving_so_id
               FROM data_access_request dar
               LEFT JOIN dar_collection collection on collection.collection_id = dar.collection_id
               LEFT JOIN dar_dataset dd on dd.reference_id = dar.reference_id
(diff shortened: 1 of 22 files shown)


src/test/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAOTest.java
```diff
@@ -7,6 +7,7 @@
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.sql.Timestamp;
+import java.time.Instant;
```
rushtong: This looks unused

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -391,6 +391,25 @@ public Response uploadIrbDocument(
     }
   }
 
+  @PUT
+  @RolesAllowed({SIGNINGOFFICIAL})
+  @Path("/approve_closeout/{referenceId}")
+  public Response approveCloseout(@Auth AuthUser authUser,
```
rushtong: You could use `DuosUser` here if you want to avoid the find call on line 401

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -280,6 +280,39 @@ public DataAccessRequest createProgressReport(User user, DataAccessRequest progr
...
+
+    try {
+      User submitter = userService.findUserById(dataAccessRequest.getUserId());
+      if (!submitter.getInstitutionId().equals(signingOfficial.getInstitutionId())) {
+        throw new BadRequestException("Signing Officials must be in the same institution as the creator of the closeout request.");
+      }
+
+    } catch (NotFoundException e) {
+      // do nothing.  we'll allow the SO to process a closeout even if the  user can't be found.
+    }
```
rushtong: I think we should at least log this at a warning level so we can follow up on these cases. I don't think we should log this at error level which would trigger a sentry error in the alerts channel since this doesn't necessarily require immediate action.

rushtong (approved):
Looks good, see minor comments inline 👍🏽

src/test/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAOTest.java
otchet-broad: Surprised Sonar didn't catch that.  Thanks!

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -401,21 +414,28 @@ WHERE reference_id IN (<referenceIds>)
       """)
   void archiveByReferenceIds(@BindList(value = "referenceIds", onEmpty = EmptyHandling.NULL_STRING) List<String> referenceIds);
 
+  @SqlUpdate(
+    """
+        UPDATE data_access_request
+          SET closeout_approving_so_id = :id, closeout_so_approval_timestamp = now()
+        WHERE reference_id = :referenceId
+    """)
+  void updateDarCloseoutSO(@Bind("id") Integer signingOfficialUserId, @Bind("referenceId") String referenceId);
```
rushtong, line 423: For a future implementation conversation - we're storing this on a single DAR, but a closeout approval actually applies to all DARS in a collection.

src/main/resources/changesets/changelog-consent-2025-06-05-save-so-closeout-approval.xml
```diff
@@ -0,0 +1,12 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+  <changeSet id="changelog-consent-2025-06-05-save-so-closeout-approval" author="otchet-broad">
+    <addColumn tableName="data_access_request">
+      <column name="closeout_approving_so_id" type="bigint">
+        <constraints foreignKeyName="fk_so_user_id" referencedTableName="users" referencedColumnNames="user_id" nullable="true"/>
+      </column>
+      <column name="closeout_so_approval_timestamp" type="timestamp"/>
```
fboulnois, line 9: If we are starting to create closeout specific objects, should we have a `DAR -> closeout` table along with a separate `closeout` table?

fboulnois: Also, is the relationship always 1-1? I'm not sure that is the case.

fboulnois (approved):
Looks good, one suggestion:

src/main/resources/changesets/changelog-consent-2025-06-05-save-so-closeout-approval.xml
snf2ye: > Also, is the relationship always 1-1? 

With the validation logic added on the endpoint, it looks like it's only allowing one approval per progress report/DAR. This seems reasonable -- When are you thinking it wouldn't be 1-1?

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
snf2ye: I think it makes sense that we're storing this on the particular DAR/progress report that included the closeout - we know then which particular closeout/PR/DAR was approved. Although, it does make it more complicated to query if the DAR collection has been closed.

snf2ye (approved):

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
rushtong: > I think it makes sense that we're storing this on the particular DAR/progress report that included the closeout - we know then which particular closeout/PR/DAR was approved. Although, it does make it more complicated to query if the DAR collection has been closed.

I agree, this is the right solution for the SO approval case. What I think we need is a more expansive update to all DARs when the final closeout is approved by the DAC. I think we'll need to store this somewhere on all DARs (DAR status?) or the collection (new field?) so we can reduce the query complexity we have now.

otchet-broad: @rushtong - exactly.  @raejohanek and I have been discussing that as well.

src/main/resources/changesets/changelog-consent-2025-06-05-save-so-closeout-approval.xml
otchet-broad: @fboulnois
As currently defined, the relationship is 1-1.
Given that we'll need this information whenever we return a DAR, Collection, or Collection Summary, I think it makes sense to keep this on the DAR level.

Our current requirements for closeouts are that only one DAC chairperson has to acknowledge the closeout.  I think I'm also a fan of keeping that on the DAR.  If we ever got into the state of needing multiple chairs to acknowledge a closeout, we might want to switch the design of that work to a different table structure .  If we get there, I think that might lead us down a discussion about having elections again.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -391,6 +392,24 @@ public Response uploadIrbDocument(
...
+  @Path("/approve_closeout/{referenceId}")
+  public Response approveCloseout(@Auth DuosUser duosUser,
+      @Context Request request,
+      @PathParam("referenceId") String referenceId) {
+    try {
+      dataAccessRequestService.approveDataAccessRequestCloseout(duosUser.getUser(), referenceId);
+      DataAccessRequest dar = getDarById(referenceId);
+      List<Dataset> datasets = datasetService.findDatasetsByIds(dar.getDatasetIds());
+      ComplianceLogger.logCloseoutApprovalBySigningOfficial(duosUser.getUser(), datasets,
+          (ContainerRequest) request, HttpStatusCodes.STATUS_CODE_OK);
```
rjohanek, line 406: maybe this is more of a consideration for the PR with DAC Chair closeout approval, but does the list of datasets on the DAR suffice? What if this progress report only included two of an original three datasets from the parent dar and the parent dar is not expired yet? isn't this a closeout for that third dataset as well?

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -280,6 +280,43 @@ public DataAccessRequest createProgressReport(User user, DataAccessRequest progr
...
+      throw new BadRequestException("This progress report closeout has already been approved by a signing official.");
+    }
+
+    if (!signingOfficial.getUserId().equals(dataAccessRequest.getData().getCloseoutSupplement().signingOfficialId())) {
+      throw new BadRequestException("This request can only be approved by the signing official selected in the closeout request.");
+    }
+
+    try {
+      User submitter = userService.findUserById(dataAccessRequest.getUserId());
+      if (!submitter.getInstitutionId().equals(signingOfficial.getInstitutionId())) {
```
rjohanek, line 307: do we validate this when the DAR closeout is created? I feel like we should, if we do it there, do we need it here too?

rjohanek (approved):
this looks good thanks just a couple questions

rjohanek (commented):
I think we need to update the yaml schemas for the dar and dar colletion summary models

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
otchet-broad: No.  This is effectively what I was talking about with Greg in standup about.  He's created a ticket for that.

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
otchet-broad: In this context, I'm acting on the specific object so I think that leaving the list as just those on the DAR is fine.  That being said, I would expect your PR to include all of the datasets in the collection since we're taking that action there.

otchet-broad:
@raejohanek - Good catch!  Thank you!
See:
image
And:
image

PR #2564, opened 2025-06-12
[DT-1586] Prevent non-draft data access requests from being deleted.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-1586](https://broadworkbench.atlassian.net/browse/DT-1586)

### Summary

When an election does not exist for a data access request and it is a draft, 
- Users should be able to delete their own draft data access requests.
- Admins should be able to delete any draft data access requests.

No longer deletes elections and votes, even if an admin takes the action.
Adds coverage.
Swagger docs updated.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -553,8 +553,7 @@ public class DataAccessRequestResource extends Resource {
   public Response deleteDar(@Auth AuthUser authUser, @PathParam("referenceId") String referenceId) {
     try {
       validateAuthedRoleUser(Collections.singletonList(UserRoles.ADMIN), authUser, referenceId);
-      User user = findUserByEmail(authUser.getEmail());
-      dataAccessRequestService.deleteByReferenceId(user, referenceId);
+      dataAccessRequestService.deleteByReferenceId(referenceId);
       return Response.ok().build();
     } catch (Exception e) {
       return createExceptionResponse(e);
--- a/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
@@ -551,8 +551,8 @@ public class DataAccessRequestResource extends Resource {
   @Produces("application/json")
   @RolesAllowed({ADMIN, RESEARCHER})
   public Response deleteDar(@Auth AuthUser authUser, @PathParam("referenceId") String referenceId) {
-    validateAuthedRoleUser(Collections.singletonList(UserRoles.ADMIN), authUser, referenceId);
     try {
+      validateAuthedRoleUser(Collections.singletonList(UserRoles.ADMIN), authUser, referenceId);
       User user = findUserByEmail(authUser.getEmail());
       dataAccessRequestService.deleteByReferenceId(user, referenceId);
       return Response.ok().build();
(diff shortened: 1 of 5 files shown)


rushtong (approved):
Thank you 👍🏽

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -105,21 +105,14 @@ public List<DataAccessRequest> findAllDraftDataAccessRequestsByUser(Integer user
     return dataAccessRequestDAO.findAllDraftsByUserId(userId);
   }
 
-  public void deleteByReferenceId(User user, String referenceId) throws NotAcceptableException {
+  public void deleteByReferenceId(String referenceId) throws NotAcceptableException {
     List<Election> elections = electionDAO.findElectionsByReferenceId(referenceId);
     if (!elections.isEmpty()) {
```
rjohanek, line 117: is it just when there are elections open that we don't want to delete? I thought if it was submitted, even if no elections have been opened, we still don't want to be able to delete it. Why do we even need to check elections at all?

src/main/java/org/broadinstitute/consent/http/resources/DataAccessRequestResource.java
```diff
@@ -700,6 +699,9 @@ private DataAccessRequest getDarById(String referenceId) {
   private void validateAuthedRoleUser(final List<UserRoles> allowableRoles, AuthUser authUser,
       String referenceId) {
     DataAccessRequest dataAccessRequest = getDarById(referenceId);
+    if (!dataAccessRequest.getDraft()) {
+      throw new BadRequestException("Only draft data access requests can be deleted");
+    }
```
rjohanek: why does this check happen in the validate authed role method? that seems out of place to me

otchet-broad: Doing it there here as opposed to somewhere else saves extra lookups.  Line 701 is the first line of code that gets the DAR.  Additional lookups are performed after that check to determine if other criteria are met.
Would you like me to rename the method?  How does `validateDarStateAndAuthedRoleUser` sound?

rjohanek: yeah renaming is cool or you could move the getDarId call outside of this method, either way is good

fboulnois (approved):
great 👍

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
otchet-broad: My understanding is that we don't want to delete a DAR any time any election exists on it (open or closed).

The only caller to this is the delete API.  Let me try to rework a few things to see if I can shuffle some logic around and change the API of this method.  It will allow me to move the draft check here.

Ultimately, I'd like to move this logic onto the data classes themselves, but that's a bigger shift in mindset.

otchet-broad: Shuffled logic around and the delete API is updated at the service layer.
Validation to check if it's a draft is added to this method.  Since we now need to inspect the DAR at this point, I've updated the API and name here to reflect the parameter change.

otchet-broad:
Investigating import complaint.  Code builds locally and tests pass without it.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
rjohanek: but if you can only delete drafts there will never be elections, right?

otchet-broad: I can't speak to the old state machine logic, so I'm being conservative here and checking to make sure we didn't allow users to convert things with elections into drafts.

PR #2565, opened 2025-06-12
DT-1774: Remove actions on summaries for Admins and Chairs when there is a closeout

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1774

### Summary
This PR prevents any DAR Summary actions for admins and chairs when there is a closeout present.

Due to the current complexity of `processDarCollectionSummariesForChair`, I have refactored it into two parts. All action-related calculations are extracted to `updateSummaryActionsForChair` and all status calculations are maintained in `processDarCollectionSummariesForChair`

Note that there is some very similar (but not exact) calculations in `processDarCollectionSummariesForAdmin`. I have chosen not to tackle that refactor now.

### Testing
 Start out with a DAR that will be updated to have a closeout. The Admin and Chair screens appropriately show a Re-open action:

Screenshot 2025-06-12 at 1 19 51 PM

Screenshot 2025-06-12 at 1 19 42 PM

After submitting a closeout on DAR-672, the Admin and Chair consoles no longer have an action:

Screenshot 2025-06-12 at 1 21 33 PM

Screenshot 2025-06-12 at 1 21 19 PM

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
@@ -73,7 +73,8 @@ public class DarCollectionService implements ConsentLogger {
   public DarCollectionService(DarCollectionDAO darCollectionDAO,
       DarCollectionServiceDAO collectionServiceDAO, DatasetDAO datasetDAO, ElectionDAO electionDAO,
       DataAccessRequestDAO dataAccessRequestDAO, EmailService emailService, VoteDAO voteDAO,
-      MatchDAO matchDAO, DarCollectionSummaryDAO darCollectionSummaryDAO, UserDAO userDAO, DacDAO dacDAO) {
+      MatchDAO matchDAO, DarCollectionSummaryDAO darCollectionSummaryDAO, UserDAO userDAO,
+      DacDAO dacDAO) {
     this.darCollectionDAO = darCollectionDAO;
     this.collectionServiceDAO = collectionServiceDAO;
     this.datasetDAO = datasetDAO;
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -132,6 +132,9 @@ private void processDarCollectionSummariesForAdmin(List<DarCollectionSummary> su
         });
         determineCollectionStatus(s, statusCount, s.getDatasetCount(), s.getElections().size());
       }
+      if (s.getCloseoutSupplement() != null) {
```
rjohanek, line 135: another way to do this would be on line 121 where the OPEN action is added, only add that action if the closeout supplement is not null. To me, that is clearer because it is proactive and deals only with the specific case we need to worry about. it is sufficient because we don't need to worry about the CANCEL action since there will never be elections. not a blocker, just a thought

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -233,46 +236,70 @@ private void processDarCollectionSummariesForMember(List<DarCollectionSummary> s
...
+          break;
+        case OPEN:
+          summary.addAction(DarCollectionActions.VOTE);
+          break;
+        default:
+          break;
       }
     });
+
+    // Add cancel if there are no closed elections and at least one open election
```
rjohanek, line 299: I know this isn't a change in your code, but I am wondering why we don't always want to show cancel when there are open elections. If there is a closed election and an open election, is the open election really not cancellable?

rjohanek (approved):
overall looks good, just some questions

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rushtong: I did actually start with that approach, but the check is also needed for line 130 and line 127. That added to the complexity and triggered a sonar warning. I had to refactor the chair version of this method for the same sonar warning even with the simplification due to the many nested conditional checks.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rushtong: I think that's a great question and I think you might be right about this. I'm hesitant to make that change here - I suspect that will have an outsized impact on tests. I'll file a ticket to look at this question in more detail.

rushtong: See https://broadworkbench.atlassian.net/browse/DT-1827

fboulnois (approved):
looks good 👍

PR #2567, opened 2025-06-13
[DT-1819] Require country of operation for PI, internal collaborators, lab staff

otchet-broad:
### Addresses
[https://broadworkbench.atlassian.net/browse/DT-1819](https://broadworkbench.atlassian.net/browse/DT-1819)
### Summary

- Provides a JSON file that can be used by a client to load a select UI component with ISO-derived country names with UTF-8 formatted values (as appropriate)
- Validates PI, Internal Collaborators and Lab staff include a value from the country list when submitting a DAR or a progress report.
- Test coverage for the feature.

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/models/Collaborator.java
rushtong, line 1: Should we think about making this a record class? We have already converted several other dar-data objects. If it blows out the PR, I would hold off on it.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
rushtong, line 1: Line 26 has an unused import

src/main/java/org/broadinstitute/consent/http/util/CountryValidator.java
```diff
@@ -0,0 +1,33 @@
...
+import java.lang.reflect.Type;
+import java.nio.charset.StandardCharsets;
+import java.util.HashSet;
+import java.util.Set;
+
+public class CountryValidator {
+  private static final String FILEPATH = "assets/ISO-3166-countries.json";
+  private Set<String> countries;
+  public CountryValidator() {
+    try (InputStream is = getClass().getClassLoader().getResourceAsStream(FILEPATH)) {
```
rushtong, line 18: I think this should be fine, but an alternative might be to load the file contents into a local cache.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
otchet-broad: Good catch.  IntelliJ autocomplete gone bad.

src/main/java/org/broadinstitute/consent/http/util/CountryValidator.java
otchet-broad: This class is only constructed once within the constructor of the DataAccessRequestService, so it acts like an in-memory cache within that class at startup.

src/main/resources/assets/ISO-3166-countries.json
```diff
@@ -0,0 +1,251 @@
+[
```
rjohanek, line 1: maybe this file name should be dated so it is easy to tell how old it is / when it may need updating

rjohanek (approved):
looks good to me overall

src/main/java/org/broadinstitute/consent/http/models/Collaborator.java
```diff
@@ -1,59 +1,10 @@
...
-  }
-}
+public record Collaborator(
+    Boolean approverStatus,
+    String email,
+    String eraCommonsId,
+    String name,
+    String title,
+    String uuid,
+    String countryOfOperation) {}
```
rushtong, line 10: Thank you!

rushtong (approved):
👍🏽

fboulnois (approved):
👍

eweitz (approved):

src/main/resources/assets/ISO-3166-countries.json
otchet-broad: Spoke with @rjohanek about this.  We're going to leave as-is because changing the name of the file will mean we need to release the front end every time this changes.  I'm going to see if we can get a proactive notification when the wikipedia entry for ISO-3166 is updated so we are alerted to changes in that standard that this list is derived from.

PR #2569, opened 2025-06-13
[DT-1812] Require PI and Submitter to be the same value

rjohanek:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1812
I still need to do the front end work for this.

### Summary
Enforce on DARs that the PI is the same as the submitter. I am not enforcing this on Progress Reports because currently the PI field is not changeable on a Progress Report, and if the DAR was made before this rule, the user would get stuck.

Add unit tests

----

- Verify all tests go green


fboulnois (approved):
thank you 👍

rushtong (approved):
👍🏽

eweitz (approved):

PR #2574, opened 2025-06-18
[DT-1860] Check to make sure Integer is non-zero before assigning it in the mapper

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-1860](https://broadworkbench.atlassian.net/browse/DT-1860)

### Summary

This is a quirk of JDBI and this was leading to cases where we thought DARs were progress reports but the parent IDs were set to zero instead of null.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/mapper/DataAccessRequestMapper.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/mapper/DataAccessRequestMapper.java
@@ -25,8 +25,15 @@ public class DataAccessRequestMapper implements RowMapper<DataAccessRequest>, Ro
     if (hasColumn(resultSet, "dar_code")) {
       dar.setDarCode(resultSet.getString("dar_code"));
     }
-    dar.setParentId(resultSet.getInt("parent_id"));
-    dar.setUserId(resultSet.getInt("user_id"));
+
+    if (hasNonZeroColumn(resultSet,"parent_id")) {
+      dar.setParentId(resultSet.getInt("parent_id"));
+    }
+
+    if (hasNonZeroColumn(resultSet, "user_id")) {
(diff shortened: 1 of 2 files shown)


src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
```diff
@@ -234,6 +234,7 @@ void testFindByReferenceIds() {
     assertFalse(dars.isEmpty());
     assertEquals(3, dars.size());
     dars.forEach(dar -> assertNotNull(dar.getDarCode()));
+    dars.forEach(dar -> assertNull(dar.getParentId()));
```
otchet-broad, line 237: Test confirmed this was being set.

src/main/java/org/broadinstitute/consent/http/db/mapper/DataAccessRequestMapper.java
```diff
@@ -25,8 +25,15 @@ public DataAccessRequest map(ResultSet resultSet, StatementContext statementCont
...
       dar.setDarCode(resultSet.getString("dar_code"));
     }
-    dar.setParentId(resultSet.getInt("parent_id"));
-    dar.setUserId(resultSet.getInt("user_id"));
+
+    if (hasNonZeroColumn(resultSet,"parent_id")) {
+      dar.setParentId(resultSet.getInt("parent_id"));
+    }
+
+    if (hasNonZeroColumn(resultSet, "user_id")) {
```
otchet-broad, line 33: Drive by to be consistent but database schema requires a non-empty value for this column so we don't run into this case.

rushtong (approved):
👍🏽

eweitz (approved):

MatthewBemis (approved):

PR #2577, opened 2025-06-19
[DT-1862] Dac object not populated from dao

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-1862](https://broadworkbench.atlassian.net/browse/DT-1862)

### Summary

We have objects returned from the DAC DAO that aren't fully populated.  This update will rely on an existing service method to populate the needed information so the list of chairpeople to notify can be built.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DacService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DacService.java
@@ -11,6 +11,7 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.EnumSet;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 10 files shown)


src/main/java/org/broadinstitute/consent/http/service/DacService.java
```diff
@@ -294,7 +295,9 @@ List<DataAccessRequest> filterDataAccessRequestsByDac(List<DataAccessRequest> do
   }
 
   public Set<Dac> findByDatasetId(List<Integer> datasetIds) {
-    return dacDAO.findDacsForDatasetIds(datasetIds);
+    Set<Dac> returnSet = new HashSet<>();
+    dacDAO.findDacsForDatasetIds(datasetIds).forEach(dac -> returnSet.add(findById(dac.getDacId())));
```
rushtong, line 299: I generally try to avoid nested dao calls like this in favor of doing the necessary population in the outer dao call.

otchet-broad: I generally don't like returning partial objects. Unfortunately, I now see this class relies heavily on that pattern.  I'll make a ticket to do a bigger refactor after we're across the finish line.

The other problem this points to is something I've tried in the past but have been asked to not do in pull requests: integration testing between our service layers and DAOs.  I'd like to discuss that again on Monday, because bugs like this would be caught in those types of tests - especially with these thinly populated objects.

(closed this PR because I realized I made the change on a branch that wasn't develop)

PR #2578, opened 2025-06-19
[DT-1862] Use service method to populate needed information.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-1862](https://broadworkbench.atlassian.net/browse/DT-1862)

### Summary

We have objects returned from the DAC DAO that aren't fully populated. This update will rely on an existing service method to populate the needed information so the list of chairpeople to notify can be built.

This PR is taking the limited approach of populating the necessary information using the patterns already in this class.  A ticket for a [larger refactor to stop producing thinly populated objects](https://broadworkbench.atlassian.net/browse/DT-1863) has been created.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DacService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DacService.java
@@ -295,9 +295,8 @@ public class DacService implements ConsentLogger {
   }
 
   public Set<Dac> findByDatasetId(List<Integer> datasetIds) {
-    Set<Dac> returnSet = new HashSet<>();
-    dacDAO.findDacsForDatasetIds(datasetIds).forEach(dac -> returnSet.add(findById(dac.getDacId())));
-    return returnSet;
+    return dacDAO.findDacsForDatasetIds(datasetIds).stream().map(dac -> findById(dac.getDacId())).collect(
+        Collectors.toUnmodifiableSet());
   }
 
 }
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/consent/http/service/DacService.java
```diff
@@ -294,7 +295,9 @@ List<DataAccessRequest> filterDataAccessRequestsByDac(List<DataAccessRequest> do
   }
 
   public Set<Dac> findByDatasetId(List<Integer> datasetIds) {
-    return dacDAO.findDacsForDatasetIds(datasetIds);
+    Set<Dac> returnSet = new HashSet<>();
+    dacDAO.findDacsForDatasetIds(datasetIds).forEach(dac -> returnSet.add(findById(dac.getDacId())));
```
fboulnois:
```suggestion
    Set.of(dacDAO.findDacsForDatasetIds(datasetIds).forEach(dac -> findById(dac.getDacId())));
```

Can we use this construction if there are no duplicates? Otherwise there are also a few other immutable constructions:
* `Set.copyOf(Arrays.asList(array));`
* `Arrays.stream(array).collect(Collectors.toSet());`

fboulnois (approved):
One suggestion:

src/main/java/org/broadinstitute/consent/http/service/DacService.java
otchet-broad: Slightly different implementation was needed but it's now immutable.

rushtong (approved):
👍🏽

PR #2583, opened 2025-06-24
DT-1849: Refactor Approved Dataset Query

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1849

### Summary
TODO

----

- Verify all tests go green

### DIFF
--- a/src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
@@ -631,16 +631,16 @@ class DataAccessRequestDAOTest extends DAOTestHelper {
     dataAccessRequestDAO.insertDARDatasetRelation(testDar1.getReferenceId(), dataset2.getDatasetId());
     dataAccessRequestDAO.insertDARDatasetRelation(testDar1.getReferenceId(), dataset3.getDatasetId());
 
-    Election e1 = createDataAccessElection(testDar1.getReferenceId(), dataset1.getDatasetId());
+    // Set up all elections for a single DAR on multiple datasets
+    Date now = new Date();
+    Election e1 = createDataAccessElection(testDar1.getReferenceId(), dataset1.getDatasetId(), now);
     Vote v1 = createFinalVote(dataset1.getCreateUserId(), e1.getElectionId());
 
-    Election e2 = createDataAccessElection(testDar1.getReferenceId(), dataset2.getDatasetId());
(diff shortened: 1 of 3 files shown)


src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
```diff
@@ -631,17 +631,17 @@ void testFindAllDatasetApprovalsByDar() {
...
+    Date now = new Date();
+    Election e1 = createDataAccessElection(testDar1.getReferenceId(), dataset1.getDatasetId(), now);
     Vote v1 = createFinalVote(dataset1.getCreateUserId(), e1.getElectionId());
 
-    Election e2 = createDataAccessElection(testDar1.getReferenceId(), dataset2.getDatasetId());
+    Election e2 = createDataAccessElection(testDar1.getReferenceId(), dataset2.getDatasetId(), now);
     Vote v2 = createFinalVote(dataset2.getCreateUserId(), e2.getElectionId());
 
-    Election e3 = createDataAccessElection(testDar1.getReferenceId(), dataset3.getDatasetId());
+    Election e3 = createDataAccessElection(testDar1.getReferenceId(), dataset3.getDatasetId(), now);
```
rushtong, line 642: Note that previously, this code was creating elections with different dates which implicitly means there were 3 separate times when a chairperson initiated the "Open" election operation. The application doesn't allow this behavior and enforces that a chairperson can only open elections for cases where there are no existing open elections.

When elections have the same creation date, that means the chairperson initiated a single "Open" election operation.

otchet-broad: I wonder if it might make sense to refactor `createDataAccessElection` to `createDataAccessElections` to a map of referenceId -> Set<datasetIds> so that this can be created in one transaction in the database.

rushtong: > I wonder if it might make sense to refactor `createDataAccessElection` to `createDataAccessElections` to a map of referenceId -> Set so that this can be created in one transaction in the database.

This sounds like a reasonable approach. What I described above isn't technically true in the multi-DAC case. In that case, there can be multiple elections created at different times (one election group per DAC). The test case is actually correct, but I think it was written assuming a single DAC chair was creating elections, so it is accidentally correct.

Separately, I'm likely going to abandon this PR in favor of a different approach. I'm not confident that we can rely on this query to do everything it is trying to do.

rushtong:
Closing in favor of the approach taken in https://github.com/DataBiosphere/consent/pull/2587

PR #2585, opened 2025-06-24
DT-1872: Handle case insensitive domain matching

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1872

### Summary
Ignore case when looking for domain mappings

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/InstitutionDomainMap.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/InstitutionDomainMap.java
@@ -19,11 +19,17 @@ public class InstitutionDomainMap {
     return institutionDomainMap.get(institution);
   }
 
+  /**
+   * Returns the lower-cased institution name for a given email address.
+   * @param email the email address to check
+   * @return the institution name if found, otherwise null
+   */
   public String getInstitutionForEmail(String email) {
-    String domain = email.substring(email.indexOf('@') + 1);
+    String domain = email.substring(email.indexOf('@') + 1).toLowerCase();
     return institutionDomainMap.entrySet().stream()
         .filter(entry -> entry.getValue().contains(domain))
         .map(Map.Entry::getKey)
+        .map(String::toLowerCase)
         .findFirst()
         .orElse(null);
   }
--- a/src/main/java/org/broadinstitute/consent/http/models/InstitutionDomainMap.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/InstitutionDomainMap.java
@@ -20,7 +20,7 @@ public class InstitutionDomainMap {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/consent/http/models/InstitutionDomainMap.java
```diff
@@ -19,10 +19,16 @@ public Set<String> getDomainsForInstitution(String institution) {
...
   }
 
+  /**
+   * Returns the institution name for a given email address.
+   * @param email the email address to check
+   * @return the institution name if found, otherwise null
+   */
   public String getInstitutionForEmail(String email) {
-    String domain = email.substring(email.indexOf('@') + 1);
+    String domain = email.substring(email.indexOf('@') + 1).toLowerCase();
```
otchet-broad: [https://www.rfc-editor.org/rfc/rfc6531.html](https://www.rfc-editor.org/rfc/rfc6531.html) and [https://www.rfc-editor.org/rfc/rfc6530](https://www.rfc-editor.org/rfc/rfc6530) are relevant.

We might want to avoid calling toLowerCase because the domain is permitted to have UTF-8 in it and toLowerCase() follows locale rules that can break things.

src/test/java/org/broadinstitute/consent/http/models/InstitutionDomainMapTest.java
```diff
@@ -80,4 +82,34 @@ void testGetDomainsForInstitution() {
...
+    map.setInstitutionDomainMap(testMap);
+    String institution = map.getInstitutionForEmail(email);
+    assertEquals("Broad Institute", institution);
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {
+      "test@B_roadInstitute.ORG",
+      "test@Broad.*mIt.eDu",
+      "TesT-User@BROAD1NSTITUTE.ORG"
```
otchet-broad, line 126: It might be good to include a test with UTF-8.

src/main/java/org/broadinstitute/consent/http/models/InstitutionDomainMap.java
rushtong: Good point, this is redundant with the lambda check.

otchet-broad (approved):
Thanks for the changes!  Looks good!

fboulnois (approved):
👍

fboulnois (commented):
Is it possible for whitespace to creep in? If so, should we trim the email?

rushtong:
> Is it possible for whitespace to creep in? If so, should we trim the email?

I'm not sure if that is a possible scenario, but I think it would be good to protect against that.

PR #2598, opened 2025-07-02
DT-1925: Simplify dataset/study lookup

rushtong:
### Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DT-1925

### Summary
This PR simplifies the lookup of the study associated with a dataset for the get-registration-by-dataset-identifier API. Previously, we made two queries, one for the dataset and one for the study. The dataset query actually populates the associated study in this case, so the second study query is unnecessary. Local testing shows roughly a 20%-30% performance improvement.

----

- Verify all tests go green


eweitz (approved):
Code looks good.  Nice performance improvement!

MatthewBemis (approved):
nice performance win! 👍

PR #2599, opened 2025-07-02
DT-1934: Add Institution Domains

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1934

### Summary
This PR adds infrastructure to store unique domains in a join table on institutions. 
* New transactional insert/update/delete queries
* All queries are migrated to text block format
* Mapper updated/simplified and Reducer added to handle the association
  * Also fixes a bug in the mapper where we cached results. Caching in mappers can lead to obscure data presentation bugs.
* Service methods updated to call the new insert/update methods
* API docs updated to include the new field

In local testing, one can add an entry to the institution_domain table and see the values returned in the get all API:

```
...
  {
    "id": 1576,
    "name": "Université Libre de Bruxelles",
    "dunsNumber": 0,
    "domains": [
      "ubx.edu"
    ],
    "createDate": "Dec 6, 2024",
    "createUserId": 4628,
    "createUser": {
      "userId": 4628,
      "email": "pbretsch@broadinstitute.org",
      "displayName": "Pamela Bretscher",
      "createDate": 1621952634219,
      "emailPreference": true,
      "eraCommonsId": "PAMELA_BRETSCHER"
    },
    "updateUser": {}
  },
...
```

----

- Verify all tests go green


s-rubenstein (approved):

src/main/java/org/broadinstitute/consent/http/db/InstitutionDAO.java
```diff
@@ -82,76 +119,147 @@ void updateInstitutionById(@Bind("institutionId") Integer institutionId,
...
-       WHERE i.institution_id = :institutionId
-  """)
+               WHERE ur.role_id = 7) so ON i.institution_id = so.so_institution_id
+      WHERE i.institution_id = :institutionId
+      """)
   Institution findInstitutionWithSOById(@Bind("institutionId") Integer institutionId);
 
-  @SqlUpdate("DELETE FROM institution WHERE create_user = :userId OR update_user = :userId")
-  void deleteAllInstitutionsByUser(@Bind("userId") Integer userId);
+  default void deleteAllInstitutionsByUser(@Bind("userId") Integer userId) throws SQLException {
```
otchet-broad, line 250: I think we might want to revisit the utility of this method.  It's only ever used when we delete a user.  Could you imagine the impact if this were done to @rushtong or @solideoglori?

src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
```diff
@@ -55,23 +45,12 @@ public Institution updateInstitutionById(Institution institutionPayload, Integer
...
-        institutionPayload.getOrgChartUrl(),
-        institutionPayload.getVerificationUrl(),
-        institutionPayload.getVerificationFilename(),
-        (Objects.nonNull(institutionPayload.getOrganizationType())
-            ? institutionPayload.getOrganizationType().getValue() : null),
-        userId,
-        updateDate
-    );
-    return institutionDAO.findInstitutionById(id);
+    try {
```
otchet-broad: If you let this throw won't the resource layer wrap the error appropriately?

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -268,7 +268,13 @@ public void deleteUserByEmail(String email) {
           String.format("Unable to delete all drafts and files for userId %d. Error was: %s",
               userId, e.getMessage()), e);
     }
-    institutionDAO.deleteAllInstitutionsByUser(userId);
+    try {
```
otchet-broad, line 271: I think we might want to remove this entire block.

src/main/java/org/broadinstitute/consent/http/db/InstitutionDAO.java
rushtong: I think that is a great point, but it's a distinctly different effort. This should be dealt with in https://broadworkbench.atlassian.net/browse/DT-1470 where we can ensure that user removal is not so destructive.

src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
rushtong: It should, yes. Will refactor.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
rushtong: See https://github.com/DataBiosphere/consent/pull/2599/files#r2192733990
Until we address the underlying issue, we still need to support this path.

eweitz (approved):

PR #2601, opened 2025-07-03
DT-1936: New API to import allow list

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-1936

## Summary
Depends on https://github.com/DataBiosphere/consent/pull/2599

New API intended to be used once to read an `allowlist.json` file and update the domains for all institutions found in the system.

### Swagger View
Screenshot 2025-07-03 at 2 30 38 PM

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/resources/InstitutionResource.java
```diff
@@ -116,4 +118,36 @@ public Response deleteInstitution(@Auth DuosUser duosUser, @PathParam("id") Inte
       return createExceptionResponse(e);
     }
   }
+
+  @POST
+  @Consumes("application/json")
+  @Produces("application/json")
+  @Path("/domains")
+  @RolesAllowed(ADMIN)
+  public Response updateInstitutionDomains(@Auth DuosUser duosUser, String institutionDomainMap) {
```
s-rubenstein, line 127: Do we do openapi codegen in Consent? I'm assuming no given the tags, but I figured I would check, since it would be nice to have it do the object mapping for us.

s-rubenstein (approved):

src/main/java/org/broadinstitute/consent/http/resources/InstitutionResource.java
rushtong: > Do we do openapi codegen in Consent? I'm assuming no given the tags, but I figured I would check, since it would be nice to have it do the object mapping for us.

We do not. While I agree it might be nice to have that do our object mapping for us, I've found it to be very restrictive for corner cases where we want to customize what an API will or will not accept.

otchet-broad:
@rushtong I might be missing something here but I don't see how this will cover a deployment in a k8s cluster with nodes reading the data from a file in GCS.  Am I missing something here?  What happens when a node is restarted?

rushtong:
> @rushtong I might be missing something here but I don't see how this will cover a deployment in a k8s cluster with nodes reading the data from a file in GCS. Am I missing something here? What happens when a node is restarted?

It populates the database with domains, not the GCS file.

otchet-broad:
> > @rushtong I might be missing something here but I don't see how this will cover a deployment in a k8s cluster with nodes reading the data from a file in GCS. Am I missing something here? What happens when a node is restarted?
> 
> It populates the database with domains, not the GCS file.

Ah... I missed the destination branch.  I'll back up and start reviewing from that end.

rushtong:
> > > @rushtong I might be missing something here but I don't see how this will cover a deployment in a k8s cluster with nodes reading the data from a file in GCS. Am I missing something here? What happens when a node is restarted?
> > 
> > 
> > It populates the database with domains, not the GCS file.
> 
> Ah... I missed the destination branch. I'll back up and start reviewing from that end.

I will be rebasing this PR and the PATCH institution one when the base branch is merged.

otchet-broad (approved):

PR #2603, opened 2025-07-10
DT-1941: Migrate user email domain lookups to database

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1941

### Summary
The primary goal of this PR is to look up user email domains from the domains associated to institutions in the database as opposed to the static JSON file we have in GCS. Along the way, the following updates are included:

* New DAO method and tests
* Remove JSON lookup code in GCS service code
* Simplify admin role checks on user in `InstitutionResource`
* Cleanup `InstitutionDomainMap` methods and tests
* Cleanup `InstitutionUtil` methods and tests
* Expand visible fields to non-admins in `InstitutionUtil`
* Fix reducer bug where signing official institution id was not getting set in the list of SOs
* Split the enforce user institution code paths to use a simpler id-only query.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/InstitutionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/InstitutionDAO.java
@@ -263,11 +263,34 @@ public interface InstitutionDAO extends Transactional<InstitutionDAO> {
     });
   }
 
-  @UseRowReducer(InstitutionReducer.class)
+  @RegisterBeanMapper(value = User.class, prefix = "u")
+  @RegisterBeanMapper(value = SimplifiedUser.class, prefix = "so")
+  @UseRowReducer(InstitutionWithUsersReducer.class)
   @SqlQuery("""
-      SELECT i.*
+      SELECT i.*,
+        d.domain,
+        u.user_id AS u_user_id,
+        u.email AS u_email,
+        u.display_name AS u_display_name,
+        u.create_date AS u_create_date,
+        u.email_preference AS u_email_preference,
+        u.era_commons_id AS u_era_commons_id,
+        u2.user_id AS u2_user_id, u2.email AS u2_email,
+        u2.display_name AS u2_display_name, u2.create_date AS u2_create_date,
+        u2.email_preference AS u2_email_preference,
+        u2.era_commons_id AS u2_era_commons_id,
+        so.so_user_id, so.so_email, so.so_display_name
       FROM institution i
       INNER JOIN institution_domains d ON d.institution_id = i.institution_id
+      LEFT JOIN users u ON u.user_id = i.create_user
+      LEFT JOIN users u2 ON u2.user_id = i.update_user
+      LEFT JOIN
+           (SELECT
+               so.user_id AS so_user_id, so.email AS so_email,
+               so.display_name AS so_display_name, so.institution_id AS so_institution_id
+               FROM users so
+               LEFT JOIN user_role ur ON ur.user_id = so.user_id
+               WHERE ur.role_id = 7) so ON i.institution_id = so.so_institution_id
       WHERE LOWER(d.domain) = LOWER(:domain)
       LIMIT 1
       """)
--- a/src/main/java/org/broadinstitute/consent/http/db/InstitutionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/InstitutionDAO.java
@@ -294,4 +294,12 @@ public interface InstitutionDAO extends Transactional<InstitutionDAO> {
       WHERE LOWER(d.domain) = LOWER(:domain)
       """)
   Institution findInstitutionByDomain(@Bind("domain") String domain);
+
+  @SqlQuery("""
+      SELECT distinct i.institution_id
+      FROM institution i
+      INNER JOIN institution_domains d ON d.institution_id = i.institution_id
+      WHERE LOWER(d.domain) = LOWER(:domain)
+      """)
+  Integer findInstitutionIdByDomain(@Bind("domain") String domain);
 }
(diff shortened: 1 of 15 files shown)


src/main/java/org/broadinstitute/consent/http/util/InstitutionUtil.java
```diff
@@ -49,7 +34,16 @@ public boolean shouldSkipField(FieldAttributes field) {
...
+            || fieldName.equals("email")
+            || fieldName.equals("itDirectorName")
+            || fieldName.equals("itDirectorEmail")
+            || fieldName.equals("institutionUrl")
+            || fieldName.equals("dunsNumber")
+            || fieldName.equals("orgChartUrl")
+            || fieldName.equals("verificationUrl")
+            || fieldName.equals("verificationFilename")
+            || fieldName.equals("organizationType")
+            || fieldName.equals("domains"));
```
rushtong, line 46: This bit of code should allow for us to get past a [production-env bug](https://broadworkbench.atlassian.net/browse/DT-1964) where fields are being filtered regardless of user admin status. Future UI development that depends on the get-by-id functionality will need this to see all fields in production.

src/main/java/org/broadinstitute/consent/http/db/InstitutionDAO.java
```diff
@@ -262,4 +262,36 @@ WHERE institution_id IN (SELECT institution_id FROM institution WHERE create_use
       handle.commit();
     });
   }
+
+  @RegisterBeanMapper(value = User.class, prefix = "u")
+  @RegisterBeanMapper(value = SimplifiedUser.class, prefix = "so")
```
rushtong, line 267: Pulling in the user and signing officials was not strictly necessary for this functionality but I chose to include it so we can be more consistent about returning fully populated entities.

src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
```diff
@@ -65,25 +60,10 @@ public Institution findInstitutionById(Integer id) throws NotFoundException {
...
-    String name = getInstitutionDomainMap().getInstitutionForEmail(email);
-    if (name != null) {
-      var institutions = institutionDAO.findInstitutionsByName(name);
-      if (institutions.size() == 1) {
-        return institutions.get(0);
-      }
-    }
-    return null;
+    String trimmedEmail = email.trim();
+    String domain = trimmedEmail.substring(trimmedEmail.indexOf('@') + 1);
```
rushtong: This is lifted straight from the previous code in `InstitutionDomainMap`

src/main/java/org/broadinstitute/consent/http/resources/InstitutionResource.java
```diff
@@ -47,7 +48,7 @@ public InstitutionResource(InstitutionService institutionService) {
   @PermitAll
   public Response getInstitutions(@Auth DuosUser duosUser) {
     try {
-      Boolean isAdmin = institutionUtil.checkIfAdmin(duosUser.getUser());
+      Boolean isAdmin = duosUser.getUser().hasUserRole(UserRoles.ADMIN);
```
rushtong, line 51: Here and below on line 66, I chose to update to the newer method. `institutionUtil.checkIfAdmin` was pretty outdated.

src/main/java/org/broadinstitute/consent/http/db/mapper/InstitutionWithUsersReducer.java
```diff
@@ -23,25 +23,26 @@ public void accumulate(Map<Integer, Institution> map, RowView rowView) {
...
-    institution.setCreateUser(create_user);
-    institution.setUpdateUser(update_user);
+    institution.setCreateUser(createUser);
+    institution.setUpdateUser(updateUser);
 
     if (Objects.nonNull(rowView.getColumn("so_user_id", Integer.class))) {
-      SimplifiedUser so_user = rowView.getRow(SimplifiedUser.class);
-      institution.addSigningOfficial(so_user);
+      SimplifiedUser signingOfficial = rowView.getRow(SimplifiedUser.class);
+      signingOfficial.setInstitutionId(institution.getId());
```
rushtong, line 44: Bug fix: the SO user object was not getting their institution id set since it is not in the query. We already know the value here since we have the institution row data available.

src/main/java/org/broadinstitute/consent/http/db/InstitutionDAO.java
otchet-broad: UserService.enforceInstitutionAndLibraryCardRules (called on **every** request) will end up invoking this a lot. I'm wondering if we should have two methods: one that returns just the ID and the other that returns this full object.

UserService.enforceInstitutionAndLibraryCardRules only really needs to lookup by domain and get the ID back for comparison with other objects.

src/test/java/org/broadinstitute/consent/http/db/InstitutionDAOTest.java
```diff
@@ -375,4 +377,31 @@ void testUpdateFullInstitution() throws Exception {
         user.getUserId());
     assertNull(reloadedInstitution.getDomains(), "Domains should be empty after update");
   }
+
+  @ParameterizedTest
+  @ValueSource(strings = {"testdoMain.com", "AnotherDomain.com", "uniÇodé.c©m", "Broad.mちt.eDu"})
```
otchet-broad, line 382: Nice!

src/main/java/org/broadinstitute/consent/http/db/InstitutionDAO.java
rushtong: That's a good point. Let me noodle on the usages here.

rushtong: @otchet-broad PTAL - I split the user service enforcement code to use an institution id specific query. Other code paths were using the full institution.

MatthewBemis (approved):
👍

otchet-broad (approved):
Nice work and optimization!

PR #2609, opened 2025-07-16
[DT-1958] Delete Delete User API

rjohanek:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1958

### Summary
Remove delete user endpoint from resource and api docs
Keep service layer intact

### Testing
Manually verified the endpoint is not included in APIs anymore nor on Swagger

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
@@ -371,15 +371,6 @@ public class UserResource extends Resource {
     }
   }
 
-  @DELETE
-  @Produces(MediaType.APPLICATION_JSON)
-  @Path("/{email}")
-  @RolesAllowed(ADMIN)
-  public Response delete(@Auth AuthUser authUser, @PathParam("email") String email, @Context UriInfo info) {
-    userService.deleteUserByEmail(email);
-    return Response.ok().build();
-  }
-
   @GET
   @Produces(MediaType.APPLICATION_JSON)
   @Path("/signing-officials")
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -371,15 +371,6 @@ public Response createResearcher(@Context UriInfo info, @Auth AuthUser authUser)
     }
   }
 
-  @DELETE
-  @Produces(MediaType.APPLICATION_JSON)
-  @Path("/{email}")
-  @RolesAllowed(ADMIN)
-  public Response delete(@Auth AuthUser authUser, @PathParam("email") String email, @Context UriInfo info) {
-    userService.deleteUserByEmail(email);
```
rushtong, line 379: This service class method `deleteUserByEmail` can also be removed along with some of the more destructive DAO delete methods like `institutionDAO.deleteAllInstitutionsByUser` and `userDAO.deleteUserById`.

Separately, `libraryCardDAO.deleteAllLibraryCardsByUser` looks particularly bad as it deletes by create user OR update user but we can tackle that separately.

rjohanek: I agree and I'm happy to remove any unused code, but my understanding from refinement today was that we were going to leave service and DAO layer code in place until the second part of this work is done.

rushtong (approved):
👍🏽

otchet-broad (approved):
Thank you!

PR #2612, opened 2025-07-21
DT-2000: Simplify Summary By Role Lookup

rushtong:
### Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DT-2000

### Summary
This is just phase one of optimizing the performance of user's DAR Console pages. Future PRs will look more closely at the DAO layer to improve the underlying query. My local testing trims the overall API call for a chairperson call with lots of collections from ~4 seconds to about ~3.5 seconds. This should be a noticeable improvement in dev where the same call can take ~18-20 seconds.

This PR does two small-ish things. 

1. Update query to find Dataset ids by DAC Ids. Previously, we were making a looped DAO call that populated a basic dataset object only to then pull out the ids. This fix queries for the ids directly and removes the looped call. 

2. Simplify the return object in list context. Previously, we were blowing out the payload with elections and votes which the UI doesn't need to see.

----

- Verify all tests go green

### DIFF
--- a/src/main/resources/assets/schemas/DarCollectionSummary.yaml
+++ b/src/main/resources/assets/schemas/DarCollectionSummary.yaml
@@ -46,8 +46,6 @@ properties:
   progressReport:
     type: boolean
     description: Indicator true when the collection contains a progress report.
-  closeoutSupplement:
-    $ref: '../schemas/CloseOutSupplement.yaml'
   closeoutSigningOfficialApprovedDate:
     type: number
     description: When the Signing Official approved the closeout, in milliseconds since the epoch.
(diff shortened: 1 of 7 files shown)


src/main/resources/assets/schemas/DarCollectionSummary.yaml
```diff
@@ -46,8 +46,6 @@ properties:
   progressReport:
     type: boolean
     description: Indicator true when the collection contains a progress report.
-  closeoutSupplement:
-    $ref: '../schemas/CloseOutSupplement.yaml'
```
rushtong, line 50: Not used in the UI

eweitz (approved):
Code looks good -- nice performance improvement and API response simplification.

src/main/resources/assets/schemas/DarCollectionSummary.yaml
rjohanek: do we have any intentions/plans to use this field in the future?

rjohanek (approved):
this looks great!

src/main/resources/assets/schemas/DarCollectionSummary.yaml
rushtong: I hope that we do NOT have plans for this. The idea behind the summary object is to provide the minimal amount of information required to display the table row in the user's console. If we do need more closeout information, I'd like to be as judicious as possible to keep the payload small and the processing time fast.

PR #2613, opened 2025-07-21
[DT-1978] Remove Deprecated Methods from DACAutomationRuleDAO

MatthewBemis:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1978

### Summary

Removes the deprecated `useAuditedFoo` methods and inlines their SQL.
Also refactors how transactions/rollbacks are managed, leaning more into JDBI-provided utilities.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DACAutomationRuleDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DACAutomationRuleDAO.java
@@ -10,9 +10,7 @@ import org.broadinstitute.consent.http.rules.RuleAuditAction;
 import org.jdbi.v3.core.Handle;
 import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
 import org.jdbi.v3.sqlobject.customizer.Bind;
-import org.jdbi.v3.sqlobject.statement.GetGeneratedKeys;
 import org.jdbi.v3.sqlobject.statement.SqlQuery;
-import org.jdbi.v3.sqlobject.statement.SqlUpdate;
 import org.jdbi.v3.sqlobject.transaction.Transactional;
 
 @RegisterRowMapper(DACAutomationRuleMapper.class)
@@ -27,26 +25,26 @@ public interface DACAutomationRuleDAO extends Transactional<DACAutomationRuleDAO
       Instant activationDate) {
     Handle handle = getHandle();
     Integer id;
-    String auditSql = """
+    String insertAuditSql = """
         INSERT INTO dac_rule_audit (action, dac_id, rule_id, user_id, action_date)
         VALUES (:auditType::rule_audit_action, :dacId, :ruleId, :auditUserId, :actionDate)
         """;
-    String insertSql = """
+    String insertRuleSql = """
         INSERT INTO dac_rule_settings (dac_id, rule_id, user_id, activation_date)
         VALUES (:dacId, :ruleId, :userId, current_timestamp)
         """;
     try (
-        var audit = handle.createUpdate(auditSql);
-        var insert = handle.createUpdate(insertSql)
+        var insertAudit = handle.createUpdate(insertAuditSql);
+        var insertRule = handle.createUpdate(insertRuleSql)
     ) {
-      audit
+      insertAudit
           .bind("dacId", dacId)
           .bind("ruleId", ruleId)
           .bind("auditUserId", userId)
           .bind("auditType", RuleAuditAction.ADD)
           .bind("actionDate", activationDate)
           .execute();
-      id = insert
+      id = insertRule
           .bind("dacId", dacId)
           .bind("ruleId", ruleId)
           .bind("userId", userId)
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/consent/http/db/DACAutomationRuleDAO.java
```diff
@@ -39,15 +29,29 @@ default Integer auditedInsertDACRuleSetting(int dacId, int ruleId, int userId,
...
+      id = insertRule
+          .bind("dacId", dacId)
+          .bind("ruleId", ruleId)
+          .bind("userId", userId)
+          .executeAndReturnGeneratedKeys("id")
+          .mapTo(Integer.class)
+          .one();
       handle.commit();
     } catch (Exception e) {
       handle.rollback();
```
MatthewBemis: From testing, it seemed like this rollback code wasn't working as intended. In the case of the `insertRule` query failing, a stray audit record was left behind. I think the issue was that a transaction was never opened, so each query was immediately committed. Calling `handle.begin()` opens a transaction, so we can roll that transaction back if anything fails. I also added a test for this particular case.

There are other ways to manage transactions with JDBI (i.e. using `inTransaction`), happy to refactor but wasn't sure if that was out of scope.

rushtong: Yes, the current code is not handling autocommit correctly. The default connection that comes with a `handle` has autocommit set to true so any statements are committed directly. The way to do this is to wrap the [handle in a transaction](https://jdbi.org/#_handle) that sets that to false, then commits if there are no errors, and rollback if there are. We have some more examples in our codebase:
* https://github.com/DataBiosphere/consent/blob/develop/src/main/java/org/broadinstitute/consent/http/service/dao/DaaServiceDAO.java#L29
* https://github.com/DataBiosphere/consent/blob/develop/src/main/java/org/broadinstitute/consent/http/service/dao/DacServiceDAO.java#L30
* etc.

rushtong: Screenshot 2025-07-22 at 3 58 46 PM

MatthewBemis: Thanks, Greg! I've updated this PR to be more consistent with those existing examples in the codebase.

src/main/java/org/broadinstitute/consent/http/db/DACAutomationRuleDAO.java
```diff
@@ -23,135 +21,166 @@ public interface DACAutomationRuleDAO extends Transactional<DACAutomationRuleDAO
...
-      handle.commit();
-    } catch (Exception e) {
-      handle.rollback();
-      throw e;
-    } finally {
-      handle.close();
-    }
-    return id;
-  }
+    return withHandle(handle -> {
```
rjohanek, line 26: I'm having a hard time following some of these changes. What is the withHandle addition to these methods for?

rjohanek (approved):
looking at the jacoco report, it looks like test coverage can be improved by adding tests for the exception/rollback case for each of these methods

src/main/java/org/broadinstitute/consent/http/db/DACAutomationRuleDAO.java
MatthewBemis: @rushtong what are your thoughts on purely using jdbi's `inTransaction` instead? It looks like we don't need to manually set autocommit to false in this case (I double checked this with the debugger)- jdbi does that for us. And we also wouldn't have to manually manage the rollbacks. Here's what `auditedInsertDACRuleSetting` would look like:

```
  default Integer auditedInsertDACRuleSetting(int dacId, int ruleId, int userId, Instant activationDate) {
    return inTransaction(d -> {
      Handle handle = getHandle();
      String auditSql = """
        INSERT INTO dac_rule_audit (action, dac_id, rule_id, user_id, action_date)
        VALUES (:auditType::rule_audit_action, :dacId, :ruleId, :auditUserId, :actionDate)
        """;
      String insertRuleSql = """
        INSERT INTO dac_rule_settings (dac_id, rule_id, user_id, activation_date)
        VALUES (:dacId, :ruleId, :userId, current_timestamp)
        """;

      handle.createUpdate(auditSql)
          .bind("dacId", dacId)
          .bind("ruleId", ruleId)
          .bind("auditUserId", userId)
          .bind("auditType", RuleAuditAction.ADD)
          .bind("actionDate", activationDate)
          .execute();

      return handle.createUpdate(insertRuleSql)
          .bind("dacId", dacId)
          .bind("ruleId", ruleId)
          .bind("userId", userId)
          .executeAndReturnGeneratedKeys("id")
          .mapTo(Integer.class)
          .one();
    });
  }
```

rushtong: I think that's probably a better approach then what we might be doing now if it handles autoCommit better.

s-rubenstein (approved):
This looks great, thank you!

src/test/java/org/broadinstitute/consent/http/db/DACAutomationRuleDAOTest.java
```diff
@@ -53,6 +53,27 @@ void testAuditedInsertDACRuleSetting() {
...
+  void testAuditedInsertDACRuleSettingRollback() {
+    Integer dacId = createRandomDAC();
+    List<DACAutomationRule> rulesByDacId = dacAutomationRuleDAO.findAll();
+    List<DACAutomationRuleAudit> auditRecords = dacAutomationRuleDAO.findAutomationAuditsForDac(
+        dacId, 5, 0
+    );
+    Assertions.assertEquals(0, auditRecords.size());
+
+    // Use -1 userId to force a failure and trigger a rollback
+    Assertions.assertThrows(Exception.class, () ->
```
s-rubenstein: Is it possible to check a more specific exception?

MatthewBemis: Good call! Updated in https://github.com/DataBiosphere/consent/pull/2613/commits/be127931030dc4e457ad69e719610f719d3778ad

otchet-broad (approved):
Nice!

PR #2614, opened 2025-07-21
DT-2000: Optimize DAR Collection Summary Query for DAC Chairs/Members

rushtong:
### Addresses
More completely addresses https://broadworkbench.atlassian.net/browse/DT-2000
See also: https://github.com/DataBiosphere/consent/pull/2612

### Summary
This PR updates the underlying query to populate DAR Collection Summary objects for DAC Chairs and Members. Local testing shows about a 75% performance improvement (e.g. 3.5-4 seconds per Chairperson call down to just under 1 second).

Previously, we were first querying on dataset ids and then subsequently on DARs and related election/vote info. Now we combine the dataset and user role restriction into the single query.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
@@ -22,19 +22,37 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
   @RegisterBeanMapper(value = Election.class)
   @UseRowReducer(DarCollectionSummaryReducer.class)
   @SqlQuery("""
-      SELECT c.collection_id as dar_collection_id, c.dar_code, latest_dar.submission_date, latest_dar.reference_id as latest_dar_reference_id, latest_dar.parent_id as latest_dar_parent_id,
-      latest_dar.closeout_approving_so_id as latest_dar_closeout_approving_so_id,  latest_dar.closeout_so_approval_timestamp AS latest_dar_closeout_so_approval_timestamp,
-       u.display_name as researcher_name, i.institution_name, e.election_id, e.status, e.dataset_id, e.reference_id, v.voteid as v_vote_id, dd.dataset_id as dd_datasetid,
-        v.user_id as v_user_id, v.vote as v_vote, v.electionid as v_election_id, v.createdate as v_create_date, v.updatedate as v_update_date, v.type as v_type,
+      SELECT c.collection_id as dar_collection_id, c.dar_code,
+        latest_dar.submission_date, latest_dar.reference_id as latest_dar_reference_id,
+        latest_dar.parent_id as latest_dar_parent_id,
+        latest_dar.closeout_approving_so_id as latest_dar_closeout_approving_so_id,
+        latest_dar.closeout_so_approval_timestamp AS latest_dar_closeout_so_approval_timestamp,
+        researcher.display_name as researcher_name, i.institution_name,
+        e.election_id, e.status, e.dataset_id, e.reference_id,
+        v.voteid as v_vote_id, dd.dataset_id as dd_datasetid,
+        v.user_id as v_user_id, v.vote as v_vote, v.electionid as v_election_id,
+        v.createdate as v_create_date,v.updatedate as v_update_date, v.type as v_type,
         (regexp_replace(latest_dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'projectTitle' AS name,
         (regexp_replace(latest_dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'status' AS dar_status,
         (regexp_replace(latest_dar.data #>> '{}', '\\\\u0000', '', 'g'))::jsonb ->> 'closeoutSupplement' AS closeout,
         ARRAY_AGG(dar_all.reference_id) AS reference_ids
       FROM dar_collection c
-      INNER JOIN users u
-        ON u.user_id = c.create_user_id
+      -- DAR Collection Researcher join
+      INNER JOIN users researcher
+        ON researcher.user_id = c.create_user_id
       LEFT JOIN institution i
-        ON i.institution_id = u.institution_id
+        ON i.institution_id = researcher.institution_id
+      -- DAC User join
+      INNER JOIN users dacUser
+        ON dacUser.user_id = :currentUserId
+      INNER JOIN user_role ur
+        ON dacUser.user_id = ur.user_id AND ur.role_id = :roleId AND ur.dac_id IS NOT NULL
+      INNER JOIN dac dac
+        ON ur.dac_id = dac.dac_id
+      -- Datasets available to DAC
+      INNER JOIN dataset d
+        ON d.dac_id = dac.dac_id
+      -- Restrict DARs to the most recent submission per collection
       INNER JOIN (
         SELECT DISTINCT ON (collection_id) *
         FROM data_access_request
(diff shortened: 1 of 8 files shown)


src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -22,53 +22,76 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
...
         AND dar_all.submission_date IS NOT NULL
         AND (LOWER(dar_all.data->>'status') != 'archived' OR dar_all.data->>'status' IS NULL)
+      -- Most recent Open and Closed Data Access Elections for DAC User datasets
+      -- Archived, Canceled, and Final elections are not used for status or action calculations
       LEFT JOIN (
         SELECT election.*, MAX(election.election_id) OVER(PARTITION BY election.reference_id, election.dataset_id) AS latest
         FROM election
-        WHERE LOWER(election.election_type) = 'dataaccess' AND election.dataset_id IN (<datasetIds>)
+        WHERE LOWER(election.election_type) = 'dataaccess'
+        AND (LOWER(election.status) = 'open' OR LOWER(election.status) = 'closed')
```
rjohanek, line 74: why do we add this filter for status 'open' and 'closed' - we just never needed/ just ignored this information so it's more efficient to just exclude it in the first place?

rushtong: That is the exact reason. In the status and action calculations, we only consider open and closed elections. All other types are not relevant so this reduces the overall row count and speeds up the reducer processing.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -22,53 +22,76 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
...
       ) AS e
         ON e.reference_id = latest_dar.reference_id
+        AND e.dataset_id = d.dataset_id
+      -- Votes for DAC User
       LEFT JOIN vote v
         ON e.election_id = v.electionid
+        AND (LOWER(v.type) = 'final' OR (v.user_id = :currentUserId OR v.voteid IS NULL))
```
rjohanek: can you explain this - I see it was in the code already, but I don't quite understand why we are collecting these specific votes, specifically why do we want the votes with null voteid? And do we want different votes for different users? like should chair people see all votes while members only see their own or anything like that?

rushtong: You bring up a good point that I completely forgot to address here. This condition, `v.voteid IS NULL` should never happen so I should remove that and double check functionality. What we really want here are the DAC member's votes so we can determine voting action, and we want the final vote to understand what the election status is.

rjohanek: so do we just want the votes of the current user or we want the member votes as well?

rushtong: > so do we just want the votes of the current user or we want the member votes as well?

I'm looking more closely at the summary process functions. I don't think we care about user votes at all so I think we might be able to remove them.

rushtong: > > so do we just want the votes of the current user or we want the member votes as well?
> 
> I'm looking more closely at the summary process functions. I don't think we care about user votes at all so I think we might be able to remove them.

Ah ... we do care about votes for Members. The processing code reviews them for this case and only needs to know if the user has voted or not.

rjohanek: So does this query need to be updated to select all member votes instead of just votes for the current user?

rushtong: > So does this query need to be updated to select all member votes instead of just votes for the current user?

No, we only need the current user votes to know if that user has voted or not. Other user votes are not relevant to what we show in the Chair/Member console.

rjohanek (approved):
thanks for clarifying some things, looks good! I like the comments in the query

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -22,53 +22,76 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
...
+        ON ur.dac_id = dac.dac_id
+      -- Datasets available to DAC
+      INNER JOIN dataset d
+        ON d.dac_id = dac.dac_id
+      -- Restrict DARs to the most recent submission per collection
       INNER JOIN (
         SELECT DISTINCT ON (collection_id) *
         FROM data_access_request
         WHERE submission_date IS NOT NULL
         AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL)
```
otchet-broad, line 60: Since we're not unwinding the json value, is the AND LOWER->>'status' even needed?

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -22,53 +22,76 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
...
-        WHERE LOWER(election.election_type) = 'dataaccess' AND election.dataset_id IN (<datasetIds>)
+        WHERE LOWER(election.election_type) = 'dataaccess'
+        AND (LOWER(election.status) = 'open' OR LOWER(election.status) = 'closed')
       ) AS e
         ON e.reference_id = latest_dar.reference_id
+        AND e.dataset_id = d.dataset_id
+      -- Votes for DAC User
       LEFT JOIN vote v
         ON e.election_id = v.electionid
+        AND (LOWER(v.type) = 'final' OR v.user_id = :currentUserId)
```
otchet-broad, line 81: Index on vote.type?

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
rushtong: Great question ... will dig into that

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
rushtong: Looking into that, thank you!

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
rushtong: We do have a test specifically for the case of ensuring that archived DARs are not in the result set: `testGetDarCollectionSummaryArchivedNotIncluded`. Removing the OR condition triggers a new set of test errors that make me hesitant to introduce a regression here.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
rushtong: We could use one on election type as well - we use both very frequently in queries.

rushtong: Added in latest [commit](https://github.com/DataBiosphere/consent/pull/2614/commits/46882531636bdbafa7c8a4067656fcbc860f9312)

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
otchet-broad: I see why that test passes and why the value works.  We're not doing the encoding when we archive through the DAO (see: archiveByReferenceIds).  That being said, we only archive in tests!  Maybe a follow-up ticket to remove the archive feature and update this query?

otchet-broad (approved):
Nice!

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
rushtong: > That being said, we only archive in tests! Maybe a follow-up ticket to remove the archive feature and update this query?

Yes, it's very likely we've refactored that code path away some time ago. Filing a [ticket](https://broadworkbench.atlassian.net/browse/DT-2020) so we can clean this up.

PR #2615, opened 2025-07-24
[DT-1984] Prevent the removal of the chairperson and member roles through UserService

rjohanek:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-1984

### Summary
The work for this ticket was already done. I just added a test for the case and updated the documentation.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
@@ -300,7 +300,7 @@ public class UserResource extends Resource {
       } else if (activeUser.hasUserRole(UserRoles.SIGNINGOFFICIAL)) {
         if (!UserRoles.isValidSoAdjustableRoleId(targetRole)) {
           throw new ForbiddenException(
-              "A Signing Official may only remove the following role ids: [6, 7, 8] ");
+              "A Signing Official may only remove the following role ids: [7, 8, 9] ");
         }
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -300,7 +300,7 @@ public Response deleteRoleFromUser(@Auth AuthUser authUser, @PathParam("userId")
       } else if (activeUser.hasUserRole(UserRoles.SIGNINGOFFICIAL)) {
```
rjohanek, line 300: see above, line 296 already contains check that this is a non dac role id

MatthewBemis (approved):
👍

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -300,7 +300,7 @@ public Response deleteRoleFromUser(@Auth AuthUser authUser, @PathParam("userId")
...
         if (!UserRoles.isValidSoAdjustableRoleId(targetRole)) {
           throw new ForbiddenException(
-              "A Signing Official may only remove the following role ids: [6, 7, 8] ");
```
rushtong, line 303: Interesting that we don't have a role id of 6 any more ... that might have been the defunct `DATA_OWNER` role.

rushtong (approved):
👍🏽

PR #2624, opened 2025-07-29
[DT-1991] Validate institution names and domains on create and update

MatthewBemis:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1991

### Summary

Adds institution name validation/sanitization:

* Trims whitespace
* Replaces many different types of quotations with straight single quotes
* There was existing validation in place to make sure that names were non-empty and unique -- this is moved into the service layer, but the logic remains the same

Adds institution domain validation:

* Checks validity of domains
* Checks for global uniqueness of domains

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
@@ -9,6 +9,7 @@ import java.util.List;
 import java.util.Objects;
 import org.broadinstitute.consent.http.db.InstitutionDAO;
 import org.broadinstitute.consent.http.db.UserDAO;
+import org.broadinstitute.consent.http.exceptions.ConsentConflictException;
 import org.broadinstitute.consent.http.models.Institution;
 import org.broadinstitute.consent.http.service.UserService.SimplifiedUser;
 import org.broadinstitute.consent.http.util.InstitutionUtil;
@@ -28,6 +29,7 @@ public class InstitutionService {
     checkUserId(userId);
     // Name validation
     checkForEmptyName(institution);
+    checkNameUniqueness(institution);
     String canonicalName = InstitutionUtil.canonicalizeInstitutionName(institution.getName());
     institution.setName(canonicalName);
 
@@ -49,6 +51,7 @@ public class InstitutionService {
 
     // Name validation
     checkForEmptyName(institutionPayload);
+    checkNameUniqueness(institutionPayload);
     String canonicalName = InstitutionUtil.canonicalizeInstitutionName(institutionPayload.getName());
     institutionPayload.setName(canonicalName);
 
@@ -128,6 +131,17 @@ public class InstitutionService {
   private void isInstitutionNull(Institution institution) {
     if (Objects.isNull(institution)) {
       throw new NotFoundException("Institution not found");
+    }
+  }
+
+  private void checkNameUniqueness(Institution institution) {
+    List<Institution> conflicts = findAllInstitutionsByName(institution.getName());
+    // We need to make sure that any found institutions are not the same as the one being updated
+    conflicts.removeIf(existingInstitution ->
+        institution.getId() != null && existingInstitution.getId().equals(institution.getId()));
+    if (!conflicts.isEmpty()) {
+      throw new ConsentConflictException(
+          "An institution exists with the name of '" + institution.getName() + "'");
     }
   }
 
--- a/src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
@@ -25,8 +25,13 @@ public class InstitutionService {
   }
 
   public Institution createInstitution(Institution institution, Integer userId) {
-    checkForEmptyName(institution);
     checkUserId(userId);
+    // Name validation
+    checkForEmptyName(institution);
+    String canonicalName = InstitutionUtil.canonicalizeInstitutionName(institution.getName());
+    institution.setName(canonicalName);
+
+    // Domain validation
     InstitutionUtil.validateInstitutionDomains(institution);
     checkDomainUniqueness(institution);
     try {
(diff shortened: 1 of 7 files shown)


src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
```diff
@@ -35,10 +43,18 @@ public Institution createInstitution(Institution institution, Integer userId) {
...
       Integer userId) throws SQLException {
+    checkUserId(userId);
     Institution targetInstitution = institutionDAO.findInstitutionById(id);
     isInstitutionNull(targetInstitution);
-    checkUserId(userId);
+
+    // Name validation
     checkForEmptyName(institutionPayload);
+    String canonicalName = InstitutionUtil.canonicalizeInstitutionName(institutionPayload.getName());
+    institutionPayload.setName(canonicalName);
```
MatthewBemis, line 56: A side effect of this that we should be aware of:

If any institution in prod already exists with a name that doesn't meet these rules, the name will be automatically updated the next time someone saves any update to that institution. Which is probably what we want, but wanted to call it out.

src/test/java/org/broadinstitute/consent/http/service/InstitutionServiceTest.java
```diff
@@ -191,6 +194,93 @@ void testFindAllInstitutions() {
     assertTrue(service.findAllInstitutions().isEmpty());
   }
 
+  @Test
+  void testCreateInstitutionDomainUniquenessAllUnique() {
+    Institution mockInstitution = initMockModel();
+    mockInstitution.setDomains(List.of("broadinstitute.org", "broad.mit.edu"));
```
otchet-broad: Add testing for IDNs

```suggestion
    mockInstitution.setDomains(List.of("broadinstitute.org", "broad.mit.edu", "café.com"));
```

src/test/java/org/broadinstitute/consent/http/util/InstitutionUtilTest.java
```diff
@@ -55,4 +65,99 @@ void testGsonBuilderNonAdmin() {
...
+  }
+
+  @Test
+  void testGetInvalidInstitutionDomainsMixedValidity() {
+    Institution institution = new Institution();
+    institution.setDomains(Arrays.asList(
+        "broadinstitute.org", // valid
+        "uconn.edu",          // valid
+        "mail.google.com",    // valid
+        "www.uconn.edu",      // valid
```
otchet-broad:
```suggestion
        "www.uconn.edu",      // valid
        "café.com",      //valid
```

src/test/java/org/broadinstitute/consent/http/util/InstitutionUtilTest.java
```diff
@@ -55,4 +65,99 @@ void testGsonBuilderNonAdmin() {
...
+  }
+
+  @Test
+  void testValidateInstitutionDomainsContainsInvalid() {
+    Institution institution = new Institution();
+    institution.setDomains(Arrays.asList(
+        "broadinstitute.org", // valid
+        "uconn.edu",          // valid
+        "sub.example.com",    // valid
+        "www.test.edu",       // valid
```
otchet-broad:
```suggestion
        "www.test.edu",       // valid
        "café.com",     //valid
```

otchet-broad (approved):
Looks good!  Thank you!

I added some small suggestions to include IDNs to domain validation testing.  They should work with the Apache Commons utility this leverages.

src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
```diff
@@ -24,8 +25,15 @@ public InstitutionService(InstitutionDAO institutionDAO, UserDAO userDAO) {
...
-    checkForEmptyName(institution);
     checkUserId(userId);
+    // Name validation
+    checkForEmptyName(institution);
+    String canonicalName = InstitutionUtil.canonicalizeInstitutionName(institution.getName());
+    institution.setName(canonicalName);
+
+    // Domain validation
+    InstitutionUtil.validateInstitutionDomains(institution);
+    checkDomainUniqueness(institution);
```
rjohanek, line 38: Is it weird that we are checking for the name uniqueness in the resource but domain uniqueness in the service? Maybe this logic should live together

src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
```diff
@@ -114,4 +130,29 @@ private void isInstitutionNull(Institution institution) {
...
+      return;
+    }
+
+    List<String> conflictingDomains = institution.getDomains().stream()
+        .map(domain -> {
+          Integer existingInstitutionId = institutionDAO.findInstitutionIdByDomain(domain);
+          if (existingInstitutionId != null && !existingInstitutionId.equals(institution.getId())) {
+            // Return the domain if it conflicts with another institution.
+            // If the domain is already associated with the institution being updated, it's not a conflict.
+            return domain;
```
rjohanek, line 166: comment makes sense, but I'm wondering if this allows for the case where you add the same domain multiple times for the same institution which will cause a db error bc of the uniqueness constraint, and maybe that is fine or maybe we want to do something else

src/test/java/org/broadinstitute/consent/http/service/InstitutionServiceTest.java
MatthewBemis: Good call!

src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
MatthewBemis: That's a great point- it probably makes sense to move the existing name uniqueness check into the service layer. I'll update this.

src/main/java/org/broadinstitute/consent/http/service/InstitutionService.java
MatthewBemis: Good catch- I'm going to add another check in `checkDomainUniqueness` to handle this. We could lean on the DB constraint error to do the lifting here, but it just gives a generic "Database conflict" error message so this would be a UX improvement.

src/test/java/org/broadinstitute/consent/http/util/InstitutionUtilTest.java
```diff
@@ -55,4 +65,101 @@ void testGsonBuilderNonAdmin() {
...
+    Institution institution = new Institution();
+    institution.setDomains(Arrays.asList(
+        "broadinstitute.org", // valid
+        "uconn.edu",          // valid
+        "mail.google.com",    // valid
+        "www.uconn.edu",      // valid
+        "café.com",           // valid
+        "invalid",            // invalid
+        "",                   // invalid
+        null                  // invalid
```
fboulnois: You may want to pull these out into a static variable to avoid repeating it in multiple tests.

fboulnois (approved):
small suggestion:

PR #2626, opened 2025-07-31
DT-2052, DT-2054: Move test only method to test helper

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2052
https://broadworkbench.atlassian.net/browse/DT-2054

### Summary
* `VoteDAO.updateVote` is only used in tests but is necessary as a helper function. This PR moves it out of application code into test infrastructure.
* Includes some sonarqube cleanup to `VoteDAOTest`
* Drive by cleanup to ensure that vote inserts get a default create date
* Make the same changes in `DarCollectionServiceDAO` to be consistent with `VoteDAO`
* Added index on vote create date so the approved datasets is more efficient
* Addresses an issue where approved datasets query was returning incorrect values due to null vote create dates

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/VoteDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/VoteDAO.java
@@ -57,18 +57,6 @@ public interface VoteDAO extends Transactional<VoteDAO> {
   @SqlUpdate("DELETE FROM vote v WHERE electionId IN (SELECT election_id FROM election WHERE reference_id IN (<referenceIds>)) ")
   void deleteVotesByReferenceIds(@BindList(value = "referenceIds", onEmpty = EmptyHandling.NULL_STRING) List<String> referenceIds);
 
-
-  @SqlUpdate("update vote set vote = :vote, updateDate = :updateDate, rationale = :rationale, reminderSent = :reminderSent, createDate = :createDate, has_concerns = :hasConcerns where voteId = :voteId")
-  void updateVote(@Bind("vote") Boolean vote,
-      @Bind("rationale") String rationale,
-      @Bind("updateDate") Date updateDate,
-      @Bind("voteId") Integer voteId,
-      @Bind("reminderSent") boolean reminder,
-      @Bind("electionId") Integer electionId,
-      @Bind("createDate") Date createDate,
-      @Bind("hasConcerns") Boolean hasConcerns);
-
-
   @SqlUpdate("update vote set reminderSent = :reminderSent where voteId = :voteId")
   void updateVoteReminderFlag(@Bind("voteId") Integer voteId,
       @Bind("reminderSent") boolean reminderSent);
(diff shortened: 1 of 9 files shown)


src/main/java/org/broadinstitute/consent/http/db/VoteDAO.java
```diff
@@ -45,7 +44,7 @@ Vote findVoteByElectionIdAndUserId(@Bind("electionId") Integer electionId,
   Integer checkVoteById(@Bind("referenceId") String referenceId,
       @Bind("voteId") Integer voteId);
 
-  @SqlUpdate("insert into vote (user_id, electionId, type, reminderSent) values (:userId, :electionId, :type, false)")
+  @SqlUpdate("INSERT INTO vote (user_id, electionid, type, remindersent, createdate) VALUES (:userId, :electionId, :type, false, current_timestamp)")
```
rushtong, line 47: Note that this was not setting the create date which left some code paths creating votes with null values.

otchet-broad: Null values were in turn the root cause of the issue we were seeing with approvals.
Manually tested with Greg to confirm the fix.

otchet-broad (approved):
Thanks for pairing on this!  Glad we fixed both issues at the same time!

src/main/java/org/broadinstitute/consent/http/db/VoteDAO.java
fboulnois: Can we add a test which covers this issue?

fboulnois (commented):
One comment:

src/main/java/org/broadinstitute/consent/http/db/VoteDAO.java
rushtong: We expanded `VoteDAOTest.testFindVoteById()` to check for the create date on line 53.

src/test/java/org/broadinstitute/consent/http/db/VoteDAOTest.java
```diff
@@ -43,15 +41,19 @@ void testFindVoteById() {
...
         dataset.getDatasetId(), user.getUserId());
     Election election = createDataAccessElection(dar.getReferenceId(), dataset.getDatasetId());
     Vote vote = createDacVote(user.getUserId(), election.getElectionId());
 
     Vote foundVote = voteDAO.findVoteById(vote.getVoteId());
     assertNotNull(foundVote);
+    assertNotNull(foundVote.getCreateDate());
+    assertEquals(user.getUserId(), foundVote.getUserId());
+    assertEquals(election.getElectionId(), foundVote.getElectionId());
+    assertEquals(VoteType.DAC.getValue(), foundVote.getType());
```
rushtong, line 56: Expanded this test to ensure the new insert is populating create date. Note that line 49 (`createDacVote`) calls a helper method that exercises the modified insert statement.

eweitz (approved):

PR #2629, opened 2025-08-04
[DT-2043] Backend API for Voting History Tab

rjohanek:
### Addresses
Need to show all elections not just the latest election for the voting history tab
https://broadworkbench.atlassian.net/browse/DT-2043

### Summary
Add new endpoint at {collectionId}/election history that includes all elections for each dar in the collection
Includes new service layer code and new DAO query that is a modified version of the query for findDARCollectionByCollectionId
Adds unit tests for reosurce, service, and DAO code

Other potential approach: use the existing query and then add on the elections in a separate query similar to how the datasets are added on in addDatasetsToCollection in the DarCollectionService

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
```diff
@@ -203,6 +203,58 @@ List<DarCollection> findDARCollectionByCollectionIds(
...
+  @RegisterBeanMapper(value = LibraryCard.class, prefix = "lc")
+  @UseRowReducer(DarCollectionReducer.class)
+  @SqlQuery(
+      // nosemgrep
+      "SELECT c.*, "
+          + User.QUERY_FIELDS_WITH_U_PREFIX + QUERY_FIELD_SEPARATOR
+          + Institution.QUERY_FIELDS_WITH_I_PREFIX + QUERY_FIELD_SEPARATOR
+          + UserProperty.QUERY_FIELDS_WITH_UP_PREFIX + QUERY_FIELD_SEPARATOR
+          + LibraryCard.QUERY_FIELDS_WITH_LC_PREFIX + QUERY_FIELD_SEPARATOR
+          + "dd.dataset_id, "
```
rushtong: Suggestion: The query string blocks in the model classes are somewhat problematic. I suggest copying the queries and using a text block. That should help with legibility, debugging, and semgrep warnings.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
```diff
@@ -203,6 +203,58 @@ List<DarCollection> findDARCollectionByCollectionIds(
...
+          + "INNER JOIN data_access_request dar ON c.collection_id = dar.collection_id "
+          + "LEFT JOIN dar_dataset dd on dd.reference_id = dar.reference_id "
+          + "LEFT JOIN election e "
+          + "ON (dar.reference_id = e.reference_id AND dd.dataset_id = e.dataset_id) "
+          + "LEFT JOIN vote v "
+          + "ON v.electionid = e.election_id "
+          + "LEFT JOIN users du "
+          + "ON du.user_id = v.user_id "
+          + "WHERE c.collection_id = :collectionId "
+          + "AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL )"
```
rushtong: We have this restriction in a lot of queries, but I'm not sure we want it here. In this use case, I think we might want to see all elections for DARs in a collection regardless of status.

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
```diff
@@ -111,20 +112,39 @@ public Response getCollectionById(
...
       return Response.ok().entity(collection).build();
+    } catch (Exception e) {
+      return createExceptionResponse(e);
+    }
+  }
 
+  @GET
+  @Path("{collectionId}/electionHistory")
+  @Produces("application/json")
+  @PermitAll
```
rushtong: Is this right? In the validate method, we're specifically looking for Admins, Chairs, Members, and SOs.

Edit: I also see that create user can see this - so it looks intended for researchers to see the voting history for their collection. If that's the case, PermitAll is the right choice.

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
```diff
@@ -111,20 +112,39 @@ public Response getCollectionById(
...
+      DarCollection collection = darCollectionService.getCollectionWithAllElectionsByCollectionId(collectionId);
+      validateRequestingUserForDarCollection(authUser, collection);
+      return Response.ok().entity(collection).build();
     } catch (Exception e) {
       return createExceptionResponse(e);
     }
   }
 
+  @VisibleForTesting
+  protected void validateRequestingUserForDarCollection(AuthUser authUser, DarCollection collection) {
```
rushtong: We have a similar methods in `Resource.validateAuthedRoleUser` and `Resource.validateUserHasRoleName` which might simplify this a bit.

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
```diff
@@ -111,20 +112,39 @@ public Response getCollectionById(
...
+      return createExceptionResponse(e);
+    }
+  }
 
+  @GET
+  @Path("{collectionId}/electionHistory")
+  @Produces("application/json")
+  @PermitAll
+  public Response getCollectionWithAllElectionsByCollectionId(
+      @Auth AuthUser authUser,
```
rushtong: If you change this to a `DuosUser`, you'll get a full `User` object that you don't have to re-query for in the validation method

rushtong (commented):
Looks good so far, see comments inline

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
rjohanek: good point, I checked with Victoria and she confirmed that the page should just be for admins chairs and members. She even said researchers should not be able to see individual votes at all meaning we should update our existing collection endpoints so I made this ticket: https://broadworkbench.atlassian.net/browse/DT-2079

src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
rjohanek: I'm hesitant to remove this line because I think we want the results of this query to match the results of the findCollectionById query (with the exception of the additional elections). So, that the same information/DARs in the collection are being shown regardless of who is viewing the darCollection page. If we do want to remove this line, I think we should do it in both places. I'm also a little confused about how the statuses are being used. I think the null status represents a draft DAR which we don't want to include (looking at updateCollectionToDraftStatus in DarCollectionService), and I'm not sure we still use archived status.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
```diff
@@ -203,6 +203,59 @@ List<DarCollection> findDARCollectionByCollectionIds(
...
+    INNER JOIN data_access_request dar ON c.collection_id = dar.collection_id
+    LEFT JOIN dar_dataset dd on dd.reference_id = dar.reference_id
+    LEFT JOIN election e
+      ON (dar.reference_id = e.reference_id AND dd.dataset_id = e.dataset_id)
+    LEFT JOIN vote v
+      ON v.electionid = e.election_id
+    LEFT JOIN users du
+      ON du.user_id = v.user_id
+    WHERE c.collection_id = :collectionId
+    AND (LOWER(data->>'status') != 'archived' OR data->>'status' IS NULL )
```
rushtong, line 254: I had previously commented on this line. I think it's fine to leave it here - we should look into deprecating all usages of this field since we are not setting the value anywhere.

src/test/java/org/broadinstitute/consent/http/resources/DarCollectionResourceTest.java
```diff
@@ -268,6 +272,101 @@ void testGetCollectionByIdMultipleRoles() {
...
+  void testGetCollectionWithAllElectionsByCollectionId() {
+    DarCollection collection = mockDarCollection();
+    User admin = new User(2, authUser.getEmail(), "Admin User", new Date(),
+        List.of(UserRoles.Admin()));
+    DuosUser duosAdmin = new DuosUser(authUser, admin);
+    collection.setCreateUser(researcher);
+    collection.setCreateUserId(researcher.getUserId());
+
+    when(darCollectionService.getCollectionWithAllElectionsByCollectionId(collection.getDarCollectionId())).thenReturn(collection);
+    initResource();
```
rushtong: We should have made this a `@BeforeEach` so every test doesn't need to call it.

rushtong (approved):
Looks good, thank you 👍🏽

src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
otchet-broad: Already noted in [https://broadworkbench.atlassian.net/browse/DT-1563](https://broadworkbench.atlassian.net/browse/DT-1563) and in the current sprint!  Huzzah!

src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
```diff
@@ -203,6 +203,59 @@ List<DarCollection> findDARCollectionByCollectionIds(
...
+"""
+  )
+  DarCollection findCollectionWithAllElectionsByCollectionId(@Bind("collectionId") Integer collectionId);
```
otchet-broad, line 257: Not part of my expectations for this PR, but one of the more interesting experiments to do is to see what performance is like and if we should retire the endpoint that only returns the latest election.  It would allow us to simplify our backend code.

otchet-broad (approved):
Nice work @rjohanek!

src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
rjohanek: I just did a little test with two collections I have in my local database but it will be better to test on dev where collections have lots of progress reports and elections. But, for what its worth, these are the results: 

electionHistory vs getCollectionById
Collection 605
3962 ms vs 734 ms
567 ms vs 950 ms

Collection 604
500 ms vs 759 ms
487 ms vs 988 ms

I think the slow first API call was a result of being the first call I made after authing in swagger and maybe is not reflective of that specific API. If we do find performance is a concern, Florian had an idea of adding the elections on in a separate query much like we do for datasets.

PR #2631, opened 2025-08-05
[DT-2028] Address TODO in DacAutomationRuleResource.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2028](https://broadworkbench.atlassian.net/browse/DT-2028)

### Summary

Uses already tested feature in User class to determine if the user is a chair of the specified DAC.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DACAutomationRuleResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DACAutomationRuleResource.java
@@ -118,7 +118,6 @@ public class DACAutomationRuleResource extends Resource {
   }
 
   private void validateAdminOrChairForDAC(User user, Integer dacId) {
-    //TODO: harmonize this with the logic in the DACResource so we're consistent.
     boolean isAdminOrChair = user.hasUserRole(UserRoles.ADMIN) || isChairOfDAC(user, dacId);
     if (!isAdminOrChair) {
       throw new ForbiddenException("User does not have access to the specified DAC ID");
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DACAutomationRuleResource.java
```diff
@@ -118,15 +118,16 @@ public Response toggleRule(@Auth AuthUser authUser, @PathParam("dacId") Integer
...
+        user.hasUserRole(UserRoles.ADMIN)
+            || user.checkIfUserHasRole(UserRoles.CHAIRPERSON.getRoleName(), dacId);
     if (!isAdminOrChair) {
       throw new ForbiddenException("User does not have access to the specified DAC ID");
     }
   }
 
   private void validateIsChairOfDAC(User user, Integer dacId) {
-    if (!isChairOfDAC(user, dacId)) {
+    if (Boolean.FALSE.equals(user.checkIfUserHasRole(UserRoles.CHAIRPERSON.getRoleName(), dacId))) {
```
rushtong: Not your code, but I wanted to call this out perhaps for a future refactor. The method name, `checkIfUserHasRole`, is somewhat misleading. There are only two roles that have dac ids, so it could be misused to check for admin with a dac id, or SO with a dac id, etc., and never return anything. Perhaps `verifyDACRole`, or something similar.

rushtong (approved):
Looks good, thank you 👍🏽

src/main/java/org/broadinstitute/consent/http/resources/DACAutomationRuleResource.java
fboulnois: Instead of `Boolean.FALSE.equals`, consider switching `checkIfUserHasRole` to return a `boolean`, and then you can do `!user.checkIfUserHasRole`

fboulnois (commented):
Suggestion:

src/main/java/org/broadinstitute/consent/http/resources/DACAutomationRuleResource.java
otchet-broad: I looked at doing that initially and wanted to refactor the underlying method a bit.  I'm going to do that as part of this PR since I'll be in the method signature.

otchet-broad: Realized this as well.  Since I'm going to change the method signature, I'm going to rename it now.

eweitz (approved):

rushtong (approved):
Thank you for the updates 👍🏽

PR #2633, opened 2025-08-08
DT-1563: Update how we're storing and retrieving jsonb data in Data Access Requests

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-1563

## Summary
This PR updates how we're storing jsonb data for `DataAccessRequestData` objects. Previously, we were doing two things that made the underlying data difficult to work with:
* Using `to_jsonb` in inserts/updates. This had the effect of double encoding the json being passed into the sql
* Using `regexp_replace` when querying for fields due to unsupported characters in the data. This made deserialization complicated.

Instead, we take a slightly different approach now.
* Prevent double encoding by removing `to_jsonb` usages.
* Prevent unsupported characters on insert/update. This makes it possible to query the json data more naturally using standard postgres json operators.

Additional work includes:
* Updating all queries on jsonb fields to use standard postgres json operators
* Adding indexes for the fields we commonly query on

As of 8/11, tested against latest dev, staging, and prod databases.

### Examples
You can see the before and after in these db views. In both cases, the `data` column now contains clean data and the `data_bak` column contains the original value for reference. A future PR/story will remove this column.
Screenshot 2025-08-08 at 12 31 18 PM

Note that now we can use standard postgres json operators on the raw `data` column:
Screenshot 2025-08-08 at 12 31 09 PM

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
@@ -137,7 +137,7 @@ public interface DarCollectionDAO extends Transactional<DarCollectionDAO> {
         dar.closeout_so_approval_timestamp AS dar_closeout_signing_official_approved_date,
         dar.closeout_approving_so_id AS dar_closeout_signing_official_approved_user_id,
         dar.create_date AS dar_create_date, dar.sort_date AS dar_sort_date, dar.submission_date AS dar_submission_date,
-        dar.update_date AS dar_update_date, dar.data as data, dd.dataset_id
+        dar.update_date AS dar_update_date, dar.data AS data, dd.dataset_id
         FROM dar_collection c
         INNER JOIN users u ON c.create_user_id = u.user_id
         LEFT JOIN user_property up ON u.user_id = up.user_id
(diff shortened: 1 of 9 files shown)


kevinmarete (commented):
Looks good, but I was surprised not to see any changes to existing tests or there isn't  any coverage?

rushtong:
> Looks good, but I was surprised not to see any changes to existing tests or there isn't any coverage?

Thank you. There is quite a bit of coverage, especially in our DAO layer where every query has at least one test. One of my goals here was to minimize the impact across the codebase.

kevinmarete (approved):

fboulnois (commented):
What happens to the existing invalid data in the db?

rushtong:
> What happens to the existing invalid data in the db?

@fboulnois It's [copied into a backup column](https://github.com/DataBiosphere/consent/pull/2633/files#diff-8655a45c74df671386f5ccb997ec733175437db0817547dcaba00f65a4587967R9) for now. I imagine that in a future PR, we'd drop it once we're comfortable that we don't need it any longer

fboulnois (approved):
👍

PR #2636, opened 2025-08-11
DT-2091: Update DAC Approve Dataset performance

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-2091

## Summary
There are two main parts to this PR.

1. The current find dataset by id query is very expensive. The query itself is not great, and the assembly of the dataset object in the mapper/reducer makes this much worse. The big slowdown relates to `FileStorageObject` so this PR creates a new DAO method to get the dataset/study info w/out FSOs.

2. When we update any aspect of a dataset, we need to publish it to ES. This is currently run in a blocking call which slows down the response to the user. To speed this up, we create an async version of the same call that completes in a separate thread.

### Other details:
In the affected API call, we were indexing the dataset twice, once in the Approve Dataset service layer code path and then again at the Resource layer code path. Instead, we just remove the second one in favor of the first.

The `DacResource` and `DacResourceTest` classes were pretty out of date. While here, we bring them up to current standards.

Refactored how we're using `ExecutorService` by pulling it into a new class, `ThreadUtils`, and adding shutdown hooks whenever it is called.

--- 

### DB Performance
Note, for example, the difference in the explain plan for a dataset in a study that has a lot of other datasets in it. On the right, we have the original query with FSO joins. On the left, we have the same query without FSO information. There is about a 10x difference in timing. This improves API performance by about 3x:

**API timing for Approving Dataset ID 2560 in DAC 3:**
DEV:  `24905 ms`
LOCAL: `7151 ms`

Screenshot 2025-08-11 at 2 45 49 PM

---

### Elastic Search Indexing Improvements
In the logs, it is clear that the response is returned to the user before the ES indexing completes:

```
proxy    | [Mon Aug 11 19:14:34.884257 2025] [deflate:debug] [pid 44:tid 140735324083968] mod_deflate.c(856): [client 192.168.65.1:19302] AH01384: Zlib: Compressed 6046 to 2275 : URL /api/dac/3/dataset/2560, referer: https://local.dsde-dev.broadinstitute.org:27443/
consent  | 172.18.0.3 - - [11/Aug/2025:19:14:32 +0000] "PUT /api/dac/3/dataset/2560 HTTP/1.1" 200 6046 "https://local.dsde-dev.broadinstitute.org:27443/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/139.0.0.0 Safari/537.36" 2616
proxy    | 192.168.65.1 - 103740509117808340318 "grushton@broadinstitute.org" "-" [2025-08-11T19:14:32.261+0000] 2623278 + "PUT /api/dac/3/dataset/2560 HTTP/1.1" 200 2293 "https://local.dsde-dev.broadinstitute.org:27443/" "https://local.dsde-dev.broadinstitute.org:27443" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/139.0.0.0 Safari/537.36"
proxy    | [Mon Aug 11 19:14:34.884922 2025] [oauth2:debug] [pid 44:tid 140735324083968] src/server/apache.c(332): [client 192.168.65.1:19302] oauth2_apache_request_context_free: dispose request context: 0x7fff18012980, referer: https://local.dsde-dev.broadinstitute.org:27443/
consent  | INFO  [2025-08-11 19:14:37,174] org.broadinstitute.consent.http.service.ElasticSearchService: Successfully synchronized dataset in ES index: DUOS-001058
```

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/resources/DacResource.java
```diff
@@ -273,13 +272,6 @@ public Response approveDataset(@Auth AuthUser authUser, @PathParam("dacId") Inte
         throw new BadRequestException("Invalid request payload");
       }
       Dataset updatedDataset = datasetService.approveDataset(dataset, user, payload.getApproval());
-      try (Response indexResponse = elasticSearchService.indexDataset(updatedDataset.getDatasetId(), user))  {
-        if (indexResponse.getStatus() >= Status.BAD_REQUEST.getStatusCode()) {
-          logWarn("Non-OK response when reindexing dataset with id: " + datasetId);
-        }
-      } catch (Exception e) {
-        logException("Exception re-indexing datasets from dataset id: " + datasetId, e);
-      }
```
rushtong, line 282: This is also happening in `datasetService.approveDataset` so we don't need it here.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -74,6 +82,7 @@ public ElasticSearchService(
       DatasetServiceDAO datasetServiceDAO,
       StudyDAO studyDAO,
       LibraryCardDAO libraryCardDAO) {
+    this.executorService = Executors.newCachedThreadPool();
```
otchet-broad: CachedThreadPool and system performance can suffer from resource exhaustion if too many threads are created.

```suggestion
    this.executorService = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors());
```

otchet-broad: This code should also probably hook JVM Shutdown and decide what to do with threads that are running (if you don't take the suggestion above) or are running/queued if you do.  :-)

otchet-broad (changes requested):
Overall, makes sense and the DatasetDAO optimizations seem reasonable.

The ElasticSearchService should be updated to handle graceful shutdowns with the Executor.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
rushtong: I lifted this from `SamDAO` so will make the same change there.

src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
```diff
@@ -44,7 +44,7 @@ public class SamDAO implements ConsentLogger {
   public final Integer readTimeoutMilliseconds;
 
   public SamDAO(HttpClientUtil clientUtil, ServicesConfiguration configuration) {
-    this.executorService = Executors.newCachedThreadPool();
+    this.executorService = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors());
```
rushtong: Applied the same feedback here as was requested at https://github.com/DataBiosphere/consent/pull/2636/files#diff-29eb97de32bb49020551ab465ca025fa23206d83ceb762a456e4f64d3524ff16R85

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -74,6 +82,7 @@ public ElasticSearchService(
       DatasetServiceDAO datasetServiceDAO,
       StudyDAO studyDAO,
       LibraryCardDAO libraryCardDAO) {
+    this.executorService = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors());
```
rushtong: FYI @otchet-broad - LMK if you're not seeing this.

otchet-broad: Thank you.  I see that.  What I think you also want is a hook to tell this.executorService to shut down.

[https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html#shutdownNow--](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html#shutdownNow--)

Pseudocode below:

Thread shutdownThread = new Thread(() -> {
try {
    this.executorSerivce.shutdownNow()
catch (Exception e) {
// log the exception but do not rethrow because the JVM is shutting down.
}
 );
Runtime.getRuntime().addShutdownHook(shutdownThread);

otchet-broad (approved):
Nice!

Thanks for the new Utils method and for adding the shutdown hooks to clean up the Executors.

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -44,12 +50,15 @@
...
 import org.broadinstitute.consent.http.service.dao.DatasetServiceDAO;
 import org.broadinstitute.consent.http.util.ConsentLogger;
+import org.broadinstitute.consent.http.util.ThreadUtils;
 import org.broadinstitute.consent.http.util.gson.GsonUtil;
 import org.elasticsearch.client.Request;
 import org.elasticsearch.client.RestClient;
 
 public class ElasticSearchService implements ConsentLogger {
 
+  private final ExecutorService executorService = new ThreadUtils().getExecutorService(
```
fboulnois, line 60: Do we need an entire executor service to run this task instead of a simple `Thread`?

fboulnois (approved):
One main question around the implementation of the async indexing:

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
rushtong: This is refactored from `SamDAO` which is under heavy load fairly consistently which is the primary reason we went with `ExecutorService`. `ElasticSearchService`, too, is heavily used with every dataset update, DAR approval, DAR expiration, etc., so I think it makes sense to stick with the current pattern.

PR #2637, opened 2025-08-11
DT-2101: Restrict study API to publicly visible/create user

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2101

### Summary
If a study is not publicly visible, check to see if the authenticated user is the create user and only return if that is true.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/StudyResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/StudyResource.java
@@ -76,7 +76,7 @@ public class StudyResource extends Resource {
   @Produces(MediaType.APPLICATION_JSON)
   @RolesAllowed({ADMIN})
   public Response convertToStudy(@Auth AuthUser authUser,
-    @PathParam("datasetIdentifier") String datasetIdentifier, String json) {
+      @PathParam("datasetIdentifier") String datasetIdentifier, String json) {
     try {
       User user = userService.findUserByEmail(authUser.getEmail());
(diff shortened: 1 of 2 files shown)


rjohanek (approved):
looks good thanks

otchet-broad (approved):

src/main/java/org/broadinstitute/consent/http/resources/StudyResource.java
```diff
@@ -178,10 +181,11 @@ public Response deleteStudyById(@Auth AuthUser authUser, @PathParam("studyId") I
   @Produces(MediaType.APPLICATION_JSON)
   @PermitAll
   @Timed
-  public Response getRegistrationFromStudy(@Auth AuthUser authUser,
+  public Response getRegistrationFromStudy(@Auth DuosUser duosUser,
```
fboulnois, line 184: Can you describe the difference here?

fboulnois (approved):
Looks good, one question for my knowledge:

src/main/java/org/broadinstitute/consent/http/resources/StudyResource.java
rushtong: `AuthUser` doesn't have a `User` object but `DuosUser` does. Since we needed it for the `checkPublicVisibilityForUser` method, and we weren't using `authUser`, I chose to switch it out here. If I didn't do that, I'd need to look up the user through `UserService` so this just saves us a method call.

PR #2640, opened 2025-08-12
DT-1636: Sync User Linked Account Status with ECM

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-1636

## Summary
New API to synchronize a user's RAS state in Consent with what is known in ECM. This returns a full user object with properties that will include the updated values from ECM. NOTE: This syncs in both directions with ECM so that a linked account can be created and removed in ECM. Since DUOS and Terra both use ECM as a source of truth, a user link/de-link in Terra affects DUOS and vice-versa.

For example (actual values redacted for demonstration):

```
{
  "userId": 1,
  "email": "my email",
  "displayName": "my display name",
  "createDate": 1,
  "roles": [...],
  "properties": [
    {
      "propertyId": 1,
      "propertyKey": "eraAuthorized",
      "propertyValue": "<Synced value from ECM>"
    },
    {
      "propertyId": 2,
      "propertyKey": "eraExpiration",
      "propertyValue": "<Synced value from ECM>"
    }
  ],
  "emailPreference": true,
  "institutionId": 1,
  "eraCommonsId": "<Synced value from ECM>",
  "institution": {...},
  "libraryCard": {...}
}
```

This is the required information that the UI needs to do any work necessary to update a user's RAS display status.

Under the hood, we re-purpose the [ECM `LinkInfo`](https://externalcreds.dsde-prod.broadinstitute.org/#/oauth/getLink) object to how we've been storing this [locally in `NIHUserAccount`](https://github.com/DataBiosphere/consent/blob/fde9d35395dd9ddbfa086f4709dbd391e25c341d/src/main/java/org/broadinstitute/consent/http/models/NIHUserAccount.java) so all existing functionality that relied on the existing data should be preserved without need for changes.

This PR also fixes a bug in the delete NIH Account Information code path. Previously, if a user chose to delete their linked account, we only deleted the properties, not the 

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/NihService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/NihService.java
@@ -1,14 +1,19 @@
 package org.broadinstitute.consent.http.service;
 
+import com.google.api.client.http.GenericUrl;
+import com.google.api.client.http.HttpRequest;
+import com.google.api.client.http.HttpResponse;
 import com.google.inject.Inject;
 import jakarta.ws.rs.BadRequestException;
 import jakarta.ws.rs.NotFoundException;
+import jakarta.ws.rs.ServerErrorException;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.List;
 import java.util.Objects;
 import org.apache.commons.lang3.StringUtils;
+import org.broadinstitute.consent.http.configurations.ServicesConfiguration;
 import org.broadinstitute.consent.http.db.UserDAO;
 import org.broadinstitute.consent.http.db.UserPropertyDAO;
 import org.broadinstitute.consent.http.enumeration.UserFields;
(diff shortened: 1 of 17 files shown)


src/main/java/org/broadinstitute/consent/http/service/dao/NihServiceDAO.java
```diff
@@ -42,4 +42,17 @@ public void updateUserNihStatus(User user, NIHUserAccount nihAccount)
...
+
+  public void deleteNihAccountById(Integer userId) {
+    jdbi.useTransaction(handler -> {
+      UserDAO userDAO = handler.attach(UserDAO.class);
+      UserPropertyDAO userPropertyDAO = handler.attach(UserPropertyDAO.class);
+      Collection<UserProperty> properties = List.of(
+          new UserProperty(userId, UserFields.ERA_EXPIRATION_DATE.getValue()),
+          new UserProperty(userId, UserFields.ERA_STATUS.getValue())
+      );
+      userDAO.updateEraCommonsId(userId, null);
```
rushtong, line 54: This is a bug fix. Previously, we were not deleting this value when a user tried to delete their eRA Commons/RAS account link.

src/main/java/org/broadinstitute/consent/http/service/NihService.java
```diff
@@ -1,36 +1,67 @@
...
+      HttpClientUtil clientUtil, ServicesConfiguration configuration) {
     this.userDAO = userDAO;
     this.userPropertyDAO = userPropertyDAO;
     this.serviceDAO = serviceDAO;
+    this.clientUtil = clientUtil;
+    this.configuration = configuration;
+  }
+
+  public User syncAccount(AuthUser authUser) throws Exception {
+    User user = userDAO.findUserByEmail(authUser.getEmail());
```
rjohanek: can we make this a duos user to avoid having to findByEmail

src/main/java/org/broadinstitute/consent/http/service/NihService.java
```diff
@@ -65,14 +96,26 @@ public List<UserProperty> authenticateNih(NIHUserAccount nihAccount, AuthUser au
...
-    properties.add(new UserProperty(userId, UserFields.ERA_EXPIRATION_DATE.getValue()));
-    properties.add(new UserProperty(userId, UserFields.ERA_STATUS.getValue()));
-    if (userDAO.findUserById(userId) == null) {
-      throw new NotFoundException("User with id: " + userId + " does not exist");
+  public void deleteNihAccountById(AuthUser authUser) {
+    User user = userDAO.findUserByEmail(authUser.getEmail());
+    // Delete linkage locally
+    serviceDAO.deleteNihAccountById(user.getUserId());
+    try {
+      // Delete linkage from ECM
```
rjohanek, line 105: why the change to this delete method? why do we want to delete from ECM?

src/main/java/org/broadinstitute/consent/http/service/NihService.java
rushtong: I think we can ... thank you. The underlying call needs an `AuthUser` to make API calls to ECM, so that could work.

src/main/java/org/broadinstitute/consent/http/service/NihService.java
rushtong: This service call is triggered when a user manually chooses to remove their linked account. Since both Terra and DUOS now use ECM as the source of truth, we need to ensure that whatever a user does in DUOS is also reflected in Terra and vice versa.

rjohanek: I wonder if a user who unlinks their account on DUOS would expect it unlink their account on Terra as well.

rushtong: > I wonder if a user who unlinks their account on DUOS would expect it unlink their account on Terra as well.

I am hoping to add some UI text that will help users understand that detail in https://github.com/DataBiosphere/duos-ui/pull/3014

rjohanek: Okay if this is doable in terra too we should add some text there too

rushtong: > Okay if this is doable in terra too we should add some text there too

I'll bring this subject up with the RAS folks. Their adoption of RAS is not on the same timetable as ours is.

rjohanek (approved):
I'm good with this as long as Victoria is good with this (unlinking for both terra and duos)

otchet-broad:
@rushtong Do we need to clear out all of the values stored since they might not be from ECM (or valid)

src/main/resources/assets/paths/nih.yaml
```diff
@@ -1,14 +1,14 @@
...
+  description: Removes Linked Account information for the authenticated user
   tags:
     - User
-    - eRA Commons
     - Nih
   responses:
     200:
       description: Returns an empty response if the operation was successful
 post:
+  deprecated: true
```
otchet-broad, line 11: Doesn't this endpoint need to go away entirely?

rushtong: When we completely move to RAS, yes. If we remove it now, the prod eRA Commons auth path will be broken until the UI part of this effort is complete. See https://broadworkbench.atlassian.net/browse/DT-2116 for that work.

rushtong:
> @rushtong Do we need to clear out all of the values stored since they might not be from ECM (or valid)

@otchet-broad - I think that is a separate effort that needs to be coordinated with the final enabling of RAS in production. I can file a ticket for this: https://broadworkbench.atlassian.net/browse/DT-2129

otchet-broad (approved):
Nice work!

Agree with @rjohanek about wanting product to confirm the delete behavior across apps.

rushtong:
This approach has been discussed and approved with Product and AppSec.

PR #2652, opened 2025-08-22
DT-2105: When fetching dataset/study, verify public visibility with create user or custodian status

rushtong:
### Addresses
Partially addresses [DT-2105](https://broadworkbench.atlassian.net/browse/DT-2105)

### Summary
This PR modifies the following list of APIs such that we check for two conditions before returning the requested dataset(s):
1. Admins are not restricted by `publicVisibility` and functionality available to admins is not changed.
2. We check that the study property of `publicVisibility` is **NOT** set to **FALSE**. Previously, datasets and studies in this category were being returned to any user regardless of this value.
3. If `publicVisibility` is **FALSE**, then we further check if the user is a create user for either dataset or study, or a data custodian for the study.

The following APIs are currently addressed in this PR:

- `GET /api/tdr/{identifier}`
- `GET /api/dataset/v2/` - Removed: not used in the UI and not necessary for API users
- `GET /api/dataset/v2/{datasetId}`
- `PUT /api/dataset/v3/{datasetId}`
- `GET api/dataset/batch`
- `GET /api/dataset/registration/{datasetIdentifier}`
- `PUT /api/study/convert/{datasetIdentifier}`
- `GET /api/tdr/{identifier}/approved/users`
- `PATCH /api/dataset/{datasetId}`
- `DELETE /api/dataset/{datasetId}`
- `PUT /api/dataset/{id}/datause`
- `GET /api/dataset/study/{studyId}`
- `GET /api/dataset/study/registration/{studyId}`

These APIs are mentioned in the ticket, but are not addressed with this PR:

- `GET api/dataset/v3`
  - Not used by the UI, but used by data submitters to review existing dataset/study summary information. I propose that we keep this unchanged since it is (currently) necessary to maintain dataset and study name uniqueness.
- `GET api/dataset/studyNames`
  - Used for validation. [DT-2141](https://broadworkbench.atlassian.net/browse/DT-2141) will update this since we can use `GET api/dataset/v3` for this 
- `GET api/dataset/datasetNames`
  - Used for validation. [DT-2141](https://broadworkbench.atlassian.net/browse/DT-2141) will update this since we can use `GET api/dataset/v3` for this 
- `GET api/dataset/autocomplete`
  - Currently used for validation. [DT-2142](https://broadworkbench.atlassian.net/browse/DT-2142) will remove the need for this altogether.

Follow on work:

- `PUT /api/dataset/study/convert/{datasetIdentifier}` is a candidate for removal. All production datasets have studies. This is admin-only, so there is no risk of unintended data leakage. See [DT-2146](https://broadworkbench.atlassian.net/browse/DT-2146) for more details.

----

- Verify all tests go green

[DT-2105]: https://broadworkbench.atlassian.net/browse/DT-2105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DT-2141]: https://broadworkbench.atlassian.net/browse/DT-2141?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -100,7 +99,61 @@ public Dataset findDatasetByIdentifier(String datasetIdentifier) throws IllegalA
...
+
+  protected Dataset verifyPublicVisibilityAccess(Dataset dataset, User user) {
+    // Admins
+    if (user.hasUserRole(UserRoles.ADMIN)) {
+      return dataset;
+    }
+    // If there is no study, we can't verify visibility, so return the dataset
+    if (dataset.getStudyId() == null) {
+      return dataset;
+    }
```
eweitz, line 113: Can you help me understand why we return `dataset` and not `null` here?

In other words, why is it the case that if a dataset has no studies, then it's best to have it publicly visible by default?

This seems like it might be inconsistent with a check in `isCreatorOrCustodian`:
```
    if (dataset.getStudy() == null) {
      return false;
    }
```

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -116,17 +169,14 @@ public List<String> findAllDatasetNames() {
...
+  public Dataset findDatasetById(User user, Integer id) {
+    Dataset dataset = datasetDAO.findDatasetById(id);
+    return verifyPublicVisibilityAccess(dataset, user);
   }
 
   /**
-   * Find the dataset without files by its ID. This method is intended to return a minimal
-   * dataset for performance reasons, avoiding the retrieval of full FSO information.
+   * Find the dataset without files by its ID. This method is intended to return a minimal dataset
+   * for performance reasons, avoiding the retrieval of full FSO information.
```
eweitz, line 180: Note to self and other newer maintainers: 

"FSO" means FileStorageObject

(I don't think this necessarily warrants expanding in the code comment; it takes < 20 seconds to look up.)

eweitz (commented):
Nice substantial update!  

I inquire below about some logic in `verifyPublicVisibilityAccess` -- see "Can you help me understand".  I'm not confident if there's a lurking bug, but it seems plausible.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
rushtong: The `publicVisibility` property only exists on studies, not datasets. I chose to default to the current behavior of returning the dataset as opposed to changing functionality when there is no dataset-level property that would indicate that it should be private.

rushtong: > This seems like it might be inconsistent with a check in `isCreatorOrCustodian`:
> 
> ```
>     if (dataset.getStudy() == null) {
>       return false;
>     }
> ```

Let me look at that again - I may have missed something while refactoring this multiple times

rushtong: @eweitz - PTAL ... that null check was in the wrong place, thanks for the callout!

eweitz (approved):
Thanks for the extra context and refinement!

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -100,7 +99,62 @@ public Dataset findDatasetByIdentifier(String datasetIdentifier) throws IllegalA
...
+    // If there is no study, we can't verify visibility, so return the dataset
+    if (dataset.getStudyId() == null) {
+      return dataset;
+    }
+    // Study isn't always populated, so fetch it if necessary
+    if (dataset.getStudy() == null) {
+      dataset.setStudy(studyDAO.findStudyById(dataset.getStudyId()));
+    }
+    // If not visible, check that the user is authorized to see it
+    if (Boolean.FALSE.equals(dataset.getStudy().getPublicVisibility())) {
```
otchet-broad, line 119: @rushtong Can publicVisibility be mutated?

otchet-broad: If so, should we remove the study from Elastic if the value switches from ON -> OFF?

rushtong: > If so, should we remove the study from Elastic if the value switches from ON -> OFF?

Yes, study updates happen here: https://github.com/DataBiosphere/consent/blob/6c46cbeb42732a1b994e75ee50c00cab24e17594/src/main/java/org/broadinstitute/consent/http/resources/StudyResource.java#L210C19-L210C44

That code re-indexes the study here: https://github.com/DataBiosphere/consent/blob/6c46cbeb42732a1b994e75ee50c00cab24e17594/src/main/java/org/broadinstitute/consent/http/resources/StudyResource.java#L233

otchet-broad: We have a parking lot item to discuss this, but that doesn't need to stop this work from landing.

otchet-broad (approved):
Nice!
👍

PR #2653, opened 2025-08-26
[DT-2128] Fix electionHistory endpoint to only show DAC 1 elections for requesting user in DAC 1

rjohanek:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2128
Drive by fix: https://broadworkbench.atlassian.net/browse/DT-2151

### Summary
getCollectionWithAllElectionsByCollectionId endpoint
The modified endpoint is for Admins, Chairs, and Members only. The functionality for Admins remains unchanged. For DAC members and chairs, the endpoint is modified to filter elections by DAC (by datasets that belong to the DACs the user is a part of).
This involves a new DAO query and service method with unit tests.

Also fixes serialization issue with dataset create date. In the mapper it was being retrieved as a SQL date instead of a timestamp to be converted into a java util date. 

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetMapper.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetMapper.java
@@ -22,7 +22,7 @@ public class DatasetMapper implements RowMapper<Dataset>, RowMapperHelper {
     dataset.setObjectId(r.getString("object_id"));
     dataset.setName(r.getString("name"));
     if (hasColumn(r, "create_date")) {
-      dataset.setCreateDate(r.getDate("create_date"));
+      dataset.setCreateDate(r.getTimestamp("create_date"));
     }
     if (hasNonZeroColumn(r, "create_user_id")) {
       dataset.setCreateUserId(r.getInt("create_user_id"));
--- a/src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
@@ -9,7 +9,6 @@ import jakarta.annotation.security.PermitAll;
 import jakarta.annotation.security.RolesAllowed;
 import jakarta.ws.rs.BadRequestException;
 import jakarta.ws.rs.Consumes;
-import jakarta.ws.rs.DELETE;
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.NotFoundException;
 import jakarta.ws.rs.POST;
@@ -31,6 +30,7 @@ import org.broadinstitute.consent.http.models.DarCollectionSummary;
 import org.broadinstitute.consent.http.models.Dataset;
 import org.broadinstitute.consent.http.models.DuosUser;
 import org.broadinstitute.consent.http.models.User;
+import org.broadinstitute.consent.http.models.UserRole;
 import org.broadinstitute.consent.http.service.DarCollectionService;
 import org.broadinstitute.consent.http.service.UserService;
 import org.broadinstitute.consent.http.util.ComplianceLogger;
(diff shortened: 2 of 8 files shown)


src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetMapper.java
```diff
@@ -22,7 +22,7 @@ public Dataset map(ResultSet r, StatementContext ctx) throws SQLException {
     dataset.setObjectId(r.getString("object_id"));
     dataset.setName(r.getString("name"));
     if (hasColumn(r, "create_date")) {
-      dataset.setCreateDate(r.getDate("create_date"));
+      dataset.setCreateDate(r.getTimestamp("create_date"));
```
rushtong, line 25: Nice catch, that column is, indeed, a timestamp

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
```diff
@@ -135,8 +135,16 @@ public Response getCollectionWithAllElectionsByCollectionId(
       @Auth DuosUser authUser,
       @PathParam("collectionId") Integer collectionId) {
     try {
-      DarCollection collection = darCollectionService.getCollectionWithAllElectionsByCollectionId(collectionId);
-      validateRequestingUserForElectionHistory(authUser, collection);
+      List<UserRole> roles = authUser.getUser().getRoles();
+      boolean isAdmin = roles.stream().anyMatch(r -> r.getName().equals(UserRoles.ADMIN.name()));
```
rushtong: User has a convenience method, `hasUserRole` that does all of this work for you

src/main/java/org/broadinstitute/consent/http/db/mapper/DatasetMapper.java
otchet-broad: It's interesting how this bug appeared.  It was in the serialization/deserialization logic for the TypeAdapter not getting invoked.  Sure enough, we were setting the subtype java.sql.Date and the GsonBuilder TypeAdapter class matcher (as set) won't match the subclass.  It took a bit to unwind, but I think it was worth the effort.
image

Our frontend tests seemed to have modeled this as millis since epoc
image


I'll make a ticket for our frontend model because it should probably be cleaned up:
image

otchet-broad: [https://broadworkbench.atlassian.net/browse/DT-2153](https://broadworkbench.atlassian.net/browse/DT-2153)

rjohanek: thanks Elliot!

src/test/java/org/broadinstitute/consent/http/service/ElasticSearchServiceTest.java
```diff
@@ -264,7 +265,7 @@ void testAsyncESIndexUpdate() {
     // Call the async method ...
     elasticSearchSpy.asyncDatasetInESIndex(datasetRecord.dataset.getDatasetId(), datasetRecord.createUser, true);
     // Ensure that the synchronous method was called with the expected parameters
-    verify(elasticSearchSpy).synchronizeDatasetInESIndex(datasetRecord.dataset, datasetRecord.dataset.getCreateUser(), true);
+    verify(elasticSearchSpy, timeout(1000)).synchronizeDatasetInESIndex(datasetRecord.dataset, datasetRecord.dataset.getCreateUser(), true);
```
otchet-broad, line 268: Fixed the flaky test here.  We can't assume the scheduler will run the thread before the verify is called.  Added up to a 1 second timeout.

otchet-broad: @rushtong - FYSA

src/test/java/org/broadinstitute/consent/http/resources/DarCollectionResourceTest.java
```diff
@@ -278,6 +278,23 @@ void testGetCollectionWithAllElectionsByCollectionId() {
...
+  void testGetCollectionWithAllElectionsByCollectionIdDacMember() {
+    DarCollection collection = mockDarCollection();
+    List<Integer> userDatasetIds = List.of(1, 2);
+    User admin = new User(2, authUser.getEmail(), "Dac Member User", new Date(),
+        List.of(UserRoles.Member()));
+    DuosUser duosAdmin = new DuosUser(authUser, admin);
+    collection.setCreateUser(researcher);
+    collection.setCreateUserId(researcher.getUserId());
+
+    when(darCollectionService.findDatasetIdsByDACUser(any())).thenReturn(userDatasetIds);
```
otchet-broad: As an aside, I've started to see where I can remove the any() matchers in favor of more specific ones (e.g. eq(admin)).

otchet-broad (approved):

otchet-broad:
Do we need to update any swagger docs on this to be clear that we're filtering?

rushtong (approved):
Thank you! 👍🏽

PR #2654, opened 2025-08-26
[DT-1522] Limit approved users to admins and service accounts

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-1522

### Summary

Limit approved users to admins and service accounts. The service account also needs to be added to DUOS. To do this:

1. Add a user in DUOS with the email of the service account, and assign it the `SERVICE_ACCOUNT` flag.
2. Add the service account's domain to the appropriate institution, e.g. Broad Institute.
3. Query an endpoint using the service account to assign the institution.
    1. To do this, first get the service account JSON (which is likely stored in GSM).
    2. Next, activate the service account: `gcloud auth activate-service-account <account> --key-file <file>`
    3. Print the access token: `gcloud auth print-access-token`
    4. Use the access token to query an endpoint, for example `/api/user/me`
4. If everything works as expected, you should see output similar to the following:

```
{
    "userId": 1,
    "email": "redacted",
    "displayName": "Terra Data Repo Service Account",
    "createDate": 2,
    "roles": [
        {
            "userRoleId": 3,
            "userId": 1,
            "roleId": 4,
            "name": "ServiceAccount"
        },
        {
            "userRoleId": 5,
            "userId": 1,
            "roleId": 6,
            "name": "Researcher"
        }
    ],
    "emailPreference": false,
    "institutionId": 7,
    "institution": {
        "id": 7,
        "name": "The Broad Institute of MIT and Harvard",
        "createDate": 2,
        "updateDate": 2
    },
    "properties": [],
    "userStatusInfo": {
        "adminEnabled": true,
        "userSubjectId": "redacted",
        "userEmail": "redacted",
        "enabled": true
    }
}
```

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
@@ -4,6 +4,7 @@ import com.codahale.metrics.annotation.Timed;
 import com.google.inject.Inject;
 import io.dropwizard.auth.Auth;
 import jakarta.annotation.security.PermitAll;
+import jakarta.annotation.security.RolesAllowed;
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.NotFoundException;
 import jakarta.ws.rs.Path;
@@ -33,7 +34,7 @@ public class TDRResource extends Resource {
 
   @GET
   @Produces("application/json")
-  @PermitAll
+  @RolesAllowed({ADMIN, SERVICE_ACCOUNT})
   @Path("/{identifier}/approved/users")
   @Timed
   public Response getApprovedUsers(@Auth DuosUser duosUser,


src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
```diff
@@ -53,7 +53,7 @@ public Response getApprovedUsers(@Auth AuthUser authUser,
 
   @GET
   @Produces("application/json")
-  @PermitAll
+  @RolesAllowed({ADMIN, SERVICE_ACCOUNT})
   @Path("/{identifier}")
```
rushtong: Is there a reason to add this restriction here?

fboulnois: Consistency, it seems odd to me that we would have TDR specific APIs that wouldn't have similar permissions. If it's not actually a TDR API then I think we should reclassify it.

eweitz (approved):

rjohanek (approved):

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
rushtong: TDR, per se, doesn't actually use this endpoint, I believe it's just developers that hit it. It would be nice to move this to the dataset resource.

rushtong (approved):
👍🏽

src/main/java/org/broadinstitute/consent/http/resources/TDRResource.java
fboulnois: I'll remove the restriction and file a ticket to move it to the dataset resource so we don't break things.

fboulnois: https://broadworkbench.atlassian.net/browse/DT-2161

PR #2657, opened 2025-09-03
[DT-2115] Delegate checking authorization of a dataset - proof of concept.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2115](https://broadworkbench.atlassian.net/browse/DT-2115)

### Summary

This proof of concept adds a new table to hold the list of authorized access checkers for each individual dataset so that we can allow "SERVICE ACCOUNT" users to check to see which users are authorized for a given dataset.  The code adds new endpoints to facilitate adding and removing authorizations along with an endpoint to invoke the same business logic authorization check as performed by our TDRResource.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DatasetAuthorizationReaderDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DatasetAuthorizationReaderDAO.java
@@ -3,7 +3,6 @@ package org.broadinstitute.consent.http.db;
 import java.util.List;
 import org.broadinstitute.consent.http.models.DatasetAuthorizationReader;
 import org.jdbi.v3.sqlobject.config.RegisterConstructorMapper;
-import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
 import org.jdbi.v3.sqlobject.customizer.Bind;
 import org.jdbi.v3.sqlobject.statement.GetGeneratedKeys;
 import org.jdbi.v3.sqlobject.statement.SqlQuery;
--- a/src/main/java/org/broadinstitute/consent/http/db/DatasetAuthorizationReaderDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DatasetAuthorizationReaderDAO.java
@@ -38,9 +38,9 @@ INSERT INTO dataset_authorized_readers (dataset_id, user_id, created_by) VALUES 
 """)
   @GetGeneratedKeys
   long addAuthorizedReaderToDataset(
-      @Bind("datasetId") int datasetId,
-      @Bind("userId") int userId,
-      @Bind("operatorId") int operatorId);
+      @Bind("datasetId") long datasetId,
+      @Bind("userId") long userId,
+      @Bind("operatorId") long operatorId);
 
   @SqlUpdate(
 """
--- a/src/main/java/org/broadinstitute/consent/http/db/DatasetAuthorizationReaderDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DatasetAuthorizationReaderDAO.java
@@ -1,0 +1,63 @@
+package org.broadinstitute.consent.http.db;
+
+import java.util.List;
+import org.broadinstitute.consent.http.models.DatasetAuthorizationReader;
+import org.jdbi.v3.sqlobject.config.RegisterConstructorMapper;
+import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
+import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.statement.GetGeneratedKeys;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+import org.jdbi.v3.sqlobject.transaction.Transactional;
+
+@RegisterConstructorMapper(DatasetAuthorizationReader.class)
+public interface DatasetAuthorizationReaderDAO
+    extends Transactional<DatasetAuthorizationReaderDAO> {
+  @SqlQuery(
+"""
+SELECT * from dataset_authorized_readers WHERE dataset_id = :datasetId
+""")
+  List<DatasetAuthorizationReader> findAuthorizedReadersByDatasetId(
+      @Bind("datasetId") long datasetId);
+
+  @SqlQuery(
+"""
+SELECT * from dataset_authorized_readers WHERE dataset_id = :datasetId AND user_id = :userId
+""")
+  DatasetAuthorizationReader findAuthorizedReadersByDatasetIdAndUserId(
+      @Bind("datasetId") long datasetId, @Bind("userId") long userId);
+
+  @SqlQuery(
+"""
+SELECT * from dataset_authorized_readers WHERE id = :recordId
+""")
+  DatasetAuthorizationReader findAuthorizedReaderByRecordId(@Bind("recordId") long recordId);
+
+  @SqlUpdate(
+"""
+INSERT INTO dataset_authorized_readers (dataset_id, user_id, created_by) VALUES (:datasetId, :userId, :operatorId)
+""")
+  @GetGeneratedKeys
+  long addAuthorizedReaderToDataset(
+      @Bind("datasetId") int datasetId,
+      @Bind("userId") int userId,
+      @Bind("operatorId") int operatorId);
+
+  @SqlUpdate(
+"""
+DELETE FROM dataset_authorized_readers WHERE dataset_id = :datasetId
+""")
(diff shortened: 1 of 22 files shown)


otchet-broad:
Coverage report reflects lines reformatted in DatasetService.  Newly added code for this feature is covered.

src/main/java/org/broadinstitute/consent/http/db/DatasetAuthorizationReaderDAO.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.List;
+import org.broadinstitute.consent.http.models.DatasetAuthorizationReader;
+import org.jdbi.v3.sqlobject.config.RegisterConstructorMapper;
+import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.statement.GetGeneratedKeys;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+import org.jdbi.v3.sqlobject.transaction.Transactional;
+
+@RegisterConstructorMapper(DatasetAuthorizationReader.class)
```
otchet-broad, line 12: This is new to consent.  JDBI has mapping infrastructure for Java `records`

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -554,23 +618,44 @@ private Integer updateStudyFromConversion(User user, Dataset dataset,
...
-              p.getValue().toString()));
+          .forEach(
+              p ->
+                  studyDAO.updateStudyProperty(
+                      studyId, p.getKey(), p.getType().toString(), p.getValue().toString()));
     }
     return studyId;
   }
 
+  public DatasetAuthorizationReader addAuthorizedReader(
```
otchet-broad, line 646: Meat of the changes begin here.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -57,9 +60,18 @@ public class DatasetService implements ConsentLogger {
   private final UserDAO userDAO;
 
   @Inject
-  public DatasetService(DatasetDAO dataSetDAO, DaaDAO daaDAO, DacDAO dacDAO, ElasticSearchService
-      elasticSearchService, EmailService emailService, OntologyService ontologyService, StudyDAO
-      studyDAO, DatasetServiceDAO datasetServiceDAO, UserDAO userDAO) {
+  public DatasetService(
```
otchet-broad, line 63: Most of the changes from here to my next comment below are largely just formatting changes.

This specific change adjusts the constructor for the new dependency.

src/main/java/org/broadinstitute/consent/http/db/DatasetAuthorizationReaderDAO.java
rushtong: I like this, it should be helpful to us for other simple objects. Library Cards come to mind here.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -519,4 +524,72 @@ private void validateDatasetDacAccess(User user, Dataset dataset) {
...
+        return Response.status(Response.Status.NOT_FOUND).build();
+      }
+      if (!datasetService.isAuthorizedToListUsers(dataset.getDatasetId(), duosUser.getUser().getUserId())) {
+        return Response.status(Response.Status.FORBIDDEN).build();
+      }
+      return Response.ok(tdrService.getApprovedUsersForDataset(duosUser, dataset)).build();
+    } catch (Exception e) {
+      return createExceptionResponse(e);
+    }
+  }
```
rushtong: I have some questions here.
* `authorizedUsers` - we're introducing additional terminology for the same entity, `ApprovedUsers`. 
* This is essentially a more locked down version of `/api/tdr/{identifier}/approved/users`. There's nothing to stop the same SA from accessing that API and bypassing the additional controls here.
* We have another ticket ([DT-2161](https://broadworkbench.atlassian.net/browse/DT-2161)) to move that api to this class which would add confusion.

src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
```diff
@@ -35,14 +36,20 @@
...
   private final DatasetDAO datasetDAO;
   private final StudyDAO studyDAO;
 
   @Inject
-  public DatasetServiceDAO(Jdbi jdbi, DatasetDAO datasetDAO, StudyDAO studyDAO) {
+  public DatasetServiceDAO(
+      Jdbi jdbi,
+      DatasetDAO datasetDAO,
+      StudyDAO studyDAO,
+      DatasetAuthorizationReaderDAO datasetAuthorizationReaderDAO) {
```
rushtong, line 48: One of the sonar warnings we have is to reduce the complexity of constructors. I've been thinking it might be possible to instantiate DatasetDAO, StudyDAO, and DatasetAuthorizationReaderDAO directly through Jdbi and reduce the need for them here.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -519,4 +524,72 @@ private void validateDatasetDacAccess(User user, Dataset dataset) {
...
+  public Response getAuthorizedUsers(@Auth DuosUser duosUser, @PathParam("identifier") String datasetIdentifier) {
+    try {
+      Dataset dataset = datasetService.findDatasetByIdentifier(duosUser.getUser(), datasetIdentifier);
+      if (Objects.isNull(dataset)) {
+        return Response.status(Response.Status.NOT_FOUND).build();
+      }
+      if (!datasetService.isAuthorizedToListUsers(dataset.getDatasetId(), duosUser.getUser().getUserId())) {
+        return Response.status(Response.Status.FORBIDDEN).build();
+      }
+      return Response.ok(tdrService.getApprovedUsersForDataset(duosUser, dataset)).build();
```
rushtong, line 590: This method, `tdrService.getApprovedUsersForDataset` has a narrow set of restrictions it places to determine who an `Approved User` is. Is it possible that we might want to show users who, for example, do not have a library card or are not a DUOS/Terra user but still be listed as a user on an approved DAR?

rushtong (commented):
Still reviewing, but some thoughts in the meantime.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
otchet-broad: This lays the groundwork by computing the answer the same way we are doing so today.  It's entirely conceivable that we change this in the future.  Given this is a proof of concept, I expect we'll get further requests to iterate on the capability.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
otchet-broad: - I'll update this method to be renamed to `getApprovedUsers` so it's consistent with the existing terminology.  Good catch.  Thank you.
- True that the caller right now could call the tdr endpoint, but I think this might be moving in a different direction.  This could replace the TDR endpoint and be defined on a dataset by dataset basis.  Seems to me that more will shake out of this proof of concept.  If we did that, I think that DT-2161 might be OBE.

src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
otchet-broad: I'm wondering if that actually solves the underlying concern of the complexity in classes.  For this particular one, I'm wondering if there's a more generic way to scope transactions to a request - more in line with the Spring model.  That would allow us to get out of the business of these wrappers by making the request boundary the scope of a transaction.

src/main/resources/assets/schemas/DatasetAuthorizedReader.yaml
```diff
@@ -0,0 +1,17 @@
...
+    description: The identifying number of the dataset
+  userId:
+    type: number
+    description: The identifying number of the user allowed to check access
+  createdBy:
+    type: number
+    description: The identifying number of the user granting this authorization
+  createdDate:
+    type: number
+    description: When this rule was created in, milliseconds since the epoch
```
eweitz:
```suggestion
    description: When this rule was created, in milliseconds since the epoch
```

eweitz (approved):

rushtong (approved):
Thank you for the updates 👍🏽

PR #2658, opened 2025-09-10
DT-2079: Filter votes from Researcher, SO Collection by ID calls

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-2079

## Summary
The meat of this PR is in `DarCollectionService.filterCollectionForUser`. This service layer method ensures that the votes populated in a `DarCollection` are only visible when the user has **Admin**, **Chairperson**, or **Member** roles. All public methods that return a `DarCollection` now filter through that method.

The DAO method, `DarCollectionDAO.findDARCollectionByCollectionIds` also populates votes, but there are no service methods that expose those entities directly to clients. `MetricsService.generateDatasetMetrics` is the only non-test method that calls this and returns a different object derived from `DarCollection` that does not contain `Vote` entities.

Another aspect of this PR is the simplification of the constructor. By passing in an instance of `Jdbi`, we can instantiate any DAO classes we might need instead of passing them all in individually. This is a new pattern for us that should address a common sonar warning across the codebase.

### Testing:
Admin/Chair/Member
* First, authenticate as a user with Admin, Chair, or Member roles
* Visit the DAR Collection GET /api/collections/{collectionId} swagger endpoint
* Use any collection id that has elections created for it.
* The response entity **SHOULD** include DARs, elections, and votes

Non-Admin/Chair/Member
* Next, authenticate as a generic researcher user
* Visit the DAR Collection GET /api/collections/{collectionId} swagger endpoint
* Use any collection id that you've created and has elections in process or complete
* The response entity **SHOULD** include DARs and elections, but **SHOULD NOT** include votes

----

- Verify all tests go green


fboulnois (approved):
👍

src/main/java/org/broadinstitute/consent/http/ConsentModule.java
```diff
@@ -221,19 +221,8 @@ DuosUserAuthenticator providesDuosUserOAuthAuthenticator() {
...
-        providesElectionDAO(),
-        providesDataAccessRequestDAO(),
-        providesEmailService(),
-        providesVoteDAO(),
-        providesMatchDAO(),
-        providesDarCollectionSummaryDAO(),
-        providesUserDAO(),
-        providesDacDAO()
-    );
+    return new DarCollectionService(providesJdbi(), providesDarCollectionServiceDAO(),
```
otchet-broad, line 228: 👍

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -48,39 +47,35 @@
...
-      MatchDAO matchDAO, DarCollectionSummaryDAO darCollectionSummaryDAO, UserDAO userDAO,
-      DacDAO dacDAO) {
-    this.darCollectionDAO = darCollectionDAO;
-    this.collectionServiceDAO = collectionServiceDAO;
-    this.datasetDAO = datasetDAO;
-    this.electionDAO = electionDAO;
-    this.dataAccessRequestDAO = dataAccessRequestDAO;
+  public DarCollectionService(Jdbi jdbi, DarCollectionServiceDAO darCollectionServiceDAO,
+      EmailService emailService) {
+    this.dacDAO = jdbi.onDemand(DacDAO.class);
```
otchet-broad, line 69: Good start here.  I appreciate the change.

This makes me think about Spring Boot and the AOP it offers here.
We're satisfying the complexity requirements of the constructor parameters, but not the complexity of setting all of the DAOs up seems like something we could decorate with an aspect so that we don't need to do this in each class.  Something to iterate on in the future.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -163,7 +158,8 @@ private void processDarCollectionSummariesForResearcher(List<DarCollectionSummar
       s.addAction(DarCollectionActions.REVIEW);
       //if the latest DAR in the collection has at least one approved dataset,
       //include the create progress report action
-      Set<Integer> datasetIds = dataAccessRequestDAO.findDatasetApprovalsByDar(s.getLatestReferenceId());
+      Set<Integer> datasetIds = dataAccessRequestDAO.findDatasetApprovalsByDar(
```
otchet-broad, line 161: We've got some tech debt in that findDatasetApprovalsByDar should really be returning a Set<Long> as the underlying database column best maps to Java Long instead of Int.  I've started to convert items to long/Long as I encounter them.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -464,46 +463,71 @@ public List<Integer> findDatasetIdsByDACUser(User user) {
...
+   * Given a DarCollection, remove all votes from elections for roles that should not see them. This
+   * method mutates the given collection.
+   *
+   * @param user       The User requesting the collection
+   * @param collection DarCollection to filter votes from
+   * @return DarCollection
+   */
+  private DarCollection filterCollectionVotesForUser(User user, DarCollection collection) {
+    // Individual votes are only visible to CHAIRPERSON, MEMBER, and ADMIN roles
+    List<UserRoles> voteViewRoles = List.of(UserRoles.CHAIRPERSON, UserRoles.MEMBER,
```
otchet-broad, line 494: Are members supposed to see how other members voted?  My understanding was only Chairs and Admins were allowed to see how everyone in a DAC voted.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rushtong: Yes, I have been trying to iterate on that idea locally, but no luck with something useful yet.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rushtong: > Are members supposed to see how other members voted? My understanding was only Chairs and Admins were allowed to see how everyone in a DAC voted.

I understand this a little differently. It is true that we only show Member votes on the Chair page, but that is only due to convention, not security - we just never implemented a Member-specific view. I'm open to reducing this to Chairs and Admins so will follow up with product tomorrow.

eweitz: Related question for product: 

If Members should be able to see individual votes, then Signing Officials should presumably be able to as well, right?

eweitz (approved):

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rushtong: Follow-ups from conversation:
1. We do, in fact, currently show individual member votes to other members so we need to keep them in the results entity.
2. SOs do not currently have a view of votes. Until we need that functionality, we can keep it restricted to Admins, Chairs, and Members.

otchet-broad (approved):
👍

PR #2674, opened 2025-09-18
[DT-2245] bugfix save s4 files properly

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2245](https://broadworkbench.atlassian.net/browse/DT-2245)

### Summary

This method was always throwing an exception, making it difficult to debug and demonstrate a working solution.  The solution to this problem is to use leverage the StudyDAO out of the DatasetService to find the study by the studyID.  Code was expecting the study to be embedded in the dataset, but refactoring in prior commits removed that since it was an expensive call. 

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
@@ -118,7 +118,8 @@ public class DatasetResource extends Resource {
           registration,
           user,
           files);
-      Study study = datasets.get(0).getStudy();
+      Integer studyId = datasets.get(0).getStudyId();
+      Study study = datasetService.findStudy(studyId);
       DatasetRegistrationSchemaV1Builder builder = new DatasetRegistrationSchemaV1Builder();
       DatasetRegistrationSchemaV1 createdRegistration = builder.build(study, datasets);
       URI uri = UriBuilder.fromPath(String.format("/api/dataset/study/%s", study.getStudyId()))
--- a/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
@@ -123,18 +123,16 @@ public class DatasetResource extends Resource {
       Study study = datasetService.findStudy(studyId);
       DatasetRegistrationSchemaV1Builder builder = new DatasetRegistrationSchemaV1Builder();
       DatasetRegistrationSchemaV1 createdRegistration = builder.build(study, datasets);
-      if (study != null) {
-        URI uri = UriBuilder.fromPath(String.format("/api/dataset/study/%s", study.getStudyId()))
-            .build();
-        String entity = GsonUtil.buildGsonNullSerializer().toJson(createdRegistration);
-        return Response.created(uri).entity(entity).build();
-      } else {
+      if (study == null) {
         Exception entityException =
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -118,13 +119,22 @@ public Response createDatasetRegistration(
...
+        URI uri = UriBuilder.fromPath(String.format("/api/dataset/study/%s", study.getStudyId()))
+            .build();
+        String entity = GsonUtil.buildGsonNullSerializer().toJson(createdRegistration);
+        return Response.created(uri).entity(entity).build();
+      } else {
+        Exception entityException =
+            new UnprocessableEntityException("Study was not found after it was created");
+        logException(entityException);
+        throw entityException;
+      }
```
kevinmarete: I think you could invert this if_else block to simplify this 
```suggestion
      if (study == null) {
          Exception entityException = new UnprocessableEntityException("Study was not found after it was created");
          logException(entityException);
          throw entityException;
      } 
      
      DatasetRegistrationSchemaV1Builder builder = new DatasetRegistrationSchemaV1Builder();
      DatasetRegistrationSchemaV1 createdRegistration = builder.build(study, datasets);
      String entity = GsonUtil.buildGsonNullSerializer().toJson(createdRegistration);
      URI uri = UriBuilder.fromPath(String.format("/api/dataset/study/%s", study.getStudyId())).build();
     
      return Response.created(uri).entity(entity).build();
```

kevinmarete (approved):
All good, just a small suggestion.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -118,13 +119,22 @@ public Response createDatasetRegistration(
...
-      return Response.created(uri).entity(entity).build();
+      if (study != null) {
+        URI uri = UriBuilder.fromPath(String.format("/api/dataset/study/%s", study.getStudyId()))
+            .build();
+        String entity = GsonUtil.buildGsonNullSerializer().toJson(createdRegistration);
+        return Response.created(uri).entity(entity).build();
+      } else {
+        Exception entityException =
+            new UnprocessableEntityException("Study was not found after it was created");
+        logException(entityException);
```
rushtong, line 129: There is logging in `createExceptionResponse` - not sure if we need it here. Logging exceptions mean we send a message to Sentry which also triggers an alert to the prod alerts channel, so I'm always very careful about calling this method. We warn in `createExceptionResponse` which is usually enough.

rushtong (approved):
Looks good, pending the logging question, thank you 👍🏽

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
otchet-broad: Specifically called the logException here to trigger Sentry.  If we can't get the study back after everything worked as expected (no SQL Exception raised), we've got a data integrity issue.  I felt that was worth the Sentry alert.

This is an edge case that I think we want to know about.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
otchet-broad: Good eye!  Thank you.

PR #2676, opened 2025-09-19
[DT-2202] Download NIH Institutional Certification doc

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2202](https://broadworkbench.atlassian.net/browse/DT-2202)

### Summary

Allows admins and other authorized users to download the NIH Institutional Certification document that was uploaded when the data submitter created the dataset.
image

image

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -572,4 +588,33 @@ public Response getApprovedUsers(@Auth DuosUser duosUser, @PathParam("identifier
...
+
+  @GET
+  @Produces(MediaType.APPLICATION_OCTET_STREAM)
+  @Path("/{id}/nihInstitutionalCertification")
+  @RolesAllowed({ADMIN, DATASUBMITTER, CHAIRPERSON, MEMBER})
+  public Response getNihInstitutionalCertification(@Auth DuosUser duosUser, @PathParam("id") Integer id) {
+    User requestingUser = duosUser.getUser();
+    Dataset dataset = datasetService.findDatasetById(requestingUser, id);
+    if (dataset == null) {
+      return Response.status(Response.Status.NOT_FOUND).build();
```
rushtong: Throw instead of return. The response should be wrapped in a try/catch with a createExceptionResponse in the catch. That will translate all known error types into a consistent error response object.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -572,4 +588,33 @@ public Response getApprovedUsers(@Auth DuosUser duosUser, @PathParam("identifier
...
+        || dataset.isCreator(requestingUser)
+        || dataset.isCustodian(requestingUser)
+        || requestingUser.verifyDACRole(CHAIRPERSON, dataset.getDacId())
+        || requestingUser.verifyDACRole(MEMBER, dataset.getDacId()))) {
+      InputStream fileStream = gcsService.getDocument(nihFile.getBlobId());
+      StreamingOutput streamOutput = createStreamingOutput(fileStream);
+      return Response.ok(streamOutput).header(HttpHeaders.CONTENT_DISPOSITION,
+          String.format("attachment; filename=\"%s\"", nihFile.getFileName())).build();
+    } else {
+      return Response.status(Status.NOT_FOUND).build();
```
rushtong: Same feedback here, throw instead of return.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -255,6 +255,10 @@ public void deleteStudy(Study study, User user) throws Exception {
     datasetServiceDAO.deleteStudy(study, user);
   }
 
+  public Study findStudy(Integer studyId) {
+    return studyDAO.findStudyById(studyId);
```
rushtong, line 259: Don't know where we landed on this subject, but should this throw an NFE if null?

src/main/resources/assets/paths/datasetNihInstitutionalCertification.yaml
```diff
@@ -0,0 +1,26 @@
...
+        application/octet-stream:
+          schema:
+            type: string
+            format: binary
+    401:
+      description: User not authorized to get file
+    404:
+      description: The dataset and/or the file cannot be found.
+    500:
+      description: An unrecoverable error occurred when attempting to return this attachment.
```
rushtong: nit: newline

src/test/java/org/broadinstitute/consent/http/models/UserCheckRoleInDacTest.java
```diff
@@ -41,4 +41,17 @@ void testVerifyDACRole() {
     assertTrue(isUserAdmin);
   }
 
+  @Test
+  void testVerifyDACRole_NullDacId() {
```
rushtong, line 45: Nice 👍🏽

rushtong (commented):
Looks pretty good, see comments inline. Love the boost in test coverage 👍🏽

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
otchet-broad: That would be fine if the UI processed our specific error types, but it doesn't in this case.  These don't return JSON, they return application/octet-stream.  This is one of those rare cases where I think we do want to return just error codes and not the DUOS formatted error messages.

otchet-broad: @rushtong - Code updated to catch any exception.

rushtong: That's a great point. There are two benefits to `createExceptionResponse`, one is to standardize the formatted result (which seems like we do NOT want that here) and the second one is to translate various exceptions that should not go to clients and translate them to something more client friendly, for example, per compliance, we should never send the user a sql exception which might be possible if internal code hits a `Throwable`. Let's see if we can come up with an approach that serves this need.

rushtong: > @rushtong - Code updated to catch any exception.

Thank you - didn't see this when writing the above ☝🏽

rushtong (approved):

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
otchet-broad: I landed at returning null for now.  We don't return NFE's for empty lists being returned, do we?

kevinmarete (approved):
Looks good!

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
rushtong: Correct, for empty lists, we generally return the empty list. This is returning a full study object though, not a list.

PR #2677, opened 2025-09-22
[DT-2228] Validate query window size

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2228

### Summary

Prevents the following issue which crashes the service:

```
Caused by: org.elasticsearch.search.query.QueryPhaseExecutionException: Result window is too large, from + size must be less than or equal to: [10000] but was [1000000000]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting.
```

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
@@ -86,6 +86,7 @@ public class ElasticSearchService implements ConsentLogger {
     this.studyDAO = studyDAO;
   }
 
+  private static final int MAX_RESULT_WINDOW = 10000;
 
   private static final String BULK_HEADER = """
       { "index": {"_type": "dataset", "_id": "%d"} }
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -86,6 +86,7 @@ public ElasticSearchService(
     this.studyDAO = studyDAO;
   }
 
+  private static final int MAX_RESULT_WINDOW = 10000;
```
rushtong, line 89: Would be great if this value existed as a constant in the ES library, but we're only pulling the ES Rest Client which does not know about this value.

rushtong (approved):
Love the validation tests 👍🏽

eweitz (approved):
Looks good, nice tests!

kevinmarete (approved):

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
fboulnois: Yes, I was hoping we could add this as a parameter to the existing query validation API, but unfortunately ES does not support this.

PR #2683, opened 2025-09-25
DT-2279: Allow Sort in ES Queries

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2279

### Summary
Filter out the `sort` field from provided ES queries when validating the query. Like `from` and `size`, `sort` triggers a failure in the built-in ES validate function.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
@@ -157,8 +157,9 @@ public class ElasticSearchService implements ConsentLogger {
       return false;
     }
 
-    // Remove `size` and `from` parameters from query, otherwise validation will fail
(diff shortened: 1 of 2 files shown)


eweitz (approved):
Looks good!

Do we also need to allow something like `search`, since the frontend will (I assume) no longer have all table data to query in client memory?  If so, I defer to you on whether to implement that here or in a separate ticket.

rushtong:
> Looks good!
> 
> Do we also need to allow something like `search`, since the frontend will (I assume) no longer have all table data to query in client memory? If so, I defer to you on whether to implement that here or in a separate ticket.

That's a great question - we may indeed need that. This subject will require some additional tickets so we should plan those out in a refinement.

otchet-broad (approved):
Looks good!

PR #2717, opened 2025-10-17
[DT-2320] Verify the user has access to the study or dataset before responding.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2320](https://broadworkbench.atlassian.net/browse/DT-2320)

### Summary

If the user is an admin, always return the summary object;
If the study is marked public, return the summary object;
If the study is private and the user is the creator of the dataset, creator of the study, or a custodian of the study, return it;
In all other cases, don't return the summary.

----

- Verify all tests go green

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -28,13 +28,11 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-maven-
       - name: Test Only
-        if: github.actor == 'dependabot[bot]'
         env:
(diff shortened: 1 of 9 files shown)


otchet-broad:
Coverage GHA will now fail when tests fail but not when SONAR can't be reached.

rushtong (approved):
Nice updates to coverage. Overall, this looks reasonable. Given that this is not in use in the UI, it could be simpler to either remove it altogether or restrict it to admin usage. Either way, this solves the stated goal.

otchet-broad:
@rushtong Thanks for the feedback!  I thought about removing it but when looked at our environments I noticed that the other user isn't an admin.

kevinmarete (approved):

PR #2725, opened 2025-10-29
[DT-2390] Address OpenAPI schema issues

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2390

### Summary

This addresses a number of issues with our OpenAPI setup:

* The OpenAPI yaml is now bundled as a single file, which improves performance in the Swagger UI and allows consumers of the spec to more easily transform the data
* The Swagger UI configuration is refactored, the class inherits the `ConsentLogger`, `/swagger` is no longer special cased, loading the `swagger.ui.path` is simplified, content is delivered as bytes where applicable, and a custom index.html is no longer required.
* The schema has a number of fixes:
  * required `path`s are indicated as such
  * spaces are not allowed in schema names
  * incomplete migration from OpenAPI 2 to 3 is resolved, for example `parameters` to `requestBody`
  * mistyping of arrays as objects
  * mistyping of objects as arrays (`items` should be `additionalProperties` for objects)

Also, as a drive-by fix, the Postgres image is also updated to 16.10.

----

- Verify all tests go green


rushtong (approved):
Nice work 👍🏽

kevinmarete (approved):

PR #2751, opened 2025-11-18
DT-2263: Create User API for Admins/Chairs

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2263

### Summary
This PR covers the following:
- Copy/move the current admin create user feature to `POST /api/user/create` in `UserResource`. This is the last piece of migration away from the deprecated `DACUserResource`. See [follow-on ticket](https://broadworkbench.atlassian.net/browse/DT-2496) to clean that up when this is fully transitioned. Uses a new model class, `CreateDuosUserRequest`, which [mirrors what we have in UI code](https://github.com/DataBiosphere/duos-ui/blob/c0da050e16660182e8c75b686a816b60d5f25842/src/types/requestTypes.ts#L3) to implement this API.
- Consolidate all database interactions into a single transaction. Previously, these were all separate transactions, each one potentially failing and leaving the user in an error state. Transactions consolidated include:
  - User insert
  - Email preference update
  - Library Card update
- New tests. There were no existing tests for this functionality
- Expand the role permissions for Admin users to allow for all non-DAC roles. This will allow for the admin interface to update users with additional roles such as Service Account, Data Submitter, etc.
- Restrict Chairpersons to creating users with the Researcher role.

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/models/CreateDuosUserRequest.java
```diff
@@ -0,0 +1,27 @@
+package org.broadinstitute.consent.http.models;
+
+import java.util.List;
+import org.broadinstitute.consent.http.enumeration.UserRoles;
+
+public record CreateDuosUserRequest(
+    String displayName, String email, boolean emailPreference, List<UserRole> roles) {
+
+  public void validate() {
```
otchet-broad, line 18: Why not create a canonical constructor that validates?  This way it will be impossible to build a record of type CreateDuosUserRequest without those fields.
```suggestion
public CreateDuosUserRequest(String displayName, String email, boolean emailPreference, List<UserRole> roles) {
    this.displayName = displayName;
    this.email = email;
    this.emailPreference = emailPreference;
    this.roles = roles;
    validate()
}

  public void validate() {
```

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -486,4 +491,38 @@ public Response getApprovedDatasets(@Auth DuosUser duosUser) {
       return createExceptionResponse(e);
     }
   }
+
+  @POST
+  @Consumes(MediaType.APPLICATION_JSON)
+  @Produces(MediaType.APPLICATION_JSON)
+  @Path("/create")
+  @RolesAllowed({ADMIN, CHAIRPERSON})
```
otchet-broad, line 499: This makes me wonder if we should eventually add signing officials with the same rules as chairs and then extend the CreateDuosUserRequest to issue a library card.

src/main/java/org/broadinstitute/consent/http/service/dao/UserServiceDAO.java
```diff
@@ -42,4 +45,34 @@ public void updateInstitutionAndClearLibraryCardForUser(Integer userId, Integer
...
+          UserRoleDAO userRoleDAOT = transactionHandle.attach(UserRoleDAO.class);
+          LibraryCardDAO libraryCardDAOT = transactionHandle.attach(LibraryCardDAO.class);
+          Integer userId =
+              userDAOT.insertUser(
+                  user.getEmail(), user.getDisplayName(), user.getInstitutionId(), new Date());
+          userDAOT.updateEmailPreference(userId, user.getEmailPreference());
+          for (UserRole role : user.getRoles()) {
+            userRoleDAOT.insertSingleUserRole(role.getRoleId(), userId);
+          }
+          LibraryCard libraryCard = libraryCardDAOT.findLibraryCardByUserEmail(user.getEmail());
```
otchet-broad, line 65: If we eventually refactor how we allow signing officials to issue Library Cards (so they also create the underlying user) it simplifies this code as well.

src/test/java/org/broadinstitute/consent/http/resources/UserResourceTest.java
```diff
@@ -984,6 +998,79 @@ void testGetApprovedDatasets() {
...
+        new CreateDuosUserRequest(null, "test@test.com", true, List.of(UserRoles.Researcher()));
+    try (var response = userResource.createNewUser(duosUser, gson.toJson(request))) {
+      assertEquals(Status.BAD_REQUEST.getStatusCode(), response.getStatus());
+    }
+  }
+
+  @Test
+  void testCreateNewUserInvalidEmail() {
+    CreateDuosUserRequest request =
+        new CreateDuosUserRequest("New User", null, true, List.of(UserRoles.Researcher()));
```
otchet-broad: This test should blow up earlier if we add the constructor to the record with the validation check there.

otchet-broad (approved):
Looks good overall.

I think it might be good to move the validation logic into the constructor for the record, but that's not required.

kevinmarete (approved):

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
rushtong: > This makes me wonder if we should eventually add signing officials with the same rules as chairs and then extend the CreateDuosUserRequest to issue a library card.

Yes, this is a potential line of thought. We could refactor the current LC issuance process to go through a similar/same process.

src/test/java/org/broadinstitute/consent/http/resources/UserResourceTest.java
rushtong: > This test should blow up earlier if we add the constructor to the record with the validation check there.

It does!

PR #2753, opened 2025-11-25
[DT-2461] Add DAC Setting for Automatic Opened DARs

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2461

### Summary

This PR adds a new backend Boolean that lets DACs choose whether new DARs automatically open to all members or stay restricted until the Chair opens them. It wires this setting into the existing DAR opening logic (used by RADAR) and ensures notifications are sent when a DAR is opened. The setting is persisted for future DARs and does not change existing ones.

----

- Verify all tests go green


src/test/java/org/broadinstitute/consent/http/service/DarCollectionServiceTest.java
```diff
@@ -1601,44 +1605,100 @@ private DarCollectionSummary createDarCollectionSummaryWithElections() {
...
+    when(dacDAO.findDacsForDatasetIds(anyList())).thenReturn(Set.of(dac));
+    when(dacAutomationRuleService.findAllByDacId(anyInt())).thenReturn(List.of());
+    when(userDAO.findUsersByRoleId(UserRoles.ADMIN.getRoleId())).thenReturn(List.of());
+    when(userDAO.findUsersForDatasetsByRole(anyList(), anyList())).thenReturn(Set.of(chair));
+    service.sendNewDARCollectionMessage(1);
+    verify(emailService).sendNewDARRequestEmail(any(), any(), any(), any());
+  }
+
+  // Only auto-open DACs
+  @Test
```
otchet-broad, line 1668: You probably want coverage for the mixed case where there are both auto-open and manual to ensure both cases are properly handled for the same DAR.

src/main/resources/changesets/changelog-consent-2025-11-24-add-open-dar-rule.xml
```diff
@@ -0,0 +1,36 @@
...
+            value="Automatically open Data Access Requests (DARs) for all DAC members upon submission, **without requiring Chair to open manually**."/>
+    <column name="state" value="AVAILABLE"/>
+  </insert>
+  <rollback>
+    <sql>
+      DELETE FROM dac_automation_rules WHERE rule = 'AUTO_OPEN_DAR_FOR_ALL_MEMBERS';
+    </sql>
+  </rollback>
+</changeSet>
+</databaseChangeLog>
```
otchet-broad: nit: could use a newline at the end of the file.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -773,49 +786,241 @@ private void cancelElectionsForReferenceIds(List<String> referenceIds) {
...
+      } else { // manual
+        if (isChair || isAdmin) {
+          targetSet.add(user);
+        }
+      }
+    }
+  }
+
+  /** Creates elections and votes for DACs with auto-open rules. */
+  private void createElectionsAndVotesForAutoOpenDacs(
```
otchet-broad, line 1003: This method would need to implement all of the features that https://github.com/DataBiosphere/consent/blob/d56d5edbfefd5a186ff3abc0b98ce73804669b18/src/main/java/org/broadinstitute/consent/http/service/dao/DarCollectionServiceDAO.java#L50 does.  Multiple elections are needed (DATA_ACCESS and RP) along with multiple vote types per user.

otchet-broad (changes requested):

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -773,49 +787,321 @@ private void cancelElectionsForReferenceIds(List<String> referenceIds) {
...
       throws IOException, TemplateException {
-    DarCollection collection = darCollectionDAO.findDARCollectionByCollectionId(collectionId);
-    if (collection == null) {
-      logWarn(
-          "Sending new DAR Collection message: Could not find collection for specified collection id: "
-              + collectionId);
+    // Retrieve the DAR collection by its ID
+    DarCollection darCollection = darCollectionDAO.findDARCollectionByCollectionId(collectionId);
+    if (darCollection == null) {
+      logWarn("Could not find DAR collection for collection id: " + collectionId);
```
eweitz, line 837: Nice trim!

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -773,49 +787,321 @@ private void cancelElectionsForReferenceIds(List<String> referenceIds) {
...
+        dataAccessElectionId, user.getUserId(), VoteType.CHAIRPERSON);
+    dacAutomationRuleService.createVoteForElection(
+        rpElectionId, user.getUserId(), VoteType.CHAIRPERSON);
+
+    dacAutomationRuleService.createVoteForElection(
+        dataAccessElectionId, user.getUserId(), VoteType.FINAL);
+
+    if (!dar.requiresManualReview()) {
+      dacAutomationRuleService.createVoteForElection(
+          dataAccessElectionId, user.getUserId(), VoteType.AGREEMENT);
```
eweitz, line 1090: Does this auto-approve the election and close it, or just auto-open it?

If it's the former, would it be preferable to change our naming around this from "auto-open" to "auto-approve"?  The four RADAR settings we have in the DAC configurations page all speak in terms of "Automatically approve" rather than e.g. "Automatically open".

I don't feel strongly about it.  But I was thrown off by the "auto-open" naming here and the "auto-approve" talk in the UI, after not doing much with RADAR for a month or so.

eweitz (approved):
Code looks good!  I ask a non-blocking question about naming.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
kevinmarete: For this, we are doing an "auto-open" operation.

src/main/java/org/broadinstitute/consent/http/service/DACAutomationRuleService.java
```diff
@@ -182,15 +182,8 @@ protected Vote openElectionAndApprove(
...
-        electionDAO.insertElection(
-            ElectionType.DATA_ACCESS.getValue(),
-            ElectionStatus.OPEN.getValue(),
-            new Date(),
-            dar.getReferenceId(),
-            dataset.getDatasetId());
-    int voteId =
-        voteDAO.insertVote(rule.enabledByUserId(), electionId, VoteType.RADAR_APPROVE.getValue());
+    int electionId = createOpenElectionForDAR(dar, dataset, ElectionType.DATA_ACCESS);
+    int voteId = createVoteForElection(electionId, rule.enabledByUserId(), VoteType.RADAR_APPROVE);
```
rushtong: It would be nice if these two db operations were wrapped in a single transaction

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -773,49 +787,321 @@ private void cancelElectionsForReferenceIds(List<String> referenceIds) {
...
-    // Only do this for the DAR... dacDAO.findDacsForDatasetIds(dar.getDatasetIds())
-    Collection<Dac> dacsInDAR = dacDAO.findDacsForDatasetIds(dar.getDatasetIds());
-    // Use only the datasets from the dar
-    List<Integer> datasetIds = dar.getDatasetIds();
-    List<Dataset> datasetsInDAR =
+
+    // Get DACs and datasets for the DAR
+    List<Integer> datasetIds = latestDar.getDatasetIds();
+    List<Dataset> datasetsForDar =
         datasetIds.isEmpty() ? List.of() : datasetDAO.findDatasetsByIdList(datasetIds);
```
rushtong: In all of our DAO calls that take a `List` argument, we've made updates so they return an empty list if a null or empty argument is passed in so the ternary condition here is not strictly necessary.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -773,49 +787,321 @@ private void cancelElectionsForReferenceIds(List<String> referenceIds) {
...
+    for (Dataset dataset : classification.autoOpenDatasets) {
+      if (hasOpenElection(latestDar, dataset)) {
+        continue;
+      }
+
+      archiveOldElections(latestDar, dataset);
+
+      int dataAccessElectionId =
+          dacAutomationRuleService.createOpenElectionForDAR(latestDar, dataset, DATA_ACCESS);
+      int rpElectionId = dacAutomationRuleService.createOpenElectionForDAR(latestDar, dataset, RP);
```
rushtong: Here, again, we should strive to wrap these in a single transaction.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -773,49 +787,321 @@ private void cancelElectionsForReferenceIds(List<String> referenceIds) {
...
+
+    dacAutomationRuleService.createVoteForElection(
+        dataAccessElectionId, user.getUserId(), VoteType.CHAIRPERSON);
+    dacAutomationRuleService.createVoteForElection(
+        rpElectionId, user.getUserId(), VoteType.CHAIRPERSON);
+
+    dacAutomationRuleService.createVoteForElection(
+        dataAccessElectionId, user.getUserId(), VoteType.FINAL);
+
+    if (!dar.requiresManualReview()) {
```
rushtong, line 1088: I'm not confident that this method is complete and will do what we want it to do - it looks outdated. I don't see, for example, AI/LLMs, IRB or for profit states which, in theory, should also be prohibitive in this case.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -773,49 +787,321 @@ private void cancelElectionsForReferenceIds(List<String> referenceIds) {
...
+  private boolean hasAutoOpenRule(Integer dacId) {
+    return dacAutomationRuleService.findAllByDacId(dacId).stream()
+        .anyMatch(
+            r ->
+                r.ruleType() == DACAutomationRuleType.AUTO_OPEN_DAR_FOR_ALL_MEMBERS
+                    && r.enabledByUserId() != null);
+  }
+
+  /** Retrieves the user ID associated with the auto-open rule for a given DAC ID. */
+  private Integer getAutoOpenRuleUserId(Integer dacId) {
```
otchet-broad: It feels like the method name should be more explicit about the value being returned.

```suggestion
  private Integer getAutoOpenRuleEnabledByUserId(Integer dacId) {
```

src/main/java/org/broadinstitute/consent/http/service/DACAutomationRuleService.java
otchet-broad: I'm wondering if it's time to bring in something like [https://github.com/isopropylcyanide/dropwizard-jdbi-unitofwork](https://github.com/isopropylcyanide/dropwizard-jdbi-unitofwork) so we can annotate and be done.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
otchet-broad: @rushtong - so if I'm understanding you correctly, there are rules you believe we need to implement for cases where we shouldn't automatically  open elections.  Do we have those codified somewhere?

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -773,49 +787,321 @@ private void cancelElectionsForReferenceIds(List<String> referenceIds) {
...
+    List<Dataset> datasetsForDar =
         datasetIds.isEmpty() ? List.of() : datasetDAO.findDatasetsByIdList(datasetIds);
+    Collection<Dac> dacsForDar = dacDAO.findDacsForDatasetIds(datasetIds);
+
+    // Classify DACs and users by automation rules
+    DacUserClassification classification =
+        classifyDacsAndUsers(dacsForDar, datasetsForDar, adminAndChairUsers);
+
+    // Create elections and votes for auto-open DACs
+    createElectionsAndVotesForAutoOpenDacs(classification, latestDar);
```
rushtong: This method call is inside of `sendNewDARCollectionMessage` which was confusing when reading tests because this method is also creating elections and votes. Looking at usages of this method, it's only called from a resource class method, `sendNewDarCollectionMessage`. Election/Vote creation seems like it should be its own code path as opposed to being embedded inside of a message sending method.

src/main/java/org/broadinstitute/consent/http/service/DACAutomationRuleService.java
rushtong: I'm open to suggestions. After looking at more of this PR, there are a LOT of cases that should be wrapped in transactions.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
rushtong: @otchet-broad - Correct, as far as I understand it now. If you look at all of the DAR questions, any selections in the radio options in **Step 3: Research Purpose Statement**, in theory, are what `requiresManualReview()` is intending to capture. That method has not been updated as we've made many changes to the DAR form.

otchet-broad: @rushtong - Should we make a separate ticket for that since it seems like that impacts some other code as well?

kevinmarete: Added ticket here: https://broadworkbench.atlassian.net/browse/DT-2520

otchet-broad (approved):
Looks good!  Thanks for pairing, the changes, and the follow-up ticket!

rushtong (approved):

PR #2755, opened 2025-11-25
[DT-2471] Upgrade to OpenSearch 3

fboulnois:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2471

### Summary

Upgrades from ElasticSearch 5 to OpenSearch 3. Here's what is needed in the `docker-compose.yaml`:

```
  elastic:
    image: opensearchproject/opensearch:3
    ports:
      - "9200:9200"
    container_name: elastic
    volumes:
      - ./data:/usr/share/opensearch/data
    environment:
      - DISABLE_SECURITY_PLUGIN=true
      - discovery.type=single-node
      - cluster.routing.allocation.disk.threshold_enabled=false
```

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
@@ -23,8 +23,8 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.ExecutorService;
-import org.apache.http.entity.ContentType;
-import org.apache.http.nio.entity.NStringEntity;
+import org.apache.hc.core5.http.ContentType;
+import org.apache.hc.core5.http.io.entity.StringEntity;
 import org.broadinstitute.consent.http.configurations.ElasticSearchConfiguration;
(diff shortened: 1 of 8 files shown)


src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -193,12 +193,12 @@ public Response searchDatasets(String query) throws IOException {
   }
 
   public InputStream searchDatasetsStream(String query) throws IOException {
-    if (!validateQuery(query)) {
+    if (invalidResultWindow(query)) {
```
fboulnois, line 196: The `/_validate/query` endpoint can **only** validate the query part, but if you have aggregations this marks the query as invalid, hence for the `v2` endpoint only the result window size is checked.

otchet-broad:
@fboulnois - Had you given any more thought to how we might migrate to this and have both running at the same time?

What's in this PR looks great for the cutover, but it necessitates an all-in migration where both the FE and BE need to convert simultaneously.  I wasn't sure where that topic landed.

eweitz (approved):
Nice consistency in the port!

PR #2758, opened 2025-12-02
DT-2519: Patch API for selected study fields

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-2519

## Summary
  This PATCH API allows for admins, chairs, and data submitters to patch a set of selected values on a study. The top level field values patchable are:
* Study name
* Study description
* Study datatypes (list of strings)
* Study PI Name
* Study public visibility (boolean)

Additionally, these study properties can be modified:
* Study Type (enum that must match a value in the current [schema](https://github.com/DataBiosphere/consent/blob/develop/src/main/resources/dataset-registration-schema_v1.json#L113))
* Phenotype Indication
* Data Custodian Email (this replicates and makes redundant `PUT /api/dataset/study/{studyId}/custodians`)
* Target Delivery Date
* Target Public Release Date

### Reviewer Notes
* I chose to use Jackson to strictly validate the input json because similar configuration functionality is missing from Gson.
* The current UI shows the two target dates as unrelated to alternative data file sharing plan, but the fields are saved as `alternativeDataSharingPlanTargetDeliveryDate` and `alternativeDataSharingPlanTargetPublicReleaseDate`:
Screenshot 2025-12-03 at 9 48 02 AM

----

- Verify all tests go green

### DIFF
--- a/src/main/resources/assets/schemas/StudyPatch.yaml
+++ b/src/main/resources/assets/schemas/StudyPatch.yaml
@@ -7,19 +7,7 @@ properties:
     type: string
     description: A human-readable description of the study.
   studyType:
-    type: string
-    enum:
-      - Observational
-      - Interventional
-      - Descriptive
-      - Analytical
-      - Prospective
-      - Retrospective
-      - Case report
-      - Case series
-      - Cross-sectional
-      - Cohort study
-    description: The study type
+    $ref: './StudyType.yaml'
   dataTypes:
     type: array
     items:
--- a/src/main/resources/assets/schemas/StudyPatch.yaml
+++ b/src/main/resources/assets/schemas/StudyPatch.yaml
@@ -6,14 +6,45 @@ properties:
   description:
     type: string
     description: A human-readable description of the study.
+  studyType:
+    type: string
+    enum:
+      - Observational
+      - Interventional
+      - Descriptive
+      - Analytical
+      - Prospective
+      - Retrospective
+      - Case report
+      - Case series
+      - Cross-sectional
+      - Cohort study
+    description: The study type
   dataTypes:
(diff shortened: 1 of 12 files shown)


src/main/resources/assets/schemas/StudyPatch.yaml
```diff
@@ -0,0 +1,50 @@
+type: object
+properties:
+  name:
+    type: string
+    description: The name of the study.
+  description:
+    type: string
+    description: A human-readable description of the study.
+  studyType:
```
kevinmarete, line 9: Is StudyType enum centrally defined somewhere so that we can reuse and avoid duplication?

kevinmarete (approved):

src/main/resources/assets/schemas/StudyPatch.yaml
```diff
@@ -0,0 +1,50 @@
...
+    type: string
+    enum:
+      - Observational
```
otchet-broad: Can we put these in a ref object so we don't keep repeating this list?  I think it's used elsewhere already.

src/main/resources/assets/schemas/StudyPatch.yaml
rushtong: It is defined here: https://github.com/DataBiosphere/consent/blob/develop/src/main/resources/assets/schemas/DatasetRegistrationSchemaV1.yaml#L6
but that's also not separate. Let me look into moving it to a separate yaml.

otchet-broad (approved):
Nice!

Left one thought about the OpenAPI docs that is implementer's choice.  Everything else looks good!
Thank you!

PR #2761, opened 2025-12-03
[DT-2528] Removes the attribute from the list of required fields.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2528](https://broadworkbench.atlassian.net/browse/DT-2528)

### Summary

This attribute should not be required from the client since the value is determined by the user performing the action in the backend.  Removing this from the 'required' lists doesn't change that it can still be provided and ignored for backwards compatibility.

----

- Verify all tests go green

### DIFF
--- a/src/main/resources/dataset-registration-schema_v1.json
+++ b/src/main/resources/dataset-registration-schema_v1.json
@@ -8,7 +8,6 @@
     "studyName",
     "studyDescription",
     "dataTypes",
-    "dataSubmitterUserId",
     "publicVisibility",
     "nihAnvilUse",
     "piName",


src/main/resources/dataset-registration-schema_v1.json
```diff
@@ -8,7 +8,6 @@
     "studyName",
     "studyDescription",
     "dataTypes",
-    "dataSubmitterUserId",
```
otchet-broad, line 11: createDatasetsFromRegistration in the DatasetRegistrationService uses the acting user as the submitter.
https://github.com/DataBiosphere/consent/blob/e5075f387f5bbf0fa4d02ab49e578d4ca4adc57d/src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java#L177

kevinmarete (approved):

src/main/resources/dataset-registration-schema_v1.json
rushtong: IIRC, we needed to specify this in the schema so that front end validation didn't break when updating a study/dataset. Does that functionality still work with this removed?

rushtong (approved):
Pending question, 👍🏽

src/main/resources/dataset-registration-schema_v1.json
otchet-broad: @rushtong I'll reach out so we can pair and validate on my local.

fboulnois (approved):
👍 assuming @rushtong 's feedback is addressed.

src/main/resources/dataset-registration-schema_v1.json
rushtong: After reviewing in person, this looks good 👍🏽

otchet-broad:
My frontend PR will also include a schema validation test update to sync this change in DUOS-UI

PR #2764, opened 2025-12-05
[DT-2542] NMDS fixes.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2542](https://broadworkbench.atlassian.net/browse/DT-2542)

### Summary

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/builder/ConsentGroupFromDataset.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/builder/ConsentGroupFromDataset.java
@@ -41,10 +41,7 @@ public class ConsentGroupFromDataset {
         consentGroup.setDiseaseSpecificUse(dataUse.getDiseaseRestrictions());
         consentGroup.setPoa(dataUse.getPopulationOriginsAncestry());
         consentGroup.setOtherPrimary(dataUse.getOther());
-        consentGroup.setNmds(
-            dataUse.getMethodsResearch() != null && dataUse.getMethodsResearch() == false
-                ? true
-                : null);
+        consentGroup.setNmds(Boolean.FALSE.equals(dataUse.getMethodsResearch()) ? true : null);
         consentGroup.setGso(dataUse.getGeneticStudiesOnly());
         consentGroup.setPub(dataUse.getPublicationResults());
         consentGroup.setCol(dataUse.getCollaboratorRequired());
--- a/src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/builder/ConsentGroupFromDataset.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/builder/ConsentGroupFromDataset.java
@@ -41,7 +41,10 @@ public class ConsentGroupFromDataset {
         consentGroup.setDiseaseSpecificUse(dataUse.getDiseaseRestrictions());
         consentGroup.setPoa(dataUse.getPopulationOriginsAncestry());
         consentGroup.setOtherPrimary(dataUse.getOther());
-        consentGroup.setNmds(dataUse.getMethodsResearch());
+        consentGroup.setNmds(
+            dataUse.getMethodsResearch() != null && dataUse.getMethodsResearch() == false
+                ? true
+                : null);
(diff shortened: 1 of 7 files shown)


src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/builder/ConsentGroupFromDataset.java
```diff
@@ -41,7 +41,7 @@ public ConsentGroup build(Dataset dataset) {
         consentGroup.setDiseaseSpecificUse(dataUse.getDiseaseRestrictions());
         consentGroup.setPoa(dataUse.getPopulationOriginsAncestry());
         consentGroup.setOtherPrimary(dataUse.getOther());
-        consentGroup.setNmds(dataUse.getMethodsResearch());
+        consentGroup.setNmds(Boolean.FALSE.equals(dataUse.getMethodsResearch()) ? true : null);
```
otchet-broad, line 44: We invert the incoming value from true to false so it makes sense that we would need to flip it on the way out as well.

[https://github.com/DataBiosphere/consent/blob/89d46ba4f69e888f682b152ec6f941a3ee4a6eee/src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java#L360](https://github.com/DataBiosphere/consent/blob/89d46ba4f69e888f682b152ec6f941a3ee4a6eee/src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java#L360)

fboulnois: (perhaps not in this PR) Since this has caused ongoing confusion, should we add/update/refactor this property so that we don't need to invert the values?

fboulnois (commented):
See below:

src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/builder/ConsentGroupFromDataset.java
otchet-broad: @fboulnois - agree with your thoughts here and had similar ones while I was going through the code.  I wanted to do the first implementation showing all of the places that were impacted by the original decisions to make sure I didn't miss anything and then come back and refactor this (and migrate the data in the database as well).  I also wanted to catch up with @rushtong on this to see if we could better understand the history of why it was treated differently (and kept that way in his earlier refactor).

rushtong: A bit of history that might help inform a future decision:
* We originally asked researchers (we do not ask this any longer) if they would be doing methods development research. See the history on [DataAccessRequestApplication.js](https://github.com/DataBiosphere/duos-ui/blob/218a2ea46247075f20f26a0390eda2697633a27e/src/pages/DataAccessRequestApplication.js#L1016C1-L1017C1)
* This is the inverse of the DUO term, NMDS: https://github.com/EBISPOT/DUO
* To store those in a Data Use object, we capture these as ["methodsResearch"](https://github.com/DataBiosphere/consent-ontology/blob/develop/src/main/resources/data-use-v4.json#L48)
* We have internal business logic (see [the V4 algorithm](https://github.com/DataBiosphere/consent-ontology/blob/develop/docs/Algorithms.md)) in ontology that matches the research purpose on this field but does NOT take the corresponding NMDS dataset value into consideration (see code reference in [DataUseMatchCasesV3.java](https://github.com/DataBiosphere/consent-ontology/blob/357b352ad6008ce996f39dc8488f3b229fa76a19/src/main/java/org/broadinstitute/dsde/consent/ontology/datause/DataUseMatchCasesV4.java#L172)).
* Since we don't collect this in the DAR, and it isn't evaluated on the Dataset for matching, I think this field is entirely moot.

otchet-broad: Maybe we fix this with this PR and add a ticket to go back to product to see if we can remove this from our codebase entirely?

rushtong (approved):
This is a confusing field, thanks for tracing this through 👍🏽

kevinmarete (approved):

PR #2770, opened 2025-12-15
[DT-2750] Erase old study properties before applying new registration to a study.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2750](https://broadworkbench.atlassian.net/browse/DT-2750)

### Summary

Removes all study properties when receiving a new registration for a study.  This allows patching to work for a particular study property through the new patch endpoint, but cleans up old study properties that may no longer be relevant when submitting a new registration and applying it to an existing study.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/StudyDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/StudyDAO.java
@@ -133,6 +133,12 @@ public interface StudyDAO extends Transactional<StudyDAO> {
       """)
   void deleteStudyByStudyId(@Bind("studyId") Integer studyId);
 
+  @SqlUpdate(
+      """
+      DELETE FROM public.study_property where study_id = :studyId
+      """)
+  void deleteStudyPropertiesByStudyId(@Bind("studyId") Integer studyId);
+
   @UseRowReducer(StudyReducer.class)
   @SqlQuery(
       """
--- a/src/main/java/org/broadinstitute/consent/http/db/StudyDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/StudyDAO.java
@@ -135,7 +135,7 @@ public interface StudyDAO extends Transactional<StudyDAO> {
 
   @SqlUpdate(
       """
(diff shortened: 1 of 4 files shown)


src/test/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAOTest.java
```diff
@@ -586,7 +586,7 @@ void testUpdateStudyWithPropUpdates() throws Exception {
     // Updated prop
     Optional<StudyProperty> updatedProp1 =
         updatedStudy.getProperties().stream()
-            .filter(p -> p.getStudyPropertyId().equals(prop1.getStudyPropertyId()))
+            .filter(p -> p.getKey().equals(prop1.getKey()))
```
otchet-broad, line 589: Since we're deleting and reinserting the property now, the ID can change.  Using the property Key ensures that we're finding the property we expect.

kevinmarete (approved):

src/main/java/org/broadinstitute/consent/http/db/StudyDAO.java
```diff
@@ -133,6 +133,12 @@ DELETE FROM study WHERE study_id in (select study_id from property_deletes)
       """)
   void deleteStudyByStudyId(@Bind("studyId") Integer studyId);
 
+  @SqlUpdate(
+      """
+      DELETE FROM public.study_property where study_id = :studyId
```
rushtong: We usually don't specify the schema, is there a reason it's needed here? Locally, tests pass without it.

```suggestion
      DELETE FROM study_property where study_id = :studyId
```

rushtong (approved):
Minor comment inline, address as you see fit. 👍🏽

src/main/java/org/broadinstitute/consent/http/db/StudyDAO.java
otchet-broad: Good catch.  Not needed.  I think I have the local DB set up in autocomplete and it did it when I pressed tab on the table name.

PR #2776, opened 2026-01-14
DT-2791, DT-2792: Allow SOs to create elections

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-2791
https://broadworkbench.atlassian.net/browse/DT-2792

## Summary
Allow SOs to create elections for a restricted set of datasets
* Requires the dataset's DAC to have a specific rule enabled
* Requires the SO be listed on the DAR

----
Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

PR #2780, opened 2026-01-16
DT-2801: Remove the OPEN action for chairs when SO approval required

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2801

### Summary
Depends on https://github.com/DataBiosphere/consent/pull/2776

New logic for action calculations for Chairs when processing DAR Collection Summary objects.
* If the DAR submitted has recorded that it requires SO approval, we don't show the `Open` action.
* If all datasets in a collection require SO approval, we don't show the `Open` action.
* If none, or only some datasets require SO approval, we continue to show the `Open` action.
* The election creation process in [#2776](https://github.com/DataBiosphere/consent/pull/2776) will correctly filter new elections for only the datasets that are allowable by the Chair

----

- Verify all tests go green


eweitz (approved):

otchet-broad:
@rushtong I think we're going to need to combine your chocolate here with some changes in my peanut butter [(https://github.com/DataBiosphere/consent/pull/2782)](https://github.com/DataBiosphere/consent/pull/2782) to account for the case where we have to toggle the rule from on->off->on again so that we only do this when the signing officials were notified that they needed to take action while the rule was enabled.

fboulnois (approved):
👍

rushtong:
@eweitz @fboulnois - FYI, re-requesting review based on recent decisions around SO approval storage on DARs.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -272,11 +272,18 @@ private void processDarCollectionSummariesForMember(
...
   private void processDarCollectionSummariesForChair(List<DarCollectionSummary> summaries) {
+    Set<Integer> soApprovalDatasetIds =
+        datasetDAO.findAllDatasetIdsByAutomationRuleType(
+            DACAutomationRuleType.REQUIRE_SO_DAR_APPROVAL.name());
     summaries.forEach(
         s -> {
+          // DarCollections with ALL datasets requiring SO approval should not show the OPEN action.
+          // DarCollections with a requiresSOApproval setting should also not show the OPEN action.
+          boolean requiresSOApproval =
+              soApprovalDatasetIds.containsAll(s.getDatasetIds()) || s.requiresSOApproval();
```
otchet-broad: If you update DarCollectionSummaryDAO.getDarCollectionSummariesForDACRole to include the SO related properties like I've done for the SigningOfficial version of the query, you'll be able to look at the DarCollectionSummary object for this now.  If the DAR requires SO approval, requiresSOApproval will be set to `true`.  If it's been approved, `soApproverId` will be a non-null value.  I don't think you'll need to query for the datasets.

rushtong: I'll re-dig into the queries to ensure `requiresSOApproval` is populated.
I still think we need to query on datasets if this sequence happens:

- No DAC rules/setting are enabled
- User submits DAR against datasets in DAC 1, DAC 2
- Chair from DAC 1 checks SO approval required

I don't think any other chairperson from either DAC 1 or 2 should see an OPEN button for a DAR in this state.

otchet-broad: I thought the rule was that we took truth about a DAR requiring SO approval at time the DAR was submitted.  If the DAC turns on the rule after the DAR is submitted, it wasn't required at DAR submission.  Maybe I'm misunderstanding or misinterpreting this?  I would expect both DAC 1 and DAC 2 would be able to see an OPEN button.

rushtong: > I thought the rule was that we took truth about a DAR requiring SO approval at time the DAR was submitted. If the DAC turns on the rule after the DAR is submitted, it wasn't required at DAR submission. Maybe I'm misunderstanding or misinterpreting this? I would expect both DAC 1 and DAC 2 would be able to see an OPEN button.

Great question - I'm not clear on what decisions have been made. Added a topic for discussion today.

rushtong: Question resolved in standup - we only need to look at SO approval at DAR submission time.

otchet-broad (approved):
Nice work!  Thanks for the great conversation on the feature and this implementation!

fboulnois (approved):
👍

PR #2784, opened 2026-01-22
[DT-2823] Refactor to remove rule exclusivity requirement.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2823](https://broadworkbench.atlassian.net/browse/DT-2823)

### Summary
* Removes Automation Rule exclusivity requirement
* Fix actions returning when the user isn't the one who can take them
* Adds coverage to methods previously uncovered.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
@@ -121,6 +121,7 @@ public interface DarCollectionSummaryDAO extends Transactional<DarCollectionSumm
                latest_dar.data ->> 'projectTitle' AS name,
                latest_dar.data ->> 'status' AS dar_status,
                latest_dar.data ->> 'closeoutSupplement' AS closeout,
+               latest_dar.data ->> 'signingOfficialEmail' AS signingOfficialEmail,
                ARRAY_AGG(dar_all.reference_id) AS reference_ids
               FROM dar_collection c
               INNER JOIN users u
--- a/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
@@ -732,6 +732,29 @@ public class DarCollectionService implements ConsentLogger {
   }
 
   /**
+   * @param user The User initiating new elections for a collection
+   * @param collection The DarCollection
+   * @param request The request context
+   * @return The updated DarCollection
+   * @throws ConsentConflictException Can be thrown when the collection no longer requires approval
+   *     or the underlying data access request does not support this operation.
+   * @throws BadRequestException Can be thrown when a collection required approval but is already
+   *     approved
+   * @throws ForbiddenException The user is not the specified signing official required to approve
+   *     the collection
+   */
+  public DarCollection approveDarCollection(
+      User user, DarCollection collection, ContainerRequest request)
+      throws ConsentConflictException, BadRequestException, ForbiddenException {
+    DataAccessRequest dar = collection.getMostRecentDar();
+    if (dar.getRequiresSOApproval()) {
+      approveDataAccessRequestBySigningOfficial(user, dar, request);
+      return darCollectionDAO.findDARCollectionByCollectionId(dar.getCollectionId());
+    }
+    throw new ConsentConflictException("This collection does not require approval.");
+  }
+
+  /**
    * DarCollections with no elections, or with previously canceled elections, are valid for
    * initiating a new set of elections. Elections in open, closed, pending, or final states are not
    * valid.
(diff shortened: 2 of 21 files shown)


src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -121,6 +121,7 @@ List<DarCollectionSummary> getDarCollectionSummariesForDACRole(
                latest_dar.data ->> 'projectTitle' AS name,
                latest_dar.data ->> 'status' AS dar_status,
                latest_dar.data ->> 'closeoutSupplement' AS closeout,
+               latest_dar.data ->> 'signingOfficialEmail' AS signingOfficialEmail,
```
otchet-broad, line 124: We need to know the signing official selected in the DAR so we can match the requester to the Approve action.

src/main/java/org/broadinstitute/consent/http/service/dao/DarCollectionServiceDAO.java
```diff
@@ -78,7 +46,7 @@ public List<Integer> findActionableDatasetIdsForUser(User user, DataAccessReques
   public List<String> createElectionsForDarByUser(User user, DataAccessRequest dar)
       throws SQLException {
     List<String> createdElectionReferenceIds = new ArrayList<>();
-    List<Integer> actionableDatasetIds = findActionableDatasetIdsForUser(user, dar);
+    List<Integer> actionableDatasetIds = datasetDAO.findDatasetIdsByDACUserId(user.getUserId());
```
rushtong, line 49: I'm unclear on how (or where, if has been moved) an SO would create an election. `datasetDAO.findDatasetIdsByDACUserId` specifically restricts the results to users with a DAC role.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -740,43 +743,52 @@ private DarCollection cancelDarCollectionElectionsAsChair(DarCollection collecti
...
-        } else {
-          emailService.sendDarNewCollectionElectionMessage(voteUsers, collection.getDarCode());
+        List<String> createdElectionReferenceIds =
+            collectionServiceDAO.createElectionsForDarByUser(user, dar);
+        if (createdElectionReferenceIds.isEmpty()) {
+          var e =
+              new IllegalStateException(
+                  "No elections were created for DAR Collection: %s %s"
+                      .formatted(collection.getDarCode(), dar.getReferenceId()));
+          logException(e);
```
rushtong: This method triggers a Sentry alert which we generally avoid unless developer action is urgently required. We are already doing this in other blocks, so the pattern is established, but I do think we should be doing something different here.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -340,25 +340,28 @@ private void updateSummaryActionsForChair(
...
     if (summary.getCloseoutSupplement() != null
-        && summary.getCloseoutSigningOfficialApprovalDate() == null) {
+        && summary.getCloseoutSigningOfficialApprovalDate() == null
+        && user.getUserId().equals(summary.getCloseoutSupplement().signingOfficialId())) {
       summary.addAction(DarCollectionActions.REVIEW_PROGRESS_REPORT);
     }
-    if (summary.requiresSOApproval() && summary.getSOApprover() == null) {
+    if (summary.requiresSOApproval()
+        && summary.getSOApprover() == null
+        && user.getEmail().equalsIgnoreCase(summary.getSigningOfficialEmail())) {
```
rushtong, line 364: Nice catch

rushtong (approved):
Code wise, this looks reasonable, but I think I have a fundamental misunderstanding of the feature as the requirements have evolved. Assuming this is the correct approach - 👍🏽

src/main/java/org/broadinstitute/consent/http/service/dao/DarCollectionServiceDAO.java
otchet-broad: SOs won't be creating the elections.  They'll trigger the notice to DACs and we'll continue to have DACs (or their automation rule settings) do that.  This allows us to avoid the need to change the DACs workflow or behavior at all.  To them, it's business as usual.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
otchet-broad: Would you like to see the logException in this block converted to a logWarn?

rushtong: > Would you like to see the logException in this block converted to a logWarn?

I would be more comfortable with that approach. I think what I'd really like to see is something that triggers an alert to our notifications channel but that's very much out of scope here.

src/main/java/org/broadinstitute/consent/http/service/dao/DarCollectionServiceDAO.java
rushtong: OK, I can see that change in `createElectionsForDarCollection` where we return after the `approveDataAccessRequestBySigningOfficial` step. Conceptually, mixing these two different actions in a `create elections` method is confusing to me. It would less confusing if we called different methods from the resource class based on the role so `createElectionsForDarCollection` can be tightened up.

otchet-broad: I can do that, but the thing I was thinking about is if we should make a new endpoint for SO approval and move the SO logic to that part since the SO isn't creating elections.  I was going to bring that up today at standup.

fboulnois (approved):
Nice cleanup, the additions are mostly tests!

rushtong (approved):
👍🏽

rushtong (approved):
👍🏽

kevinmarete (approved):
Great job!

PR #2786, opened 2026-01-22
[DT-2824] Feature flags

fboulnois:
### Summary

Adds feature flags to Consent. The API is:

* `GET /api/feature` - List all feature flags (unauthenticated)
* `GET /api/feature/id` - Get a specific flag (unauthenticated)
* `POST /api/feature/id` - Create/update value (admin-only), body: `{"value": "..."}`
* `DELETE /api/feature/id` - Delete a feature flag (admin-only)

Some helper methods exist that consent itself can use to determine if a feature flag is enabled. I kept the `value` as a string, as we may want multiple states beyond enabled or disabled.

Unit tests cover all functionality.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/FeatureFlagResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/FeatureFlagResource.java
@@ -1,0 +1,130 @@
+package org.broadinstitute.consent.http.resources;
+
+import com.google.inject.Inject;
+import io.dropwizard.auth.Auth;
+import jakarta.annotation.security.PermitAll;
+import jakarta.annotation.security.RolesAllowed;
+import jakarta.ws.rs.Consumes;
+import jakarta.ws.rs.DELETE;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.POST;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.PathParam;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.Context;
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+import jakarta.ws.rs.core.UriInfo;
+import java.net.URI;
+import java.util.List;
+import java.util.Map;
+import org.broadinstitute.consent.http.models.AuthUser;
+import org.broadinstitute.consent.http.models.FeatureFlag;
+import org.broadinstitute.consent.http.service.FeatureFlagService;
+
+@Path("api/feature")
+public class FeatureFlagResource extends Resource {
+
+  private final FeatureFlagService featureFlagService;
+
+  @Inject
+  public FeatureFlagResource(FeatureFlagService featureFlagService) {
+    this.featureFlagService = featureFlagService;
+  }
+
+  /**
+   * Get all feature flags
+   *
+   * @return List of all feature flags
+   */
+  @GET
+  @Produces(MediaType.APPLICATION_JSON)
+  @PermitAll
+  public Response getAllFeatureFlags() {
+    try {
+      List<FeatureFlag> flags = featureFlagService.getAllFeatureFlags();
+      return Response.ok(flags).build();
+    } catch (Exception e) {
+      return createExceptionResponse(e);
+    }
+  }
+
+  /**
+   * Get a specific feature flag by id
+   *
+   * @param id The feature flag id
+   * @return The feature flag
+   */
+  @GET
+  @Path("/{id}")
+  @Produces(MediaType.APPLICATION_JSON)
+  @PermitAll
+  public Response getFeatureFlagById(@PathParam("id") String id) {
+    try {
+      FeatureFlag flag = featureFlagService.getFeatureFlagById(id);
+      return Response.ok(flag).build();
+    } catch (Exception e) {
+      return createExceptionResponse(e);
+    }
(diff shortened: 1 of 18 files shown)


src/main/java/org/broadinstitute/consent/http/resources/FeatureFlagResource.java
```diff
@@ -0,0 +1,130 @@
...
+
+  private final FeatureFlagService featureFlagService;
+
+  @Inject
+  public FeatureFlagResource(FeatureFlagService featureFlagService) {
+    this.featureFlagService = featureFlagService;
+  }
+
+  /**
+   * Get all feature flags (unauthenticated)
```
rushtong: Anything behind `/api` requires an auth token, so this does require an authenticated user.

rushtong: One way to solve this is to have two different resource classes, one that handles `@Path("feature")` and an admin one that handles `@Path("api/feature")`

src/test/java/org/broadinstitute/consent/http/db/FeatureFlagDAOTest.java
```diff
@@ -0,0 +1,135 @@
...
+  @Test
+  void testFindById_NotFound() {
+    String id = "non-existent-" + RandomStringUtils.secureStrong().nextAlphanumeric(10);
+    FeatureFlag flag = featureFlagDAO.findById(id);
+    assertNull(flag);
+  }
+
+  @Test
+  void testFindAll() {
+    String id1 = "feature-1-" + RandomStringUtils.secureStrong().nextAlphanumeric(10);
```
rushtong: FYI, there are shortcuts to `RandomStringUtils.secureStrong().nextAlphanumeric()` (and some others as well) in `AbstractTestHelper`

src/main/java/org/broadinstitute/consent/http/resources/FeatureFlagResource.java
fboulnois: Let's discuss this in standup tomorrow. I'm wondering if we should defer this complexity into the future, with the understanding that we may not be able to feature flag everything we want.

rushtong: I'm open to any approach here. Another potential approach is that `@Path` can take a regex: https://eclipse-ee4j.github.io/jersey.github.io/documentation/latest3x/jaxrs-resources.html  We used to use them in consent when we supported basic auth. Here's an example from long ago: https://github.com/DataBiosphere/consent/blob/ec544d1c62498d10f84a91172e406d572b6b8d1d/src/main/java/org/broadinstitute/consent/http/resources/DACUserResource.java#L36

fboulnois: Future approach: maybe split into `/api/feature` and `/feature` . Our decision based on today's conversation was to continue with the authenticated approach for now.

rushtong (approved):
Nice work, thank you 👍🏽

src/main/java/org/broadinstitute/consent/http/service/FeatureFlagService.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  /**
+   * Create or update a feature flag
+   *
+   * @param id The feature flag id
+   * @param value The feature flag value
+   * @return The created or updated feature flag
+   */
+  public FeatureFlag createOrUpdateFeatureFlag(String id, String value) {
```
otchet-broad: Since we're not recording who enabled or disabled a feature flag in the database, do you think we should log the user who's adding/removing them?

otchet-broad (approved):
Looks god!
One small item: I think we might want to add some audit logging for adding/removing feature flags so we can track who turned them on or off in a specific environment.

src/main/java/org/broadinstitute/consent/http/service/FeatureFlagService.java
fboulnois: That's a good idea.

src/main/java/org/broadinstitute/consent/http/service/FeatureFlagService.java
```diff
@@ -0,0 +1,129 @@
...
+      throw new IllegalArgumentException("Feature flag id cannot be empty");
+    }
+    if (value == null) {
+      throw new IllegalArgumentException("Feature flag value cannot be null");
+    }
+
+    if (featureFlagDAO.exists(id)) {
+      featureFlagDAO.update(id, value);
+      featureFlagDAO.insertAudit(userId, id, "UPDATE");
+    } else {
```
rushtong: Minor suggestion - it would be nice to wrap the dao calls in a single transaction. For features that see more usage, this would be more important, but for this, I don't think it's critical.

rushtong (approved):
Nice addition 👍🏽

src/main/java/org/broadinstitute/consent/http/service/FeatureFlagService.java
fboulnois: done!

otchet-broad (approved):
👍
Looks good!

rushtong (approved):
Nice, thank you 👍🏽

PR #2788, opened 2026-01-27
[DT-2779] Remove validation for data location fields

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2779

### Summary

This PR adds support for an `Other` Data Location option and removes the requirements and validation for the Data Location dropdown and the URL fields, making both optional. 

----

- Verify all tests go green

### DIFF
--- a/src/main/resources/dataset-registration-schema_v1.json
+++ b/src/main/resources/dataset-registration-schema_v1.json
@@ -669,6 +669,7 @@
           "format" : "uri",
           "label" : "Free text field for entering URL of data",
           "description" : "Free text field for entering URL of data",
+          "minLength" : 1
         },
         "numberOfParticipants" : {
           "type" : "integer",
--- a/src/main/resources/dataset-registration-schema_v1.json
+++ b/src/main/resources/dataset-registration-schema_v1.json
@@ -684,7 +684,6 @@
           "format" : "uri",
           "label" : "Free text field for entering URL of data",
           "description" : "Free text field for entering URL of data",
-          "minLength" : 1
         },
         "numberOfParticipants" : {
           "type" : "integer",
--- a/src/main/resources/dataset-registration-schema_v1.json
+++ b/src/main/resources/dataset-registration-schema_v1.json
@@ -460,21 +460,6 @@
       ],
       "allOf" : [
         {
-          "if" : {
-            "properties" : {
-              "dataLocation" : {
-                "const" : "Not Determined"
-              }
-            }
-          },
-          "then" : {},
-          "else" : {
-            "required" : [
-              "url"
-            ]
-          }
-        },
-        {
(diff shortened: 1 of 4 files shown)


eweitz (approved):

src/main/resources/dataset-registration-schema_v1.json
otchet-broad, line 657: I thought we were going to get rid of the 'enum' and allow this to be any string value.

As I read the ticket:
"Neither Data Location Dropdown nor URL should be required or validated in the backend"

The Dropdown values are the enum - am I reading the ticket incorrectly?

fboulnois (approved):
👍 modulo Elliot's question.

src/main/resources/dataset-registration-schema_v1.json
kevinmarete: The ticket says to add 'OTHER' as an option and both data location and url should not be required. It does not say remove the dropdown. Where would we add OTHER?

otchet-broad: If we remove this enum entirely and treat the content as a string, there's nothing in the backend that needs "other" added to it there.  The frontend can send it and the backend should happily store it.  :-)

kevinmarete: What about the logic to display the URL when OTHER is selected in the Data Library.

kevinmarete: What about the logic to display the URL when OTHER is selected in the Data Library.

otchet-broad: The backend will still need to return whatever value was stored so that the front end can perform this action.  The change here is that the backend will no longer validate the strings the frontend sends.  This will work much like the way the 'assets' field now works.  The backend doesn't validate the content in the schema, it just stores it and returns it when the client asks for a dataset/consent group.  This one does get a little different treatment in that the backend does make use of a specific value in this field, but I'd expect for those cases, we'll make a constant (or enum) and use the values for the logic the backend needs.  I'll make a parking lot item to discuss further at standup.

kevinmarete: The next ticket I am working on is to bring validation to the assets. This is a present gap not a feature. Why didn't this come up when refining the ticket. I totally disagree with your suggestion as this is not what has been specified in the ticket. We cannot change the AC during PR review if the implementation solves the existing AC.

 Does the current implementation not resolve the requirements/issue on the ticket?

otchet-broad: Thanks for the discussion at standup around this!

otchet-broad (approved):
👍
Thanks for the discussion at standup today.

PR #2793, opened 2026-02-03
DT-2875: Ontology indexing improvements

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-2875

## Summary
Addresses some minor bugs and adds a new feature to ontology indexing
* New admin feature to delete by ontology type. Useful for developer testing so an ontology can be re-indexed
* Use a different reasoner that is slightly (10-20%) faster which is necessary to find superclasses for terms
* Move indexing to an async service method. Indexing can be slow so return to the client once the index has been kicked off. Additional logging added for both success and failure cases
* Minor bug fix for searching on multiple terms to trim white space from terms before querying on them.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/ontology/OntologyDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/ontology/OntologyDAO.java
@@ -13,6 +13,7 @@ import org.jdbi.v3.sqlobject.customizer.BindMethods;
 import org.jdbi.v3.sqlobject.statement.BatchChunkSize;
 import org.jdbi.v3.sqlobject.statement.SqlBatch;
 import org.jdbi.v3.sqlobject.statement.SqlQuery;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
 import org.jdbi.v3.sqlobject.transaction.Transactional;
 
 public interface OntologyDAO extends Transactional<OntologyDAO> {
@@ -29,6 +30,9 @@ public interface OntologyDAO extends Transactional<OntologyDAO> {
 
   @SqlQuery("SELECT COUNT(*) FROM ontology_index")
   int countTerms();
+
+  @SqlUpdate("DELETE FROM ontology_index where ontology = :ontology")
+  void deleteByOntology(@Bind("ontology") String ontology);
 
   @Json
   default StreamingOutput findByTermIds(@Bind("ids") String[] ids) {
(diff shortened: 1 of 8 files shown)


src/main/java/org/broadinstitute/consent/http/service/ontology/OntologyIndexService.java
```diff
@@ -64,7 +64,7 @@ public Collection<OntologyTerm> generateTerms(OntologyType ontologyType)
     Optional<IRI> versionIRI = ontologyID.getVersionIRI();
     String version = versionIRI.map(IRI::toString).orElse("unknown_version");
     OWLReasonerFactory reasonerFactory = new StructuralReasonerFactory();
-    OWLReasoner reasoner = reasonerFactory.createNonBufferingReasoner(ontology);
+    OWLReasoner reasoner = reasonerFactory.createReasoner(ontology);
```
rushtong, line 67: The standard reasoner is roughly 10-20% faster in processing parent classes for terms.

src/main/java/org/broadinstitute/consent/http/service/ontology/OntologyDAO.java
```diff
@@ -39,7 +43,8 @@ default StreamingOutput findByTermIds(@Bind("ids") String[] ids) {
             WHERE LOWER(id) = ANY (:ids)
                OR LOWER(obo_id) = ANY (:ids)
             """;
-    String[] lowerIds = Arrays.stream(ids).map(String::toLowerCase).toArray(String[]::new);
+    String[] lowerIds =
+        Arrays.stream(ids).map(String::toLowerCase).map(String::trim).toArray(String[]::new);
```
rushtong, line 47: `String::trim` was inadvertently removed from the original PR that introduced this code.

fboulnois (approved):
👍

kevinmarete (approved):
👍🏾

src/test/java/org/broadinstitute/consent/http/service/OntologyServiceTest.java
```diff
@@ -169,13 +170,35 @@ void testTranslateDataUseError() {
...
 
+  @Test
+  @SuppressWarnings("java:S2925") // Thread.sleep needed to wait for async callback
+  void testIndexOntologyFailure() throws Exception {
+    User user = new User();
+    user.setUserId(1);
+    String message = "Failed to generate terms";
+    when(indexService.generateTerms(OntologyType.DUO)).thenThrow(new RuntimeException(message));
+
+    service.indexOntology(user, OntologyType.DUO);
```
otchet-broad, line 193: I'm wondering if there's a better way we can expose the underlying thread (or pool) to allow us to know when these complete so we can avoid the sleep.

otchet-broad (approved):
👍

src/test/java/org/broadinstitute/consent/http/service/OntologyServiceTest.java
rushtong: Great idea, I'll look more into this.

rushtong: I'm going to look into this separately so the blocked tickets can move forward.

PR #2802, opened 2026-02-11
[DT-2759] Don't save empty files; clean up empty files.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2759](https://broadworkbench.atlassian.net/browse/DT-2759)

### Summary

Feature: prevent the (long-term) storage of empty files.  This is implemented in such a way that files are first uploaded to GCS and then checked if they have a file size of zero so that consent doesn't need to read the file into memory and rewire it into an input stream to upload to GCS.  If the file is zero bytes after it is uploaded to GCS, the file is deleted and the FileStorageObject is not persisted.

Feature: Adds an endpoint that cleans up empty NIH Institutional Certification and Alternate Sharing Plan files that may have been included in study or consent group registrations.

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
```diff
@@ -395,13 +404,17 @@ private List<FileStorageObject> uploadFilesForDataset(
...
               uploadedFileCache,
               user,
               String.format(NIH_INSTITUTIONAL_CERTIFICATION_NAME, consentGroupIdx),
-              FileCategory.NIH_INSTITUTIONAL_CERTIFICATION));
+              FileCategory.NIH_INSTITUTIONAL_CERTIFICATION);
+      if (!gcsService.hasBytes(fso.getBlobId())) {
+        gcsService.deleteDocument(fso.getBlobId().getName());
+      }
+      consentGroupFSOs.add(fso);
     }
```
kevinmarete, line 418: I see a common pattern of allowing the upload to proceed, then using `gcsService` to determine the size of the file, and then deleting it if it has no bytes. This can help clean up failed uploads as well. However, wouldn't it be easier to check the file size before in the uploadFile function, to just prevent upload if the file has no bytes?

```
FormDataContentDisposition contentDisposition = bodyPart.getFormDataContentDisposition();
if (contentDisposition != null) {
  long size = contentDisposition.getSize();
  if (size == 0) {
    throw new BadRequestException("File cannot be empty");
  }
}
```

kevinmarete (approved):
Your implementation is great, just a comment on a possible improvement.

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
otchet-broad: The size parameter is optional and user agents aren't required to send it.  [https://www.rfc-editor.org/in-notes/rfc2183.html](https://www.rfc-editor.org/in-notes/rfc2183.html).  In practice, most (including the version of Chrome I'm running) don't.

It would be entirely appropriate to also ask 'why can't you use the .available() method on the InputStream' in question and see if there are bytes?  Turns out, that option also isn't reliable as it always returns 0 for this type of input stream.  You would have to read some part of the input stream and then convert that part back so that the gcs client could get a handle to the full input stream.

Given all of the pain with doing this with the input stream, I decided against that approach as it complicates our code.

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
```diff
@@ -635,4 +635,14 @@ public Response getNihInstitutionalCertification(
       return Response.status(Response.Status.INTERNAL_SERVER_ERROR).build();
     }
   }
+
+  @PUT
+  @Produces(MediaType.APPLICATION_OCTET_STREAM)
+  @Path("/cleanupEmptyCertificationAndAlternativeSharingFiles")
```
rushtong, line 641: Do we anticipate that we'll need this more than once on production? If not, maybe we can ticket this for removal.

src/main/java/org/broadinstitute/consent/http/service/DatasetRegistrationService.java
rushtong: I had similar thoughts as Kevin, but the current approach makes sense.

rushtong (approved):
Nice work, thank you 👍🏽

src/main/java/org/broadinstitute/consent/http/resources/DatasetResource.java
otchet-broad: I think the resource can likely be removed after we run this in all of the environments.  I've made a ticket and attached it to tech health.

PR #2806, opened 2026-02-12
DT-2945: Handle null datasets and return 404 if none found

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2945

### Summary
In the case where there are no datasets found to index, return a 404 instead of throwing an NPE when attempting to convert to a `DatasetTerm`

----

- Verify all tests go green


kevinmarete (approved):

otchet-broad (approved):
👍
Thanks for adding this to the swagger docs as well!

fboulnois (approved):
👍

PR #2810, opened 2026-02-13
[DT-2946] Switch to using election date from vote date in query + defense

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2946](https://broadworkbench.atlassian.net/browse/DT-2946)

### Summary

In production, it's observed that votes may not have 'create_date' set to a value, but it was confirmed that the election create date is always set.  This PR switches the query to use the election date.

Other changes:
Adds defensive logic to ensure that create date or dar code is present before passing a reminder to the template.
Ensures an election create_date has better resolution so that the underlying database value can be compared between queries

----

- Verify all tests go green


kevinmarete (approved):
👍🏾

rushtong (approved):
👍🏽

PR #2811, opened 2026-02-17
DT-2531: Replicate the Translate feature from Ontology to Consent

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-2531

## Summary
This PR replicates the current [Ontology Translate](https://consent-ontology.dsde-prod.broadinstitute.org/#/Translate/post_translate) functionality internally to Consent. Usages of this feature described in the Testing section. There is an additional translation feature, `/translate/summary` that will be addressed in a separate ticket. 

### TranslationUtil
This class is a partial copy of [`TextTranslationServiceImpl`](https://github.com/DataBiosphere/consent-ontology/blob/3c0c70804f692424733d8725d23596107fb1a64a/src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java). The functions brought over are:
1. [`translate`](https://github.com/DataBiosphere/consent-ontology/blob/3c0c70804f692424733d8725d23596107fb1a64a/src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java#L269C17-L269C26)
2. [`translatePurpose`](https://github.com/DataBiosphere/consent-ontology/blob/3c0c70804f692424733d8725d23596107fb1a64a/src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java#L117C17-L117C33) is fully replaced by `translate`
3. [`translateDataset`](https://github.com/DataBiosphere/consent-ontology/blob/3c0c70804f692424733d8725d23596107fb1a64a/src/main/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImpl.java#L76C17-L76C33) is fully replaced by `translate`

There are some `DataUse` conditions in current Consent code that are NOT covered in this code. New tickets will be generated for that work.

### TranslationUtilTest
This class is a partial copy of [`TextTranslationServiceImplTest`](https://github.com/DataBiosphere/consent-ontology/blob/3c0c70804f692424733d8725d23596107fb1a64a/src/test/java/org/broadinstitute/dsde/consent/ontology/datause/services/TextTranslationServiceImplTest.java#L38). There are some additional tests to cover cases not originally covered, but all existing logic tests are mostly unchanged aside from minor syntax and variable updates.

### Testing
This code path can be tested in these ways:
- Update a dataset's data use via the Admin API: `/api/dataset/{id}/datause`
[...]

PR #2815, opened 2026-02-18
DT-2531: Migrate translate usages from UseRestrictionConverter to OntologyService

rushtong:
### Addresses
Follow up work for https://broadworkbench.atlassian.net/browse/DT-2531

### Summary
This is follow up work to https://github.com/DataBiosphere/consent/pull/2811. In that PR, we replicated the translate feature. Unfortunately, I missed one of the usages which was in `UseRestrictionConverter` which called the same remote endpoint. This PR removes that and replaces all usages with the same method in `OntologyService`.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/UseRestrictionConverter.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/UseRestrictionConverter.java
@@ -1,16 +1,8 @@
 package org.broadinstitute.consent.http.service;
 
-import jakarta.ws.rs.InternalServerErrorException;
-import jakarta.ws.rs.client.Client;
-import jakarta.ws.rs.client.Entity;
-import jakarta.ws.rs.client.WebTarget;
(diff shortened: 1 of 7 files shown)


src/main/java/org/broadinstitute/consent/http/service/UseRestrictionConverter.java
```diff
@@ -130,22 +116,4 @@ public DataUse parseDataUsePurpose(DataAccessRequest dar) {
     }
     return dataUse;
   }
-
-  public String translateDataUse(DataUse dataUse, DataUseTranslationType type) {
```
rushtong, line 134: This is the meat of the change. All services that called this method now call the newer version in `OntologyService`

kevinmarete (approved):

otchet-broad (approved):
👍
Nice!

PR #2817, opened 2026-02-20
DT-2918: Add user status info to the user in the GET /me call

rushtong:
### Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DT-2918
See front end PR here: https://github.com/DataBiosphere/duos-ui/pull/3337

### Summary
This PR makes a different call to Sam to get the user's combined state when any API call is made (during the initial OAuth filter process). It then tacks on an additional field to `UserStatusInfo` to reflect the current `tosAccepted` that we would normally get from the diagnostics call. The generated `userStatusInfo` is then added to the user response object in the current `GET /api/user/me` call.

Example response snippet from `GET /api/user/me`:
```
...
  "userStatusInfo": {
    "userSubjectId": "103..................318",
    "userEmail": "gr......@broadinstitute.org",
    "enabled": true,
    "tosAccepted": true
  },
...
```

The reason to do this is to reduce the number of API calls the UI has to make when a user signs in. Currently, it calls the me API and then after waiting for that, it makes a call to the diagnostics API which returns terms of service information:

Screenshot 2026-02-20 at 1 01 28 PM

In a future UI PR, we can trim that down and skip the diagnostics API call altogether because we'll have terms of service information stored directly on the user.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
+++ b/src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
@@ -72,7 +72,7 @@ public class AuthorizationHelper implements ConsentLogger {
    */
   protected UserStatusInfo getUserStatusInfo(DuosUser duosUser) {
     try {
-      return samService.getRegistrationInfo(duosUser);
+      return samService.getCombinedUserStatusInfo(duosUser);
     } catch (NotFoundException e) {
       try {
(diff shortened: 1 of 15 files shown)


kevinmarete (approved):
Nice

fboulnois (approved):
thank you 👍

PR #2827, opened 2026-03-04
DT-2993: Clean up missing User entities and FKs

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-2993

## Summary
This PR addresses the following issues:
1. Current queries often (but not always) expect that a user id exists. In production, there are two instances of deleted users such that `INNER JOIN` queries on those user ids will fail. This PR repopulates those missing user ids with a special case `redacted` user.
2. There are a number of missing FK constraints which allowed for this condition to exist. This PR adds new constraints to prevent this from happening in the future.
3. `UserService.deleteUserByEmail` is the original source of these problems so removing it and all supporting code should prevent this from happening again.

There still exist in the codebase `LEFT JOIN`s on the `users` table that we're not trying to fix here. New tickets will be filed to look more closely into the application logic that assumes the possibility of null user ids. After reviewing each production table, we've discovered two completely empty and unused tables (`datarequest` and `researchpurpose`) which will also be ticketed for removal.

### TODO:
* New tickets to address follow-on tasks
  * Remove unused tables: [DT-3066](https://broadworkbench.atlassian.net/browse/DT-3066)

----

- Verify all tests go green

[DT-3066]: https://broadworkbench.atlassian.net/browse/DT-3066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/org/broadinstitute/consent/http/db/DarCollectionDAO.java
```diff
@@ -69,51 +69,6 @@ List<DarCollection> findDARCollectionByCollectionIds(
...
-          + "INNER JOIN data_access_request dar on c.collection_id = dar.collection_id "
-          + "LEFT JOIN dar_dataset dd on dd.reference_id = dar.reference_id "
-          + "LEFT JOIN institution i ON i.institution_id = u.institution_id "
-          + "LEFT JOIN ("
-          + "   SELECT election.*, MAX(election.election_id) OVER (PARTITION BY election.reference_id, election.election_type, election.dataset_id) AS latest FROM election "
-          + "   WHERE LOWER(election.election_type) = 'dataaccess' OR LOWER(election.election_type) = 'rp' "
-          + ") AS e "
-          + "   ON (dar.reference_id = e.reference_id AND dd.dataset_id = e.dataset_id) AND (e.latest = e.election_id OR e.latest IS NULL) "
-          + "WHERE (LOWER(data->>'status')!='archived' OR data->>'status' IS NULL) ")
-  List<DarCollection> findAllDARCollections();
```
rushtong, line 115: Only usages were in tests

fboulnois (approved):
👍

fboulnois:
I wonder if the users should get a completely random ID, that way their user ID does not appear in the UI.

otchet-broad:
Looks good overall!  Nice work.

I'd recommend deleting the UserService.[deleteUserByEmail](https://github.com/DataBiosphere/consent/blob/e02bfa4da2d8658eeb4be4be2247ff45129c5f32/src/main/java/org/broadinstitute/consent/http/service/UserService.java#L262) method.  We already removed it from the API but it will blow up with these changes.

otchet-broad:
Along with Florian's thinking, should we prioritize DT-1470?

otchet-broad (approved):
👍

rushtong:
> I wonder if the users should get a completely random ID, that way their user ID does not appear in the UI.

Great question - digging into this. There is a unique constraint on that column which is why I chose the current approach.

rushtong:
> Looks good overall! Nice work.
> 
> I'd recommend deleting the UserService.[deleteUserByEmail](https://github.com/DataBiosphere/consent/blob/e02bfa4da2d8658eeb4be4be2247ff45129c5f32/src/main/java/org/broadinstitute/consent/http/service/UserService.java#L262) method. We already removed it from the API but it will blow up with these changes.

As a follow up to this, there are a good number of methods to clean up as a result 👍🏽

rushtong:
@fboulnois @otchet-broad - I made some additional changes here based on additional feedback, PTAL, thank you.
* Updated the user email value to look like a random string which is just a base64 encoded version of the missing user id
* Significant cleanup of unused code after removing `UserService.deleteUserByEmail`
* Minor updates based on co-pilot feedback.

fboulnois:
> @fboulnois @otchet-broad - I made some additional changes here based on additional feedback, PTAL, thank you.

Base64 has a signature like `==` and other characters. Do we need to treat the user id as a secret? If so, we probably need to hash or encrypt the value with a salt so it can't be reconstructed after the fact.

rushtong:
> > @fboulnois @otchet-broad - I made some additional changes here based on additional feedback, PTAL, thank you.
> 
> Base64 has a signature like `==` and other characters. Do we need to treat the user id as a secret? If so, we probably need to hash or encrypt the value with a salt so it can't be reconstructed after the fact.

In general, database IDs are not secret - they are visible in a variety of ways. The only place these values will be visible is in the Admin user management screen since these users will not have a valid email domain or institution.

fboulnois (approved):
👍

src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -259,58 +230,11 @@ public List<SimplifiedUser> getUsersByDaaId(Integer daaId) {
     return users.stream().map(SimplifiedUser::new).toList();
   }
 
-  public void deleteUserByEmail(String email, Integer auditUserId) {
```
otchet-broad, line 262: Nice!

otchet-broad (approved):

PR #2830, opened 2026-03-11
DT-2963: Add data location URL to emails that go to Researchers

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-2963

## Summary
This adds the `dataLocationUrl` dataset property (if it exists) to the underlying `DatasetMailDTO` object that populates emails. This model is used in email sent to the following users in the following conditions:

* **Researcher**: alerted when a DAR is approved via DAC or RADAR
* **Researcher**: alerted when a PR is approved via DAC or RADAR
* DAC Chairs, Members: alerted that a dataset was approved via RADAR
* Data Custodian: alerted that a researcher has been approved

However, we are only modifying the emails that go out to researchers. The DAC and Data Custodian emails are not modified, but could be if needed in the future. 

Drive-by improvements:
1. We migrate `VoteService` to use a `DAOContainer` to populate DAO classes instead of directly injecting them one-by-one.
2. We convert the modified html templates to ftl templates instead.
3. Convert `DatasetMailDTO` to a record class.

### Example Email
Screenshot 2026-03-12 at 3 46 51 PM

----
Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

PR #2832, opened 2026-03-17
DT-3091: Update freemarker template accessors for record model class

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-3091

## Summary
In https://github.com/DataBiosphere/consent/pull/2830, we converted the model class used in several email templates to a record class. In local functional and unit testing, record class field accessors work as expected. In production, we now see a template processing exception on these model class fields:
```
freemarker.core.NonStringOrTemplateOutputException: For "${...}" content: Expected a string or something automatically convertible to string (number, date or boolean), or "template output" , but this has evaluated to a method+sequence (wrapper: f.e.b.SimpleMethodModel):
==> dataset.identifier  [in template "researcher-dar-approved.ftl" at line 43, column 39]

----
Tip: Maybe using obj.something() instead of obj.something will yield the desired value
----

----
FTL stack trace ("~" means nesting-related):
	- Failed at: ${dataset.identifier}  [in template "researcher-dar-approved.ftl" at line 43, column 37]
----
```

This PR migrates all usages to method based field accessors for the templates that use a `DatasetMailDTO` model class.

### Screens
Researcher DAR Approved (email type 8) with null data location
Screenshot 2026-03-17 at 3 58 15 PM

---
Researcher DAR Approved (email type 8) with valid data location
Screenshot 2026-03-17 at 3 58 19 PM

---
Data Custodian Approval (email type 7)
Screenshot 2026-03-17 at 4 19 49 PM

---
Researcher PR Approved (email type 24)
Screenshot 2026-03-17 at 4 29 33 PM

---
DAC RADAR Approval (email type 32)
Screenshot 2026-03-17 at 4 27 04 PM

---

- Verify all tests go green


kevinmarete (approved):

fboulnois (approved):
Good catch 👍

fboulnois (approved):
Can we add a test for this behavior so we do not see regressions?

rushtong:
> Can we add a test for this behavior so we do not see regressions?

I'm unable to reproduce the prod behavior locally or in a unit test. In fact, template processing in unit tests work for both cases. The only freemarker configuration we have is here: https://github.com/broadinstitute/terra-helmfile/blob/b7ffb7e6fd7c3bc0556b640e6cadb49a0628bf1e/charts/consent/templates/_consent.yaml.tpl#L63 and that does not explain why prod would behave differently than a local instance.

PR #2845, opened 2026-03-27
[DT-2863] Add json bucket to user object so that we can store profile links.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2863](https://broadworkbench.atlassian.net/browse/DT-2863)

### Summary

Backend facility to store arbitrary data on the user object needed by the front end.  Like patterns on other objects, this will hold a map of keys and values so that the front end can store and retrieve the types of profile information this ticket requires.

Building the User object from the persistence tier is a complex operation handled by two patterns bean mapping and row mapping.  To facilitate bean mapping, jdbi3-json has been added.  This facility makes it trivial to map a json column in postgres to a Java bean.  Row mapping uses pre-existing features to translate between the persistence tier and the object model.

----

- Verify all tests go green


eweitz (approved):
Convenient!

src/main/java/org/broadinstitute/consent/http/db/mapper/UserWithRolesMapper.java
```diff
@@ -18,6 +19,7 @@
 public class UserWithRolesMapper implements RowMapper<User>, RowMapperHelper {
 
   private final Map<Integer, User> users = new HashMap<>();
+  private final Gson gson = new Gson();
```
rushtong, line 22: Any reason not to use the GsonUtil class here?

src/main/java/org/broadinstitute/consent/http/db/UserDAO.java
```diff
@@ -407,11 +413,13 @@ SELECT DISTINCT ON (u.user_id)
...
-          + " LEFT JOIN roles r ON r.role_id = ur.role_id "
-          + " WHERE LOWER(r.name) = 'signingofficial' "
-          + " AND u.institution_id = :institutionId")
+      """
+          SELECT u.user_id, u.display_name, u.email, u.user_data FROM users u
+          LEFT JOIN user_role ur ON ur.user_id = u.user_id
+          LEFT JOIN roles r ON r.role_id = ur.role_id
+          WHERE LOWER(r.name) = 'signingofficial'
+          AND u.institution_id = :institutionId
+        """)
```
rushtong, line 422: Nice

rushtong (approved):
👍🏽

fboulnois (approved):
Looks good 👍

PR #2850, opened 2026-04-02
DT-3047: Save DAA IDs when submitting a DAR or PR

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-3047

### Summary
This PR adds the DAA IDs for the datasets in a DAR or PR when the entity is saved. These ids are stored in the DAR's json data package and can be referred to in other workflows.

----

- Verify all tests go green


kevinmarete (approved):

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequestData.java
```diff
@@ -161,6 +161,7 @@ public class DataAccessRequestData {
   private String piName;
   private String piEmail;
   private String piCountryOfOperation;
+  private List<Integer> daaIds;
```
otchet-broad: I'm wondering if this should be a Set.  Lists can have duplicate values, Sets won't.
Why would we want to record duplicate ids?

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -235,6 +238,11 @@ public DataAccessRequest createDataAccessRequest(
     }
     String referenceId;
     List<Integer> datasetIds = dataAccessRequest.getDatasetIds();
+
+    // Save the agreed upon DAAs at the time of DAR submission
+    List<Integer> daaIds = daaDAO.findDaaIdsByDatasetIds(datasetIds);
```
otchet-broad: Could we add the DAA IDs that were agreed to to the DataAccessRequestData object and store those values?

We have other tickets to verify the value in DataAccessRequestData contains the full list needed.  I expect that we'll call something in DataAccessRequestService on submission (not draft) that verifies the list contains the needed elements or we'll reject the creation.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -280,6 +288,12 @@ public DataAccessRequest createProgressReport(
     List<Integer> progressReportDatasetIds = progressReport.getDatasetIds();
     Set<Integer> darDatasetIds =
         dataAccessRequestDAO.findDatasetApprovalsByDar(parentDar.getReferenceId());
+
+    // Save the agreed upon DAAs at the time of PR submission
+    DataAccessRequestData darData = progressReport.getData();
+    List<Integer> daaIds = daaDAO.findDaaIdsByDatasetIds(new ArrayList<>(progressReportDatasetIds));
```
otchet-broad: Same here.  Why not just save the DAAs provided in DataAccessRequestData?

src/test/java/org/broadinstitute/consent/http/models/DataAccessRequestDataTest.java
```diff
@@ -101,4 +102,554 @@ void testPopulateDARDataWithNullJson() {
...
+
+  // String fields
+
+  @Test
+  void testGetSetRus() {
+    DataAccessRequestData data = new DataAccessRequestData();
+    assertNull(data.getRus());
+    data.setRus("rus text");
+    assertEquals("rus text", data.getRus());
+  }
```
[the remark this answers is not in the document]
otchet-broad: Thanks for adding the coverage.  I'm glad it's in place.  I think copilot is undervaluing the fact that getters and setters can transform and your tests are covering that.

There are a few methods still missing coverage (validateOntologyEntries included) that might be worth eventually adding coverage around, but they'll come with time.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
rushtong: When a DAR comes into the service layer, there are no DAA id values in the `DataAccessRequestData`. The UI does not populate that value in the POST request. A different approach would be to populate the values in the `DatasetService.enforceDAARestrictions` method, but that would be overloading (in the colloquial sense) that method's intention.

otchet-broad: I think we should add the DAAs to the DataAccessRequestData object so we know what IDs were loaded when the client submitted the DARs.  If we just add them at submission time, how can we be sure they were actually reviewed by the client?

otchet-broad (approved):

otchet-broad:
Thanks for the updates!  Looks good!

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
rushtong: Per conversation, we're removing the explicit set values and relying on the UI to populate correct values.

otchet-broad (approved):
👍

PR #2855, opened 2026-04-06
DT-3167: Bug fix for daa population in reducer

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-3167

### Summary
The reducer class was overwriting the user's LC on every row so only the last DAA processed was associated to the LC. This PR updates the logic so it only writes the LC if it doesn't already exist and then adds the DAA. Also updates a test for ensuring multiple DAAs are populated.

Sonar had quite a few warnings about this class so it is also refactored to a large degree to reduce cognitive complexity and ignored parameters.

----

- Verify all tests go green


eweitz (approved):

fboulnois (approved):
👍

PR #2859, opened 2026-04-08
[DT-3063, DT-3065] Store relationship between DARs/Progress Reports and DAAs, Flag DARs/PRs for SO approval when user not pre authorized for DAAs

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3063](https://broadworkbench.atlassian.net/browse/DT-3063)
[https://broadworkbench.atlassian.net/browse/DT-3065](https://broadworkbench.atlassian.net/browse/DT-3065)

### Summary

- Provides an API for obtaining the mapping of the DAAs to a list of datasets so that the client can display the relationship
- Validates the DAAs submitted match the DAR or Progress Report  (when not a closeout) expected DAAs and routes the DAR to either the SO or the DAC.
- Stores the relationship between data_access_request table entries with DAAs on DAR or Progress Report (when not a closeout) submission

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/db/mapper/DaaDatasetReducer.java
```diff
@@ -0,0 +1,31 @@
...
+
+  @Override
+  public void accumulate(Map<Integer, Set<Integer>> container, RowView rowView) {
+    int daaId = rowView.getColumn("daa_id", Integer.class);
+    Integer datasetId = rowView.getColumn("dataset_id", Integer.class);
+    container.computeIfAbsent(daaId, k -> new HashSet<>()).add(datasetId);
+  }
+
+  @Override
+  public Stream<Entry<Integer, Set<Integer>>> stream(Map<Integer, Set<Integer>> container) {
```
rushtong, line 28: This looks super useful 👍🏽

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -243,18 +247,20 @@ public DataAccessRequest createDataAccessRequest(
...
+      Integer darId =
+          dataAccessRequestDAO.insertDataAccessRequest(
+              collectionId,
+              referenceId,
+              user.getUserId(),
+              now,
+              now,
+              now,
+              darData,
+              user.getEraCommonsId());
```
rushtong: It would be nice if these were both in a single transaction. I have some `WITH` examples that might serve as a template for this in [one of my open PRs](https://github.com/DataBiosphere/consent/pull/2853/changes#diff-b214fb221cc07c674fa4d92cfa827bd90e7612321c2599f6d064644ad5e23471).

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
```diff
@@ -405,6 +405,11 @@ public static DataAccessRequest populateProgressReportFromJsonString(
     // object and not the original DAR.
     originalDataCopy.setReferenceId(referenceId);
 
+    // We need to set the new DAAs that were in place on the DAR because the DAAs may have been
+    // updated
+    // from the original DAR.
+    originalDataCopy.setDaaIds(newData.getDaaIds());
```
rushtong, line 420: Just clarifying that I think we're planning on the FE to populate these values on submission. If not, these will be empty.

src/main/java/org/broadinstitute/consent/http/db/DaaDAO.java
```diff
@@ -299,4 +302,25 @@ WHERE dataset.dataset_id IN (<datasetIds>)
...
+          GROUP BY daa.daa_id, dataset.dataset_id
+          ORDER BY daa.daa_id
+      """)
+  @UseRowReducer(DaaDatasetReducer.class)
+  Map<Integer, Set<Integer>> findDaaIdsByDatasetIds(
+      @BindList(value = "datasetIds", onEmpty = EmptyHandling.NULL_STRING) Set<Integer> datasetIds);
+
+  @SqlBatch(
+      """
+        INSERT INTO dar_daa (dar_id, daa_id) VALUES (:darId, :daaId)
```
rushtong, line 342: In existing code paths, it doesn't look possible to call this twice with the same arguments which would create duplicates. However, it would be nice to have a `ON CONFLICT` clause to prevent that structurally.

src/main/java/org/broadinstitute/consent/http/db/DaaDAO.java
```diff
@@ -299,4 +302,25 @@ WHERE dataset.dataset_id IN (<datasetIds>)
...
+    """)
+  void insertDarDAARelationship(@Bind("darId") Integer darId, @Bind("daaId") Set<Integer> daaIds);
```
rushtong, line 345: Copilot caught this for me - I was unaware of this detail. See current [jdbi doc](https://jdbi.org/#_singlevalue) for reference.

> In JDBI 3, all @SqlBatch parameters are expected to be iterable. A non-iterable scalar must be annotated with @SingleValue to be held constant per row. Without it, behavior is undefined across JDBI point-releases and the batch will fail if the set has more than one element.

> Fix:

```
void insertDarDAARelationship(@Bind("darId") @SingleValue Integer darId,
                              @Bind("daaId") Set<Integer> daaIds);
```

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -438,6 +456,29 @@ protected void validateCommonDarAndProgressReportElements(User user, DataAccessR
...
+        && requiredDaas.size() == dar.getData().getDaaIds().size())) {
+      throw new BadRequestException("All of the DAAs required were not acknowledged.");
+    }
+  }
+
+  private boolean isUserPreAuthorizedForAllDaas(User user, List<Integer> datasetIds) {
+    Set<Integer> datasetDaas =
+        daaDAO.findDaaIdsByDatasetIds(datasetIds).stream().collect(Collectors.toSet());
+
+    Set<Integer> userDaas = user.getLibraryCard().getDaaIds().stream().collect(Collectors.toSet());
```
rushtong, line 507: Any potential for NPEs here? Current code paths look clean but there are likely exception cases w're missing. Might be nice to log potential data errors here.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -438,6 +456,29 @@ protected void validateCommonDarAndProgressReportElements(User user, DataAccessR
...
 
+  @VisibleForTesting
+  protected void hasAcknowledgedRequiredDaas(DataAccessRequest dar) {
+    if (dar.getDatasetIds().isEmpty()) {
+      throw new BadRequestException("At least one dataset is required");
+    }
+
+    Set<Integer> requiredDaas = daaDAO.findDaaIdsByDatasetIds(dar.getDatasetIds());
+
+    if (!(requiredDaas.containsAll(dar.getData().getDaaIds())
```
rushtong, line 497: Legacy DARs/PRs will have null `daaIds`

src/main/resources/assets/paths/daaDatasets.yaml
```diff
@@ -0,0 +1,34 @@
+post:
+  summary: Mapping between DAAs and Datasets
+  description: |
+    Provides insights into which DAA is associated with a list of datasets.  Useful to understand
+    which rules apply to each dataset in a multi-dataset/multi-dac DAR.
```
rushtong: More copilot feedback:

> The description says:
> "Provides insights into which DAA is associated with a list of datasets"
> But the response map is keyed by DAA ID (not dataset ID): { "100": [1, 2, 3] } means DAA 100 governs datasets 1, 2, 3. The description implies the caller can look up a dataset → DAA, but the structure is the reverse. This will confuse API consumers.
> Fix: Clarify the description — e.g., "Returns a mapping of DAA ID → dataset IDs that fall under that DAA."

src/main/resources/assets/paths/daaDatasets.yaml
```diff
@@ -0,0 +1,34 @@
...
+              items:
+                type: integer
+            example:
+              100: [ 1, 2, 3 ]
+              101: [ 5 ]
+      description: A Map of DAA to Datasets the DAA applies to.
+    400:
+      description: Bad Request (malformed list of dataset ids)
+    401:
+      description: Unauthenticated.
```
rushtong: Could use a 500 case.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
rushtong: Sonar is flagging both of these set constructors as simplify-able with a `HashSet` initialization.

src/main/resources/changesets/changelog-consent-2026-04-08-dar-daa.xml
```diff
@@ -0,0 +1,16 @@
...
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                   https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+  <changeSet id="changelog-consent-2026-04-08-dar-daa" author="otchet">
+    <createTable tableName="dar_daa">
+      <column name="dar_id" type="bigint">
+        <constraints primaryKey="false" nullable="false" foreignKeyName="fkDaa_Dar_DarId" referencedTableName="data_access_request" referencedColumnNames="id" />
+      </column>
+      <column name="daa_id" type="bigint">
+        <constraints nullable="false" foreignKeyName="fkDaa_Daa_DaaId" referencedTableName="data_access_agreement" referencedColumnNames="daa_id" />
+      </column>
```
rushtong, line 13: Should we add some indexes on these columns?

src/main/java/org/broadinstitute/consent/http/db/DaaDAO.java
```diff
@@ -299,4 +302,25 @@ WHERE dataset.dataset_id IN (<datasetIds>)
...
+          FROM data_access_agreement daa
+          INNER JOIN dac_daa ON daa.daa_id = dac_daa.daa_id
+          INNER JOIN dac ON dac.dac_id = dac_daa.dac_id
+          INNER JOIN dataset ON dataset.dac_id = dac.dac_id
+          WHERE dataset.dataset_id IN (<datasetIds>)
+          GROUP BY daa.daa_id, dataset.dataset_id
+          ORDER BY daa.daa_id
+      """)
+  @UseRowReducer(DaaDatasetReducer.class)
+  Map<Integer, Set<Integer>> findDaaIdsByDatasetIds(
```
rushtong: There is another `findDaaIdsByDatasetIds` method that is distinctly different from this one. Can we clarify the naming here to avoid confusion?

rushtong (commented):
Still reviewing functionally, but wanted to get this first pass of feedback out in the meantime.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
otchet-broad: The getDaaIds was updated to return an empty set if it's null.  :-)

src/main/resources/changesets/changelog-consent-2026-04-08-dar-daa.xml
otchet-broad: We don't have the read use cases documented yet (e.g. from product), but we could do that now.  I guess I'd just add standard indexes to each column.  Is that what you'd like to see?

src/main/java/org/broadinstitute/consent/http/db/DaaDAO.java
otchet-broad: I want to spend some time looking at this because I think I wrote a test that ends up proving this works.  See `testStoreDarDAARelationshipForPR` in DaaDAOTest.java

src/main/java/org/broadinstitute/consent/http/models/DataAccessRequest.java
otchet-broad: Exactly.  It's the next ticket I'm grabbing.  :-)

src/main/java/org/broadinstitute/consent/http/db/DaaDAO.java
otchet-broad: I've updated the method and added the `@SingleValue` decoration.   No change in the test result.  Looking at the docs, they all are using the `@SingleValue` decoration on Collection parameters.  That's different than what I've got in the code.  darId is a single integer value.  I've added an assertion to the `testStoreDarDAARelationshipForPR` to confirm multiple rows are being inserted as expected.

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
otchet-broad: I might go with a slightly different approach to see how we like it, but it will be in one transaction.

src/main/resources/changesets/changelog-consent-2026-04-08-dar-daa.xml
rushtong: That is what I was thinking, but I'm open to waiting on the exact product needs.

rushtong (approved):
Thank you for the updates 👍🏽

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
otchet-broad: Coming back to this one.  It looks like there were already multiple database updates in this method that should effectively unwind if there's an error here, but because of the number of elements involved, I'm wondering if we should make a ticket for this and come back to it later.  Would that be acceptable?  There are also other methods in this class (and related services) that need to be linked as well, hence my hesitation to add this to the scope of this ticket.

rushtong: Sure thing - it would be nice to tighten this up separately

otchet-broad: DT-3176 created.

kevinmarete (approved):

eweitz (approved):

PR #2861, opened 2026-04-13
[DT-3111] Add list endpoint for all FileStorageObjects for an entity

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3111

### Summary

Implements a new GET /api/<entity>/{entityId}/document endpoint to return all non-deleted FileStorageObject records associated with an entity. Results are ordered by create_date DESC and return metadata only (no file bytes).

Includes DAO query using existing FSO select patterns, along with validation for entity existence (404) and read access (403). Returns an empty array when no documents are found.

After

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/resources/DocumentResource.java
```diff
@@ -0,0 +1,114 @@
...
+      return Response.ok(fileStorageObjects).build();
+    } catch (Exception e) {
+      return createExceptionResponse(e);
+    }
+  }
+
+  private String validateEntityAndReadAccess(String entity, String entityId, User user) {
+    return switch (entity.toLowerCase(Locale.ROOT)) {
+      case "dataset" -> validateDatasetReadAccess(entityId, user);
+      case "study" -> validateStudyReadAccess(entityId, user);
```
rushtong: Could we use an enum instead of raw strings?

src/main/java/org/broadinstitute/consent/http/resources/DocumentResource.java
```diff
@@ -0,0 +1,114 @@
...
+    };
+  }
+
+  private String validateDatasetReadAccess(String entityId, User user) {
+    Integer datasetId = parseNumericEntityId(entityId);
+    Dataset dataset = datasetService.findDatasetWithoutFSOInformation(datasetId);
+    if (dataset == null) {
+      throw new NotFoundException("Entity not found");
+    }
+    Dataset authorizedDataset = datasetService.findDatasetById(user, datasetId);
```
rushtong: I'm confused by the double DAO query for the dataset. Line 70 makes a DAO call. `datasetService.findDatasetById` also makes a similar DAO call, but the authorized piece is provided by `datasetService.verifyPublicVisibilityAccess`. We could save a DAO call by just making that method public.

src/main/java/org/broadinstitute/consent/http/db/FileStorageObjectDAO.java
```diff
@@ -103,4 +105,19 @@ void deleteFilesByEntityId(
...
+
+  @UseRowMapper(FileStorageObjectMapperWithFSOPrefix.class)
+  @SqlQuery(
+      """
+          SELECT
+          """
+          + FileStorageObject.QUERY_FIELDS_WITH_FSO_PREFIX
+          + " "
+          + """
+          FROM file_storage_object fso
```
rushtong, line 149: I find this style of query construction challenging to follow. Where I see these `QUERY_FIELDS_*` arguments, I've been replacing them with the actual query string and making these full queries inside the text block and avoiding string concatenation. This is a stylistic preference I have, so this PR shouldn't be held up on this. I would encourage this subject to be discussed with the team for the group preference.

src/main/java/org/broadinstitute/consent/http/resources/DocumentResource.java
```diff
@@ -0,0 +1,114 @@
...
+  @Produces(MediaType.APPLICATION_JSON)
+  @PermitAll
+  public Response findDocumentsByEntity(
+      @Auth DuosUser duosUser,
+      @PathParam("entity") String entity,
+      @PathParam("entityId") String entityId) {
+    try {
+      if (duosUser == null || duosUser.getUser() == null) {
+        throw new ForbiddenException("User does not have permission");
+      }
```
rushtong: This case should not be possible here. If the duosUser or user is null, an error is thrown upstream. Look at `DuosUserAuthenticator.authenticate`. If there is no user, we return `Optional.empty()` which will throw an error in the servlet filter process.

src/main/resources/assets/paths/entityDocumentsByEntityId.yaml
```diff
@@ -0,0 +1,32 @@
+get:
```
rushtong, line 1: I think this could use an `operationId`

rushtong (approved):
Thank you for the changes, looks great! 👍🏽

eweitz (approved):
Looks good, especially once redundant code noted in merge conflict is resolved.

PR #2865, opened 2026-04-15
[DT-3108] Add PUT endpoint to Update FileStorageObject category

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3108

### Summary

Adds a new endpoint to update the `FileCategory` of an existing FileStorageObject (`PUT` `/api/{entity}/{entityId}/document/{id}`) without requiring file re-upload. The change ensures that only the category is mutable while preserving file content and storage references.

Implements validation for valid and entity-appropriate categories, enforces entity ownership and access control, and prevents updates to soft-deleted or mismatched records. Adds DAO support for updating category metadata (`updateUserId`, `updateDate`) and includes Swagger documentation and unit tests for resource and DAO behavior.

After

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/FileStorageObjectService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/FileStorageObjectService.java
@@ -1,17 +1,22 @@
 package org.broadinstitute.consent.http.service;
 
 import com.google.cloud.storage.BlobId;
+import jakarta.ws.rs.BadRequestException;
+import jakarta.ws.rs.ForbiddenException;
 import jakarta.ws.rs.NotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.time.Instant;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.UUID;
 import org.broadinstitute.consent.http.cloudstore.GCSService;
 import org.broadinstitute.consent.http.db.FileStorageObjectDAO;
 import org.broadinstitute.consent.http.enumeration.DocumentEntity;
 import org.broadinstitute.consent.http.enumeration.FileCategory;
+import org.broadinstitute.consent.http.enumeration.UserRoles;
+import org.broadinstitute.consent.http.models.Dataset;
 import org.broadinstitute.consent.http.models.FileStorageObject;
 import org.broadinstitute.consent.http.models.Study;
 import org.broadinstitute.consent.http.models.User;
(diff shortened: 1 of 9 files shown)


src/main/java/org/broadinstitute/consent/http/service/FileStorageObjectService.java
```diff
@@ -152,6 +203,27 @@ private String resolveStudyFsoEntityIdForRead(String entityId, User user) {
     return study.getUuid().toString();
   }
 
+  private String resolveDatasetFsoEntityIdForWrite(String entityId, User user) {
+    Integer datasetId = parseNumericEntityId(entityId);
+    Dataset dataset = datasetService.findDatasetByIdForRead(user, datasetId);
+    if (!user.hasUserRole(UserRoles.ADMIN) && !user.getUserId().equals(dataset.getCreateUserId())) {
```
otchet-broad, line 209: Don't we allow custodians to also edit?

src/main/java/org/broadinstitute/consent/http/service/FileStorageObjectService.java
```diff
@@ -137,6 +169,25 @@ private String resolveFsoEntityIdForRead(User user, String entity, String entity
     };
   }
 
+  private String resolveFsoEntityIdForWrite(User user, String entity, String entityId) {
+    if ("dac".equalsIgnoreCase(entity) || "dar".equalsIgnoreCase(entity)) {
+      if (!user.hasUserRole(UserRoles.ADMIN)) {
+        throw new ForbiddenException("User does not have permission");
+      }
+      Integer numericEntityId = parseNumericEntityId(entityId);
```
otchet-broad, line 177: Don't we need additional logic here to check that if the user is editing something like a DAA, that they need to be the CHAIRPERSON role on the DAC that the FSO is associated with?

kevinmarete: I think all these permissioning issues will be handled in DT-3113

src/main/java/org/broadinstitute/consent/http/service/FileStorageObjectService.java
kevinmarete: Will be handled in DT-3113.

kevinmarete:
To be implemented in this PR: https://github.com/DataBiosphere/consent/pull/2867

PR #2866, opened 2026-04-16
DT-2845: Handle known AzureB2C exception cases

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-2845

## Summary
See companion front end PR: https://github.com/DataBiosphere/duos-ui/pull/3442

This PR adds new error handling for Sam errors wrt AzureB2C errors. When a user first uses MS as an identity provider with a google account, then subsequently uses Google as an identity provider, Sam will respond with an error similar to:
```
AzureB2C authentication Error for user user@test.com: {
    "causes [],
    "exceptionClass":"org.broadinstitute.dsde.workbench.model.WorkbenchException",
    "message":"Cannot update azureB2cId for user 123... because user does not exist or the azureB2cId has already been set for this user",
    "source":"sam",
    "stackTrace":[]
}
```
When ECM encounters this underlying error (because it proxies requests to Sam), it merely responds with a 500 Internal Server Error.

This PR does two things:
1. When ECM responds with a 500 error, we assume the worst and remove the user's current linked account status. This needs to be done to ensure that users are not improperly linked.
2. When Sam responds with an AzureB2C error, we percolate that back up to the client in the primary case of a user getting their own information: `GET /api/user/me`

## Testing
Local testing is challenging. Using swagger in Consent, one uses google sign in. When using swagger in Sam, the google sign in is OIDC. Therefore, the bearer tokens you get in Consent swagger are the more compact ones, e.g. `ya29....`. The bearer tokens you get in Sam swagger are encoded jwts, e.g. `eyJhbGci...`. When using the compact token, Sam does NOT error out. When using the jwt token, Sam DOES error out. So, when using Consent locally, a problematic user will have successful responses from Sam. To test a problematic user, one has to use the jwt token in Consent using curl (OIDC auth does not work in Consent swagger). The DUOS UI uses OIDC, so generates full jwt bearer tokens that DO fail when Consent proxies those tokens against Sam and other services like ECM.

The fact that Sam works with compact tokens but fails with full jwt tokens opens up another approach. Since most UI requests are proxied through Consent, we could extract the compact token and use that against Sam. The UI does also communicate directly with ECM so that would still be problematic unless we were able to do that same manipulation in the UI.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
+++ b/src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
@@ -33,6 +33,7 @@ import java.util.Map;
 import java.util.Objects;
 import org.broadinstitute.consent.http.configurations.ServicesConfiguration;
 import org.broadinstitute.consent.http.enumeration.UserRoles;
+import org.broadinstitute.consent.http.exceptions.SamAzureB2CException;
 import org.broadinstitute.consent.http.models.Acknowledgement;
 import org.broadinstitute.consent.http.models.ApprovedDataset;
 import org.broadinstitute.consent.http.models.AuthUser;
@@ -134,9 +135,11 @@ public class UserResource extends Resource {
     }
   }
 
-  private UserStatusInfo getUserStatusInfo(DuosUser duosUser) {
+  private UserStatusInfo getUserStatusInfo(DuosUser duosUser) throws SamAzureB2CException {
     try {
       return samService.getCombinedUserStatusInfo(duosUser);
+    } catch (SamAzureB2CException e) {
+      throw e;
     } catch (Exception ex) {
       logWarn("Unable to retrieve user status info from Sam: " + ex.getMessage());
       // Intentionally ignore Sam errors here to avoid failing /me on transient outages.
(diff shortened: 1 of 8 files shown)


kevinmarete (approved):

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
```diff
@@ -134,9 +135,11 @@ public Response getUser(@Auth DuosUser duosUser) {
     }
   }
 
-  private UserStatusInfo getUserStatusInfo(DuosUser duosUser) {
+  private UserStatusInfo getUserStatusInfo(DuosUser duosUser) throws SamAzureB2CException {
     try {
       return samService.getCombinedUserStatusInfo(duosUser);
+    } catch (SamAzureB2CException e) {
+      throw e;
```
otchet-broad, line 145: Should we still log some user details when this happens?

otchet-broad (approved):
Looks good.  Nice work.

The only question I had was about continuing to log the specific Sam exception so that we can generate metrics on how much we're seeing this specific issue.

src/main/java/org/broadinstitute/consent/http/resources/UserResource.java
rushtong: In `createExceptionResponse`, where this is ultimately handled, we do log, but it's probably clearer to add that here since we're also logging the general case on the next line.

PR #2871, opened 2026-04-21
DT-3151 Audit DAC and DAC Chair/Member changes

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-3151

### Summary
This PR adds the following features:
* New DAC Audit table and method to return all audit records
  * Currently unused, but will be necessary to show a full DAC history
* Audit DAC inserts, updates, and deletes
* Audit DAC Chair/Member additions/removals
* Consolidation of DAC/Chair/Member deletes in `DacServiceDAO` to use more isolated, more easily testable methods in `DacDAO` and to unify auditing.
* `DacResource` is updated to use `DuosUser` which simplifies user lookups and removes the need for `UserService`
* Create and update dates are no longer passed in as arguments and are now generated in sql
* Small number of sonar warnings are addressed, mostly in tests

Due to the signature changes necessary for auditing DACs/chairs/members, there are a very large number of test updates. Most of those are pure signature changes.

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/db/mapper/DacAuditMapper.java
```diff
@@ -0,0 +1,22 @@
...
+
+public class DacAuditMapper implements RowMapper<DacAudit>, RowMapperHelper {
+  @Override
+  public DacAudit map(ResultSet rs, StatementContext ctx) throws SQLException {
+    return new DacAudit(
+        rs.getLong("id"),
+        rs.getInt("dac_id"),
+        rs.getInt("user_id"),
+        hasNonZeroColumn(rs, "affected_user_id") ? rs.getInt("affected_user_id") : null,
+        hasNonZeroColumn(rs, "role_id") ? rs.getInt("role_id") : null,
```
rushtong, line 18: Copilot complains about this pattern:
> DacAuditMapper — hasNonZeroColumn still relies on a > 0 assumption
```
hasNonZeroColumn(rs, "affected_user_id") ? rs.getInt("affected_user_id") : null,
hasNonZeroColumn(rs, "role_id") ? rs.getInt("role_id") : null,
```
> The hasNonZeroColumn helper (from RowMapperHelper) reads the column and checks whether its integer value is non-zero. This works correctly for the current data (all IDs are positive), but it is semantically wrong: a SQL NULL and an integer 0 are distinct values, and rs.getInt() returns 0 for both. If a role_id of 0 were ever introduced it would be silently dropped.

> The idiomatic JDBC pattern for nullable integer columns is:
```
rs.getObject("affected_user_id", Integer.class),
rs.getObject("role_id", Integer.class),
rs.getObject(column, Integer.class) returns a boxed Integer (or null when the SQL value is NULL) without any > 0 assumption. 
```
> This is a minor robustness issue — it does not affect current behaviour — but worth a one-line fix.

---

However, I find that this introduces a different set of errors. Using `rs.getObject` with any numeric class, (Integer, Long, etc) causes either a 
```
PSQLException: conversion to class java.lang.Integer from int8 not supported
```
or 
```
PSQLException: conversion to class java.lang.Long from int4 not supported
```
I'm inclined to keep the code as is since this is a common pattern we're using to ensure we don't try to populate ID values of `0`. If we do want to address it, we should do so holistically across all mappers and reducers.

src/test/java/org/broadinstitute/consent/http/service/DacServiceTest.java
```diff
@@ -176,8 +175,8 @@ void testUpdateDac() {
     initService();
 
     try {
-      service.updateDac("name", "description", 1);
-    } catch (Exception e) {
+      service.updateDac("name", "description", 1, 1);
+    } catch (Exception _) {
       fail("Update should not fail");
     }
```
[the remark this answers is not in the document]
rushtong: This feedback is counter to Sonarqube's warning about java:S7467 which requires using the "unnamed pattern"

src/test/java/org/broadinstitute/consent/http/service/DacServiceTest.java
```diff
@@ -290,50 +289,50 @@ void testAddDacMember() {
...
     try {
       service.removeDacMember(role, chair1, dac, 1);
-    } catch (Exception e) {
+    } catch (Exception _) {
       fail();
     }
-    verify(dacDAO, atLeastOnce()).removeDacMember(anyInt());
+    verify(dacDAO, atLeastOnce()).removeDacMember(anyInt(), anyInt());
     verify(voteService, atLeastOnce()).deleteOpenDacVotesForUser(any(), any());
     verify(ruleDAO, atLeastOnce()).auditedDeleteDACRuleSettingByUser(anyInt(), anyInt(), anyInt());
```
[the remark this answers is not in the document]
rushtong: This feedback is counter to Sonarqube's warning about java:S7467 which requires using the "unnamed pattern"

kevinmarete (approved):
👍🏾 Nice work

src/main/java/org/broadinstitute/consent/http/db/DacDAO.java
```diff
@@ -163,66 +163,125 @@ public interface DacDAO extends Transactional<DacDAO> {
...
+   * @return Integer the new dac_id
    */
-  @SqlUpdate(
-      "INSERT INTO dac (name, description, create_date) VALUES (:name, :description, :createDate)")
-  @GetGeneratedKeys
+  @SqlQuery(
+      """
+      WITH new_dac AS (
+        INSERT INTO dac (name, description, create_date) VALUES (:name, :description, NOW())
+        RETURNING dac_id
```
otchet-broad, line 177: Good use of `RETURNING`

src/main/java/org/broadinstitute/consent/http/db/DacDAO.java
```diff
@@ -254,12 +313,86 @@ void updateDac(
...
+   * @param affectedUserId The user whose role is being changed (null for DAC-level events)
+   * @param roleId The role involved (null for DAC-level events)
+   * @param action The audit action string (e.g. 'CREATE', 'UPDATE', 'DELETE')
+   */
+  @SqlUpdate(
+      """
+      INSERT INTO dac_audit (dac_id, user_id, affected_user_id, role_id, action, action_date)
+      VALUES (:dacId, :userId, :affectedUserId, :roleId, :action, NOW())
+      """)
+  void insertDacAudit(
```
otchet-broad: Is this used anywhere?

otchet-broad (commented):
It looks like there's an unused (and untested) `insertDacAudit` method in DacDAO.java.  Is that being saved for later?

Everything else looks good.  Thank you!

src/main/java/org/broadinstitute/consent/http/db/DacDAO.java
rushtong: It was at one point, but now is only used in a test. Will refactor this out to the test class.

otchet-broad (approved):
👍

PR #2873, opened 2026-04-22
[DT-3152] Remove any preexisting relation before inserting a new one.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3152](https://broadworkbench.atlassian.net/browse/DT-3152)

### Summary

Deletes any DAC-DAA relationships before adding a new one.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DaaDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DaaDAO.java
@@ -181,10 +181,23 @@ public interface DaaDAO extends Transactional<DaaDAO> {
 
   @SqlUpdate(
       """
-      WITH audit AS (INSERT INTO daa_audit (daa_id, dac_id, user_id, action, action_date) VALUES (:daaId, :dacId, :userId, 'ADD', NOW()))
+      WITH previous_dac_daa AS (
+          SELECT daa_id
+          FROM dac_daa
+          WHERE dac_id = :dacId
+      ),
+      delete_audit AS (
+          INSERT INTO daa_audit (daa_id, dac_id, user_id, action, action_date)
+          SELECT daa_id, :dacId, :userId, 'REMOVE', NOW()
+          FROM previous_dac_daa
+      ),
+      add_audit AS (
+          INSERT INTO daa_audit (daa_id, dac_id, user_id, action, action_date)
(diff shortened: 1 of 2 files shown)


kevinmarete (approved):

src/main/java/org/broadinstitute/consent/http/db/DaaDAO.java
```diff
@@ -181,10 +181,23 @@ Integer createDaa(
 
   @SqlUpdate(
       """
-      WITH audit AS (INSERT INTO daa_audit (daa_id, dac_id, user_id, action, action_date) VALUES (:daaId, :dacId, :userId, 'ADD', NOW()))
+      WITH previous_dac_daa AS (
```
rushtong, line 184: I have an open PR with a similar WITH that did not include a RETURNING clause. Copilot [flagged](https://github.com/DataBiosphere/consent/pull/2871#discussion_r3120860993) that with:

> The deleteDac SQL uses a data-modifying CTE (WITH deleted AS (DELETE ...)) that is never referenced by the main statement. In Postgres, an unreferenced data-modifying CTE will not execute, so the DAC row may not be deleted while still writing a DELETE audit row. Restructure this to DELETE ... RETURNING and INSERT ... SELECT ... FROM deleted (and ideally only insert the audit row when a row was actually deleted).

I found [this in the postgres docs](https://www.postgresql.org/docs/current/queries-with.html#QUERIES-WITH-MODIFYING) that seems to indicate otherwise IIUC:

> If a data-modifying statement in WITH lacks a RETURNING clause, then it forms no temporary table and cannot be referred to in the rest of the query. Such a statement will be executed nonetheless.

In my original unit tests, WITH lacking a RETURNING clause worked just fine, but I did end up rewriting the query regardless.

rushtong (approved):
👍🏽

src/main/java/org/broadinstitute/consent/http/db/DaaDAO.java
otchet-broad: I've got tests to confirm the "REMOVE" audit and "ADD" audit entries are being recorded to catch if the behavior changes.  If these were related actions, I might refactor but there's literally nothing to return from the audit tables of use for this case.

PR #2876, opened 2026-04-24
DT-3150: Soft delete DACs

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-3150

### Summary
This PR adds new code for soft deleting DACs.
* DACs are no longer deleted, just marked as deleted similarly to how we treat FSOs
* New fields added to DAC and mappers/reducers
* DAC queries are updated to exclude deleted DACs

See related UI PR that fixes a minor bug in the same feature space: https://github.com/DataBiosphere/duos-ui/pull/3452

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DaaDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DaaDAO.java
@@ -60,7 +60,7 @@ public interface DaaDAO extends Transactional<DaaDAO> {
           FROM data_access_agreement daa
           LEFT JOIN file_storage_object fso ON daa.daa_id::text = fso.entity_id
           LEFT JOIN dac_daa dd ON daa.daa_id = dd.daa_id
-          LEFT JOIN dac ON dd.dac_id = dac.dac_id
+          LEFT JOIN dac ON dd.dac_id = dac.dac_id AND dac.deleted IS NOT TRUE
       """)
   List<DataAccessAgreement> findAll();
 
@@ -101,7 +101,7 @@ public interface DaaDAO extends Transactional<DaaDAO> {
           FROM data_access_agreement daa
           LEFT JOIN file_storage_object fso ON daa.daa_id::text = fso.entity_id
           LEFT JOIN dac_daa dd ON daa.daa_id = dd.daa_id
-          LEFT JOIN dac ON dd.dac_id = dac.dac_id
+          LEFT JOIN dac ON dd.dac_id = dac.dac_id AND dac.deleted IS NOT TRUE
           WHERE daa.daa_id = :daaId
       """)
   DataAccessAgreement findById(@Bind("daaId") Integer daaId);
@@ -265,7 +265,7 @@ public interface DaaDAO extends Transactional<DaaDAO> {
           FROM data_access_agreement daa
           LEFT JOIN file_storage_object fso ON daa.daa_id::text = fso.entity_id
           INNER JOIN dac_daa ON daa.daa_id = dac_daa.daa_id
-          INNER JOIN dac ON dac.dac_id = dac_daa.dac_id
+          INNER JOIN dac ON dac.dac_id = dac_daa.dac_id AND dac.deleted IS NOT TRUE
           INNER JOIN dataset ON dataset.dac_id = dac.dac_id
           INNER JOIN dar_dataset dd on dd.dataset_id = dataset.dataset_id
           WHERE dd.reference_id = :referenceId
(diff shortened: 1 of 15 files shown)


src/main/resources/changesets/changelog-consent-2026-04-24-dac-soft-delete.xml
```diff
@@ -0,0 +1,58 @@
...
+
+    <!-- 1c. Re-add fk_daa_initial_dac_id (dropped in DT-3059 changelog-consent-2026-03-31-daa-audit.xml).
+         Now safe to restore because DAC rows are soft-deleted, not physically removed. -->
+    <addForeignKeyConstraint
+      baseTableName="data_access_agreement"
+      baseColumnNames="initial_dac_id"
+      constraintName="fk_daa_initial_dac_id"
+      onDelete="NO ACTION" onUpdate="NO ACTION"
+      referencedTableName="dac"
+      referencedColumnNames="dac_id"/>
```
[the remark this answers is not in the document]
rushtong: I will move this feature to a separate ticket that can be isolated from the rest of this code.

src/main/java/org/broadinstitute/consent/http/db/DaaDAO.java
```diff
@@ -265,7 +265,7 @@ void deleteDacDaaRelation(
           FROM data_access_agreement daa
           LEFT JOIN file_storage_object fso ON daa.daa_id::text = fso.entity_id
           INNER JOIN dac_daa ON daa.daa_id = dac_daa.daa_id
-          INNER JOIN dac ON dac.dac_id = dac_daa.dac_id
+          INNER JOIN dac ON dac.dac_id = dac_daa.dac_id AND dac.deleted IS NOT TRUE
```
otchet-broad: Do we care that a delete is happening on a DacDAARelation when a dac is already deleted?  If so, we might need to modify other components.

I'm specifically thinking about cases where the user "deletes" the dac but we still have these documents in our system.  I'm not sure this one and the one below should care if the dac was deleted or not.

src/main/java/org/broadinstitute/consent/http/db/DaaDAO.java
```diff
@@ -304,7 +304,7 @@ void deleteDacDaaRelation(
           SELECT distinct daa.daa_id
           FROM data_access_agreement daa
           INNER JOIN dac_daa ON daa.daa_id = dac_daa.daa_id
-          INNER JOIN dac ON dac.dac_id = dac_daa.dac_id
+          INNER JOIN dac ON dac.dac_id = dac_daa.dac_id AND dac.deleted IS NOT TRUE
```
otchet-broad: Same thought here.

fboulnois (approved):
👍

src/main/java/org/broadinstitute/consent/http/db/DaaDAO.java
rushtong: > Do we care that a delete is happening on a DacDAARelation when a dac is already deleted?

Specifically for this question, I don't think we care if a DacDAARelation is deleted for a DAC that is also deleted.

I think maybe one concern is that there might be DAA documents in the system that are no longer associated to a visible DAC. I'm not sure that's a real problem, but might be persuaded.

I think a second concern might be that a DAR that had accepted a DAA for a DAC that is deleted might not be able to see the DAA in the above query condition. That edge case seems like it would need a lot more special handling than is possible here.

src/main/java/org/broadinstitute/consent/http/db/DacDAO.java
```diff
@@ -86,6 +86,7 @@ public interface DacDAO extends Transactional<DacDAO> {
       LEFT JOIN dac_daa dd ON dac.dac_id = dd.dac_id
       LEFT JOIN data_access_agreement daa ON dd.daa_id = daa.daa_id
       LEFT JOIN file_storage_object fso ON daa.daa_id::text = fso.entity_id
+      WHERE dac.deleted IS NOT TRUE
```
otchet-broad: Same here.

src/main/java/org/broadinstitute/consent/http/db/DaaDAO.java
rushtong: As discussed in standup, we'll remove these query modifiers and follow up on what to do about the following questions:
* What needs to happen with/to DARs against datasets with deleted DACs
* Do we need to do anything with orphaned DAA documents?

kevinmarete (approved):

PR #2887, opened 2026-04-29
DT-245: Standardize sql in MailMessageDAO

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-245

## Summary
This started out as a formatting PR, but there were several other errors that needed addressing so I ended up doing some significant refactoring:

* The `MailMessage` object did not have a field for `entity_reference_id` so that was added
* The create and update dates were being truncated so those are now populated from Timestamps in the mapper
* The `email_type` was being stored as an integer but coerced to a string in the model. It is now an integer
* `MailMessage` is now a record class
* Sonarqube warned about `MailMessageDAO.insert` argument count being too large so that is now a `@BindMethods` argument that takes a single populated object instead.
* While I was there, I modified the insert to return the populated object instead of just the ID.
* Many, many test updates to cover the changes.

### Testing Notes
This is fairly easily testable using Swagger. The Notifier section has a reminder API that can be used. Look for a recent vote id in the votes table. After sending a reminder, that email should show up in the email_entity table.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/MailMessageDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/MailMessageDAO.java
@@ -17,9 +17,12 @@ import org.jdbi.v3.sqlobject.transaction.Transactional;
 public interface MailMessageDAO extends Transactional<MailMessageDAO> {
 
   @SqlUpdate(
-      "INSERT INTO email_entity "
-          + "(entity_reference_id, vote_id, user_id, email_type, date_sent, email_text, sendgrid_response, sendgrid_status, create_date) VALUES "
-          + "(:entityReferenceId, :voteId, :userId, :emailType, :dateSent, :emailText, :sendGridResponse, :sendGridStatus, :createDate)")
+      """
+      INSERT INTO email_entity
+          (entity_reference_id, vote_id, user_id, email_type, date_sent, email_text, sendgrid_response, sendgrid_status, create_date)
+      VALUES
+          (:entityReferenceId, :voteId, :userId, :emailType, :dateSent, :emailText, :sendGridResponse, :sendGridStatus, :createDate)
+      """)
   @GetGeneratedKeys
   Integer insert(
       @Nullable @Bind("entityReferenceId") String entityReferenceId,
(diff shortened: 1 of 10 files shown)


src/test/java/org/broadinstitute/consent/http/db/MailMessageDAOTest.java
```diff
@@ -404,6 +404,15 @@ void testFetchByCreateDate_with_limit_and_offset() {
...
   }
 
+  @Test
+  void testFetchMessageById() {
+    MailMessage message = generateMessage(Instant.now());
+    MailMessage fetchedMessage = mailMessageDAO.fetchMessageById(message.getEmailId());
+    assertEquals(message.getEmailId(), fetchedMessage.getEmailId());
+    assertEquals(message.getUserId(), fetchedMessage.getUserId());
+    assertEquals(message.getEmailType(), fetchedMessage.getEmailType());
+  }
```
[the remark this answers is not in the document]
rushtong: Will tackle this problem.

fboulnois (approved):
👍

src/main/java/org/broadinstitute/consent/http/db/MailMessageDAO.java
```diff
@@ -1,40 +1,34 @@
...
-      @Nullable @Bind("dateSent") Instant dateSent,
-      @Bind("emailText") String emailText,
-      @Nullable @Bind("sendGridResponse") String sendGridResponse,
-      @Nullable @Bind("sendGridStatus") Integer sendGridStatus,
-      @Bind("createDate") Instant createDate);
+  @SqlQuery(
+      """
+      WITH insterted_row AS (
+        INSERT INTO email_entity
+          (entity_reference_id, vote_id, user_id, email_type, date_sent, email_text, sendgrid_response, sendgrid_status, create_date)
```
otchet-broad, line 20: What's the difference between date_sent and create_date for these intending to represent?  Why not use NOW() instead of binding to a variable?

rushtong: Date sent captures the date that the mail was sent via Sendgrid. Create date captures the record creation. `date_sent` does need to be passed in as business logic sets it. We **could** update the `create_date` to be `NOW()` but that has an impact on test logic (which mutates create date for date range searching) and is covered in a different ticket: https://broadworkbench.atlassian.net/browse/DT-2932. I chose to keep the current implementation for those reasons.

otchet-broad (approved):
👍

src/main/java/org/broadinstitute/consent/http/models/mail/MailMessage.java
fboulnois, line 1: Nice cleanup 👍

fboulnois (commented):
👍

kevinmarete (approved):

PR #2893, opened 2026-05-02
DT-1470: Admin Redact User

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-1470

## Summary
Add admin redact-user endpoint (`POST /api/user/redact?email=***`)

### Motivation
DUOS stores user PII (email address and display name) in the users table. When a user requests deletion or account redaction, there was no mechanism to anonymise that data while preserving a traceable audit trail. This PR introduces a dedicated admin-only redaction endpoint to satisfy that need.
 
### Database
- New Liquibase changeset creates the user_redaction_audit table:
  - user_id — account that was redacted (no FK; audit survives independently of the row it describes)
  - admin_user_id (FK → users) — administrator who triggered the redaction
  - original_email / original_display_name / original_institution_id — PII captured before the overwrite
  - action_date — defaulted to NOW()
  - Index on user_id; full rollback block included

### DAO
- New UserRedactionAuditDAO.redactUser(userId, adminUserId, originalEmail, originalDisplayName) executes a single atomic writeable CTE that:
- UPDATE users SET email = 'redacted_' || encode(:userId::text::bytea, 'base64'), display_name = 'redacted' — the same redaction pattern used in the 2026-03-05 FK migration
- INSERT INTO user_redaction_audit … SELECT … FROM update_user — the audit row is only written if the UPDATE matched a row, which prevents orphaned audit records

### Service
- UserService.redactUser(User adminUser, String email) — looks up the target user by email (surfacing NotFoundException if absent), then delegates to userRedactionAuditDAO.redactUser(…)
- UserRedactionAuditDAO injected directly into UserService; no Java-level transaction wrapper is needed because the CTE provides atomicity at the SQL level

### Resource
- POST /api/user/redact?email=<address> — @RolesAllowed({ADMIN}), returns 400 for blank/missing email, 404 if user not found, 200 on success

### OpenAPI
New paths/userRedact.yaml and corresponding entry in api-docs.yaml

### Tests
- UserRedactionAuditDAOTest — 6 integration tests against a real Postgres container covering: email/display-name update, audit row content, null display name, idempotent replay, and no-op when user not found
- UserServiceTest — 2 unit tests: success (verifies DAO called with correct args) and not-found propagation
- UserResourceTest — 4 unit tests: success, missing email, blank email, not-found — all using try-with-resources
- DAOTestHelper updated with userRedactionAuditDAO field

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/service/UserService.java
```diff
@@ -413,6 +417,25 @@ public User enforceInstitutionAndLibraryCardRules(String email) {
...
+   * anonymised placeholders, and institution_id is set to null. All original values are persisted
+   * atomically in the same SQL statement via a writeable CTE.
+   *
+   * @param adminUser the administrator performing the redaction
+   * @param email the email address of the user to redact
+   * @throws NotFoundException if no user exists with the given email
+   */
+  public void redactUser(User adminUser, String email) {
+    User target = findUserByEmail(email);
+    userRedactionAuditDAO.redactUser(
```
otchet-broad: Should we also turn off the email pref so we don't keep generating emails for users who are redacted?

src/main/java/org/broadinstitute/consent/http/db/UserRedactionAuditDAO.java
```diff
@@ -0,0 +1,45 @@
...
+        UPDATE users
+        SET
+          email          = 'redacted_' || encode(:userId::text::bytea, 'base64'),
+          display_name   = 'redacted',
+          institution_id = NULL
+        WHERE user_id = :userId
+        RETURNING user_id
+      )
+      INSERT INTO user_redaction_audit
+        (user_id, admin_user_id, original_email, original_display_name, original_institution_id, action_date)
```
otchet-broad, line 46: Glad we're preserving this information internally.

otchet-broad (approved):
Looks like a solid addition.  Thanks for adding this.

I've asked one question but defer to your discretion on how to handle.

src/main/java/org/broadinstitute/consent/http/service/UserService.java
rushtong: That's a good point, will add that. I believe mail sending is smart enough to ignore bad emails, but better to have 100% clear intentions in this code.

kevinmarete (approved):
👍🏾

PR #2909, opened 2026-05-14
DT-1591: Refactor domain specific message handling to domain services.

rushtong:
### Addresses
Addresses bullet point 3 in https://broadworkbench.atlassian.net/browse/DT-1591

> It might be better to make sendMessage() public and move all the custom email sending code into other service objects, so this class doesn’t need to inject DAO classes.

### Summary
This PR moves most domain-specific message sending methods to the primary service classes they are called from. This reduces the scope of EmailService and moves business logic closer to the relevant service classes. It also removes an unused code path for `NewResearcherLibraryRequestMessage`

Currently, this covers method migration to the following classes:
* `AcknowledgementService`
* `DaaService`
* `DarCollectionService`
* `DataAccessRequestService`
* `DatasetRegistrationService`
* `DatasetService`
* `LibraryCardService`
* `VoteService`

This covers all service class level calls to `EmailService`. There are also calls from `EmailNotifierResource` which does not have a similar service layer so I'm leaving calls to `sendVoteDigestMessages` and `sendNewDatasetInDUOSNotifications` where they are for now.

Additionally, there are some new tests to increase coverage.

----

- Verify all tests go green


otchet-broad (approved):
Nice refactor and cleanup.  Thank you!

kevinmarete (approved):

PR #2911, opened 2026-05-15
DT-1591: Consolidate send mail methods

rushtong:
### Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DT-1591
Specifically, bullet point 2:
> The EmailService methods saveEmailAndResponse and sendMessage could be further refactored now that saveEmailAndResponse is only used in one place.

### Summary
This PR simplifies `EmailService` by combining methods. We now delegate the creation date to the database instead of passing it in via application code. Additionally, tests are updated for coverage and comprehensiveness.

----

- Verify all tests go green

### DIFF
--- a/src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
@@ -2,6 +2,7 @@ package org.broadinstitute.consent.http.service;
 
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyInt;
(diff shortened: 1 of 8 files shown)


src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
```diff
@@ -171,16 +153,119 @@ public Integer getVoteId() {
...
+                  }
+                  return true;
+                }));
+
+    service.sendMessage(message, userId);
+
+    var captor = ArgumentCaptor.forClass(Mail.class);
+    verify(sendGridAPI).sendMessage(captor.capture(), eq(user.getEmail()));
+    var mail = captor.getValue();
+    assertTrue(mail.getCategories() == null || mail.getCategories().isEmpty());
```
otchet-broad: Could this be cleaned up to assert only one of the cases?  I'm not sure I understand why the response to a single call might be null OR empty.  Couldn't we predict the exact response here?

otchet-broad (approved):
Looks good overall.

Left one note about a test case.  Feel free to address as you see fit.

src/test/java/org/broadinstitute/consent/http/service/EmailServiceTest.java
rushtong: Yes, we know the categories are null in this case.

kevinmarete (approved):
👍🏾

PR #2891, opened 2026-04-30
DT-3208: Run Synthetic Integration Tests as Unit Tests

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-3208

## Summary
Previously, integration/smoke tests were:

- Gated behind a separate Maven profile (-P integration-tests)
- Run in a dedicated GitHub Actions workflow (smoke-tests.yaml) that spun up a full Broad BEE environment
- Excluded from the normal test run via a Surefire <excludes> rule
- Dependent on external infrastructure (Sam, ECM, etc.) components deployed alongside the app

This approach was slow, expensive, and fragile — a BEE outage meant no smoke coverage. A quick scan of [recent coverage workflow runs](https://github.com/DataBiosphere/consent/actions/workflows/coverage.yaml?query=branch%3Agr-integration+branch%3Adevelop) shows no significant difference in run times between this branch and develop which means we're saving 100% of the time spent in the [smoke test action](https://github.com/DataBiosphere/consent/actions/workflows/smoke-tests.yaml) and gaining slightly more coverage in this PR.

### What we lose
It is important to call out the one feature smoke tests did previously which was to upload test results to a big query table. This was originally built to support pact testing but never gained traction. It was cumbersome to configure correctly due to xml formatting requirements and we haven't looked at the data in years. I see no value in maintaining it.

### Test infrastructure (ContainerTests)

- New abstract base class that boots the full Dropwizard application in-process using DropwizardAppExtension against consent-ci.yaml
- All external services (Sam, ECM, GCS, OIDC) are stubbed by a WireMock server on port 9999 — no live infrastructure required
- Database is seeded once before any tests via typed DAO calls (UserDAO, InstitutionDAO, UserRoleDAO, DacDAO) with 7 synthetic CI users, an institution, role assignments, and a DAC
- DAOTestHelper (existing TestExecutionListener) automatically provides a Testcontainers Postgres instance in all environments

### Test coverage

- StatusTests — unauthenticated liveness/readiness/version smoke tests that **replace** the original smoke tests
- UserTests — **new** authenticated GET /api/user/me exercised for every seeded CI user; validates the full auth stack (OAUTH2 claim headers → ClaimsCache → DuosUserAuthenticator → Sam/ECM stubs)

### Removed
- smoke-tests.yaml GitHub Actions workflow (BEE-based)
- integration-tests and all-tests Maven profiles + the corresponding Surefire <excludes> rule
- IntegrationTestHelper (superseded by ContainerTests)

[...]

PR #2914, opened 2026-05-20
DT-3295: Re-add surefire plugin

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-3295

### Summary
In https://github.com/DataBiosphere/consent/pull/2891, we removed two conditional surefire blocks. We should have kept the first one in place and only removed the integration test profile block. Tests still run regardless, but we get a warning:
```
Mockito is currently self-attaching to enable the inline-mock-maker. This will no longer work in future releases of the JDK. Please add Mockito as an agent to your build as described in Mockito's documentation: https://javadoc.io/doc/org.mockito/mockito-core/latest/org.mockito/org/mockito/Mockito.html#0.3
```
so using the plugin explicitly gives us better control over the fine-grained jvm settings.

----

- Verify all tests go green

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,6 +55,20 @@
     <plugins>
 
       <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>${surefire.version}</version>
+        <configuration>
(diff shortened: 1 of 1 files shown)


kevinmarete (approved):

fboulnois (approved):
good catch 👍

PR #2930, opened 2026-06-11
DT-3076: Replace per-class thread pools with a single shared executor

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-3076

## Summary
`ThreadUtils.getExecutorService()` created a new fixed thread pool of availableProcessors - 1 thread — plus a JVM shutdown hook — on every call. Five classes called it from instance initializers (`SamDAO`, `OntologyService`, `DatasetRegistrationService`, `InstitutionAndLibraryCardEnforcement`, `EmailNotifierResource`), and because `ConsentModule`'s providers are unscoped and call each other directly, several of those classes were instantiated multiple times (`OntologyService` ×5, `InstitutionAndLibraryCardEnforcement` ×3, `SamDAO` ×2). The result was ~14 thread pools — roughly 100 idle platform threads on an 8-core host — and ~14 ad-hoc shutdown hooks.

### Changes
- One shared executor. `ConsentModule` now creates a single `Executors.newVirtualThreadPerTaskExecutor()`, exposed via a `providesExecutorService()` provider. All work submitted to these executors is blocking I/O (Sam HTTP calls, Elasticsearch indexing, batch DB writes, GCS cleanup), which is the ideal workload for virtual threads: no pool sizing, no thread starvation if a downstream call hangs, near-zero idle cost.
- Lifecycle-managed shutdown. The executor is registered with Dropwizard's lifecycle (`Managed.stop()`), replacing the per-pool JVM shutdown hooks.
- Constructor injection. The five consuming classes now take `ExecutorService` as a constructor parameter instead of building their own pool. `ThreadUtils` is deleted.
- Singleton scoping for the consuming services. `providesOntologyService`, `providesDatasetRegistrationService`, `providesInstitutionAndLibraryCardEnforcement`, and `providesSamDAO` are now `@Singleton` and memoized in module fields. The field memoization matters because providers in this module call each other directly, which bypasses Guice scoping; the fields guarantee one instance on both the Guice and direct-call paths.
- Tests construct a fresh `newVirtualThreadPerTaskExecutor()` per test, preserving the existing shutdown/await patterns.

### Out of scope (follow-up work)
- Module-wide `@Singleton` scoping: providers still rebuild transitive dependency graphs (DAOs, services) on each direct call; only the four executor-consuming services are deduplicated here. (See [DT-3601](https://broadworkbench.atlassian.net/browse/DT-3601))
- `EmailNotifierResourceTest` could use a same-thread executor (`MoreExecutors.newDirectExecutorService()`) to drop its `shutdown()`/`awaitTermination()` choreography.

----

- Verify all tests go green

[DT-3601]: https://broadworkbench.atlassian.net/browse/DT-3601?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kevinmarete (approved):
Excellent solution 👍🏾

src/main/java/org/broadinstitute/consent/http/ConsentModule.java
```diff
@@ -167,6 +181,28 @@ public class ConsentModule extends AbstractModule {
...
+    this.executorService = Executors.newVirtualThreadPerTaskExecutor();
+    environment
+        .lifecycle()
+        .manage(
+            new Managed() {
+              @Override
+              public void stop() {
+                executorService.shutdown();
+                try {
+                  if (!executorService.awaitTermination(30, TimeUnit.SECONDS)) {
```
otchet-broad, line 199: According to the [docs](https://javadoc.io/doc/io.dropwizard/dropwizard-project/latest/io/dropwizard/lifecycle/Managed.html), incoming requests stop being processed by Dropwizard before this call will be invoked.

The idea is to drain the executor after the application is no longer accepting requests.

Do you want to log information about the things that were interrupted?

src/main/java/org/broadinstitute/consent/http/resources/EmailNotifierResource.java
```diff
@@ -24,10 +23,12 @@ public class EmailNotifierResource extends Resource {
...
   @Inject
   public EmailNotifierResource(
-      DataAccessRequestService dataAccessRequestService, EmailService emailService) {
+      DataAccessRequestService dataAccessRequestService,
+      EmailService emailService,
+      ExecutorService executor) {
     this.dataAccessRequestService = dataAccessRequestService;
     this.emailService = emailService;
-    this.executor = new ThreadUtils().getExecutorService(EmailNotifierResource.class);
+    this.executor = executor;
```
otchet-broad, line 31: This has me concerned about the node running the "new data in DUOS" messages because it will put all X,000 messages into the same thread pool as everything else.  I'm wondering if we should keep emails in their own pool.  If I remember correctly, that queue also paused so that we didn't exceed rate limits.

otchet-broad: After reviewing the code, it's the thread that's running the loop over all users that will pause, not the entire thread pool.  The consequences of this are that there's periodically a very long running thread occupied in the common pool on a somewhat infrequent basis.  That long running thread will very likely sleep periodically to mitigate throttling limits.

otchet-broad: The virtual thread infrastructure should handle this nicely as it will no longer lock a thread in the operating system during the thread sleep.

src/main/java/org/broadinstitute/consent/http/ConsentModule.java
rushtong: Great point, let me add that.

otchet-broad (approved):
👍

PR #2931, opened 2026-06-11
[DT-3132] Fix NPE in DatasetPatch when properties is null 

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3132

### Summary

 `validateProperties()` threw a `NullPointerException` when a PATCH request was sent without a `properties` field, because Gson deserializes the missing field as `null` and the method called `properties.stream()`         
  directly.                                                                                                                                                                                                         
                                                                                                                                                                                                                    
  Changes:                                                                                                                                                                                                          
  - `validateProperties()` — returns `false` for null properties, rejecting the request with a `400` before reaching the DAO layer                                                                                        
  - `isPatchable()` — guards against null properties to avoid a second latent NPE                                                                                                                                     
  - `DatasetPatchTest` — adds tests for null and empty properties on both methods                                                                                                                                     
  - `DatasetResourceTest` — adds an end-to-end test confirming a null-properties PATCH returns `400`    

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/DatasetPatch.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/DatasetPatch.java
@@ -25,6 +25,9 @@ public record DatasetPatch(String name, List<DatasetProperty> properties) {
     if (name() != null && !name().equals(dataset.getName()) && !name().isBlank()) {
       return true;
     }
+    if (properties() == null) {
+      return false;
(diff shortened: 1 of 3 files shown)


rushtong (approved):
👍🏽

fboulnois (approved):
👍

PR #2937, opened 2026-06-16
[DT-3614] Fix DAR chair summary status and Open actions

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3614

### Summary

Fixes chair DAR collection summaries so Open / Re-Open is not shown while active elections exist, and terminal election states are handled consistently.
Changes include:

- Only show `Open` when there are no open elections.
- Treat closed/canceled elections as terminal for summary status.
- Calculate `Complete` by dataset election coverage instead of raw election count.
- Include canceled latest `DataAccess` elections in chair summary queries.
- Ignore older elections by filtering to the latest election per dataset.
- Preserve archived DAR filtering from summary results.

Adds regression coverage for closed `DataAccess` / open `RP` and all-terminal election cases.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
@@ -285,10 +285,6 @@ public class DarCollectionService implements ConsentLogger {
         s -> {
           Map<String, Integer> statusCount = new HashMap<>();
           Map<Integer, Election> elections = s.getElections();
-          if ((!s.requiresSOApproval() || s.getSOApprover() != null)
-              && elections.size() < s.getDatasetCount()) {
-            s.addAction(DarCollectionActions.OPEN);
-          }
           elections
               .values()
               .forEach(election -> updateStatusCount(statusCount, election.getStatus()));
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -110,23 +110,30 @@ private void updateStatusCount(Map<String, Integer> statusCount, String status)
...
-      Map<String, Integer> statusCount,
-      Integer datasetCount,
-      Integer electionCount) {
-    // If there are no elections, status is unreviewed
-    // if there are some elections open, status is in process
-    // if all elections are closed or canceled and electionCount == datasetCount, status is complete
-    if (electionCount.equals(0)) {
+      DarCollectionSummary summary, Map<String, Integer> statusCount) {
+    // If there are no elections, status is unreviewed.
+    // If there are any open elections, status is in process.
```
rushtong, line 129: I think this is logic is not correct any longer. Since we stopped populating votes for RP elections, we can have a closed/completed Data Access election and also have an open RP election. In that case, the status should NOT be in process from a user's point of view.

rushtong: Responding to my own question. Reviewing the query, we're only populating `dataaccess` elections, so my concern above is moot.

rushtong (approved):
👍🏽

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
otchet-broad: This is another reason in the list to remove RP elections.  :-)

otchet-broad (approved):
Nice work!

PR #2940, opened 2026-06-18
[DT-3075] Standardize on jdbi.onDemand across all services.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3075](https://broadworkbench.atlassian.net/browse/DT-3075)

### Summary

With Claude.ai

Summary
Standardizes how Jdbi SQL object DAOs are instantiated across every service in the codebase, matching the pattern already established in DarCollectionService. Removes the intermediate DAOContainer bundling class entirely.

Major changes
DAO instantiation pattern — Every service now calls jdbi.onDemand(XxxDAO.class) inside its own constructor rather than receiving pre-built DAO instances from the outside. Jdbi is injected as a constructor parameter in place of individual DAOs. This is the standard Jdbi idiom for short-lived, connection-pooled proxies.

DAOContainer deleted — The DAOContainer class (a plain holder that aggregated raw DAO instances) is removed. The one remaining call site (DataAccessRequestService) was migrated to the jdbi.onDemand() pattern, after which the class had no purpose.

@Inject and private final hygiene — @Inject was added to all service constructors that were missing it, and all injected fields were tightened to private final.

Constructor parameter ordering — Established and applied a consistent convention across all service constructors:

Jdbi jdbi — always first
XxxServiceDAO composite — immediately after Jdbi
Other injected services — in any order
Configuration objects — last
ConsentModule singleton correctness — @Provides methods call each other directly, which bypasses Guice's @Singleton scope. Every provider that constructs a new object now uses a @Singleton synchronized null-guard field to guarantee a single instance on both the Guice resolution path and the direct inter-provider call path. Five previously unguarded providers (OntologyIndexService, SamService, OidcAuthorityDAO, OidcService, SupportRequestService) were also brought into line. Spurious @Inject annotations on private final module fields were removed.

Test improvements — Tests across 25 test classes were updated to match the new constructors. The initService() anti-pattern (a setup method that tests had to call manually) was converted to @BeforeEach setUp() in four test classes to eliminate the risk of silent NPEs. DaaServiceTest retains per-test initialization with an explanatory comment because its tests configure stubs before constructing the service.

Documentation — CLAUDE.md records the constructor ordering convention, the jdbi.onDemand() rule, and the ConsentModule singleton pattern with examples. service/dao/package-info.java was updated to note that the orchestration classes in that package are plain Java objects and must not be instantiated via jdbi.onDemand().

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/service/dao/DarCollectionServiceDAO.java
```diff
@@ -29,10 +29,10 @@ public class DarCollectionServiceDAO {
 
   @Inject
   public DarCollectionServiceDAO(
-      DatasetDAO datasetDAO, ElectionDAO electionDAO, Jdbi jdbi, UserDAO userDAO) {
+      Jdbi jdbi, DatasetDAO datasetDAO, ElectionDAO electionDAO, UserDAO userDAO) {
+    this.jdbi = jdbi;
     this.datasetDAO = datasetDAO;
     this.electionDAO = electionDAO;
```
rushtong: I think this is a good first step for service DAOs given that the test infrastructure might not support direct instantiation. I do think we should try to tackle them, maybe as a follow-on task.

src/main/java/org/broadinstitute/consent/http/service/dao/DarCollectionServiceDAO.java
rushtong, line 1: I think other service DAOs fall into the same pattern as this class does:
* DaaServiceDAO
* DatasetServiceDAO
* DraftFileStorageServiceDAO
* DraftServiceDAO
* VoteServiceDAO

docs/ai/CLAUDE.md
rushtong, line 1: Should this live in the docs/ai directory?

src/main/java/org/broadinstitute/consent/http/ConsentModule.java
rushtong, line 1: The `@Singleton` changes here address https://broadworkbench.atlassian.net/browse/DT-3601

src/main/java/org/broadinstitute/consent/http/ConsentModule.java
```diff
@@ -322,112 +307,159 @@ DuosUserAuthenticator providesDuosUserOAuthAuthenticator() {
...
   }
 
   @Provides
-  SendGridAPI providesSendGridAPI() {
-    return new SendGridAPI(config.getMailConfiguration(), providesUserDAO());
+  @Singleton
+  synchronized SendGridAPI providesSendGridAPI() {
+    if (sendGridAPI == null) {
+      // SendGridAPI predates the Jdbi migration and owns only a single read; migrating it
+      // requires its own constructor change and is deferred to a separate ticket.
```
rushtong: I'm good with deferring this. `SendGridAPI` used to be `MailService` which didn't have any database dependency, so didn't need `Jdbi`, but `Jdbi` predates both `SendGridAPI` and `MailService`.

rushtong (commented):
This is a great step forward, love the consistency. See comments inline.

fboulnois (approved):
I agree with @rushtong 's comments, otherwise looks good.

docs/ai/CLAUDE.md
otchet-broad: Happy to move it.  I was working from the recommendations here:

[https://code.claude.com/docs/en/claude-directory](https://code.claude.com/docs/en/claude-directory)

rushtong: Maybe we need to coordinate on the right home for these docs

src/main/java/org/broadinstitute/consent/http/ConsentModule.java
otchet-broad: I'm not sure that this addresses all of the things we want to cover in DT-3601 so I'll leave that open for now.

src/main/java/org/broadinstitute/consent/http/service/dao/DarCollectionServiceDAO.java
otchet-broad: I'm looking at what changing the service DAOs does to the scope of the PR.  If it's similar in kind without mushrooming the number of files, I'll add it to this one.  If not, I'll make a ticket.

kevinmarete (approved):
Looks good, also agree with Greg's observations

src/main/java/org/broadinstitute/consent/http/service/dao/DarCollectionServiceDAO.java
otchet-broad: These are all now converted as well.

src/main/java/org/broadinstitute/consent/http/service/dao/DarCollectionServiceDAO.java
otchet-broad: Got it in.

src/main/java/org/broadinstitute/consent/http/ConsentModule.java
otchet-broad: Refactored for pattern, pr updated.

rushtong (approved):
Thank you for the updates! 👍🏽

PR #2941, opened 2026-06-18
[DT-2869] Add `externalIdentifier` and `externalIdentifierType` to study   

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2869

### Summary
                                                                                                                                                                                                                    
  - Adds `externalIdentifier` and `externalIdentifierType` as optional study fields to support linking DUOS studies to pages in external systems (e.g. `SCP1671` → `Single Cell Portal`)                            
  - Both fields are stored as `study_property` entries, consistent with the existing pattern for optional study metadata like `studyType`, `dbGaPPhsID`, etc.                                                       
  - No database migration required                                                                                                                                                                                  
                                                                                                                                                                                                                    
  ## Changes                                                                                                                                                                                                        
                                                                                                                                                                                                                    
  - **`StudyPatch`** — adds the two fields so they can be set via `PATCH /api/dataset/study/{studyId}`                                                                                                              
  - **`DatasetRegistrationSchemaV1` / `DatasetRegistrationSchemaV1Builder`** — adds fields to the registration schema so they flow through `POST /api/dataset/v3` and `PUT /api/dataset/study/{studyId}`            
  - **`DatasetRegistrationService`** — new `StudyPropertyExtractor` entries persist both fields as study properties on create/update                                                                                
  - **`SchemaFromStudy`** — reads the properties back when building a registration schema from a stored study (`GET /api/dataset/study/registration/{studyId}`)                                                     
  - **`DatasetServiceDAO`** — `convertToStudyUpdate` maps patch fields to `StudyProperty` objects                                                                                                                   
  - **`StudyTerm` / `ElasticSearchService`** — indexes both fields as named fields on the ES study document                                                                                                         
  - **`StudyPatch.yaml` / `DatasetRegistrationSchemaV1.yaml`** — OpenAPI schema updates for all affected endpoints    

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
@@ -3,6 +3,8 @@ package org.broadinstitute.consent.http.service.dao;
 import static org.broadinstitute.consent.http.models.StudyPatch.ALTERNATIVE_DATA_SHARING_PLAN_TARGET_DELIVERY_DATE;
 import static org.broadinstitute.consent.http.models.StudyPatch.ALTERNATIVE_DATA_SHARING_PLAN_TARGET_PUBLIC_RELEASE_DATE;
 import static org.broadinstitute.consent.http.models.StudyPatch.DATA_CUSTODIAN_EMAIL;
+import static org.broadinstitute.consent.http.models.StudyPatch.EXTERNAL_IDENTIFIER;
+import static org.broadinstitute.consent.http.models.StudyPatch.EXTERNAL_IDENTIFIER_TYPE;
 import static org.broadinstitute.consent.http.models.StudyPatch.PHENOTYPE_INDICATION;
 import static org.broadinstitute.consent.http.models.StudyPatch.SPECIES_KEY;
 import static org.broadinstitute.consent.http.models.StudyPatch.STUDY_TYPE;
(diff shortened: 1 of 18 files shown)


rushtong (approved):
The copilot feedback could be more extensive if we applied it to the rest of the statics in this class 👍🏽

src/main/java/org/broadinstitute/consent/http/service/dao/DatasetServiceDAO.java
```diff
@@ -500,6 +502,15 @@ private StudyUpdate convertToStudyUpdate(Study study, User user, StudyPatch patc
               patch.alternativeDataSharingPlanTargetPublicReleaseDate(),
               PropertyType.String));
     }
+    if (patch.externalIdentifier() != null) {
```
otchet-broad: Do we need to check for empty strings and only whitespace as well?
How do we support removing these attributes?

kevinmarete: Good catch — updated to handle both. Blank strings are now rejected (not stored), and sending `""` removes the property. Extended the fix to all string-typed study properties since the same gap existed there.  
                                                                                                                                                                                                                    
  Semantics: `null` = no-op, `""` / whitespace = delete, non-blank = upsert.

otchet-broad (approved):
Nice work!
👍

PR #2943, opened 2026-06-18
[DT-3601] Migrate to @Singleton for provider methods.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3601](https://broadworkbench.atlassian.net/browse/DT-3601)

### Summary

`ConsentModule` contained a mix of scoped and unscoped providers. Several providers were annotated `@Singleton` but bypassed Guice's scoping by calling sibling provider methods directly rather than through the injector — meaning a second call to a provider method would construct a second object even though the method carried `@Singleton`. Other providers constructed managed objects (`ClaimsCache`, `HttpClientUtil`, `DataUseMatcherV4`, etc.) with `new` inside downstream service constructors, producing unmanaged instances outside Guice's lifecycle entirely.

---

#### `ConsentModule` — provider consistency

All providers that were missing `@Singleton` now have it. The four config-getter providers (`providesElasticSearchConfiguration`, `providesMailConfiguration`, `providesServicesConfiguration`, `providesOidcConfiguration`) gain `synchronized` null-check memoization fields to close the scoping gap that exists when `@Provides @Singleton` methods call each other as plain Java methods.

Nine new singleton providers are added for previously-unmanaged leaf types:

| New provider | Type |
|---|---|
| `providesOntologyDAO()` | `OntologyDAO` |
| `providesClaimsCache()` | `ClaimsCache` |
| `providesTranslationUtil()` | `TranslationUtil` |
| `providesDataUseUtil()` | `DataUseUtil` |
| `providesDataUseMatcherV4()` | `DataUseMatcherV4` |
| `providesCountryValidator()` | `CountryValidator` |
| `providesInstitutionUtil()` | `InstitutionUtil` |
| `providesJsonSchemaUtil()` | `JsonSchemaUtil` |
| `providesTicketFactory()` | `TicketFactory` |

Existing providers that were constructing those types with `new` are updated to receive them as injected parameters.

#### `ClaimsCache` — static singleton removed

`ClaimsCache` previously held a static `INSTANCE` field with no `volatile` and no synchronization — a race-prone pattern. The static field and `getInstance()` are removed; the constructor is changed from `private` to `@Inject public`, making `ClaimsCache` a proper Guice-managed singleton.

#### `RequestHeaderCacheFilter` — constructor injection + registration fix

The filter was acquiring its `ClaimsCache` via the (now-deleted) `ClaimsCache.getInstance()` field initializer. It now receives `ClaimsCache` through an `@Inject` constructor.

The registration in `ConsentApplication` is updated from:

```java
env.jersey().register(RequestHeaderCacheFilter.class);
```

to:

```java
env.jersey().register(injector.getInstance(RequestHeaderCacheFilter.class));
```

Registering by class let Jersey/HK2 create its own unmanaged instance, which would have been a different `ClaimsCache` object from the one used by the auth pipeline. Registering the Guice instance ensures the filter and all auth components share the same singleton.

#### Service and resource constructors — stop using `new`

| Class | Before | After |
|---|---|---|
| `OntologyService` | `new TranslationUtil(jdbi.onDemand(...))` | `TranslationUtil` injected |
| `DataUseMatcherV4` | `new DataUseUtil(ontologyService)` | `DataUseUtil` injected |
| `MatchService` | `new DataUseMatcherV4(ontologyService)` | `DataUseMatcherV4` injected |
| `DataAccessRequestService` | `new CountryValidator()` | `CountryValidator` injected |
| `SupportRequestService` | `new HttpClientUtil(config)`, `new TicketFactory()` | both injected |
| `DatasetResource` | `new JsonSchemaUtil()` | `JsonSchemaUtil` injected |
| `SchemaResource` | `new JsonSchemaUtil()` | `JsonSchemaUtil` injected |
| `InstitutionResource` | `new InstitutionUtil()` | `InstitutionUtil` injected |
| `AuthorizationHelper` | `ClaimsCache.getInstance()` | `ClaimsCache` injected |

#### Leaf classes — `@Inject` annotations

`DataUseUtil`, `TranslationUtil`, `CountryValidator`, `InstitutionUtil`, `JsonSchemaUtil`, and `TicketFactory` each gain `@Inject` on their existing constructors so Guice can construct and wire them.

#### Gson — deferred TODO comments - see [https://broadworkbench.atlassian.net/browse/DT-3689](https://broadworkbench.atlassian.net/browse/DT-3689) for follow-on work.

`VoteResource`, `DaaResource`, `StudyResource`, `DatasetResource`, `SamHealthCheck`, and `SendGridHealthCheck` contain `new Gson()` calls. Switching to `GsonUtil.buildGson()` requires further investigation of Gson configuration options; each site is annotated with a TODO rather than changed:

```java
// TODO: Replace new Gson() with GsonUtil.buildGson() — deferred pending Gson configuration investigation
```

---

### Tests

- **`ConsentModuleTest`** — new `testNewSingletonProvidersReturnTheSameInstance` asserts `assertSame` across all nine new providers; updates the four existing config-getter singleton assertions.
- **`ClaimsCacheTest`** _(new)_ — covers constructor initialisation, cache population, no-overwrite on repeat token, non-OAuth header filtering, and the null-value-list branch (documents that `Collectors.toMap` throws `NullPointerException` for null header value lists).
- **`RequestHeaderCacheFilterTest`** _(new)_ — covers the token-present path (verifies `loadCache` is called with the stripped bearer value) and the no-token path (verifies `loadCache` is never called).
- All existing tests that construct the modified services and resources are updated to supply the now-required injected parameters directly.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/OntologyService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/OntologyService.java
@@ -19,6 +19,7 @@ import org.broadinstitute.consent.http.service.ontology.OntologyDAO;
 import org.broadinstitute.consent.http.service.ontology.OntologyIndexService;
 import org.broadinstitute.consent.http.service.ontology.OntologyTerm;
 import org.broadinstitute.consent.http.util.ConsentLogger;
+import org.jdbi.v3.core.Jdbi;
 import org.jspecify.annotations.NonNull;
 
 public class OntologyService implements ConsentLogger {
@@ -30,11 +31,11 @@ public class OntologyService implements ConsentLogger {
 
   @Inject
   public OntologyService(
-      OntologyDAO ontologyDAO,
+      Jdbi jdbi,
       OntologyIndexService indexService,
       ExecutorService executorService,
       TranslationUtil translationUtil) {
-    this.ontologyDAO = ontologyDAO;
+    this.ontologyDAO = jdbi.onDemand(OntologyDAO.class);
     this.indexService = indexService;
     this.executorService = executorService;
     this.translationUtil = translationUtil;
--- a/src/main/java/org/broadinstitute/consent/http/service/OntologyService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/OntologyService.java
@@ -19,7 +19,6 @@ import org.broadinstitute.consent.http.service.ontology.OntologyDAO;
 import org.broadinstitute.consent.http.service.ontology.OntologyIndexService;
 import org.broadinstitute.consent.http.service.ontology.OntologyTerm;
 import org.broadinstitute.consent.http.util.ConsentLogger;
-import org.jdbi.v3.core.Jdbi;
 import org.jspecify.annotations.NonNull;
 
 public class OntologyService implements ConsentLogger {
@@ -31,11 +30,14 @@ public class OntologyService implements ConsentLogger {
 
   @Inject
   public OntologyService(
-      Jdbi jdbi, OntologyIndexService indexService, ExecutorService executorService) {
-    this.ontologyDAO = jdbi.onDemand(OntologyDAO.class);
+      OntologyDAO ontologyDAO,
+      OntologyIndexService indexService,
+      ExecutorService executorService,
+      TranslationUtil translationUtil) {
+    this.ontologyDAO = ontologyDAO;
     this.indexService = indexService;
     this.executorService = executorService;
-    this.translationUtil = new TranslationUtil(this.ontologyDAO);
+    this.translationUtil = translationUtil;
   }
 
   public DataUseSummary translateDataUseSummary(DataUse dataUse) {
--- a/src/main/java/org/broadinstitute/consent/http/ConsentApplication.java
+++ b/src/main/java/org/broadinstitute/consent/http/ConsentApplication.java
@@ -215,7 +215,7 @@ public class ConsentApplication extends Application<ConsentConfiguration> {
     env.jersey().register(feature);
     env.jersey().register(binder);
 
-    env.jersey().register(RequestHeaderCacheFilter.class);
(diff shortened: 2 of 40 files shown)


src/main/java/org/broadinstitute/consent/http/service/OntologyService.java
```diff
@@ -31,11 +30,14 @@ public class OntologyService implements ConsentLogger {
 
   @Inject
   public OntologyService(
-      Jdbi jdbi, OntologyIndexService indexService, ExecutorService executorService) {
-    this.ontologyDAO = jdbi.onDemand(OntologyDAO.class);
```
rushtong, line 35: I can see where this came from - the fact that we added an `OntologyDAO` in `ConsentModule` means it can be injected here but that is counter to the [new advice around DAO instantiation](https://github.com/DataBiosphere/consent/blob/develop/docs/ai/CLAUDE.md) that we added in recent work:

> All Jdbi SQL object DAOs (interfaces in `org.broadinstitute.consent.http.db`) must be
instantiated inside the service constructor via `jdbi.onDemand(XxxDAO.class)`. Never pass
a raw DAO as a constructor parameter to a service.

rushtong (commented):
This is fantastic work, thank you! See one comment inline.

src/main/java/org/broadinstitute/consent/http/service/OntologyService.java
otchet-broad: Updated.  Good catch.  Thank you!

I think this confirms that sometimes Claude.ai will follow the references in the project's root (CLAUDE.md) and sometimes it won't.  From what I've been reading, there's reasoning behind it.  That means we might need to consider building agentic advice tailored for each vendor.

rushtong (approved):
Thanks for the updates 👍🏽

fboulnois (approved):
nice work 👍

kevinmarete (approved):

PR #2944, opened 2026-06-22
[DT-3690] Apply ClaimsCache and RequestHeaderCacheFilter feedback adjustments

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3690](https://broadworkbench.atlassian.net/browse/DT-3690)

### Summary

ClaimsCache.java

- Race condition (check-then-put): Replaced the non-atomic getIfPresent + put pattern with cache.get(token, callable), which serialises loads per key under Guava's own lock. Two threads racing on the same new token now produce a single cache entry.

- NPE / auth bypass: The previous fallback stored "" for null or empty OAUTH2_CLAIM_* value lists. Collectors.toMap threw NullPointerException on null, and the "" sentinel would have bypassed the email == null guard in AuthorizationHelper. Both problems are fixed by filtering out null/empty-list entries before collecting — absent claims produce no map entry, so existing null-guards fire correctly.

- Unnecessary lock contention: Removed cache.cleanUp() from the insertion path. Guava performs expiry lazily during normal cache operations; calling cleanUp() on every write added unnecessary segment-level locking.

- Dead exception handler: Changed catch (ExecutionException _) to catch (Exception _). Guava re-throws RuntimeExceptions from the callable as UncheckedExecutionException (a RuntimeException subtype), which ExecutionException never intercepts. The original catch block was dead for all realistic failure paths; the intent — don't crash the request filter if the header map is malformed — now actually works.

RequestHeaderCacheFilter.java

- Case-sensitive Bearer stripping (RFC 7235): Replaced token.replace("Bearer ", "") with token.replaceFirst("(?i)^Bearer ", ""). OAuth2 clients that send bearer <token> (lowercase) were storing the full prefixed string as the cache key, causing a permanent cache miss for every request from those clients.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/filters/ClaimsCache.java
+++ b/src/main/java/org/broadinstitute/consent/http/filters/ClaimsCache.java
@@ -4,8 +4,6 @@ import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.inject.Inject;
 import jakarta.ws.rs.core.MultivaluedMap;
-import java.util.AbstractMap;
-import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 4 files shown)


rushtong (approved):
These are great improvements, thank you 👍🏽

kevinmarete (approved):
Very Nice 👍🏾

PR #2953, opened 2026-06-25
DT-3019: Update DAR Metrics to use Summary objects

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-3019

## Summary

The DAR metrics endpoint was materializing full `DataAccessRequest` objects — running each row through `DataAccessRequestReducer` and parsing every DAR's `data` JSON blob — only to copy six fields into a `DarMetricsSummary`. This reworks the query to select exactly the fields the summary needs and map them straight into the DTO.

**Changes**

- **`DataAccessRequestDAO.findSummaryMetricApprovedDARsByDatasetIdIncludesExpired`** now returns `List<DarMetricsSummary>` directly via a new lightweight `DarMetricsSummaryMapper`, instead of `List<DataAccessRequest>`. The query selects only the six columns the summary uses (`dar_code`, `submission_date`, `reference_id`, `update_date`, `projectTitle`, `nonTechRus`) rather than the full DAR row + JSON.
- **Result shape change:** results are now **one summary per qualifying DAR collection**, sourced from the most recently submitted DAR in that collection, instead of one row per approved DAR. The set of qualifying collections (approved-by-last-vote or carrying a closeout supplement, including expired) is unchanged — only the per-collection de-duplication is new.
- **`MetricsService.generateDarSummaries`** returns the DAO result directly; the `.stream().map(DarMetricsSummary::new)` mapping step is gone.
- **`DarMetricsSummary`** is now a plain record — the `DataAccessRequest`-based constructor (and its `DarMetricsSummaryTest`) are removed, since mapping happens in the DAO mapper.
- **Bug fix in `RowMapperHelper.hasOptionalColumn`:** switched `Optional.of(...)` → `Optional.ofNullable(...)` so a column that is present but `NULL` returns `Optional.empty()` instead of throwing an NPE that gets swallowed and logged as a misleading "RowView does not contain column" debug line (Copilot finding).
- Tests updated: rewrote the DAO integration test for the new return type and one-per-collection behavior, and updated the service/resource tests.

**Security risk modifier:** no

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
```diff
@@ -88,7 +90,8 @@ AND LOWER(election.status) IN ('open', 'closed', 'canceled')
         AND (e.latest = e.election_id OR e.election_id IS NULL)
       GROUP BY
         c.collection_id, c.dar_code, latest_dar.submission_date, latest_dar.reference_id, latest_dar.parent_id,
-        latest_dar.requires_so_approval, latest_dar.approving_so_id, latest_dar.approving_so_timestamp, researcher.display_name, i.institution_name,
+        latest_dar.requires_so_approval, latest_dar.approving_so_id, latest_dar.approving_so_timestamp,
```
otchet-broad: If I'm understanding correctly, this query can return expired DARs.  Do we want those in this response?

otchet-broad: I'm also thinking a bit more broadly about this and wonder if we should check in with product.

I wonder if we really want to reveal this information for all datasets in the system.  It effectively tells everyone who is authorized for what without being a service account, right?  We added new endpoints to specifically allow other users to see this type of information.  I realize this might be a question best answered by @jlaw-codes

otchet-broad (commented):
Code looks good, but left a comment about a thought for @jlaw-codes about the general concept here.

src/main/java/org/broadinstitute/consent/http/db/DarCollectionSummaryDAO.java
rushtong: I think your comment is meant not for this query, but for `getDarCollectionSummariesByDatasetId`
If you compare the query there, we are pulling the same DARs that we were pulling in the old method using the `approved_collections` CTE. That's what we're then filtering on to generate the list of summaries.

> I wonder if we really want to reveal this information for all datasets in the system. It effectively tells everyone who is authorized for what without being a service account, right?

I'm not seeing this. We've added two new fields to the summary objects, last update timestamp and the non-tech-rus statement.

rushtong: I'm going to pull this back into progress, I think I have a better way to do this without resorting to using DarCollectionSummary objects.

otchet-broad:
Nice refactor!

kevinmarete (approved):
That DAO query was a mouthful but otherwise nice and clean 👍🏾

src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java
```diff
@@ -916,57 +911,116 @@ void testFindSummaryMetricApprovedDARsByDatasetIdIncludesExpired() {
...
+        createDataAccessRequest(
+            collectionId, user.getUserId(), Date.from(Instant.now().minus(10, ChronoUnit.DAYS)));
+    dataAccessRequestDAO.insertDARDatasetRelation(
+        approvedDAR.getReferenceId(), queriedDataset.getDatasetId());
+    Election election =
+        createDataAccessElection(approvedDAR.getReferenceId(), queriedDataset.getDatasetId());
+    Vote vote = createFinalVote(queriedDataset.getCreateUserId(), election.getElectionId());
+    updateVote(true, "", now, vote.getVoteId(), false, election.getElectionId(), now, false);
+
+    // Later DAR in the same collection, linked only to a different dataset
```
otchet-broad, line 1012: 👍
Thanks for covering this.  I was looking at the query and this test is helpful to have!

otchet-broad (approved):
👍

PR #2957, opened 2026-06-29
[DT-3593] Study Registration validator and tests.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3593](https://broadworkbench.atlassian.net/browse/DT-3593)

### Summary

Introduces two new validator classes for the StudyRegistrationRequest and StudyUpdateRequest DTOs added in DT-3592, providing comprehensive server-side validation before those payloads reach service layer logic.

- StudyRegistrationRequestValidator — validates new study registration payloads:
  - Required fields: study name (non-blank), description, data types, public visibility, NIH AnVIL use, and PI name
  - NIH AnVIL use conditional fields: dbGaP phs ID, PI institution, and NIH grant/contract number are required depending on the selected NihAnvilUse value
  - GSR controlled access: explanation is required when controlledAccessRequiredForGenomicSummaryResultsGSR is true
  - Alternative data sharing plan: explanation and at least one reason are required when the plan flag is set
  - Email format: validates PI email and all data custodian emails via Apache Commons EmailValidator
  - Date format: validates embargo release date, alternative data sharing plan delivery/public release dates, and per-consent-group MOR dates against ISO-8601 (YYYY-MM-DD)
  - Consent groups: at least one required; each must have a name, number of participants, exactly one primary data use (open access, GRU, HMB, POA, disease-specific, or other), and a DAC if access is controlled
- StudyUpdateRequestValidator — validates update payloads against the existing study state:
  - All shared field validations above (description, data types, visibility, NIH fields, GSR, emails, dates)
  - Study name uniqueness: if the name changes, checks via DatasetService.findAllStudyNames() that it is not already taken; unchanged names skip the service call
  - Data submitter immutability is enforced structurally — dataSubmitterUserId is absent from StudyUpdateRequest by design
  - Consent group membership: existing consent groups in the payload must belong to the study being updated
  - Consent group name changes: renaming an existing consent group is blocked; only new consent groups (no datasetId) can receive a name
  - Consent group removal: the payload must include all datasets already associated with the study (omitting one is rejected)
  - New consent groups (no datasetId): validated with the same rules as the create path; existing consent groups skip data-use re-validation

#### Tests

Both validators are covered by unit test classes (StudyRegistrationRequestValidatorTest, StudyUpdateRequestValidatorTest) with ~100 test cases total, covering every required field, each conditional braf primary data use, email and date format edge cases, and the update-specific consent group lifecycle rules. The update validator uses Mockito to stub DatasetService.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/dto/registration/StudyRegistrationRequestValidator.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/dto/registration/StudyRegistrationRequestValidator.java
@@ -1,0 +1,171 @@
+package org.broadinstitute.consent.http.models.dto.registration;
+
+import jakarta.ws.rs.BadRequestException;
+import java.time.LocalDate;
+import java.time.format.DateTimeParseException;
+import java.util.List;
+import java.util.Objects;
+import org.apache.commons.validator.routines.EmailValidator;
+import org.broadinstitute.consent.http.models.dataset_registration_v1.ConsentGroup.AccessManagement;
+import org.broadinstitute.consent.http.models.dataset_registration_v1.DatasetRegistrationSchemaV1.NihAnvilUse;
+
+public class StudyRegistrationRequestValidator {
+
+  public boolean validate(StudyRegistrationRequest registration) {
+    validateRequiredStudyFields(registration);
+    validateConditionalStudyFields(registration);
+    validateEmailFields(registration);
+    validateDateFields(registration);
+    validateConsentGroups(registration.getConsentGroups());
+    return true;
+  }
+
+  private void validateRequiredStudyFields(StudyRegistrationRequest registration) {
+    if (Objects.isNull(registration.getStudyName()) || registration.getStudyName().isBlank()) {
+      throw new BadRequestException("Study Name is required");
+    }
+    if (Objects.isNull(registration.getStudyDescription())) {
+      throw new BadRequestException("Study Description is required");
+    }
+    if (Objects.isNull(registration.getDataTypes()) || registration.getDataTypes().isEmpty()) {
+      throw new BadRequestException("Data Types is required");
+    }
+    if (Objects.isNull(registration.getPublicVisibility())) {
+      throw new BadRequestException("Public Visibility is required");
+    }
+    if (Objects.isNull(registration.getNihAnvilUse())) {
+      throw new BadRequestException("NIH Anvil Use is required");
+    }
+    if (Objects.isNull(registration.getPiName())) {
+      throw new BadRequestException("Principal Investigator is required");
+    }
+  }
+
+  private void validateConditionalStudyFields(StudyRegistrationRequest registration) {
+    NihAnvilUse anvilUse = registration.getNihAnvilUse();
+    if (anvilUse.equals(NihAnvilUse.I_AM_NHGRI_FUNDED_AND_I_HAVE_A_DB_GA_P_PHS_ID_ALREADY)) {
+      if (Objects.isNull(registration.getDbGaPPhsID())) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/consent/http/models/dto/registration/StudyRegistrationRequestValidator.java
```diff
@@ -0,0 +1,180 @@
...
+import java.util.Objects;
+import org.apache.commons.validator.routines.EmailValidator;
+import org.broadinstitute.consent.http.models.dataset_registration_v1.ConsentGroup.AccessManagement;
+import org.broadinstitute.consent.http.models.dataset_registration_v1.DatasetRegistrationSchemaV1.NihAnvilUse;
+
+public class StudyRegistrationRequestValidator {
+
+  public boolean validate(StudyRegistrationRequest registration) {
+    validateRequiredStudyFields(registration);
+    validateConditionalFields(registration);
```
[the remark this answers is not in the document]
otchet-broad: This is intentional based on the scope of the ticket.

src/main/java/org/broadinstitute/consent/http/models/dto/registration/StudyRegistrationRequestValidator.java
```diff
@@ -0,0 +1,180 @@
...
+    if (Boolean.TRUE.equals(cg.getHmb())) count++;
+    if (Boolean.TRUE.equals(cg.getPoa())) count++;
+    if (cg.getDiseaseSpecificUse() != null && !cg.getDiseaseSpecificUse().isEmpty()) count++;
+    if (cg.getOtherPrimary() != null && !cg.getOtherPrimary().isBlank()) count++;
+    if (count != 1) {
+      throw new BadRequestException(
+          "Consent Group must have exactly one primary data use (open access, or one of:"
+              + " general research use, health/medical/biomedical, populations/origins/ancestry,"
+              + " disease-specific, other)");
+    }
```
rushtong, line 159: This is all correct based on current UX restrictions. I want to point out that there is an historic inconsistency that we should deal with. These conditions are not aligned with existing `DataUse` logic decisions. See, for example, [matchDiseases](https://github.com/DataBiosphere/consent/blob/5dd97d6ba13cd5ad55f9f681e338a8f0232b5f71/src/main/java/org/broadinstitute/consent/http/matching/DataUseMatchCasesV4.java#L64) where we assume that a dataset's data use **can** contain more than one primary data use condition. This is just drift from moving over data use conditions from Ontology to Consent and moving previous UI logic to Consent as well. Having all of this logic in Consent will hopefully allow us to move towards better consistency.

rushtong (approved):
Nice work 👍🏽

src/main/java/org/broadinstitute/consent/http/models/dto/registration/StudyRegistrationRequestValidator.java
otchet-broad: @rushtong - Do you have thoughts in a ticket somewhere on how to deal with this?  I'm wondering if we should bring that into the sprint and resolve the inconsistency before moving forward with the next phase of this work.

kevinmarete (approved):
Excellent 👍🏾

src/main/java/org/broadinstitute/consent/http/models/dto/registration/StudyRegistrationRequestValidator.java
rushtong: Closing the loop - see new ticket: https://broadworkbench.atlassian.net/browse/DT-3764

PR #2961, opened 2026-06-30
[DT-3594] Parallel shadow validators with timing.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3594](https://broadworkbench.atlassian.net/browse/DT-3594)

### Summary

Wires the new DTO-based StudyRegistrationRequestValidator / StudyUpdateRequestValidator (added in #2957) into the live create/update study-registration endpoints as non-authoritative shadow validators. On every create and update request, the existing (authoritative) JSON Schema / manual validation result is compared against the new validator's result, and the agreement/disagreement plus timing for each is logged. The new validators never affect the response returned to the client — they are purely observational, to build confidence before cutting over.

Changes

- RegistrationShadowValidator (new): runs the new validator against the same request JSON used by the authoritative path, times both outcomes, and logs:
  - [VALIDATOR_PARITY:create|update] AGREE accepted=... oldDurationMicros=... newDurationMicros=... when the two validators agree on accept/reject
  - [VALIDATOR_PARITY:create|update] DISAGREE old=ACCEPT/REJECT new=ACCEPT/REJECT oldMessage=... newMessage=... ... when they disagree
  - A warning and a no-op comparison if no authoritative outcome was supplied (defensive guard)
  - Catches and records any exception/deserialization failure from the new validator so it can never propagate to the caller
- DatasetResource#createDatasetRegistration: times the existing JSON-schema validation and forwards its outcome (accept or reject) to RegistrationShadowValidator.compareCreate before continuing with (or rejecting) the request as before.
- StudyResource#updateStudyByRegistration: times the existing deserialize+validate path, captures success, validation failure, or thrown exception, and forwards the outcome to RegistrationShadowValidator.compareUpdate — including the malformed-JSON/exception case — before re-throwing or proceeding exactly as the authoritative path did previously.
- Both resources now take RegistrationShadowValidator as a constructor-in

Testing

- RegistrationShadowValidatorTest: covers ValidationOutcome factory methods, agreement/disagreement on both create and update flows, malformed-JSON handling, and the null-outcome skip path.
- DatasetResourceTest / StudyResourceTest: updated for the new constructor dependency; added a test confirming the shadow comparison still runs (with the correct args) even when the update request JSON is malformed and the authoritative path rejects it.

Risk / rollout notes

- No behavior change for end users — response codes and bodies are unchanged; this only adds logging.
- Shadow validator failures (exceptions, bad JSON) are caught internally and surfaced only as a DISAGREE/rejected log entry, never as an error to the caller.
- Next step once parity is confirmed via logs: cut the new validators oveove the JSON Schema / manual path.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/RegistrationShadowValidator.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/RegistrationShadowValidator.java
@@ -1,5 +1,6 @@
 package org.broadinstitute.consent.http.service;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.inject.Inject;
 import jakarta.ws.rs.BadRequestException;
@@ -110,7 +111,7 @@ public class RegistrationShadowValidator implements ConsentLogger {
   @FunctionalInterface
   private interface ThrowingRunnable {
 
-    void run() throws Exception;
+    void run() throws JsonProcessingException;
   }
 
   private ValidationOutcome runTimed(ThrowingRunnable validation) {
@@ -120,7 +121,7 @@ public class RegistrationShadowValidator implements ConsentLogger {
       return ValidationOutcome.acceptedSince(start);
     } catch (BadRequestException e) {
       return ValidationOutcome.rejectedSince(e.getMessage(), start);
-    } catch (Exception e) {
+    } catch (JsonProcessingException e) {
       return ValidationOutcome.rejectedSince(
           "Unable to deserialize/validate: " + e.getMessage(), start);
     }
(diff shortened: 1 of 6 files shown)


src/main/java/org/broadinstitute/consent/http/service/RegistrationShadowValidator.java
```diff
@@ -0,0 +1,168 @@
...
+import org.broadinstitute.consent.http.models.dto.registration.StudyRegistrationRequestValidator;
+import org.broadinstitute.consent.http.models.dto.registration.StudyUpdateRequest;
+import org.broadinstitute.consent.http.models.dto.registration.StudyUpdateRequestValidator;
+import org.broadinstitute.consent.http.util.ConsentLogger;
+
+/**
+ * Runs the new DTO-based registration validators in shadow (non-authoritative) mode alongside the
+ * existing JSON Schema / manual validation, and logs agreement or discrepancy. This class never
+ * throws back to its caller; the result it returns is informational only and must not affect the
+ * authoritative validation outcome.
```
rushtong, line 18: Very nice

rushtong (approved):
This looks reasonable to me. Looking forward to seeing the comparisons in dev 👍🏽

kevinmarete (approved):
Excellent 👍🏾

PR #2962, opened 2026-07-03
[DT-3774] Fix missing `dac_approval` column in `DacDAO.findAll()` 

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3774

### Summary

`GET /api/dac` never selected `dac_approval` for the datasets embedded in each DAC, so `dacApproval` was always `null` on that response — causing the Manage DAC table's dataset counts to disagree with the DAC profile page (which uses a different endpoint that does select this column). Added the column to the query and mapped it in `DacReducer`, added a `Dac.getDatasets()` accessor, and added a DAO test covering approved vs. pending datasets.     

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/mapper/DacReducer.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/mapper/DacReducer.java
@@ -72,8 +72,8 @@ public class DacReducer
           dataset.setDataUse(dataUseParser.parseDataUse(duStr));
         }
 
-        Boolean dacApproval = rowView.getColumn("dataset_dac_approval", Boolean.class);
-        dataset.setDacApproval(dacApproval);
+        hasOptionalColumn(rowView, "dataset_dac_approval", Boolean.class)
+            .ifPresent(dataset::setDacApproval);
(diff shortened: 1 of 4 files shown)


fboulnois (approved):
👍

src/main/java/org/broadinstitute/consent/http/db/mapper/DacReducer.java
```diff
@@ -72,6 +72,9 @@ public void accumulate(Map<Integer, Dac> container, RowView rowView) {
           dataset.setDataUse(dataUseParser.parseDataUse(duStr));
         }
 
+        Boolean dacApproval = rowView.getColumn("dataset_dac_approval", Boolean.class);
+        dataset.setDacApproval(dacApproval);
```
rushtong: There is a helper method that will conditionally check if the column exists before trying to populate the value. It's a touch safer and will let you handle optional cases a little more smoothly.

rushtong (approved):
Nice catch, minor suggested improvement inline.

PR #2974, opened 2026-07-14
 [DT-3595] Switch dataset/study registration to DTO/domain validators

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3595

### Summary

  - Make the phase-2 DTO validators (`StudyRegistrationRequestValidator`, `StudyUpdateRequestValidator`) authoritative for `POST /api/dataset/v3` and `PUT /api/dataset/study/{studyId}`, replacing JSON Schema     
  validation and the legacy manual update validator.                                                                                                                                                                
  - Create aggregates all violations into one response (preserving today's multi-error UX); update stays fail-fast, matching its existing behavior.                                                                 
  - Removes `RegistrationShadowValidator` now that both endpoints are authoritative.                                                                                                                                
  - Aligns validator error messages with existing user-facing terminology (e.g. "Datasets" instead of "Consent Group(s)", full "Principal Investigator Name/Institution" instead of abbreviated/raw field names).   
              

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/models/dto/registration/StudyRegistrationRequestValidator.java
```diff
@@ -20,24 +23,80 @@ public boolean validate(StudyRegistrationRequest registration) {
     return true;
   }
 
+  /**
+   * Runs the same checks as {@link #validate(StudyRegistrationRequest)}, but collects every
```
rushtong: That method (`validate(StudyRegistrationRequest)`) looks dead now, it's only used in tests.

rushtong (approved):
Looks good, thank you 👍🏽

otchet-broad (approved):
+1 on @rushtong 's comment about removing the code that's only under test.

I've also been in favor of changing the encapsulation of private methods to protected so that they can be directly tested where possible.  I think we've been using the `@VisibleForTesting` IDE device to mark methods this way so that we can warn other developers they're not for external consumption but still directly validate each encapsulated function.  YMMV.

PR #2976, opened 2026-07-15
[DT-3211] Add Per-User Rate Limits to APIs

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-3211

## Summary
Adds per-user API rate limiting via a bucket4j token-bucket filter, reworked for our multi-pod (varies by env — 2 in dev/staging, 3 in prod) Kubernetes deployment where rate-limit state can't be shared in-memory across pods.

- New `RateLimitFilter` (`ContainerRequestFilter`, runs at `Priorities.USER`), scoped to requests under `api/` only — infrastructure endpoints like `/status` and `/liveness` (hit continuously by Kubernetes probes) and `swagger/` docs traffic are never subject to it, so the limiter can't cause probe failures or pod restarts.
- Keyed by the authenticated user's email, falling back to the first IP in `X-Forwarded-For` (correctly handling multi-hop proxy chains and blank values) for unauthenticated requests. Returns 429 with a dynamically computed Retry-After (derived from Bucket4j's actual refill wait time, not a hardcoded value) when the bucket is exhausted.
- Bucket storage is a bounded Guava `LoadingCache` (`expireAfterAccess + maximumSize`), not an unbounded map, so long-running pods don't accumulate memory from distinct callers indefinitely.
- New `RateLimitConfiguration`: enabled, `requestsPerMinute`, `podCount` are all required (`@NotNull`, `@Positive`) — no hardcoded Java defaults, so every environment must explicitly declare them. A cross-field `@AssertTrue` constraint rejects `requestsPerMinute < podCount`, which would otherwise let the per-pod share round up to a 1-request-per-minute floor and silently exceed the configured deployment-wide aggregate.
- `requestsPerMinute` is the intended limit for the whole deployment; each pod enforces `requestsPerMinute / podCount`, so the deployment-wide aggregate stays close to the configured value instead of being multiplied by the number of pods.
- Wired into `ConsentConfiguration`/`ConsentModule`/`ConsentApplication` following the existing `ClaimsCache`/`RequestHeaderCacheFilter` patterns.
- Test fixtures (`consent-config.yml`, `consent-ci.yaml`) updated with the now-required `rateLimit` block.
- `RateLimitFilterTest` (15 cases) covers: within/over-capacity behavior, dynamic Retry-After, enabled: false short-circuiting, independent buckets per user/IP, the full key-extraction fallback chain including multi-hop X-Forwarded-For and blank-value edge cases, and that `/status`, `/liveness`, `/swagger/*` are never rate-limited even under sustained load. `RateLimitConfigurationTest` (3 cases) covers the Bean Validation constraints directly.

[Companion PR required: terra-helmfile](https://github.com/broadinstitute/terra-helmfile/pull/6466) — adds `rateLimit` to the rendered `consent.yaml` config, with `podCount` driven by `.Values.replicas` (not duplicated) so the divisor always matches the actual per-environment pod count (bees run a single pod, dev/staging run 2 replicas, prod runs 3). `rateLimitEnabled`/`rateLimitRequestsPerMinute` are set per-env across dev, staging, prod, and bee.

### Local Testing
This can be tested by updating your local `requestsPerMinute` value to something low, like 5, and hitting any endpoint under `/api` repeatedly. You'll eventually get a 429:
```
{
  "origin": "jersey",
  "message": "Too Many Requests",
  "url": "http://app:8080/api/user/me",
  "status": "429"
}
```
Non-`/api` calls should not hit this error. The content of this error is outside the scope of this PR and needs to be addressed in a separate task (TODO)

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/filters/RateLimitFilter.java
```diff
@@ -0,0 +1,99 @@
...
+
+  private String extractKey(ContainerRequestContext requestContext) {
+    SecurityContext securityContext = requestContext.getSecurityContext();
+    Principal principal = securityContext == null ? null : securityContext.getUserPrincipal();
+    if (principal instanceof AuthUser authUser
+        && authUser.getEmail() != null
+        && !authUser.getEmail().isBlank()) {
+      return authUser.getEmail();
+    }
+    String forwardedFor = requestContext.getHeaderString("X-Forwarded-For");
```
otchet-broad: If we're not allowing any data API access without being a valid user, do we need to do this?
I worry that infrastructure could change around us and suddenly everyone is locked out.
The other thing I'm thinking about is the filter order.  What's the pattern like when we need to return a 401 because the token has expired?

rushtong: The API point is a great catch, thank you - will rework that. Auth filtering is handled before this filter sees the request (see `OAuthCustomAuthFilter`) so it should never see a 401.

rushtong: Code is updated and includes additional documentation to explain current behavior.

otchet-broad (approved):
👍
Getting rid of the 'by-ip' limiting relieves all of the other concerns or things I was thinking about.  This is great.  Thank you!

kevinmarete (approved):

PR #2979, opened 2026-07-17
[DT-3788] Migrate eligible AuthUser resource endpoints to DuosUser

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-3788

### Summary

Migrates 51 of the 52 `@Auth AuthUser` resource endpoints to `@Auth DuosUser` across 12 resource
classes (`EmailNotifierResource`, `DatasetResource`, `DacResource`, `MatchResource`,
`StudyResource`, `DraftResource`, `MailResource`, `UserResource`, `VoteResource`,
`FeatureFlagResource`, `DataAccessRequestResource`, `DACAutomationRuleResource`), removing the
redundant `UserService.findUserByEmail(authUser.getEmail())` lookup from each migrated method in
favor of `duosUser.getUser()` / `duosUser.getUserId()`.

`UserResource.createResearcher` is the one endpoint left on `AuthUser` — it's the registration
flow and explicitly depends on the caller *not* having a Consent user yet, which `DuosUser`
doesn't allow.

No business logic moved into the resource layer as part of this cleanup; only the principal type
and the now-redundant lookup changed. Response bodies and success status codes are unchanged.

**Behavior change (documented, not a bug):** for role-restricted endpoints, an authenticated
caller who has no Consent user row now gets **401** from `DuosUserAuthenticator` at the
authentication layer, instead of whatever status the resource/authorizer produced before (usually
403, from `AuthorizationHelper.authorize()` silently swallowing the lookup failure). Role
authorization already performed its own independent `findUserByEmail` lookup regardless of the
resource's principal type, so this migration doesn't introduce new lookup risk — it only shifts
*when* the "no Consent user" case is caught. `src/main/resources/assets/paths/*.yaml` (and the
inline paths in `api-docs.yaml`) have been updated to document `401` on the ~24 affected operations
that didn't already list it.

**Tests:** updated ~15 resource test files to use `DuosUser` fixtures in place of `AuthUser`,
dropping now-dead `findUserByEmail` stubs (would otherwise trip Mockito's
`UnnecessaryStubbingException`) and the handful of tests that specifically exercised the
resource-level `NotFoundException` catch — that scenario is now handled upstream by the
authenticator and is already covered generically in `AuthorizationHelperTest`. Added a
`filterUsers` test to `DacResourceTest` (previously untested). Full suite: 3379 tests, 0
failures/errors.

[DT-3788]: https://broadworkbench.atlassian.net/browse/DT-3788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kevinmarete (approved):
Very clean, handle the nit sonar warnings and all good!

otchet-broad (approved):
👍 Nice cleanup!

fboulnois (approved):
One minor suggestion, otherwise looks good! 

Currently, there is no direct unit test for the 401-on-unregistered path. The 401 behavior is now the crux of these endpoints' contract, but there's no DuosUserAuthenticatorTest. This is a pre-existing gap, but a small test asserting authenticate() returns Optional.empty() on NotFoundException would verify this behavior.

rushtong:
> One minor suggestion, otherwise looks good!
> 
> Currently, there is no direct unit test for the 401-on-unregistered path. The 401 behavior is now the crux of these endpoints' contract, but there's no DuosUserAuthenticatorTest. This is a pre-existing gap, but a small test asserting authenticate() returns Optional.empty() on NotFoundException would verify this behavior.

Excellent point, tests added!

PR #2983, opened 2026-07-22
[DT-2257] Replace ErrorResource/ErrorPageErrorHandler with JAX-RS ExceptionMappers for 403/404

rushtong:
### Addresses
https://broadworkbench.atlassian.net/browse/DT-2257

### Summary
- Replaces the old `ErrorResource` + Jetty `ErrorPageErrorHandler` 404 forwarding with proper JAX-RS `ExceptionMapper`s: `NotFoundExceptionMapper` and `ForbiddenExceptionMapper`. Both normalize responses into the app's standard `Error(message, code)` JSON shape instead of leaking Jetty/Jersey's own default error format (e.g. `{"servlet":"jersey","message":"Forbidden","url":...,"status":"403"}`).
- Adds `JsonErrorHandler`, a thin Jetty-level `ErrorHandler` that's registered as a last-resort net for requests that never reach Jersey's dispatch at all (e.g. a path matching no `@Path` template) — JAX-RS `ExceptionMapper`s can't intercept that case since it never becomes a JAX-RS exception. Keeps every error response in the same JSON shape regardless of which layer produced it.
- Updates `docs/ARCHITECTURE.md`, `.github/copilot-instructions.md`, and `docs/ai/prompts/api-endpoint.md` to point at the new mapper convention instead of the removed `ErrorResource`.

## Why

401/403 error responses were surfacing Jetty's raw default JSON body instead of the app's normal error shape, because `RolesAllowedDynamicFeature` throws a `ForbiddenException` with no response entity, and nothing was writing a body before it hit Jetty's fallback handler. Fixing this at the JAX-RS layer alone (`ExceptionMapper`) covers in-pipeline cases (explicit resource/service throws, `@RolesAllowed` failures) but not genuinely unmatched routes — those bypass Jersey's exception-mapping entirely, which is why `JsonErrorHandler` still exists as a Jetty-level fallback for that one case.

401 (dropwizard-auth's `NotAuthorizedException`) is intentionally out of scope here — it already returns a non-empty body (`DefaultUnauthorizedHandler` sets an entity), so it isn't hitting the Jetty fallback bug and doing it right would require also swapping in a custom `UnauthorizedHandler`.

## Test plan

- [x] `NotFoundExceptionMapperTest` / `ForbiddenExceptionMapperTest` — unit tests covering messaged and message-less exceptions
- [x] `NotFoundTests` (integration, boots the full app via `DropwizardAppExtension`) — confirms an unmatched path now returns a JSON `Error` body instead of Jetty's HTML error page
- [x] Full existing test suite passes (`mvn clean test`)

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/mappers/JsonErrorHandler.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * A request only reaches this Jetty-level handler when it never entered Jersey's JAX-RS dispatch at
+ * all - most notably, a path that matches no {@code @Path} template. Jersey's own {@link
+ * jakarta.ws.rs.ext.ExceptionMapper}s (see {@link NotFoundExceptionMapper}, {@link
+ * ForbiddenExceptionMapper}) cannot intercept that case, since it never becomes a JAX-RS exception;
+ * Jetty resolves it directly via {@code HttpServletResponse.sendError}. This handler is the
+ * last-resort net that keeps that response in the app's standard {@link Error} JSON shape instead
+ * of Jetty's own error page format, ignoring the client's Accept header since this is a JSON API
+ * with no HTML or plain-text representation to negotiate to.
+ */
```
otchet-broad, line 24: Some feedback on this from Claude:
1. JsonErrorHandler's documented trigger is likely inaccurate — and it has no integration coverage (Low–Medium)
The Javadoc says it handles "a path that matches no @Path template … never reach[ing] Jersey's dispatch." But with applicationContextPath: / and no custom rootPath, the Jersey servlet is mapped at /*, so unmatched paths do reach Jersey and throw NotFoundException → handled by NotFoundExceptionMapper. That means the NotFoundTests integration test (GET /where/am/i) almost certainly exercises NotFoundExceptionMapper, not JsonErrorHandler. Keeping JsonErrorHandler as defense-in-depth is fine, but consider:
- Reword the comment to describe what actually reaches it (errors Jetty raises outside Jersey — e.g. malformed requests, filter/sendError failures before dispatch), rather than the @Path-miss example.
- If you want real coverage of it, an integration test that triggers a pre-Jersey error would be needed (harder to construct); otherwise note that the unit test is the only coverage.

src/main/java/org/broadinstitute/consent/http/mappers/JsonErrorHandler.java
```diff
@@ -0,0 +1,57 @@
...
+
+  @Override
+  protected void writeErrorJson(
+      Request request, PrintWriter writer, int code, String message, Throwable cause) {
+    String finalMessage = message;
+    if (code == 404
+        && (message == null || message.isBlank() || message.equals(HttpStatus.getMessage(code)))) {
+      finalMessage = "Unable to find requested path: " + request.getHttpURI().getPath();
+    }
+    writer.write(GsonUtil.getInstance().toJson(new Error(finalMessage, code)));
```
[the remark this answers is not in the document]
otchet-broad: More from Claude related to this:
NotFoundExceptionMapper uses uriInfo.getRequestUri().getPath() (URI.getPath() returns a decoded path), but JsonErrorHandler uses request.getHttpURI().getPath(), which returns the raw/encoded path. The old ErrorResource explicitly URLDecoder.decode(...)-ed. For encoded paths the two 404 messages will differ (encoded vs decoded). Consider getHttpURI().getDecodedPath() (or canonical path) for consistency.

otchet-broad (approved):
Looks good.  Claude found a couple of low priority issues I left in the comments.

src/main/java/org/broadinstitute/consent/http/mappers/JsonErrorHandler.java
rushtong: Quick reply to point 1 - this finding is incorrect. I manually tested this at multiple levels and the NFE mapper does not handle paths like `GET /where/am/i`. This, btw, was my exact test case that bypasses Jersey altogether.

rushtong: Addressed feedback around documentation and path encoding, thank you!

kevinmarete (approved):

PR #2984, opened 2026-07-22
[DT-3144] Adds VODAR plan with SWOT analysis.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3144](https://broadworkbench.atlassian.net/browse/DT-3144)

### Summary

Adds a proposed implementation plan for **VODAR** — a constrained, RADAR auto-approved Data Access Request that lets a researcher request *view-only* access to a dataset (no analysis, no publication) to decide whether to later pursue full access. Modeled as a normal `DataAccessRequest` with a view-only flag and fixed RUS, auto-approved when the dataset's DAC opts in — no manual vote. Spans `duos-ui` and `consent`, extending existing surfaces rather than adding a new entity.

### Changes

- **`docs/plans/vodar-plan.md`** (new) — the full plan.
- **`docs/plans/README.md`** — index entry.

### Contents

Documents current behavior, a requirements→behavior mapping and decisions table, **7 Jira-ready tickets** (4 `consent`, 2 `duos-ui`, 1 cross-repo E2E) with sizing/deps/AC/tests, plus refactoring/risk, Future Directions, and a SWOT analysis.

### Notes for reviewers

- **Documentation only** — no code or schema changes.
- Key tension flagged in the plan: "view only" is an **attestation** in the MVP (downstream TDR/auth API unchanged); true enforcement is deferred to external teams.

----

- Verify all tests go green

### DIFF
--- a/docs/plans/vodar-plan.md
+++ b/docs/plans/vodar-plan.md
@@ -20,8 +20,8 @@ auth-list push) rather than extensions of existing behavior.
 ## Background and Scope
 
 A VODAR is not a new entity. It is a `DataAccessRequest` whose data-use payload declares view-only
-intent and whose approval is automated. The work therefore extends the three existing surfaces the
-research below documents, rather than introducing a parallel request type.
+intent and whose approval is automated. The work therefore extends the three existing surfaces this
+plan documents, rather than introducing a parallel request type.
 
 ### Terminology
 
--- a/docs/plans/vodar-plan.md
+++ b/docs/plans/vodar-plan.md
@@ -18,6 +18,16 @@ This plan covers the minimum viable feature. Several product requirements are ex
 Future Directions section because they require net-new infrastructure (timed access revocation, active
 auth-list push) rather than extensions of existing behavior.
 
+**Product-confirmed framing — VODAR is a policy control, not a technical one.** Once a VODAR DAR is
+approved, downstream systems do not need to know or care that it was a VODAR. VODAR constrains the
+*user* (they attest to view-only intent and agree not to analyze or publish); it is not a technical
+access restriction imposed on downstream consumers of DUOS authorization grants. Consequently, a VODAR
+approval is intentionally identical to any other approval from the consumer contract's point of view,
+and this plan does **not** need to (and does not) modify the outward-facing approved-users / TDR / Sam
+APIs to carry a view-only scope. This resolves what earlier drafts treated as an "enforcement gap": there
+is no gap, because enforcement of view-only intent lives with the user and the DAC's policy, not with the
+data plane.
+
 ## Background and Scope
 
 A VODAR is not a new entity. It is a `DataAccessRequest` whose data-use payload declares view-only
@@ -35,7 +45,9 @@ plan documents, rather than introducing a parallel request type.
 - **Ethical questions** — for VODAR purposes, *all* of the Section 3 / Step 3
   (`ResearchPurposeStatement.tsx`) questions: the study-characteristic radios and the ethics-concern
   radios. If any Section 3 question is answered Yes, the request cannot be submitted or auto-approved as
-  a VODAR (see the decision table for the exact field list).
+  a VODAR (see the decision table for the exact field list). **Product-confirmed:** per the ticket, "do
+  not allow submission of a VODAR if any Ethical (Step 3) question is 'Yes'." This supersedes any narrower
+  legacy reading of "Section 3" — the gate is the entire current Step 3, both radio groups.
 
 ## Current Behavior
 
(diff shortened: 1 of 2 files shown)


otchet-broad:
Requirements in ticket were missing updates to plans.  I'm pulling this back to adjust.

kevinmarete (changes requested):
Overall, looks good but here are some comments from agent:

## Required changes

### 1. Resolve or explicitly approve the scope/enforcement gap

The MVP sends VODAR approvals through the existing `approvedUsers` contract, where they are indistinguishable from full DAR approvals and remain eligible for up to one year. This does not technically enforce view-only access or the intended short lifetime.

Please either:

- include VODAR-aware authorization scope and expiry enforcement in the MVP;
- exclude VODARs from the legacy `approvedUsers` result and define a separate authorization path; or
- document explicit Product and Compliance approval of the attestation-only MVP and its known over-grant risk.

### 2. Confirm the meaning of “Ethical (Step 3) questions”

The plan treats every Step 3 “Yes” as disqualifying, including study characteristics such as `controls`, `population`, `forProfit`, `oneGender`, and `pediatric`. DT-3144 could instead mean only the ethics-concern subset of Step 3.

Please obtain and record explicit PO confirmation of which fields should block VODAR submission before this becomes frontend and backend validation.

#### 3. Include OpenAPI updates

The implementation plan changes documented API contracts but does not include the required OpenAPI work.

Please add acceptance criteria to:

- add `vodar` to `src/main/resources/assets/schemas/DataAccessRequest.yaml`; and
- add `VODAR_V1` to the `ruleType` enum in `src/main/resources/assets/schemas/DacAutomationRule.yaml`.

These updates are required by the repository’s API-change guidance.

## Non-blocking refinements

- **RADAR UI:** The management UI already exists in `DACBotComponent.tsx` and `ManageRadar.tsx`. `VODAR_V1` must be added explicitly to `RULE_GROUP_LABELS`; otherwise it will appear under “Other.” The surrounding GRU/HMB-only explanatory copy also needs updating.
- **Cross-repository contract:** Specify how CI will compare the boilerplate across independently built repositories. Two separate tests containing the same expected literal do not prevent drift.
- **VODAR deselection:** Decide whether toggling VODAR off restores the previous RUS or clears the boilerplate. It should not leave VODAR text in a normal DAR.
- **Console labeling:** Researcher and DAC consoles consume `DarCollectionSummary`, which currently has no `vodar` field. Add the marker to that projection or create a follow-up ticket for console labeling.
- **E2E framework:** `duos-ui` uses Playwright, so Ticket 6 should say Playwright rather than “Cypress/Playwright.”
- **PR consistency:** The PR description says seven tickets, while the plan currently contains six. The title should also include the required risk assessment, such as `[risk=low]`.

otchet-broad:
@kevinmarete - For items 1 and 2, these are by design and I confirmed with product that the expected behavior is:
1) when a VODAR DAR is approved, downstream systems do not need to be concerned with the fact that this is VODAR.  VODAR is not a technical control, rather a policy one placed on the user, not downstream consumers of DUOS authorization grants.
2) When any question in Section 3 of the DAR application is true, VODAR approval is not possible.  We have a legacy definition of section three the AI is picking up on.  Per the ticket: **Do not allow submission of a VODAR if any Ethical (Step 3) questions are ‘Yes’**, per the ticket.

Good catch on the OpenAPI miss and the other elements for refinement.  I'll update the plan with those.  Thanks for the review!  I've added the first to points to the plan in hopes that others will benefit without having to consult the PR.

docs/plans/vodar-plan.md
```diff
@@ -0,0 +1,884 @@
...
+  for substantive reviews.
+- **Free test-coverage uplift** on legacy RADAR/matching code from the characterization-tests-first
+  mitigation.
+
+### Threats (external, negative)
+
+- **Downstream dependency for a future feature.** The 12h auto-expire (a deferred item) would require
+  Terra/TDR/Sam coordination outside these repos — timelines not controlled by this plan. Note this does
+  *not* apply to view-only itself: product has confirmed VODAR is a user-facing policy control, so no
+  downstream enforcement is required for the MVP to be correct.
```
rushtong, line 862: I don't think this is true - Consent controls what is presented to external systems. The actual threat, IMO, is a more finer grained one. TDR polls hourly. If a VO/DAR is still valid at time X, it may not be valid at time X+5 minutes. For the first 5 minutes, it will technically be approved, then not approved. For the rest of the 55 minutes in that hour, TDR will consider it approved.

rushtong (approved):
This looks very solid to me. One comment inline that shouldn't hold this up.

docs/plans/vodar-plan.md
otchet-broad: Exactly.   However, the 12 hour limit is a future requirement (per the ticket).

The polling approach employed might be better replaced with something more proactive - architecturally, like pub/sub where we inform of access and revocation.  

Currently, we actually never tell a system a user's access is revoked.  We rely on the downstream system figuring things out by noticing a user is missing from a list.  IMHO, we should be more explicit with our intent.

kevinmarete (approved):
Thanks for the updates!

PR #2985, opened 2026-07-23
[DT-3325] Bulk DAA endpoint hardening.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3325](https://broadworkbench.atlassian.net/browse/DT-3325)

### Summary

The SO "Pre-Authorize Researchers" bulk **Approve All / Remove All** actions (`/api/daa/bulk/...`)
previously applied relationships in a plain loop and returned an empty `200` — a mid-batch failure
could leave a **partially applied** batch, and the UI had no count to report. This PR makes each
bulk operation **atomic**, returns a structured result, and brings the bulk path to full validation
parity with the single-relationship flow.

> Companion UI change: `duos-ui` (`otchet-dt-3325-bulk-daa-ops`). **Deploy this service first** —
> the UI reads the new response body.

#### Changes
- **`DaaBulkRelationResult`** (new): `{ requested, applied, skipped, errors }`. `applied` counts only
  relationships that actually changed; no-ops (re-add existing / remove absent) fall under `skipped`.
- **Atomic DAO ops** (`DaaServiceDAO`): all four bulk methods run in a single transaction, so a
  mid-batch failure rolls back the whole batch. Inserts are idempotent (`ON CONFLICT DO NOTHING`);
  the accurate `applied` count comes from the audit statement's row count.
- **Validation parity** (`DaaService`/`LibraryCardService`): add paths now reject before mutating in
  every case the single-link flow would — `requireSigningOfficialInstitution` (unconditional) and
  `validateNewLibraryCardCreation` (per card-less user). Single-link `addDaaToUserLibraryCard` reuses
  the same guard.
- **Post-commit email**: the "card issued" notification is sent after commit, best-effort, so it can't
  roll back a committed batch.
- **`DaaResource`**: bulk endpoints now produce/return the result body.
----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/models/DaaBulkRelationResult.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/DaaBulkRelationResult.java
@@ -1,0 +1,75 @@
+package org.broadinstitute.consent.http.models;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Result of an atomic bulk pre-authorization operation between researchers and a DAA (the SO-driven
+ * "Approve All" / "Remove All" actions).
+ *
+ * <p>Because the underlying operation runs in a single database transaction it is all-or-nothing
+ * with respect to failure: a mid-batch error rolls the whole batch back and surfaces as an
+ * exception rather than a partial result, so {@code errors} is always empty on a returned summary.
+ * {@code applied} counts only the relationships that actually changed; requested items that were
+ * already in the desired state (re-adding an existing pre-authorization, or removing one that was
+ * never present) are no-ops counted under {@code skipped}, so {@code applied + skipped ==
+ * requested} and {@code applied} may be less than {@code requested}.
+ */
+public class DaaBulkRelationResult {
+
+  @JsonProperty private Integer requested;
+
+  @JsonProperty private Integer applied;
+
+  @JsonProperty private Integer skipped;
+
+  @JsonProperty private List<String> errors;
+
+  public DaaBulkRelationResult() {}
+
+  public DaaBulkRelationResult(
+      Integer requested, Integer applied, Integer skipped, List<String> errors) {
+    this.requested = requested;
+    this.applied = applied;
+    this.skipped = skipped;
+    this.errors = errors == null ? new ArrayList<>() : errors;
+  }
+
+  /** Convenience factory for the atomic happy path where every requested item was applied. */
+  public static DaaBulkRelationResult allApplied(int requested) {
+    return new DaaBulkRelationResult(requested, requested, 0, List.of());
+  }
+
+  public Integer getRequested() {
+    return requested;
+  }
+
+  public void setRequested(Integer requested) {
+    this.requested = requested;
+  }
+
+  public Integer getApplied() {
+    return applied;
+  }
+
+  public void setApplied(Integer applied) {
+    this.applied = applied;
+  }
+
+  public Integer getSkipped() {
+    return skipped;
+  }
+
+  public void setSkipped(Integer skipped) {
+    this.skipped = skipped;
+  }
+
+  public List<String> getErrors() {
+    return errors;
+  }
+
+  public void setErrors(List<String> errors) {
+    this.errors = errors;
+  }
+}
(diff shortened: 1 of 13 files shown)


src/main/java/org/broadinstitute/consent/http/service/dao/DaaServiceDAO.java
```diff
@@ -63,4 +68,157 @@ public Integer createDaaWithFso(Integer userId, Integer dacId, FileStorageObject
...
+            daaIds.size(),
+            lcDAO -> {
+              Integer lcId =
+                  findOrCreateLibraryCardId(lcDAO, user, signingOfficial, usersWithNewCard);
+              int applied = 0;
+              for (Integer daaId : daaIds) {
+                applied +=
+                    lcDAO.createLibraryCardDaaRelation(
+                        user.getUserId(), signingOfficial.getUserId(), lcId, daaId);
+              }
```
rushtong, line 140: Claude finding:
> Confirmed regression: DaaServiceDAO.bulkAddDaasToUser drops the per-DAA existence check the old single-item path had. A bad daaId now surfaces as a raw FK-violation → generic 500 "Database error" instead of the previous 404. The new test for this path only asserts a generic Exception.class, tacitly confirming the friendlier error is gone. (Counterpart bulkAddUsersToDaa still validates correctly.)

src/main/java/org/broadinstitute/consent/http/models/DaaBulkRelationResult.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+public class DaaBulkRelationResult {
+
+  @JsonProperty private Integer requested;
+
+  @JsonProperty private Integer applied;
+
+  @JsonProperty private Integer skipped;
+
+  @JsonProperty private List<String> errors;
```
rushtong: Confirmed Claude finding:
> Minor: DaaBulkRelationResult.errors is permanently unused/empty given the atomic design — dead API surface.

All usages of the constructor pass in an empty list.

rushtong (commented):
This looks really good, thank you. See feedback inline.

src/main/java/org/broadinstitute/consent/http/service/dao/DaaServiceDAO.java
otchet-broad: Great catch.  Adding a test for the regression and fixing.

otchet-broad: See change ~L372 of [‎src/main/java/org/broadinstitute/consent/http/service/DaaService.java‎](https://github.com/DataBiosphere/consent/pull/2985/changes/376435963299b8eb0436839f536ccf314411a9f2#diff-9ca6884b52d6c055b2f5b979f6a62a840fddde450921c232b8f5ad622abdc902) in most recent commit for how it's addressed.

fboulnois (approved):
Two minor comments:
* API doc/model mismatch: `DaaBulkRelationResult.yaml` documents an errors array field ("always empty on a returned summary"), but `DaaBulkRelationResult.java` has no errors field at all.
* Redundant existence check: `DaaService.bulkAddDaasToUser` re-validates every daaId via `findById` (a DB call each), but both its callers already went through `daaService.findDAAsInJsonArray`, which already throws 404 on any missing DAA.

kevinmarete (approved):

src/main/java/org/broadinstitute/consent/http/service/DaaService.java
```diff
@@ -366,8 +366,11 @@ public DaaBulkRelationResult bulkRemoveUsersFromDaa(
   /** Atomically pre-authorize {@code user} for every DAA in {@code daaIds}. */
   public DaaBulkRelationResult bulkAddDaasToUser(
       User user, List<Integer> daaIds, User signingOfficial) {
-    // Reject before mutating anything, matching the single-link add flow: the signing official must
-    // have an institution regardless of whether a card is created.
+    // Reject before mutating anything, matching the single-link add flow. Each referenced DAA must
+    // exist: findById throws NotFoundException (404), preserving the single-link flow's friendly
+    // error rather than letting a bad id surface as a raw FK violation / generic 500 mid-batch.
+    daaIds.stream().distinct().forEach(this::findById);
```
rushtong, line 372: In practice, we should not have large numbers of DAAs to approve per user, so this is reasonable. It would be more efficient to do a find-in-list and ensure they all exist. The `findById` approach is also more expensive in that it will go through the mappers as opposed to just querying on ids.

rushtong (approved):
Thanks for the updates, these look good.

PR #2986, opened 2026-07-23
[DT-3862] Enforce canonical primary Data Use validation

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3862

### Summary

- Add a shared classifier for primary Data Use categories.
- Enforce access-management rules during registration, admin updates, and dataset-to-study conversion.
- Reject invalid or null JSON without exposing submitted Data Use content.
- Update API documentation and add regression tests.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/mapper/DataUseParser.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/mapper/DataUseParser.java
@@ -1,5 +1,8 @@
 package org.broadinstitute.consent.http.db.mapper;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+import com.google.common.hash.Hashing;
 import com.google.gson.Gson;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -22,7 +25,10 @@ public class DataUseParser implements ConsentLogger {
           try {
             return gson.fromJson(dataUseString, DataUse.class);
           } catch (Exception _) {
-            logWarn("Unable to parse data use string");
+            String hashPrefix = Hashing.sha256().hashString(s, UTF_8).toString().substring(0, 12);
+            logWarn(
+                "Unable to parse data use string (length=%d, sha256Prefix=%s)"
+                    .formatted(s.length(), hashPrefix));
           }
           return null;
         });
--- a/src/main/java/org/broadinstitute/consent/http/db/mapper/DataUseParser.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/mapper/DataUseParser.java
@@ -21,8 +21,8 @@ public class DataUseParser implements ConsentLogger {
         s -> {
           try {
             return gson.fromJson(dataUseString, DataUse.class);
-          } catch (Exception e) {
-            logWarn(String.format("Unable to parse data use string: '%s'", dataUseString));
+          } catch (Exception _) {
+            logWarn("Unable to parse data use string");
           }
           return null;
         });
--- a/src/main/java/org/broadinstitute/consent/http/db/mapper/DataUseParser.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/mapper/DataUseParser.java
@@ -25,6 +25,7 @@ public class DataUseParser implements ConsentLogger {
           try {
             return gson.fromJson(dataUseString, DataUse.class);
           } catch (Exception _) {
+            // Correlate repeated malformed rows without exposing their raw Data Use JSON.
             String hashPrefix = Hashing.sha256().hashString(s, UTF_8).toString().substring(0, 12);
             logWarn(
                 "Unable to parse data use string (length=%d, sha256Prefix=%s)"
--- a/src/main/java/org/broadinstitute/consent/http/models/datause/DataUsePrimaryClassifier.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/datause/DataUsePrimaryClassifier.java
@@ -1,6 +1,5 @@
 package org.broadinstitute.consent.http.models.datause;
 
-import java.util.Collection;
 import java.util.EnumSet;
 import java.util.List;
 import org.broadinstitute.consent.http.models.DataUse;
@@ -11,9 +10,6 @@ public final class DataUsePrimaryClassifier {
   private DataUsePrimaryClassifier() {}
 
   public static DataUsePrimaryClassification classify(DataUse dataUse) {
-    if (dataUse == null) {
-      return new DataUsePrimaryClassification(List.of());
-    }
     return classify(
         dataUse.getGeneralUse(),
         dataUse.getHmbResearch(),
@@ -26,7 +22,7 @@ public final class DataUsePrimaryClassifier {
       Boolean generalUse,
       Boolean hmbResearch,
       Boolean populationOriginsAncestry,
-      Collection<?> diseaseRestrictions,
+      List<String> diseaseRestrictions,
       String other) {
     EnumSet<DataUsePrimaryCategory> categories = EnumSet.noneOf(DataUsePrimaryCategory.class);
     if (Boolean.TRUE.equals(generalUse)) {
--- a/src/main/java/org/broadinstitute/consent/http/models/datause/DataUsePrimaryClassifier.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/datause/DataUsePrimaryClassifier.java
@@ -1,0 +1,49 @@
+package org.broadinstitute.consent.http.models.datause;
+
+import java.util.Collection;
+import java.util.EnumSet;
+import java.util.List;
+import org.broadinstitute.consent.http.models.DataUse;
+
+/** Classifies primary dataset Data Use fields without applying access-management policy. */
+public final class DataUsePrimaryClassifier {
+
+  private DataUsePrimaryClassifier() {}
+
+  public static DataUsePrimaryClassification classify(DataUse dataUse) {
+    if (dataUse == null) {
+      return new DataUsePrimaryClassification(List.of());
+    }
+    return classify(
+        dataUse.getGeneralUse(),
+        dataUse.getHmbResearch(),
+        dataUse.getPopulationOriginsAncestry(),
+        dataUse.getDiseaseRestrictions(),
+        dataUse.getOther());
+  }
+
+  public static DataUsePrimaryClassification classify(
+      Boolean generalUse,
+      Boolean hmbResearch,
+      Boolean populationOriginsAncestry,
+      Collection<?> diseaseRestrictions,
+      String other) {
+    EnumSet<DataUsePrimaryCategory> categories = EnumSet.noneOf(DataUsePrimaryCategory.class);
+    if (Boolean.TRUE.equals(generalUse)) {
+      categories.add(DataUsePrimaryCategory.GRU);
+    }
+    if (Boolean.TRUE.equals(hmbResearch)) {
+      categories.add(DataUsePrimaryCategory.HMB);
+    }
+    if (Boolean.TRUE.equals(populationOriginsAncestry)) {
+      categories.add(DataUsePrimaryCategory.POA);
+    }
+    if (diseaseRestrictions != null && !diseaseRestrictions.isEmpty()) {
+      categories.add(DataUsePrimaryCategory.DS);
+    }
+    if (other != null && !other.isBlank()) {
+      categories.add(DataUsePrimaryCategory.OTHER);
+    }
+    return new DataUsePrimaryClassification(List.copyOf(categories));
+  }
+}
--- a/src/main/java/org/broadinstitute/consent/http/models/datause/DataUsePrimaryValidator.java
+++ b/src/main/java/org/broadinstitute/consent/http/models/datause/DataUsePrimaryValidator.java
@@ -1,0 +1,35 @@
+package org.broadinstitute.consent.http.models.datause;
+
+import jakarta.ws.rs.BadRequestException;
+import org.broadinstitute.consent.http.models.DataUse;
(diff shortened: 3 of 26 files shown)


src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -551,6 +564,36 @@ public Study convertDatasetToStudy(User user, Dataset dataset, StudyConversion s
     return studyDAO.findStudyById(studyId);
   }
 
+  private AccessManagement findAccessManagement(Dataset dataset) {
```
otchet-broad: From Claude:

**This is the fifth copy of access-management extraction, and the only one that knows about the legacy prefix.**

| Location | Lookup |
|---|---|
| `ConsentGroupFromDataset.java:38` | `schemaProperty` == `accessManagement` |
| `ElasticSearchService.java:486` | `schemaProperty` == `"accessManagement"` |
| `Dataset.java:291-303` | `propertyName` == `"Access Management"` |
| `StudyDAO.java:183` | SQL `schema_property = 'accessManagement'` |
| **new** here | `accessManagement` **or** `consentGroup.accessManagement` |

For a legacy dataset carrying only `consentGroup.accessManagement=open`, this validator treats it as open while ES indexing and `ConsentGroupFromDataset` still see *no* access management — two subsystems disagreeing about the same dataset.

Please hoist to one shared accessor (`Dataset.getAccessManagement()` reads naturally; `Dataset` already imports `AccessManagement`) and switch the other readers to it. If that's too wide for this PR, the minimum is a `TODO` with the follow-up ticket plus a comment on `LEGACY_ACCESS_MANAGEMENT` (line 62) explaining where the legacy form originates.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -551,6 +564,36 @@ public Study convertDatasetToStudy(User user, Dataset dataset, StudyConversion s
...
+        .filter(Objects::nonNull)
+        .map(Object::toString)
+        .map(String::trim)
+        .map(value -> value.toLowerCase(Locale.ROOT))
+        .map(
+            value -> {
+              try {
+                return AccessManagement.fromValue(value);
+              } catch (IllegalArgumentException _) {
+                return null;
```
otchet-broad: Claude:

Unparseable access-management values are silently swallowed

A typo'd or newly-added property value silently flips a dataset from "open" to "treated as controlled" with no signal anywhere. DatasetService implements ConsentLogger — add a logWarn with the dataset id and the offending value so this is diagnosable.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -551,6 +564,36 @@ public Study convertDatasetToStudy(User user, Dataset dataset, StudyConversion s
...
+  }
+
+  private Optional<AccessManagement> findAccessManagement(
+      Set<DatasetProperty> properties, String schemaProperty) {
+    return properties.stream()
+        .filter(property -> schemaProperty.equalsIgnoreCase(property.getSchemaProperty()))
+        .map(DatasetProperty::getPropertyValue)
+        .filter(Objects::nonNull)
+        .map(Object::toString)
+        .map(String::trim)
```
otchet-broad: Claude:

Test gap: the normalization and failure branches of the property parse are untested

testUpdateDatasetDataUseAcceptsOpenWithNoPrimary parameterizes the schema property casing nicely, but nothing exercises the trim()/toLowerCase(Locale.ROOT) on the value or the IllegalArgumentException path. Please add cases for "OPEN", " open ", and an unrecognized value like "unknown" (asserting it falls back to the controlled rule).

src/main/java/org/broadinstitute/consent/http/models/datause/DataUsePrimaryClassifier.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.List;
+import org.broadinstitute.consent.http.models.DataUse;
+
+/** Classifies primary dataset Data Use fields without applying access-management policy. */
+public final class DataUsePrimaryClassifier {
+
+  private DataUsePrimaryClassifier() {}
+
+  public static DataUsePrimaryClassification classify(DataUse dataUse) {
+    if (dataUse == null) {
```
otchet-broad: Claude:

 Two entry points disagree about what a null Data Use means

src/main/java/org/broadinstitute/consent/http/models/datause/DataUsePrimaryClassifier.java:14-16 returns NONE for null (which would be valid for an open dataset), while DataUsePrimaryValidator.java:19-21 throws "Data Use is required" for null regardless of access management. Pick one contract: if null is never acceptable, the classifier's null branch is dead defensive code worth dropping; if it is acceptable for open datasets, the validator shouldn't short-circuit.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
otchet-broad: Claude: 

Two overloads named findAccessManagement with different meanings (resolve-for-dataset vs. parse-one-property). Rename the private one to parseAccessManagementProperty.

otchet-broad (commented):
Strong base.  A few things Claude helped me identify in addition to the items tagged on specific lines:

- DataUsePrimaryCategory.java:4 — enum declaration order is the canonical ordering asserted by DataUsePrimaryClassifierTest:74-79 (via EnumSet iteration), but nothing says so. One comment prevents someone reordering it and breaking a test for non-obvious reasons.
- DataUsePrimaryClassification.java:8-10 — the public constructor accepts duplicates (List.of(GRU, GRU) → MULTIPLE). Normalizing through an EnumSet in the compact constructor would make the record's invariant match what classify actually produces.
- DataUsePrimaryClassifier.java:29 — Collection<?> diseaseRestrictions is looser than needed; List<String> matches both call sites.
- DataUsePrimaryClassification.categories() is currently read only by tests — is a production consumer planned (ES facets, API response)? If not, Shape alone carries all the information the validator uses, and NONE/SINGLE/MULTIPLE could collapse to a size check.
- DataUsePrimaryClassifierTest also covers DataUsePrimaryValidator (the accessManagementValidationCases block). Worth splitting into DataUsePrimaryValidatorTest so failures point at the right class.
- DataUseParser.java:24-26 — dropping the payload from the log is right, but it removes the only breadcrumb for diagnosing bad rows. Consider logging the string's length plus a short hash prefix, which keeps correlation without leaking content.
- DatasetService.java:484 — validating only when studyConversion.getDataUse() != null means a conversion can leave a pre-existing invalid Data Use in place. I read that as intentional patch semantics; confirming.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
kevinmarete: Good catch. I moved access-management extraction into `Dataset.getAccessManagement()`. It handles canonical and legacy schema properties case-insensitively, with deterministic canonical precedence. `DatasetService`, Elasticsearch indexing, consent-group mapping, and dataset search now use this accessor. The `StudyDAO` aggregation query mirrors the same precedence. Regression tests cover the legacy form and conflicting values.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
kevinmarete: Addressed. Unrecognized access-management values now produce a warning containing the dataset ID and offending value. They still resolve to missing access management and therefore follow the existing controlled-dataset validation rule.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
kevinmarete: Added coverage for `"OPEN"`, `" open "`, and `"unknown"`. The first two resolve to `OPEN`; the unknown value exercises the failure branch, logs a warning, and follows the controlled-dataset rule.

src/main/java/org/broadinstitute/consent/http/models/datause/DataUsePrimaryClassifier.java
kevinmarete: Agreed. I chose the contract that null Data Use is never valid. The classifier’s null-to-`NONE` branch was removed, while the validator retains the explicit `"Data Use is required"` error. This behavior now has dedicated validator coverage.

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
kevinmarete: Addressed as part of the shared-accessor refactor. The overloaded methods were removed from `DatasetService`; property parsing now lives in `Dataset.parseAccessManagementProperty()` behind `Dataset.getAccessManagement()`.

kevinmarete:
> Strong base. A few things Claude helped me identify in addition to the items tagged on specific lines:
> 
> * DataUsePrimaryCategory.java:4 — enum declaration order is the canonical ordering asserted by DataUsePrimaryClassifierTest:74-79 (via EnumSet iteration), but nothing says so. One comment prevents someone reordering it and breaking a test for non-obvious reasons.
> * DataUsePrimaryClassification.java:8-10 — the public constructor accepts duplicates (List.of(GRU, GRU) → MULTIPLE). Normalizing through an EnumSet in the compact constructor would make the record's invariant match what classify actually produces.
> * DataUsePrimaryClassifier.java:29 — Collection<?> diseaseRestrictions is looser than needed; List matches both call sites.
> * DataUsePrimaryClassification.categories() is currently read only by tests — is a production consumer planned (ES facets, API response)? If not, Shape alone carries all the information the validator uses, and NONE/SINGLE/MULTIPLE could collapse to a size check.
> * DataUsePrimaryClassifierTest also covers DataUsePrimaryValidator (the accessManagementValidationCases block). Worth splitting into DataUsePrimaryValidatorTest so failures point at the right class.
> * DataUseParser.java:24-26 — dropping the payload from the log is right, but it removes the only breadcrumb for diagnosing bad rows. Consider logging the string's length plus a short hash prefix, which keeps correlation without leaking content.
> * DatasetService.java:484 — validating only when studyConversion.getDataUse() != null means a conversion can leave a pre-existing invalid Data Use in place. I read that as intentional patch semantics; confirming.

Thanks for the thorough review. I addressed the summary items as follows:
 - Documented that enum declaration order defines canonical classification and audit-report ordering.
 - Normalized `DataUsePrimaryClassification` categories through an `EnumSet`, removing duplicates and preserving canonical ordering.
 - Narrowed `diseaseRestrictions` from `Collection<?>` to `List<String>`.
 - Retained `categories()` intentionally because the ticket requires reporting individual `SINGLE(category)` and sorted `MULTIPLE` classifications, in addition to shape validation.
 - Split validator coverage into `DataUsePrimaryValidatorTest`.
 - Added invalid Data Use diagnostics using payload length and a short SHA-256 prefix without logging raw JSON.
 - Confirmed that study conversion has patch semantics: an omitted Data Use preserves the persisted value. This is now documented in the service.

rushtong (approved):
This looks pretty comprehensive, thank you 👍🏽

src/main/java/org/broadinstitute/consent/http/db/mapper/DataUseParser.java
```diff
@@ -21,8 +24,11 @@ public DataUse parseDataUse(String dataUseString) {
         s -> {
           try {
             return gson.fromJson(dataUseString, DataUse.class);
-          } catch (Exception e) {
-            logWarn(String.format("Unable to parse data use string: '%s'", dataUseString));
+          } catch (Exception _) {
+            String hashPrefix = Hashing.sha256().hashString(s, UTF_8).toString().substring(0, 12);
```
fboulnois, line 29: Why do we hash here?

src/main/java/org/broadinstitute/consent/http/models/datause/DataUsePrimaryValidator.java
```diff
@@ -0,0 +1,35 @@
...
+import jakarta.ws.rs.BadRequestException;
+import org.broadinstitute.consent.http.models.DataUse;
+import org.broadinstitute.consent.http.models.dataset_registration_v1.ConsentGroup.AccessManagement;
+import org.broadinstitute.consent.http.models.datause.DataUsePrimaryClassification.Shape;
+
+/** Applies dataset access-management policy to canonical primary Data Use classifications. */
+public final class DataUsePrimaryValidator {
+
+  public static final String VALIDATION_MESSAGE =
+      "Dataset must have exactly one primary data use (open access, or one of:"
```
fboulnois: This message is confusing to me, since `open access` is not a primary data use (it is a category of access management).

fboulnois (approved):
A few questions:

otchet-broad (approved):
Nice implementation improvements.

Curious to see if I can guess at your answer to @fboulnois' question.

src/main/java/org/broadinstitute/consent/http/db/mapper/DataUseParser.java
kevinmarete: The short hash prefix lets us correlate repeated malformed persisted values without logging raw Data Use JSON, which may contain sensitive free text. It is diagnostic only and is not used for matching or validation. I added a comment explaining this intent.

src/main/java/org/broadinstitute/consent/http/models/datause/DataUsePrimaryValidator.java
kevinmarete: Agreed—`open access` is access management, not a primary Data Use. I rewrote the message to distinguish them: open-access datasets require no primary Data Use, while controlled, external, and missing-access-management datasets require exactly one. I also updated the related API documentation.

PR #2993, opened 2026-07-30
[DT-2408] Add ECM health check

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-2408

### Summary

- Add ECM to Consent’s dependent service health checks.
- Mark Consent as degraded when ECM is unhealthy.
- Include ECM details in the /status response.
- Update API documentation and tests.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/configurations/ServicesConfiguration.java
+++ b/src/main/java/org/broadinstitute/consent/http/configurations/ServicesConfiguration.java
@@ -19,6 +19,9 @@ public class ServicesConfiguration {
   public static final String REJECT_TOS_PATH = "api/termsOfService/v1/user/self/reject";
   public static final String SAM_V1_USER_EMAIL = "api/users/v1";
   public static final String ECM_RAS_PROVIDER = "/api/oauth/v1/ras";
+  // ECM's detailed status endpoint is auth-gated at the environment proxy. This legacy endpoint
+  // remains public specifically for unauthenticated service health checks.
+  public static final String ECM_STATUS_PATH = "/status";
   // nosemgrep
   public static final String BROAD_ZENDESK_URL = "https://broadinstitute.zendesk.com";
 
@@ -78,6 +81,15 @@ public class ServicesConfiguration {
 
   public String getEcmRasProviderUrl() {
     return getEcmUrl() + ECM_RAS_PROVIDER;
+  }
+
+  public String getEcmStatusUrl() {
+    String baseUrl = getEcmUrl();
+    int end = baseUrl.length();
+    while (end > 0 && baseUrl.charAt(end - 1) == '/') {
+      end--;
+    }
+    return baseUrl.substring(0, end) + ECM_STATUS_PATH;
   }
 
   public String getV1ResourceTypesUrl() {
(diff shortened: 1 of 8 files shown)


src/main/java/org/broadinstitute/consent/http/configurations/ServicesConfiguration.java
```diff
@@ -80,6 +83,15 @@ public String getEcmRasProviderUrl() {
     return getEcmUrl() + ECM_RAS_PROVIDER;
   }
 
+  public String getEcmStatusUrl() {
+    String baseUrl = getEcmUrl();
+    int end = baseUrl.length();
+    while (end > 0 && baseUrl.charAt(end - 1) == '/') {
+      end--;
+    }
```
rushtong, line 91: The need for this is an indication of tech debt - we should have standard base urls but it looks like [we do not.](https://github.com/broadinstitute/terra-helmfile/blob/master/values/app/consent/live/prod.yaml#L59). We should file a ticket to align them all or build a utility that treats them all the same before constructing actual paths.

src/main/resources/assets/api-docs.yaml
```diff
@@ -1152,9 +1152,9 @@ paths:
       description: |
         A detailed description of the various subsystem statuses that Consent relies upon.
         Current systems include:
+          * ECM
           * Elastic Search
           * GCS
-          * Ontology
```
rushtong, line 1157: Nice catch

src/test/java/org/broadinstitute/consent/http/configurations/ServicesConfigurationTest.java
```diff
@@ -0,0 +1,22 @@
...
+import org.junit.jupiter.params.provider.CsvSource;
+
+class ServicesConfigurationTest {
+
+  @ParameterizedTest
+  @CsvSource({
+    "https://externalcreds.example.org, https://externalcreds.example.org/status",
+    "https://externalcreds.example.org/, https://externalcreds.example.org/status",
+    "https://externalcreds.example.org///, https://externalcreds.example.org/status"
+  })
```
rushtong, line 15: This would be great as a more general test for all base url tests.

rushtong (approved):
This looks good 👍🏽 Optionally, it could resolve some tech debt by standardizing the same url treatment across all base url fields.

src/main/java/org/broadinstitute/consent/http/configurations/ServicesConfiguration.java
kevinmarete: Agreed. I kept this PR scoped to the ECM health check and created [DT-3882](https://broadworkbench.atlassian.net/browse/DT-3882) to standardize base URL/path construction and add generalized coverage for all service URLs.

src/test/java/org/broadinstitute/consent/http/configurations/ServicesConfigurationTest.java
kevinmarete: Will be addressed in https://broadworkbench.atlassian.net/browse/DT-3882

otchet-broad (approved):
Agree with @rushtong 's thoughts about the URL processing being indicative of tech debt.  Seems like a ticket to resolve that would be appropriate.

PR #2994, opened 2026-07-30
[DT-3881] Ensure EntityIDs in emails are durable IDs.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3881](https://broadworkbench.atlassian.net/browse/DT-3881)

### Summary

`EmailService.sendMessage` hands the message to SendGrid **before** inserting the
`email_entity` row (`EmailService.java:93-108`). `email_entity.entity_reference_id` is
`varchar(255)`, and an over-long value doesn't truncate — it fails the insert. So any
`MailMessage` whose `getEntityReferenceId()` can exceed 255 characters sends the email and
then loses the record that it was ever sent.

Two implementations were keyed on submitter-supplied free text:

- `NewStudyRegistrationConfirmationMessage` keyed on `studyName`, sourced from
  `study.name`, which is `TEXT` — genuinely unbounded.
- `DatasetDeniedMessage` keyed on the value passed as `datasetName`. `DatasetService` was
  actually passing `dataset.getDatasetIdentifier()` into that parameter, so the name and
  identifier were conflated: the email body rendered the identifier where it meant to
  render the dataset name.

##### Study registration confirmation now keyed on the study UUID

The study UUID was previously generated inside `DatasetServiceDAO.executeInsertStudy`, so
the caller never saw it and had nothing stable to key the email on. The UUID is now
generated by the caller and threaded through:

- `DatasetServiceDAO.StudyInsert` gains a `uuid` component; `executeInsertStudy` persists
  the supplied UUID rather than generating its own.
- `DatasetRegistrationService.createStudyInsert` generates the UUID; the created study's
  UUID is passed to `sendSubmissionConfirmationEmail` /
  `sendStudySubmissionConfirmation` and on into the message.
- `NewStudyRegistrationConfirmationMessage.getEntityReferenceId()` returns
  `Objects.toString(studyUuid, null)` — the UUID when present, `null` otherwise (the
  column is nullable). The study name is still a model field, so the email body is
  unchanged.

##### Dataset denied email: name and identifier separated

`DatasetDeniedMessage` now takes both `datasetIdentifier` and `datasetName`.
`getEntityReferenceId()` returns the identifier (bounded, `DUOS-nnnnnn`), while
`createModel()` continues to name the dataset for the reader. `DatasetService` passes both
values instead of passing the identifier as the name — fixing the body text as well as the
reference id.

#### Tests

- **`EntityReferenceIdLengthTest`** (new) — one parameterized case per concrete
  `MailMessage` implementation, each constructed with the longest value its reference id
  source column can hold (255 chars for bounded sources, 1000 for the unbounded `TEXT`
  case), asserting the resulting reference id fits in the column. This catches messages
  that decorate or concatenate sources even when each source fits on its own. Two guards
  keep it honest:
  - a negative case builds an anonymous `MailMessage` keyed on 1000 chars and asserts the
    assertion actually fails, so the suite can't go vacuous;
  - `testAllMailMessageImplementationsAreCovered` discovers implementations via Guava
    `ClassPath` and fails if a new one is added without a case, forcing an explicit
    decision about what it keys on.
- **`MailMessageDAOTest`** — asserts a 255-char reference id round-trips intact, and that
  256 chars throws `UnableToExecuteStatementException` rather than truncating. The second
  test also reads `information_schema` to pin the column width that
  `EntityReferenceIdLengthTest` hard-codes, so a migration that changes the width breaks
  here rather than silently invalidating the length suite.
- **`DatasetServiceDAOTest`** — asserts the persisted study carries the UUID supplied on
  the insert (loaded via `studyDAO.findStudyById`, since a dataset's study projection
  doesn't select the UUID).
- **`DatasetRegistrationServiceTest`** — captures both the `StudyInsert` and the sent
  message to assert the confirmation email references the study by the same UUID used to
  create it; `sendSubmissionConfirmationEmail` test asserts the passed UUID reaches the
  reference id.
- **`DatasetServiceTest`** — captures the `DatasetDeniedMessage` to assert the model names
  the dataset while the reference id is the identifier.
- **`DatasetDeniedMessageTest`** / **`NewStudyRegistrationConfirmationMessageTest`** —
  cover the name/identifier split, the identifier's absence from the rendered body, and
  the null-UUID reference id.

#### Notes for reviewers

- No schema migration. `email_entity.entity_reference_id` stays `varchar(255)`;
  `study.uuid` already existed and was already populated per study.
- Existing `email_entity` rows keyed on study or dataset names are not backfilled — the
  column is a loose reference, not a foreign key.
- Nothing reads `entity_reference_id` for study registration confirmations today, so the
  switch from name to UUID doesn't break a lookup path.

----

- Verify all tests go green


rushtong (approved):
Nice fix, thank you 👍🏽

kevinmarete (approved):

PR #3002, opened 2026-08-05
[DT-3865] Migrate match identity to dataset IDs

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3865

### Summary

- Add `dataset_id` as the stable database identity for match persistence and relationships.
- Backfill existing matches from their legacy DUOS consent identifiers without truncating aliases longer than six digits.
- Retain and dual-write the legacy `consent` value during this compatibility release.
- Add compatibility triggers so old and new application versions can safely overlap during deployment.
- Replace `MAX(alias) + 1` allocation with a sequence-backed trigger that also supports legacy explicit alias writes.
- Cascade dataset deletion to associated matches and match rationales.
- Fail immediately when attempting to insert a match for a nonexistent dataset.
- Preserve canonical `DUOS-######` identifiers in API-facing behavior.
- Add migration, rollback, concurrency, cascade, compatibility, and DAO regression tests.

### Rollout notes

- Liquibase applies the migration automatically during application startup.
- Run the documented preflight queries before deployment.
- The migration stops if existing match identifiers cannot be resolved or would violate the new constraints.
- This is an expand/migrate compatibility release, so old and new application versions can overlap during rollout.
- No manual data migration is required when the preflight checks pass.
- A follow-up contract release is required after every environment is running the compatible application version. That release should remove the legacy `consent` column and constraint, remove the compatibility triggers, and complete the alias-allocation transition.

----

- Verify all tests go green

### DIFF
--- a/docs/plans/dataset-identity-migration-plan.md
+++ b/docs/plans/dataset-identity-migration-plan.md
@@ -24,7 +24,9 @@ HAVING COUNT(*) > 1;
 SELECT m.match_id, m.purpose, m.consent
 FROM match_entity m
 LEFT JOIN dataset d
-  ON m.consent = 'DUOS-' || LPAD(d.alias::text, 6, '0')
+  ON m.consent = 'DUOS-'
+    || REPEAT('0', GREATEST(0, 6 - LENGTH(d.alias::text)))
+    || d.alias::text
 WHERE d.dataset_id IS NULL;
 
 SELECT COUNT(*) AS pre_migration_total,
@@ -32,7 +34,9 @@ SELECT COUNT(*) AS pre_migration_total,
        COUNT(*) - COUNT(d.dataset_id) AS unresolved
 FROM match_entity m
 LEFT JOIN dataset d
-  ON m.consent = 'DUOS-' || LPAD(d.alias::text, 6, '0');
+  ON m.consent = 'DUOS-'
+    || REPEAT('0', GREATEST(0, 6 - LENGTH(d.alias::text)))
+    || d.alias::text;
 ```
 
 Any unresolved match rows must be explicitly corrected or isolated before the final constraint
@@ -41,33 +45,50 @@ dataset aliases or match identifiers, so it cannot resolve an unmatched alias.
 
 ## Rollout
 
-The bounded compatibility phase runs during application startup, before the new application code
-serves traffic.
+This migration uses an expand/migrate/contract rollout so old and new instances can safely overlap.
 
-Deployment note: avoid mixed-version dataset writes during the alias sequence cutover (older
-`COALESCE(MAX(alias), 0) + 1` inserts won’t advance `dataset_alias_seq` and can cause alias collisions).
+### Compatibility release (this change)
+
 1. Verify aliases are non-null, non-negative, and unique.
-2. Create `dataset_alias_seq` above the audited maximum, make it the alias default, and add database
-   uniqueness and non-null constraints.
-3. Add nullable `match_entity.dataset_id` with its foreign key and index.
-4. Backfill only exact canonical alias matches.
-5. Halt if any match is unresolved or if `(purpose, dataset_id)` would be duplicated.
-6. Make `dataset_id` non-null, replace the old uniqueness constraint, and remove `consent`.
+2. Lock dataset writes while initializing `dataset_alias_seq` above the audited maximum.
+3. Install an alias-allocation trigger that overrides both legacy `MAX(alias) + 1` values and
+   omitted aliases with the next sequence value. Add database uniqueness and non-null constraints.
+4. Add nullable `match_entity.dataset_id` with a cascading foreign key and index. Match rationales
+   also cascade when their owning match is removed.
+5. Backfill only exact canonical alias matches.
+6. Install a compatibility trigger that populates `dataset_id` for legacy match writes that provide
+   only `consent`.
+7. Halt if any match is unresolved or if `(purpose, dataset_id)` would be duplicated, then enforce
+   non-null and `(purpose, dataset_id)` uniqueness.
+8. Keep the legacy `consent` column and `(purpose, consent)` constraint for old instances. New match
+   writes populate both identities during this bounded phase.
 
-The DAO change is deployed with the migration. Match writes use `dataset_id`; match reads join to
-`dataset` only to derive the existing public `consent` value.
+Dataset deletion cascades to its matches and their rationales, preserving the supported deletion
+workflow without leaving records that refer to a deleted dataset.
+
+### Contract release (required follow-up)
+
+After every environment is running the compatibility release and reconciliation still reports zero
+unresolved or conflicting rows:
+
+1. Remove the match compatibility trigger.
+2. Remove the legacy `(purpose, consent)` constraint and `match_entity.consent` column.
+3. Replace the temporary alias-allocation trigger with `dataset_alias_seq` as the column default.
+
+The contract changes must be delivered in a later release; they must not run while an older
+application instance can still serve traffic.
 
 ## Alias Lookup Inventory
 
 - `DatasetService` identifier lookup and `TDRService` identifier lookup remain public API/integration
   boundaries and may translate a `DUOS-######` identifier to an alias.
 - Dataset and DAC response projections may format aliases for user-facing output.
-- Match persistence and election joins use only `dataset_id` after this migration.
+- Match persistence and election joins use `dataset_id`; the compatibility trigger translates only
+  writes from older application instances.
 
 ## Rollback
 
-Liquibase rollback recreates and backfills `match_entity.consent` from the referenced dataset,
-restores `(purpose, consent)` uniqueness, removes `dataset_id`, removes alias constraints/default,
-and drops the alias sequence. Because aliases remain immutable public identifiers throughout the
-rollout, the reconstructed legacy values are compatible with the previous application version.
-
+Liquibase rollback removes the compatibility triggers and new constraints, restores the original
+match-rationale foreign key behavior, removes `dataset_id`, removes the alias constraints, and drops
+the alias sequence. The legacy `consent` column and its uniqueness constraint remain present
+throughout this release, so rollback does not need to reconstruct public identifiers.
--- a/docs/plans/dataset-identity-migration-plan.md
+++ b/docs/plans/dataset-identity-migration-plan.md
@@ -1,0 +1,71 @@
+# Dataset Identity Migration
+
+## Goal
+
+Use `dataset.dataset_id` for match persistence and relational joins while keeping the canonical
+`DUOS-######` alias as the public dataset identifier. Allocate new aliases from a PostgreSQL
+sequence and enforce uniqueness in the database.
+
+## Preflight Audit
+
+Before production rollout, run the following read-only checks. The Liquibase migration repeats the
+blocking checks and halts rather than guessing when the data is ambiguous.
+
+```sql
+SELECT dataset_id, alias
+FROM dataset
+WHERE alias IS NULL OR alias < 0;
+
+SELECT alias, COUNT(*)
+FROM dataset
+GROUP BY alias
+HAVING COUNT(*) > 1;
+
+SELECT m.match_id, m.purpose, m.consent
+FROM match_entity m
+LEFT JOIN dataset d
+  ON m.consent = 'DUOS-' || LPAD(d.alias::text, 6, '0')
+WHERE d.dataset_id IS NULL;
+
+SELECT COUNT(*) AS pre_migration_total,
+       COUNT(d.dataset_id) AS exact_matches,
+       COUNT(*) - COUNT(d.dataset_id) AS unresolved
+FROM match_entity m
+LEFT JOIN dataset d
+  ON m.consent = 'DUOS-' || LPAD(d.alias::text, 6, '0');
+```
+
+Any unresolved match rows must be explicitly corrected or isolated before the final constraint
+change. The Data Use classification sample associated with the preceding ticket does not contain
+dataset aliases or match identifiers, so it cannot resolve an unmatched alias.
+
+## Rollout
+
+The bounded compatibility phase runs during application startup, before the new application code
+serves traffic:
+
+1. Verify aliases are non-null, non-negative, and unique.
+2. Create `dataset_alias_seq` above the audited maximum, make it the alias default, and add database
+   uniqueness and non-null constraints.
+3. Add nullable `match_entity.dataset_id` with its foreign key and index.
+4. Backfill only exact canonical alias matches.
+5. Halt if any match is unresolved or if `(purpose, dataset_id)` would be duplicated.
+6. Make `dataset_id` non-null, replace the old uniqueness constraint, and remove `consent`.
+
+The DAO change is deployed with the migration. Match writes use `dataset_id`; match reads join to
+`dataset` only to derive the existing public `consent` value.
+
+## Alias Lookup Inventory
+
+- `DatasetService` identifier lookup and `TDRService` identifier lookup remain public API/integration
+  boundaries and may translate a `DUOS-######` identifier to an alias.
+- Dataset and DAC response projections may format aliases for user-facing output.
+- Match persistence and election joins use only `dataset_id` after this migration.
+
+## Rollback
+
+Liquibase rollback recreates and backfills `match_entity.consent` from the referenced dataset,
+restores `(purpose, consent)` uniqueness, removes `dataset_id`, removes alias constraints/default,
+and drops the alias sequence. Because aliases remain immutable public identifiers throughout the
+rollout, the reconstructed legacy values are compatible with the previous application version.
+
--- a/docs/plans/dataset-identity-migration-plan.md
+++ b/docs/plans/dataset-identity-migration-plan.md
@@ -42,8 +42,10 @@ dataset aliases or match identifiers, so it cannot resolve an unmatched alias.
 ## Rollout
 
 The bounded compatibility phase runs during application startup, before the new application code
-serves traffic:
+serves traffic.
 
+Deployment note: avoid mixed-version dataset writes during the alias sequence cutover (older
+`COALESCE(MAX(alias), 0) + 1` inserts won’t advance `dataset_alias_seq` and can cause alias collisions).
 1. Verify aliases are non-null, non-negative, and unique.
 2. Create `dataset_alias_seq` above the audited maximum, make it the alias default, and add database
    uniqueness and non-null constraints.
--- a/src/main/java/org/broadinstitute/consent/http/db/MatchDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/MatchDAO.java
@@ -74,11 +74,13 @@ public interface MatchDAO extends Transactional<MatchDAO> {
         INSERT INTO match_entity
           (dataset_id, consent, purpose, match_entity, failed, create_date, algorithm_version,
            abstain)
-        SELECT d.dataset_id,
-               'DUOS-' || REPEAT('0', GREATEST(0, 6 - LENGTH(d.alias::text))) || d.alias::text,
-               :purposeId, :match, :failed, :createDate, :algorithmVersion, :abstain
-        FROM dataset d
-        WHERE d.dataset_id = :datasetId
+        VALUES (
+          :datasetId,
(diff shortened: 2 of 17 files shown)


src/main/resources/changesets/changelog-consent-2026-08-05-dataset-identity.xml
```diff
@@ -0,0 +1,117 @@
...
+        SELECT COUNT(*)
+        FROM (SELECT alias FROM dataset GROUP BY alias HAVING COUNT(*) &gt; 1) duplicate_aliases
+      </sqlCheck>
+    </preConditions>
+
+    <createSequence sequenceName="dataset_alias_seq"/>
+    <sql>
+      SELECT setval(
+        'dataset_alias_seq',
+        COALESCE((SELECT MAX(alias) FROM dataset), 0) + 1,
```
fboulnois, line 22: Isn't it possible that at the moment that the SQL query runs, another dataset is added to the database, causing issues? If this happen, how can we roll back safely, since the migration itself will change the underlying structure of the data?

src/main/resources/changesets/changelog-consent-2026-08-05-dataset-identity.xml
```diff
@@ -0,0 +1,117 @@
...
+    <rollback>
+      <dropIndex tableName="match_entity" indexName="idx_match_entity_dataset_id"/>
+      <dropForeignKeyConstraint baseTableName="match_entity"
+        constraintName="fk_match_entity_dataset_id"/>
+      <dropColumn tableName="match_entity" columnName="dataset_id"/>
+    </rollback>
+  </changeSet>
+
+  <changeSet id="changelog-consent-2026-08-05-backfill-match-dataset-id" author="kmarete">
+    <!-- Only canonical, exact, and unambiguous public identifiers are eligible for backfill. -->
```
fboulnois, line 86: I want to double check my understanding here. We should never overwrite existing data because scientific papers may depend on the canonical `DUOS-` identifiers. Is it the case that this approach may overwrite those identifiers to something different?

fboulnois (commented):
I understand the goal, but can you explain the issue with the current approach?

This change seems particularly risky if we have any issues because I can imagine a situation where we have partially overwritten previous data and cannot safely migrate forward or roll back.

kevinmarete:
This ticket came from [DT-3860](https://broadworkbench.atlassian.net/browse/DT-3860)’s matching-consistency work. The current implementation stores the formatted public `DUOS-######` identifier in `match_entity` and uses it for internal relationships. Dataset aliases are also allocated with `MAX(alias) + 1`, which can produce duplicates under concurrent creation.

DT-3865 separates those responsibilities: `dataset_id` becomes the internal relational identity, while the existing alias remains the permanent public identifier. The migration does not modify existing aliases.

The migration is staged: it adds a nullable foreign key, backfills only exact canonical matches, and halts before removing the legacy column if any rows are unresolved or duplicated. Until the final changeset succeeds, the original `consent` data remains present. If the migration completes, rollback reconstructs `consent` from the unchanged dataset alias. This rollback and reapplication path is covered by an integration test.

[DT-3860]: https://broadworkbench.atlassian.net/browse/DT-3860?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

src/main/resources/changesets/changelog-consent-2026-08-05-dataset-identity.xml
kevinmarete: Yes, an old application instance creating a dataset during the sequence cutover could conflict with sequence initialization. A database lock during initialization would only protect the migration itself; it would not prevent an old instance from resuming afterward and continuing to use `MAX(alias) + 1`.

The production requirement is therefore a full cutover: stop old-version dataset writes, or temporarily pause dataset creation, before starting the migrated version. This is documented in the rollout plan and PR deployment notes. Once all instances use the new code, alias allocation is sequence-backed and uniqueness is enforced by the database.

If our deployment process cannot guarantee that cutover, we should not deploy this PR as-is and would need a separate mixed-version compatibility mechanism.

src/main/resources/changesets/changelog-consent-2026-08-05-dataset-identity.xml
kevinmarete: No existing canonical identifier is overwritten. This statement only writes the new `match_entity.dataset_id` column by matching the existing `match_entity.consent` value to the existing dataset alias.

It does not update `dataset.alias` or the legacy `consent` value. The sequence is initialized above the current maximum and applies only to future datasets. API responses continue deriving the same `DUOS-######` value from the unchanged alias.

During rollback, the unchanged alias is used to reconstruct the original legacy identifier.

otchet-broad:
More feedback from codex that builds on the issues I think @fboulnois was getting at:
• ## PR review: Request changes

  ### src/main/resources/changesets/changelog-consent-2026-08-05-dataset-identity.xml:95

  [P1] Preserve compatibility during rolling deployment

  Dropping match_entity.consent in the same deployment that introduces the new application code makes mixed-version operation unsafe. As soon as
  the first new instance runs this migration, existing instances will continue selecting and inserting consent, causing match operations to fail.
  Old instances can also allocate an alias without advancing dataset_alias_seq, allowing a subsequent new-code insert to collide.

  Please use an expand/migrate/contract rollout: add and backfill dataset_id, deploy code compatible with both schemas, then remove consent in a
  later release. Otherwise this requires an explicitly enforced stop-the-world deployment, not only a runbook warning.

  ———

  ### src/main/resources/changesets/changelog-consent-2026-08-05-dataset-identity.xml:45

  [P1] Account for matches when deleting a dataset

  This NO ACTION foreign key prevents deletion of every dataset that has a match. The existing DatasetServiceDAO.deleteDataset() workflow removes
  authorized readers and properties, then calls deleteDatasetById() without deleting matches or their rationales, so the supported dataset deletion
  endpoint will now fail with a foreign-key violation.

  Please define the intended retention behavior and either remove matches and rationales before deleting the dataset or use appropriate cascading
  constraints. An integration test should cover deleting a dataset that has a match and rationale.

  ———

  ### src/main/java/org/broadinstitute/consent/http/db/MatchDAO.java:26

  [P2] Do not truncate aliases longer than six digits

  PostgreSQL LPAD(value, 6, '0') truncates values longer than six characters. For example, alias 1000000 becomes 100000, while
  Dataset.parseAliasToIdentifier() returns DUOS-1000000. This can make migration backfills fail and eventually return incorrect or colliding public
  identifiers.

  Please use padding that preserves the complete alias, such as repeat('0', greatest(0, 6 - length(alias::text))) || alias::text, consistently in
  the DAO queries, backfill, and rollback—or enforce a database maximum of 999999 if six digits is an intentional hard limit.

kevinmarete:
> More feedback from codex that builds on the issues I think @fboulnois was getting at: • ## PR review: Request changes
> 
> ### src/main/resources/changesets/changelog-consent-2026-08-05-dataset-identity.xml:95
> [P1] Preserve compatibility during rolling deployment
> 
> Dropping match_entity.consent in the same deployment that introduces the new application code makes mixed-version operation unsafe. As soon as the first new instance runs this migration, existing instances will continue selecting and inserting consent, causing match operations to fail. Old instances can also allocate an alias without advancing dataset_alias_seq, allowing a subsequent new-code insert to collide.
> 
> Please use an expand/migrate/contract rollout: add and backfill dataset_id, deploy code compatible with both schemas, then remove consent in a later release. Otherwise this requires an explicitly enforced stop-the-world deployment, not only a runbook warning.
> 
> ———
> 
> ### src/main/resources/changesets/changelog-consent-2026-08-05-dataset-identity.xml:45
> [P1] Account for matches when deleting a dataset
> 
> This NO ACTION foreign key prevents deletion of every dataset that has a match. The existing DatasetServiceDAO.deleteDataset() workflow removes authorized readers and properties, then calls deleteDatasetById() without deleting matches or their rationales, so the supported dataset deletion endpoint will now fail with a foreign-key violation.
> 
> Please define the intended retention behavior and either remove matches and rationales before deleting the dataset or use appropriate cascading constraints. An integration test should cover deleting a dataset that has a match and rationale.
> 
> ———
> 
> ### src/main/java/org/broadinstitute/consent/http/db/MatchDAO.java:26
> [P2] Do not truncate aliases longer than six digits
> 
> PostgreSQL LPAD(value, 6, '0') truncates values longer than six characters. For example, alias 1000000 becomes 100000, while Dataset.parseAliasToIdentifier() returns DUOS-1000000. This can make migration backfills fail and eventually return incorrect or colliding public identifiers.
> 
> Please use padding that preserves the complete alias, such as repeat('0', greatest(0, 6 - length(alias::text))) || alias::text, consistently in the DAO queries, backfill, and rollback—or enforce a database maximum of 999999 if six digits is an intentional hard limit.

Thanks for the detailed feedback. I’ve addressed all three concerns:

1. **Rolling-deployment compatibility**
   - This PR is now the expand/migrate compatibility release.
   - It retains `match_entity.consent` and the existing `(purpose, consent)` constraint.
   - New match writes populate both `consent` and `dataset_id`.
   - A compatibility trigger populates `dataset_id` for writes made by older application instances that provide only `consent`.
   - Alias initialization locks dataset writes, and an allocation trigger makes both old `MAX(alias) + 1` inserts and new inserts use `dataset_alias_seq`.
   - Removing `consent` and the compatibility triggers is explicitly deferred to a later contract release after all instances have migrated.

2. **Dataset deletion**
   - Dataset deletion now cascades to its matches, and match deletion cascades to its rationales.
   - An integration test verifies deletion of a dataset with an associated match and rationale.

3. **Aliases longer than six digits**
   - Replaced `LPAD` with non-truncating padding in the DAO queries, migration backfill, compatibility trigger, and other SQL projections.
   - Added coverage confirming alias `1000000` remains `DUOS-1000000`.

src/main/java/org/broadinstitute/consent/http/db/MatchDAO.java
```diff
@@ -41,28 +51,38 @@ public interface MatchDAO extends Transactional<MatchDAO> {
...
-        VALUES
-          (:consentId, :purposeId, :match, :failed, :createDate, :algorithmVersion, :abstain)
+          (dataset_id, consent, purpose, match_entity, failed, create_date, algorithm_version,
+           abstain)
+        SELECT d.dataset_id,
+               'DUOS-' || REPEAT('0', GREATEST(0, 6 - LENGTH(d.alias::text))) || d.alias::text,
+               :purposeId, :match, :failed, :createDate, :algorithmVersion, :abstain
+        FROM dataset d
+        WHERE d.dataset_id = :datasetId
       """)
```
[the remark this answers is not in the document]
kevinmarete: **Addressed. `insertMatch` now uses `INSERT ... VALUES` with a scalar dataset alias lookup. This ensures a missing dataset still attempts the insert and fails through the database constraints instead of producing a zero-row no-op and a null generated key.

I also added a regression test verifying that inserting a match for a nonexistent dataset throws `UnableToExecuteStatementException`.**

src/main/resources/changesets/changelog-consent-2026-08-05-dataset-identity.xml
rushtong: > If our deployment process cannot guarantee that cutover, we should not deploy this PR as-is and would need a separate mixed-version compatibility mechanism.

I'm fairly confident that we cannot guarantee a complete cutover. On a new deploy, liquibase updates run **_before_** the existing pods are destroyed. When that completes, the pods are cycled. I see two options.
1. Go with current approach and inform Nate of the deploy window. We can also deploy off-hours to reduce the risk.
2. In the past, when faced with this, we performed a two or three phase rollout.
   - update the code and db to store new values and maintain existing values
   - perform the back migration
   - full cutover of code - this can often be rolled into the back migration.

docs/plans/dataset-identity-migration-plan.md
```diff
@@ -0,0 +1,94 @@
+# Dataset Identity Migration
+
+## Goal
+
+Use `dataset.dataset_id` for match persistence and relational joins while keeping the canonical
+`DUOS-######` alias as the public dataset identifier. Allocate new aliases from a PostgreSQL
+sequence and enforce uniqueness in the database.
```
rushtong, line 7: These feel like two distinct efforts to me. In an effort to get to the first goal (`dataset.dataset_id` for match persistence), we need consistent canonical identifiers. If, instead, this were phased such that alias/identifier migration -> sequence happened first, this would be a lot smaller and easier to grok. Once that work is complete, the migration of `Match` related work would be a lot simpler to implement.

docs/plans/dataset-identity-migration-plan.md
```diff
@@ -0,0 +1,94 @@
...
+   non-null and `(purpose, dataset_id)` uniqueness.
+8. Keep the legacy `consent` column and `(purpose, consent)` constraint for old instances. New match
+   writes populate both identities during this bounded phase.
+
+Dataset deletion cascades to its matches and their rationales, preserving the supported deletion
+workflow without leaving records that refer to a deleted dataset.
+
+### Contract release (required follow-up)
+
+After every environment is running the compatibility release and reconciliation still reports zero
```
otchet-broad, line 71: Is a new ticket made for this?

src/main/resources/changesets/changelog-consent-2026-08-05-dataset-identity.xml
otchet-broad: Agree that we have done `#2` in the past.  Maybe we have a quick parking lot to discuss this today and decide on a way forward.  I'll add it to the notes.

kevinmarete: Agreed that we cannot depend on a complete pod cutover. I updated the migration to use the multi-phase compatibility approach you described.

During this release:

- the database alias trigger overrides both omitted aliases and legacy `MAX(alias) + 1` values with `nextval`;
- legacy match writes that only provide `consent` have `dataset_id` populated by a compatibility trigger;
- the new application dual-writes `dataset_id` and `consent`; and
- the legacy column and constraint remain in place.

The removal of `match_entity.consent` and the compatibility triggers is explicitly deferred to a later contract release after all environments are running the compatible version. This should allow Liquibase to run while old pods are still serving traffic without requiring an off-hours cutover.

docs/plans/dataset-identity-migration-plan.md
kevinmarete: I agree these are conceptually distinct changes, and splitting them would make each individual PR smaller.

The reason they are together here is that DT-3865 explicitly includes both outcomes:

1. migrate match persistence and uniqueness to `dataset_id`; and
2. replace `MAX(alias) + 1` with sequence-backed alias allocation and database uniqueness.

The alias remains the canonical public identifier, while `dataset_id` becomes the relational identity. The sequence work ensures future public identifiers continue to be allocated safely while the match migration removes those identifiers from relational ownership and joins.

The current implementation is also already phased operationally: this PR is the expand/migrate compatibility release, while removal of the legacy `consent` column and compatibility triggers is deferred to a required follow-up contract release.

Given that context, do you feel strongly that the alias-sequence portion should still be extracted into a prerequisite PR, or does documenting these as two concerns within the compatibility release address the reviewability concern?

docs/plans/dataset-identity-migration-plan.md
kevinmarete: Not yet. The separate contract release became necessary when this PR was changed to support mixed-version deployments.

I agree that it should be tracked in a dedicated follow-up ticket before this PR is merged. That ticket will cover:

- removing the match compatibility trigger;
- removing the legacy `(purpose, consent)` constraint and `match_entity.consent` column; and
- replacing the temporary alias-allocation trigger with the sequence-backed column default.

The follow-up should only be deployed after every environment is running the compatibility release and reconciliation confirms there are no unresolved or conflicting rows.

kevinmarete: I have created [DT-3940](https://broadworkbench.atlassian.net/browse/DT-3940) to track the required contract release.

It covers removing the legacy `match_entity.consent` identity and compatibility triggers, completing the sequence-default transition, and verifying reconciliation and public identifier compatibility. It is linked to DT-3865 and must not be deployed until the compatibility release is running in every environment.

fboulnois:
I think the work here can be separated into three separate changes:

- **A — alias allocation from a sequence.** Changeset 1, `DatasetDAO.insertDataset`, concurrency test.
- **B — `match_entity` keyed on `dataset_id`.** Changesets 2–5, `MatchDAO` / `MatchMapper` / `MatchReducer` / `Match` / `MatchService`, cascade delete, tests.
- **C — SQL identifier formatting** (`LPAD` → `REPEAT`). `LPAD(alias, 6, '0')` truncates 7+ digit aliases, disagreeing with Java's `StringUtils.leftPad`. Files: `DacServiceDAO`.

A and B are independent of each other. C is a hard prerequisite for B.

### Sequencing: C → A → B

- **C before B is mandatory.** Truncating `LPAD` silently breaks the backfill, surfacing three changesets downstream.
- **A before B is preferred.** `uq_dataset_alias` is what keeps the alias → `dataset_id` map single-valued. B can substitute its own precondition if it must ship first.

### Risk

**C — Low.** Read-path only, revert to fix.
**A — Medium**, concentrated at migration time. HALTs on NULL/negative alias (likely — `DacServiceDAO`'s `COALESCE` fallbacks imply legacy NULLs exist). `ACCESS EXCLUSIVE` lock plus unique index build blocks the `dataset` table. The trigger overwrites any explicit alias: app-safe (single write path), but ops exposure on data restores. Aliases will now gap. Rollback is clean.
**B — High.**

1. **Backfill coverage dominates.** `match_entity.consent` was originally an FK to `consents.consentid` (dropped in `changelog-consent-87.0.xml`), so legacy rows may hold consent UUIDs rather than `DUOS-######`. Unresolved rows HALT changeset 5 mid-migration.
2. **The half-applied state is fail-silent.** The new queries `INNER JOIN dataset`, so matches with `NULL dataset_id` vanish from results instead of erroring.
3. **`ON DELETE CASCADE` is new irreversible behavior** — deleting a dataset deletes its match history. That's a product decision riding inside a schema refactor.
4. **A behavior change is bundled in.** The election join gains `e.dataset_id = m.dataset_id`, changing what multi-dataset DARs display. Unrelated to the identity swap.

Rollback is clean (`consent` is retained, and `DatasetIdentityMigrationTest` covers rollback + reapply) — that's what makes High acceptable rather than blocking.

### Blocker on all of it

Run the preflight queries against prod **before** scoping the PRs. The unresolved-match count and the NULL-alias count decide whether A and B are three-day PRs or data-cleanup projects with a PR attached. Both currently HALT on conditions nobody has measured.

kevinmarete:
Thanks, @fboulnois. Based on your feedback and the production preflight results, I am going to close this PR without merging it.

The alias data is clean, but the match migration is not currently deterministic:

- 416 legacy UUID-based match rows require reconciliation
- `election` and `dar_dataset` disagree for 76 rows
- 6 rows have matching multi-dataset candidate sets

I split the work into smaller, independently reviewable tickets:

1. [DT-3941](https://broadworkbench.atlassian.net/browse/DT-3941) — fix public identifier formatting
2. [DT-3865](https://broadworkbench.atlassian.net/browse/DT-3865) — introduce sequence-backed alias allocation
3. [DT-3942](https://broadworkbench.atlassian.net/browse/DT-3942) — reconcile legacy matches and migrate to `dataset_id`
4. [DT-3940](https://broadworkbench.atlassian.net/browse/DT-3940) — remove compatibility mechanisms after rollout

The Jira dependencies and epic implementation order now reflect:

`DT-3941 → DT-3865 → DT-3942 → DT-3940`

I’ll restart with DT-3941 as a small focused PR. The match migration (DT-3942) will remain in the backlog until the historical mapping is recovered or an explicit reconciliation policy is approved.

[DT-3941]: https://broadworkbench.atlassian.net/browse/DT-3941?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DT-3942]: https://broadworkbench.atlassian.net/browse/DT-3942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3003, opened 2026-08-05
[DT-3826] Formal access contract

otchet-broad:
### Addresses

[DT-3826](https://broadworkbench.atlassian.net/browse/DT-3826) — Ticket A-2, formal Elasticsearch access contract. Now targets `develop` directly; the DT-3825 base has merged.

Security risk: **low** — no change to authorization behavior. Everything here is a decision record, a test harness, or a dependency bump. The only production-path change is how the admin capability report resolves the REST client's version string.

### How this branch evolved

It opened as three things: write the access contract, record the A-1 production capability measurement, and fix the client-version lookup that came back indeterminate in every deployed build (the shade plugin strips dependency manifests, so `Package.getImplementationVersion()` is null in the uber jar; it now falls back to `elasticsearch.rest.client.version` in the generated `mvn.properties`, with the version promoted to a pom property).

Review raised P1 concerns about whether the contract's claims about Elasticsearch's native controls were actually true. Rather than re-reading documentation, the branch grew a proof-of-concept harness that exercises them against a live cluster — `ElasticSearchLeakDefensePocTest`, `ElasticSearchIndexSeparationTest`, and a shared access-contract model with four enforcement modes, including `UNMEDIATED`, which reproduces today's code path.

**That measurement changed the design.** Three of the five claims the contract originally made about DLS/FLS were wrong:

- A field outside an FLS grant is not just unreadable, it is **unqueryable** — a query on it silently returns zero hits. So the grant must be *wider* than what the caller may see, and application code has to narrow the response.
- `copy_to` (and mapping-level runtime fields) reach **around** the field grant. No request-side control closes this; the mapping itself becomes an authorization surface that must be asserted against the live cluster.
- Per-document field bundles are impossible — FLS binds to the index privilege. B-2 is cancelled.

Plus sharp edges that are all **silent**: a second role descriptor on an API key yields the *union* of permissions and removes document filtering entirely; a text-field grant doesn't cover `.keyword`, so sorts return the wrong page; `significant_terms` and `explain` report index-wide statistics past DLS. Two of these were only found to be load-bearing by deliberately weakening the PoC to see whether the tests noticed — twice they did not.

Consequences recorded in the plan and contract:

- **The "fallback for unlicensed clusters" is the mainline.** Query mediation, server-owned aggregations, and response-channel filtering are required in every configuration. A licensed cluster defers two tickets, not an epic.
- **OPEN-13 is now open: build native DLS/FLS at all?** It uniquely provides credential-bound enforcement, which matters only if something other than this endpoint queries the index with the service's credentials. Default is defer, don't cancel.
- Arbitrary DSL forced the design onto a **closed allowlist of nine clauses** — `wrapper`, `terms` lookup, `_script` sort and others defeat any field-reference check.
- Two disclosure channels are recommended for **acceptance**: hit-count binary search recovers a "queryable but internal" value in ten accepted requests (OPEN-12), and relevance scores leak index-wide statistics (OPEN-10 — the index split was measured, works as a control, and breaks the catalog: 13.6× score divergence across tiers).

An external document shared outside of the PR condenses all of it for a decision-making audience. Twelve OPEN items remain, each with a preserve-today default; none blocks Phase 1.

Also here: ES REST client 9.5.0 → 9.5.1, a README for the ES tests, `ConsentModule` javadoc on the shared `RestClient` shutdown-hook ordering, and Sonar/coverage cleanup.

### Testing

`ElasticSearchCapabilityServiceTest` covers the pom wiring itself (re-inlining the version would silently return the report to "Could not determine"), absent build properties, and version matching. The PoC tests require a live cluster — see `src/test/java/org/broadinstitute/consent/integration/README-ELASTICSEARCH.md`.

No UI or workflow change, so no PO sign-off. Post-deploy check is `GET /api/elasticSearch/capabilities` against dev, which should now name the client version.

[DT-3826]: https://broadworkbench.atlassian.net/browse/DT-3826?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kevinmarete (commented):
Here are some comments:

- **[P1] Native DLS does not guarantee the stated count/aggregation invisibility.**  
  [Decision 1](https://github.com/DataBiosphere/consent/blob/739fe465dba6a30c7bf9066be34de75bf2b0d265/docs/plans/es-access-contract.md#L24-L41) says restricted documents cannot contribute to counts, facets, or aggregations and that DLS filters before scoring/counting. Elastic explicitly documents that DLS still permits search requests that infer terms and counts from inaccessible documents through aggregate information. The native path therefore needs query mediation/restrictions too, or the contract must weaken this guarantee. See [Elastic’s DLS limitations](https://www.elastic.co/docs/deploy-manage/users-roles/cluster-or-deployment-auth/controlling-access-at-document-field-level#document-level-security-limitations).

- **[P1] The fallback allowlist only filters `_source`, leaving other extraction channels open.**  
  [E-3](https://github.com/DataBiosphere/consent/blob/739fe465dba6a30c7bf9066be34de75bf2b0d265/docs/plans/elasticsearch-service-duos-ui-usage.md#L1415-L1433) filters only `hits.hits[*]._source`, while E-1 preserves caller-supplied aggregations, sorting, and highlighting. A caller could aggregate `accessPolicy.custodianEmails`, highlight a dynamic-map field, or receive an INTERNAL field through hit `sort` values. The design should validate every field reference against an allowed query-field list and filter all response channels—not just `_source`.

- **[P1] The supposedly exhaustive field audit marks several live UI dependencies INTERNAL.**  
  [The classification](https://github.com/DataBiosphere/consent/blob/739fe465dba6a30c7bf9066be34de75bf2b0d265/docs/plans/es-access-contract.md#L292-L346) says these have no consumers, but current `duos-ui/develop` uses:

  - `requestLocation` in dataset details and the library’s “Request Path”
  - `deletable` for the submissions delete action
  - `createUserId` and `study.dataSubmitterId` to construct “My Data Submissions”
  - `submitter.displayName` in the DAC datasets table

  Generating FLS/E-3 allowlists from this table would break those flows. The audit should cover all search callers and distinguish response-visible fields from fields that must remain queryable.

docs/plans/es-access-contract.md
```diff
@@ -0,0 +1,512 @@
...
+simply leave the projection.
+
+Consequently the classification in §B has **two** tiers, not three: served by search, or not served
+by search. Privileged, document-scoped data continues to reach privileged callers through the
+per-dataset endpoints, which already do document-scoped checks in `DatasetService` and are not
+affected by this contract.
+
+**If a future requirement does need a privileged field in search results**, this decision has to be
+reopened, not worked around: the answer is application projection (Epic E's mechanism, applied to
+the native path as well), and the cost is that Epic D stops being sufficient on its own.
```
rushtong, line 264: I can see this happening. For example, we removed `approvedUserIds` which had the potential for exposure of partially controlled info. The need for something similar could resurface. I think the risk is low, but non-zero and there are likely other workarounds we can come up with.

rushtong (approved):
Looks good, thank you 👍🏽

docs/plans/es-access-contract.md
otchet-broad: Agreed on both fronts.  The bigger issue I'm coming back to is aggregation across hidden documents from @kevinmarete's comments.  I had forgotten about that leaky behavior in ES and need to think more about the consequences for this.

fboulnois (approved):
Thanks for verifying the behavior 👍

kevinmarete (approved):

docs/plans/es-access-contract.md
otchet-broad: Plan updated with tests added to address my concerns about the leaky behavior of ES.

kevinmarete (commented):
Two P1 issues remain in the newly added security model.

- **[P1] Field validation can be bypassed through `query_string`.**  
  [`collectFieldReferences`](https://github.com/DataBiosphere/consent/blob/908e4720e580ea5b8dcf6b493cadd4fc1b21020c/src/test/java/org/broadinstitute/consent/integration/ElasticSearchAccessContractModel.java#L460-L498) validates the optional `fields` array but not field names embedded in the query text. For example:

  ```json
  {"query":{"query_string":{"query":"accessPolicy.custodianEmails:secret@example.org"}}}
  ```

  The mediator accepts this unchanged. On the fallback path, which uses privileged credentials without FLS, callers can probe RESPONSE-INTERNAL values by observing whether authorized documents match. Other unenumerated query shapes with a `field` member have the same problem. The mediator needs a closed allowlist of supported query shapes, or it must reject syntax whose field references cannot be exhaustively extracted.

- **[P1] Aggregation bucket keys are not filtered despite the contract promising they are.**  
  [§F.2 requires defensive projection of `buckets[*].key`](https://github.com/DataBiosphere/consent/blob/908e4720e580ea5b8dcf6b493cadd4fc1b21020c/docs/plans/es-access-contract.md#L911-L924), but [`filterChannels`](https://github.com/DataBiosphere/consent/blob/908e4720e580ea5b8dcf6b493cadd4fc1b21020c/src/test/java/org/broadinstitute/consent/integration/ElasticSearchAccessContractModel.java#L681-L708) only filters objects containing `_source`; bucket keys pass through unchanged. A drifted server aggregation over `accessPolicy.custodianEmails` would therefore return the email values, exactly the case §F.2 claims to stop. Because bucket keys do not carry their source field, the practical fix is to validate the server-owned aggregation vocabulary before execution, or retain request metadata when filtering the response. The drift test currently exercises only `top_hits`, so it misses this case.

rushtong:
The PR description is slightly out of date here:
> Stacked on otchet-dt-3825-confirm-es-in-env — review that PR first, or read this one's diff against
that branch rather than develop.

rushtong (approved):
Looks reasonable - no blocking issues that I can see 👍🏽

otchet-broad:
> The PR description is slightly out of date here:
> 
> > Stacked on otchet-dt-3825-confirm-es-in-env — review that PR first, or read this one's diff against
> > that branch rather than develop.

Good point.  I'm going to update that now.

PR #3004, opened 2026-08-06
[DT-3869] Define the Study Template CSV v1 contract and fixtures

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3869

### Summary

- Defines the canonical Study Template CSV v1 contract.
- Adds synthetic valid and invalid fixture files for future parser tests.
- Adds fixture integrity tests covering CSV structure, UTF-8, BOM handling, versioning, duplicate fields, and structured errors.
- Adds Apache Commons CSV as a test-only dependency.
- No parser, endpoint, or API behavior is introduced.

----

- Verify all tests go green

### DIFF
--- a/docs/study-template-v1.md
+++ b/docs/study-template-v1.md
@@ -123,9 +123,15 @@ field is absent when it has no rows; there is no encoding for an explicitly empt
 appear twice within one field.
 
 JSON in a `value` cell is therefore limited to `fileTypes` and asset payloads, which have no
-lossless flat representation. Both are optional and neither appears in a minimal template. This
-narrowing is pending confirmation with product; if hand-authored JSON is unacceptable for assets
-too, the remaining option is dropping non-file assets from v1 rather than re-encoding them.
+lossless flat representation. Both are optional and neither appears in a minimal template.
+
+Product confirmed that producers get the template by downloading it from the DUOS UI rather than
+building it from this contract. Those JSON cells are edited in place, not authored from scratch,
+which is what makes them acceptable in a user-facing template. That makes the download a v1
+dependency rather than a convenience: it must emit the header and the `templateVersion`,
+`recordType`, `recordId`, `parentRecordId`, and `field` columns for the fields being offered,
+leaving `value` empty for the producer to fill in, and it must escape leading `=`, `+`, and `@` as
+described under Spreadsheet compatibility.
 
 ## Study field mapping
 
--- a/docs/study-template-v1.md
+++ b/docs/study-template-v1.md
@@ -29,9 +29,9 @@ platform upload limit: the shared check in `Resource#validateFileDetails` uses t
 certification documents but far too permissive for a text template.
 
 The limit is generous for the intended client. Rows average roughly 120 bytes. A study with 40
-study-level rows, 100 consent groups at about 25 rows each, and 20 asset rows is about 2,560 rows,
-or roughly 300 KB — about a tenth of the limit. A file that exceeds 5 MiB is not a study template
-that a person authored in a spreadsheet.
+study-level rows and 100 consent groups at about 25 rows each, file types included, is around 2,540
+rows, or roughly 300 KB — about a tenth of the limit. A file that exceeds 5 MiB is not a study
+template that a person authored in a spreadsheet.
 
 ### Spreadsheet compatibility
 
@@ -82,18 +82,18 @@ Each row assigns one `field` to one logical record.
 | --- | --- | --- | --- |
 | `study` | Must be `study` | Empty | One field on the single `StudyRegistrationRequest`. |
 | `consentGroup` | Unique non-empty template ID | Must be `study` | One field on a `ConsentGroupRequest`; each ID becomes one `consentGroups[]` item. |
-| `asset` | Unique non-empty template ID | Must be `study` | One supported non-file asset; `field` names the asset collection and `value` is a JSON object. |
+| `fileType` | Unique non-empty template ID | Must be a `consentGroup` `recordId` | One field on a `fileTypes[]` item belonging to that consent group. |
 
-`recordId` is template-only grouping metadata for study and consent-group records. For assets it
-also becomes the client asset identifier (`modelId`, `workspaceId`, `publicationId`,
-`presentationId`, `clinicalTrialId`, `fundingId`, or `ipId`). Numeric study and dataset IDs remain
-server-owned.
+`recordId` is template-only grouping metadata. It never reaches the wire: numeric study and dataset
+IDs remain server-owned, and `fileTypes[]` items are positional. A `fileType` record's
+`parentRecordId` must name a `consentGroup` record declared in the same file; `fileTypes` items are
+attached to that consent group in file order.
 
 Rows may appear in any order after the header. A `(recordType, recordId, field)` tuple may occur
 only once for a single-valued field; duplicate assignments are errors even if their values match.
 Scalar array fields are the one exception: they repeat the tuple once per item, as described under
-Value encoding. Unknown headers, record types, fields, and asset properties are errors and are never
-silently ignored.
+Value encoding. Unknown headers, record types, and fields are errors and are never silently
+ignored.
 
 ## Value encoding
 
@@ -105,12 +105,12 @@ silently ignored.
 | Date | ISO local date `YYYY-MM-DD`; calendar-invalid dates are rejected. |
 | URI | Absolute `http` or `https` URI. |
 | Scalar array | One row per item, repeating the same `(recordType, recordId, field)` tuple. Each `value` cell holds one plain string or enum item, encoded exactly as a single value of that kind. |
-| Object array | JSON array in one CSV cell. Used only by `fileTypes`. |
-| Object | JSON object in one CSV cell. Used only for supported asset payloads. |
 | Empty value | An empty `value` cell means absent (`null`), not an empty string or empty array. Omit optional fields; empty required values fail validation. |
 
-Every array-typed field is a scalar array except `fileTypes`, which is an object array. So the wire
-value `["Genomic","Phenotypic"]` is two rows, and neither cell needs quoting:
+Every `value` cell is a single scalar. **v1 has no JSON encoding at all**: structured wire values
+are expressed as rows, not as embedded documents. Arrays of scalars repeat their row, and the one
+array of objects, `fileTypes`, is its own record type. So the wire value `["Genomic","Phenotypic"]`
+is two rows, and neither cell needs quoting:
 
 ```csv
 1,study,study,,dataTypes,Genomic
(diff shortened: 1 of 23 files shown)


rushtong (approved):
Nice work 👍🏽

docs/study-template-v1.md
```diff
@@ -0,0 +1,278 @@
...
+registration draft. It maps template fields to Consent's `StudyRegistrationRequest` wire contract;
+it does not define a new registration API or replace existing registration validation.
+
+Canonical synthetic fixtures live in `src/test/resources/fixtures/study-template/v1`, where Consent
+parser tests can load them directly. DT-3869 itself does not implement parsing or UI behavior.
+
+## File contract
+
+- File extension: `.csv`.
+- Maximum size: 5 MiB (5,242,880 bytes), including a UTF-8 BOM when present.
```
otchet-broad, line 23: Where does this 5MB limit come from?  Will that work for our client?

rushtong: We check it here: https://github.com/DataBiosphere/consent/blob/develop/src/main/java/org/broadinstitute/consent/http/resources/Resource.java#L291
which delegates to:
https://github.com/augustd/owasp-java-fileio/blob/master/src/main/java/org/owasp/fileio/FileValidator.java#L46

docs/study-template-v1.md
```diff
@@ -0,0 +1,278 @@
...
+- Encoding: UTF-8. A UTF-8 BOM is accepted only at the beginning of the file.
+- Dialect: RFC 4180-style comma-separated values. Double quotes enclose values; a literal quote is
```
otchet-broad, line 25: If I remember the original ticket correctly, I thought we needed to validate we supported CSVs from Excel and Google Sheets.  While RFC-4180 is a great starting point for CSVs - and I agree that it can be a great foundational way to get started, Excel [has issues](https://www.loc.gov/preservation/digital/formats/fdd/fdd000323.shtml) with RFC 4180 style CSVs.  We might need to make adjustments going forward.  If we expect European users, for example, we might want to allow the user to supply an optional separator.

docs/study-template-v1.md
```diff
@@ -0,0 +1,278 @@
...
+| `nihGrantContractNumber` | `nihGrantContractNumber` | String | Required for NHGRI-funded or AnVIL-seeking `nihAnvilUse` choices. |
+| `nihICsSupportingStudy` | `nihICsSupportingStudy` | Enum array | Optional; exact NIH institute/center abbreviations. |
+| `nihProgramOfficerName` | `nihProgramOfficerName` | String | Optional. |
+| `nihInstitutionCenterSubmission` | `nihInstitutionCenterSubmission` | Enum | Optional; exact NIH institute/center abbreviation. |
+| `nihGenomicProgramAdministratorName` | `nihGenomicProgramAdministratorName` | String | Optional. |
+| `multiCenterStudy` | `multiCenterStudy` | Boolean | Optional. |
+| `collaboratingSites` | `collaboratingSites` | String array | Optional. |
+| `controlledAccessRequiredForGenomicSummaryResultsGSR` | Same name | Boolean | Optional. |
+| `controlledAccessRequiredForGenomicSummaryResultsGSRRequiredExplanation` | Same name | String | Required when controlled GSR access is `true`. |
+| `alternativeDataSharingPlan` | `alternativeDataSharingPlan` | Boolean | Optional. |
```
otchet-broad: Since the user can't upload the alternativeDataSharingPlan file, it probably makes sense to remove all of the 'alternativeDataSharingPlan' related items from the CSV.  Might be worth confirming with @jlaw-codes.

fboulnois (approved):
Have we run these templates by Jonathan? I see some JSON in the multi-consent-group. I think these templates were meant to be user facing, so that might become complicated with embedded JSON.

otchet-broad (changes requested):
The alternativeSharingPlan should be run to ground.

kevinmarete:
@jlaw-codes — DT-3869 defines the v1 CSV contract for importing one study and its datasets into a registration draft (doc: `docs/study-template-v1.md`). Nothing is parsed or wired up yet; the contract gates that work, so I need a product call on two things before it merges.

**1. Dropping the alternative data sharing plan from the template**

The template can't carry file uploads. Rather than exclude only `alternativeDataSharingPlanFileName`, I've removed the whole `alternativeDataSharingPlan*` group — the boolean, reasons, explanation, dataSubmitted, dataReleased, both target dates, and accessManagement. The reasoning: importing the flags without the document produces a draft the user still has to open the form to finish, so the CSV saves them nothing.

Is that right — a user who needs an alternative sharing plan fills that section in on the populated draft, where the upload lives? Or would you rather the non-file fields import anyway so they only have to attach the document?

**2. How much hand-authored JSON is acceptable**

Each row is one field. Multi-value fields like `dataTypes` are one row per item:

    1,study,study,,dataTypes,Genomic
    1,study,study,,dataTypes,Phenotypic

That keeps JSON out of every field a typical study needs. It survives in two optional places — `fileTypes` on a dataset, and non-file assets (workspaces, publications, presentations, clinical trials, funding, IP), where the whole payload is one JSON object in a cell:

    1,asset,workspace-1,study,workspaces,"{""name"":""...."",""platform"":""Terra"",...}"

Neither appears in a minimal template. Is that acceptable for the people who'll actually fill these out? If not, the realistic alternative is dropping non-file assets from v1 and adding them once there's a UI to author them — there's no
lossless flat CSV form for those payloads.

**One thing that would help me judge #2**: how do you expect producers to get the template — download a pre-filled one from DUOS, or build it from this spec? If DUOS generates it, the JSON is something people edit rather than write, and it matters much less.

jlaw-codes:
Fine to exclude Alt DS plans
No JSON is preferred

docs/study-template-v1.md
kevinmarete: Good question — and @rushtong's pointer turned out not to back the number. `Resource#validateFileDetails` uses the OWASP `FileValidator` default, which is 500,000,000 bytes, and nothing in the codebase calls `setMaxFileUploadSize`. So that check is 500 MB, not 5.

The doc now says 5 MiB is template-specific and has to be enforced by the import endpoint itself, with sizing for your "will that work for our client" question: rows average ~120 bytes, so a study with 40 study-level rows and 100 consent groups at ~25 rows each is ~2,540 rows, roughly 300 KB. About a tenth of the limit.

docs/study-template-v1.md
kevinmarete: Added a Spreadsheet compatibility section. It documents what we accept from those tools by default — CRLF, the BOM Excel writes for "CSV UTF-8", minimal quoting, Sheets' comma-delimited export — and calls out two Excel deviations we reject rather than guess at:

- **Locale list separator.** Your European-users case. v1 stays comma-only, but when the header parses as a single cell containing `;` or a tab we reject with a message naming the detected character and asking for a comma-delimited re-export, instead of the misleading "missing header" you'd otherwise get. An optional producer-supplied delimiter is written down as a v2 item.
- **Value auto-formatting** — Excel turning cells into dates or scientific notation. We don't reverse it; those fail their normal type rule with row and column context.

Two fixtures back this: `valid/excel-export-valid.csv` (real BOM + CRLF, asserted byte-level) and `invalid/semicolon-delimited.csv`.

docs/study-template-v1.md
kevinmarete: Done — and @jlaw-codes confirmed ("Fine to exclude Alt DS plans"). Removed the whole group rather than just `alternativeDataSharingPlanFileName`: the CSV can't carry the plan document, so importing the flags alone produces a draft the user still has to open the form to finish. The doc records that reasoning and points users at the draft form, where the upload lives.

src/test/java/org/broadinstitute/consent/http/models/dto/registration/StudyTemplateV1FixturesTest.java
```diff
@@ -0,0 +1,169 @@
...
+      CSVFormat.RFC4180.builder().setIgnoreEmptyLines(true).get();
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+  @Test
+  void validFixturesAreParseableAndVersioned() throws Exception {
+    for (String name : VALID_FIXTURES) {
+      byte[] bytes = readResource("valid/" + name);
+      List<CSVRecord> records = parse(bytes);
+
+      assertEquals(HEADERS, records.getFirst().toList(), name);
```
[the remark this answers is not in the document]
kevinmarete: `pom.xml` sets `java.version=25`, which feeds `maven.compiler.release`, so sequenced-collection methods are fine here. No change needed.

src/test/java/org/broadinstitute/consent/http/models/dto/registration/StudyTemplateV1FixturesTest.java
```diff
@@ -0,0 +1,169 @@
...
+          assertTrue(HEADERS.contains(error.get("column").asText()), manifestName);
+        }
+      }
+    }
+  }
+
+  @Test
+  void structuralInvaliditiesRemainIntentional() throws Exception {
+    List<CSVRecord> duplicateHeader = parse(readResource("invalid/duplicate-header.csv"));
+    List<String> duplicateHeaderValues = duplicateHeader.getFirst().toList();
```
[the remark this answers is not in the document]
kevinmarete: `pom.xml` sets `java.version=25`, which feeds `maven.compiler.release`, so sequenced-collection methods are fine here. No change needed.

src/test/java/org/broadinstitute/consent/http/models/dto/registration/StudyTemplateV1FixturesTest.java
```diff
@@ -0,0 +1,169 @@
...
+  @Test
+  void utf8BomIsAcceptedAtTheBeginning() throws Exception {
+    byte[] fixture = readResource("valid/minimal-valid.csv");
+    byte[] withBom = new byte[fixture.length + 3];
+    withBom[0] = (byte) 0xEF;
+    withBom[1] = (byte) 0xBB;
+    withBom[2] = (byte) 0xBF;
+    System.arraycopy(fixture, 0, withBom, 3, fixture.length);
+
+    assertEquals(HEADERS, parse(withBom).getFirst().toList());
```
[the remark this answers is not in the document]
kevinmarete: `pom.xml` sets `java.version=25`, which feeds `maven.compiler.release`, so sequenced-collection methods are fine here. No change needed.

kevinmarete:
Pushed two rounds of changes since your reviews. Summary of what moved:

- **Alternative data sharing plan** — the whole `alternativeDataSharingPlan*` group is gone, not just the filename field. @jlaw-codes confirmed.
- **No JSON in the template.** Also from @jlaw-codes. Scalar arrays now repeat their row (`dataTypes` on two rows instead of `"[""Genomic""]"`), `fileTypes` became a `fileType` record type, and non-file assets are deferred to v2 — their payloads nest, so a JSON-free encoding would need dotted paths or second-level child records.
- **5 MiB limit** — now has a rationale section; see the thread below.
- **Excel / Google Sheets** — new Spreadsheet compatibility section and two new fixtures; see the thread below.

The plan in duos-ui is updated to match, including the ticket changes these decisions force: DataBiosphere/duos-ui#3828.

@otchet-broad this should cover your change request when you get a chance.

kevinmarete:
@fboulnois yes — took it to Jonathan, and he preferred no JSON, so it's all gone rather than justified. Scalar arrays repeat their row, `fileTypes` became its own record type, and non-file assets are deferred to v2 since their payloads nest. Even `minimal-valid.csv` no longer has `"[""Genomic""]"` in it.

He also confirmed the template gets downloaded from the DUOS UI rather than hand-built, which is now a v1 requirement on the UI side — DataBiosphere/duos-ui#3828 adds it to Ticket 4.

otchet-broad (approved):
👍

PR #3013, opened 2026-08-11
[DT-3939] DAC Dashboard API.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3939](https://broadworkbench.atlassian.net/browse/DT-3939)

### Summary

* Adds API for DAC Dashboard
* Refactors to extract concerns common to all dashboards
* Allows Members and Chairs to have visibility on publicVisibility = false because of potential cases where DACs are voting on DARs where publicVisibility = false.

----

- Verify all tests go green


rushtong (approved):
Nice work 👍🏽

kevinmarete (commented):
Nice work, one blocking issue, matching the concern in the DUOS UI PR.

**P1 — Dual-role users lose member-scoped DAR counts**

A user with both Chairperson and Member roles is always evaluated as a Chairperson here. Since `DacDashboardDAO.getCounts()` filters `user_role` by that single role ID, DARs and pending votes from DACs where the user is only a member are omitted.

This also makes the response internally inconsistent: `getDacIds(user)` supplies associations from both roles to the Elasticsearch counts, while the relational DAR counts include only Chairperson associations.

Could we aggregate both role scopes for dual-role users, deduplicating collections that overlap, or otherwise let the caller select a role? This needs to preserve the different `awaitingMyVote` semantics for chairs and members.

kevinmarete (approved):
👍🏾

PR #3014, opened 2026-08-12
[DT-3888] Surface SO authorization model on indexed datasets

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3888

### Summary

Datasets now carry the DAC automation state clients need, so duos-ui no longer resolves DAC rules itself — it was issuing one `/api/dac/{id}/rules` request per unique DAC per page of results and re-deriving the rules in JS.

`DatasetTerm` gains two fields:

- `soApprovalModel` (`PER_REQUEST` | `PRE_AUTHORIZED`), from whether the DAC has `REQUIRE_SO_DAR_APPROVAL` enabled.
- `instantApprovalEligible`, from whether the DAC has an auto-approve rule enabled that this dataset's own data use satisfies.

Eligibility runs the rule implementations themselves rather than restating the policy. The dataset-side half of each rule is split out as `datasetQualifies(Dataset)`, so `compare()` becomes `datasetQualifies(dataset) && requestIsOnlyHMB(...)` — unchanged otherwise, and the existing rule tests pass untouched. `hasCanonicalSinglePrimary` moved to `DataUsePrimaryClassifier` so `applyRule` and the indexer share one definition of the shape gate rather than each holding a copy.

**This changes which datasets can show the instant-approval badge — measured impact on prod: 3 datasets.** The client-side re-derivation matched the translated `DataUseSummary` and rejected four secondary codes (`IRB, COL, GSO, NPU`); `hasNoModifiers` rejects roughly twenty. A GRU dataset carrying a publication moratorium (`MOR`), a geographic restriction (`GS`), or required publication of results (`PUB`) could therefore advertise instant approval that no DAR against it would ever receive.

Measured against prod: **3 datasets lose the badge** — DUOS-000151, DUOS-000157 and DUOS-000266, all in DAC 1, all via `secondaryOther` (free text, rendered as secondary code `OTHER`, which the old frontend check never looked at). Dev showed 0, so this is confined to one DAC. **No approval outcome changes**: `hasNoModifiers` already rejects a non-empty `secondaryOther`, so `compare()` returns false for all three today and the engine has never auto-approved them — the UI simply stops claiming otherwise.

One of the three carries a genuine restriction ("Possible restriction on gamete research"). The other two carry an audit note rather than a restriction, and are being excluded from auto-approval today by that stray text — a pre-existing data issue this change surfaces rather than causes, raised separately with @ncalvanese1 and not a blocker here.

Rules are resolved once per index batch via `findEnabledRuleAssignments()`, returning every enabled `(dac_id, rule)` pair for AVAILABLE rules in one query. Still DAC-keyed rather than reusing `filterDatasetIdsByAutomationRuleType`, because indexing walks the entire dataset corpus and a dataset-keyed query would mean an unbounded IN list. A dataset with no DAC resolves to `PRE_AUTHORIZED` and ineligible without consulting the query at all.

Both fields are left absent, not `false`, when the rules cannot be resolved, so clients can tell "not eligible" from "not yet known" and render nothing rather than being told the wrong approval process. Two tests assert the serialized document through the same `GsonUtil` the bulk indexer uses, since those field names and values are duos-ui's only contract here and a rename on either side should fail a test rather than silently blank the Data Library.

Any rule toggle now reindexes that DAC's datasets — every rule affects an indexed field, so the previous rule-type check no longer described anything. Scoped to the DAC rather than the whole corpus, since prod is now mostly external entries with no DAC, which no rule change can affect. `findDatasetsAssociatedWithDac` rather than `findDatasetIdsByDacIds`, so datasets carrying the DAC as a `dataAccessCommitteeId` property are covered too; it returns full rows and its property joins can repeat a dataset, so ids are mapped and deduped. The IN list is bounded by DAC size rather than corpus size.

It runs off the request thread so toggle latency does not track DAC size, and toggles are queued FIFO so DACs reindex in the order they were toggled. A DAC already queued is not queued twice — the pending pass has not started and reads the newest state when it does. Failures are logged rather than raised: the toggle is already committed and audited by the time the reindex runs.

**Deploy order matters.** This PR, then `POST /api/dataset/index`, then DataBiosphere/duos-ui#3787. Unlike the SO Approval column, the instant-approval badge is an existing feature, so shipping the UI first would make it vanish rather than degrade.

----

- Verify all tests go green

[DT-3888]: https://broadworkbench.atlassian.net/browse/DT-3888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/enumeration/SoApprovalModel.java
+++ b/src/main/java/org/broadinstitute/consent/http/enumeration/SoApprovalModel.java
@@ -1,0 +1,47 @@
+package org.broadinstitute.consent.http.enumeration;
+
+import com.google.gson.annotations.SerializedName;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Which Signing Official authorization model applies to a dataset, derived from whether the
+ * dataset's DAC has the REQUIRE_SO_DAR_APPROVAL automation rule enabled. Surfaced on indexed
+ * datasets so clients do not have to resolve DAC rules themselves.
+ */
+public enum SoApprovalModel {
+  /** The SO named in each DAR must approve that request before the DAC reviews it. */
+  @SerializedName("PER_DAR")
+  PER_DAR("PER_DAR"),
+
+  /** The SO authorizes researchers in advance; no per-request SO approval is needed. */
+  @SerializedName("PRE_AUTHORIZED")
+  PRE_AUTHORIZED("PRE_AUTHORIZED");
+
+  private final String value;
+
+  public static List<String> getValues() {
+    return Stream.of(SoApprovalModel.values())
+        .map(SoApprovalModel::getValue)
+        .collect(Collectors.toList());
+  }
+
+  SoApprovalModel(String value) {
+    this.value = value;
+  }
+
+  public String getValue() {
+    return value;
+  }
+
+  public static SoApprovalModel fromValue(String value) {
+    Optional<SoApprovalModel> model =
+        EnumSet.allOf(SoApprovalModel.class).stream()
+            .filter(m -> m.getValue().equalsIgnoreCase(value))
+            .findFirst();
+    return model.orElse(null);
+  }
+}
--- a/src/main/java/org/broadinstitute/consent/http/enumeration/SoApprovalModel.java
+++ b/src/main/java/org/broadinstitute/consent/http/enumeration/SoApprovalModel.java
@@ -11,9 +11,9 @@ import com.google.gson.annotations.SerializedName;
  * change the published contract.
  */
 public enum SoApprovalModel {
-  /** The SO named in each DAR must approve that request before the DAC reviews it. */
-  @SerializedName("PER_DAR")
-  PER_DAR,
+  /** The SO named in each access request must approve that request before the DAC reviews it. */
+  @SerializedName("PER_REQUEST")
+  PER_REQUEST,
 
   /** The SO authorizes researchers in advance; no per-request SO approval is needed. */
   @SerializedName("PRE_AUTHORIZED")
--- a/src/main/java/org/broadinstitute/consent/http/enumeration/SoApprovalModel.java
+++ b/src/main/java/org/broadinstitute/consent/http/enumeration/SoApprovalModel.java
@@ -1,47 +1,21 @@
 package org.broadinstitute.consent.http.enumeration;
 
 import com.google.gson.annotations.SerializedName;
-import java.util.EnumSet;
-import java.util.List;
-import java.util.Optional;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 /**
  * Which Signing Official authorization model applies to a dataset, derived from whether the
  * dataset's DAC has the REQUIRE_SO_DAR_APPROVAL automation rule enabled. Surfaced on indexed
  * datasets so clients do not have to resolve DAC rules themselves.
+ *
+ * <p>The wire values are pinned with {@link SerializedName} so renaming a constant cannot silently
+ * change the published contract.
  */
 public enum SoApprovalModel {
   /** The SO named in each DAR must approve that request before the DAC reviews it. */
   @SerializedName("PER_DAR")
-  PER_DAR("PER_DAR"),
+  PER_DAR,
 
   /** The SO authorizes researchers in advance; no per-request SO approval is needed. */
   @SerializedName("PRE_AUTHORIZED")
-  PRE_AUTHORIZED("PRE_AUTHORIZED");
-
-  private final String value;
-
-  public static List<String> getValues() {
-    return Stream.of(SoApprovalModel.values())
-        .map(SoApprovalModel::getValue)
-        .collect(Collectors.toList());
-  }
-
-  SoApprovalModel(String value) {
-    this.value = value;
-  }
-
-  public String getValue() {
-    return value;
-  }
-
-  public static SoApprovalModel fromValue(String value) {
-    Optional<SoApprovalModel> model =
-        EnumSet.allOf(SoApprovalModel.class).stream()
-            .filter(m -> m.getValue().equalsIgnoreCase(value))
-            .findFirst();
-    return model.orElse(null);
-  }
+  PRE_AUTHORIZED
 }
--- a/src/main/java/org/broadinstitute/consent/http/service/DACAutomationRuleService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DACAutomationRuleService.java
@@ -4,6 +4,7 @@ import static java.util.Objects.isNull;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.inject.Inject;
+import jakarta.ws.rs.core.Response;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Date;
@@ -52,10 +53,15 @@ public class DACAutomationRuleService implements ConsentLogger {
   private final VoteDAO voteDAO;
   private final VoteService voteService;
   private final VoteServiceDAO voteServiceDAO;
+  private final ElasticSearchService elasticSearchService;
 
   @Inject
   public DACAutomationRuleService(
-      Jdbi jdbi, VoteServiceDAO voteServiceDAO, VoteService voteService) {
+      Jdbi jdbi,
+      VoteServiceDAO voteServiceDAO,
+      VoteService voteService,
+      ElasticSearchService elasticSearchService) {
+    this.elasticSearchService = elasticSearchService;
     this.dataAccessRequestDAO = jdbi.onDemand(DataAccessRequestDAO.class);
     this.datasetDAO = jdbi.onDemand(DatasetDAO.class);
     this.ruleDAO = jdbi.onDemand(DACAutomationRuleDAO.class);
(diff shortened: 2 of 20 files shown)


src/main/java/org/broadinstitute/consent/http/enumeration/SoApprovalModel.java
```diff
@@ -0,0 +1,21 @@
...
+ * Which Signing Official authorization model applies to a dataset, derived from whether the
+ * dataset's DAC has the REQUIRE_SO_DAR_APPROVAL automation rule enabled. Surfaced on indexed
+ * datasets so clients do not have to resolve DAC rules themselves.
+ *
+ * <p>The wire values are pinned with {@link SerializedName} so renaming a constant cannot silently
+ * change the published contract.
+ */
+public enum SoApprovalModel {
+  /** The SO named in each DAR must approve that request before the DAC reviews it. */
+  @SerializedName("PER_DAR")
```
otchet-broad: I wonder if this makes sense to call it "PER_REQUEST" vs "PER_DAR".  In my mind, DAR attaches it to a specific type of request that may/may not change in the future.

src/main/java/org/broadinstitute/consent/http/service/DACAutomationRuleService.java
```diff
@@ -89,26 +95,54 @@ public List<DACAutomationRule> findAllByDacId(Integer dacId) {
...
         ruleId, true, insertTime.toEpochMilli(), user.getDisplayName(), user.getEmail());
   }
 
+  /**
+   * Indexed datasets carry their DAC's Signing Official authorization model, so toggling
+   * REQUIRE_SO_DAR_APPROVAL leaves those documents stale. Reindex failures are logged rather than
+   * raised: the toggle itself is already committed and audited, and the next reindex corrects the
+   * documents.
+   */
+  private void reindexDatasetsForSoApprovalChange(Integer dacId, DACAutomationRule rule) {
```
otchet-broad: Why not just reindex all?  I think we've made significant performance improvements in those methods to just reindex all.  It would allow us to leave this code out.

otchet-broad (approved):
Nice work!  Left a couple of comments to consider, but solid implementation.

otchet-broad:
@kevinmarete, it my be worth extending this to decorate other RADAR rules on datasets so that the frontend can stop making calls an decorating independently.  This would allow us to fold that into filters, etc.  Is that too much to consider for this?

kevinmarete:
Not too much — done here, three commits on top of what you reviewed. `DatasetTerm` now carries `instantApprovalEligible` too, computed by running the rule implementations themselves rather than a second copy of the policy, and duos-ui drops the per-DAC fetch entirely. Filters landed in DataBiosphere/duos-ui#3787. Body updated with the details.

One thing worth your attention: this turned out to be a correctness fix as well as a request-count one. The JS re-derivation recognised four data use modifiers where `hasNoModifiers` rejects about twenty, so some datasets show the instant-approval badge today that the approval engine would never auto-approve. Those badges go away — probably wants PO sign-off rather than riding along as a refactor.

src/main/java/org/broadinstitute/consent/http/enumeration/SoApprovalModel.java
kevinmarete: Agreed — `PER_REQUEST` now, here and in duos-ui. The UI already said "Per-Request Approval", so the code was the odd one out.

src/main/java/org/broadinstitute/consent/http/service/DACAutomationRuleService.java
kevinmarete: Done — reindexes all now, which drops the DAC-scoped lookup and the IN-list concern Copilot kept raising. It also stops being rule-type specific: every rule affects an indexed field now, so that check no longer described anything.

Two deviations I'd like your read on. It runs off the request thread, since a full-corpus sweep inline makes the toggle's latency track dataset count. And passes are coalesced — going full-corpus *and* async at once removed both brakes, so two chairs toggling within a minute would otherwise put two whole-corpus reindexes on the DB at the same time. Coalescing rather than skipping, because the last toggle carries the newest state.

If the perf work you had in mind makes a synchronous call fine, happy to simplify.

kevinmarete: Ran this against a local stack (`config/docker-compose.yaml`, dev database copy, 1,154 datasets) rather than leaving it as design intent:

- **Async holds up.** Toggling `REQUIRE_SO_DAR_APPROVAL` on a DAC returned in ~2s; the reindex landed ~8s later. The request never waits on the sweep.
- **Coalescing works.** Six toggles fired back to back produced **2 reindex passes, not 6** — and the index converged on the *final* rule state, which is the reason for coalescing rather than skipping.
- **Scoping is right.** The DAC's three datasets flipped to `PER_REQUEST` while `instantApprovalEligible` stayed put, so the two rule-derived fields don't interfere.

So the answer to my own question: I'd keep it async. 8s for ~1,200 datasets is fine in the background and would be a poor toggle response, and prod is larger. Still happy to drop the coalescing if you think it's more machinery than a rare admin action warrants — but going full-corpus *and* async removed both brakes at once, and two chairs toggling within a minute would otherwise put two whole-corpus sweeps on the DB together.

Copilot also caught a real one in this method while I was here (fixed in c1b6443b): `executorService.submit` sat outside the lock after `reindexRunning` was set, so a rejected submission — an executor shutting down — both failed a toggle that was already committed and left `reindexRunning` stuck true, silently disabling reindexing for the life of the process.

kevinmarete:
Following up on the badge change I flagged above — measured against prod rather than left as a warning.

**3 datasets lose the badge**: DUOS-000151, DUOS-000157 and DUOS-000266, all in DAC 1, all via `secondaryOther` — free text, rendered as secondary code `OTHER`, which `isOnlyGRUorHMB` never checked. (Dev showed 0, so this is confined to one DAC in prod.)

**No approval outcome changes.** `hasNoModifiers` already rejects a non-empty `secondaryOther`, so `compare()` returns false for all three today and the engine has never auto-approved them. This only stops the Data Library claiming they're eligible.

One of the three carries a genuine restriction ("Possible restriction on gamete research") — correctly excluded. The other two say "Previously categorized with 'other data use terms', now reclassified to GRU", which is an audit note sitting in the restriction field rather than a restriction. Those two are clean GRU that *should* qualify and are being blocked from auto-approval today by that stray text; the badge was masking it. Raised separately below — a data change, not something to fold in here.

PR body updated with the same numbers.

kevinmarete:
@ncalvanese1 — a data gap this change surfaced, flagging since it needs a data operation rather than a code fix.

Two DAC 1 datasets have an audit note parked in the `secondaryOther` restriction field:

| Dataset | `secondaryOther` |
| --- | --- |
| DUOS-000157 | "Previously categorized with 'other data use terms', now reclassified to GRU" |
| DUOS-000266 | "Previously categorized with 'other data use terms', now reclassified to GRU" |

That field is treated as a real restriction by the automation engine — `hasNoModifiers` rejects any non-empty value — so both datasets are excluded from DACbot auto-approval today, even though they are otherwise clean GRU and their DAC has GRU auto-approval enabled. The text reads as a record of a past reclassification rather than a restriction on use.

This predates the PR; it just became visible because the Data Library was showing these as instant-approval eligible while the engine was refusing them. If the note is indeed only a historical annotation, clearing `secondaryOther` on both would make them genuinely auto-approvable.

A third dataset (DUOS-000151) also has `secondaryOther` set, but there it is a real restriction — "Possible restriction on gamete research" — so it should stay as is.

kevinmarete:
@jlaw-codes — PO sign-off needed on one small user-visible change, should be quick.

**What changes.** Three datasets in the Data Library currently show an "instant approval eligible" badge. After this ships they won't.

**Why.** The badge was wrong on those three. It's meant to mean "a request for this dataset can be auto-approved without DAC review" — but the system was never going to auto-approve them, because each carries a free-text restriction that no automated rule can evaluate. The badge has been promising something the system wouldn't deliver. This change makes the badge match what actually happens.

**What does not change.** No researcher loses access to anything. No request is approved or rejected differently. These three were already going to normal DAC review, and still will. The only difference is the Data Library stops advertising otherwise.

**Realistically this is one dataset, not three:**

| Dataset | Restriction text | Note |
| --- | --- | --- |
| DUOS-000151 | "Possible restriction on gamete research" | A genuine restriction — badge correctly removed |
| DUOS-000157 | "Previously categorized with 'other data use terms', now reclassified to GRU" | Looks like an audit note, not a restriction |
| DUOS-000266 | "Previously categorized with 'other data use terms', now reclassified to GRU" | Same |

The last two are a data-entry issue @ncalvanese1 is looking at separately. If that text is cleared, both would legitimately regain the badge — and would also start being auto-approved, which they should have been all along.

All three are in DAC 1, so the blast radius is one DAC.

**The ask:** confirm you're OK with the badge being corrected on these three. Happy to walk through it if useful, but the short version is that it removes a promise the system was never keeping.

jlaw-codes:
Approve and appreciate the clear documentation, thanks!

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
```diff
@@ -433,10 +447,69 @@ public StreamingOutput indexDatasetIds(List<Integer> datasetIds) {
...
+  private boolean isInstantApprovalEligible(Dataset dataset, Set<DACAutomationRuleType> dacRules) {
+    if (!DataUsePrimaryClassifier.hasCanonicalSinglePrimary(dataset.getDataUse())) {
+      return false;
+    }
+    return Rules.implementationList.stream()
+        .filter(rule -> dacRules.contains(rule.getRuleType()))
+        .anyMatch(rule -> rule.datasetQualifies(dataset));
+  }
+
   public DatasetTerm toDatasetTerm(Dataset dataset) {
```
rushtong: This method looks like it is only used in tests now.

src/main/java/org/broadinstitute/consent/http/service/DACAutomationRuleService.java
```diff
@@ -89,26 +102,99 @@ public List<DACAutomationRule> findAllByDacId(Integer dacId) {
...
+      logException("Unable to schedule dataset reindex after DAC rule toggle", e);
+    }
+  }
+
+  /**
+   * Runs reindex passes until none is pending. Both flags are read and written under {@code
+   * reindexLock}, so a toggle cannot observe {@code reindexRunning} as true moments before this
+   * clears it and have its request dropped.
+   */
+  private void drainPendingReindexes() {
```
rushtong, line 160: Claude feedback on this method:

> Minor: drainPendingReindexes() has no outer catch — if an Error (not Exception) escapes the inner catch, reindexRunning could stay stuck true permanently, silently halting future reindexes.

rushtong (approved):
Looks good. A couple minor comments inline 👍🏽

src/main/java/org/broadinstitute/consent/http/service/ElasticSearchService.java
kevinmarete: Correct — it lost its last production caller when `indexDatasetList` moved to the batch-resolving overload. Removed, along with its "skip the query when the dataset has no DAC" branch, which was unreachable once indexing resolved once per batch. The remaining overload and `resolveEnabledRulesByDacId` are package-private now, and the tests go through a helper that mirrors `indexDatasetList`: resolve once, then apply.

src/main/java/org/broadinstitute/consent/http/service/DACAutomationRuleService.java
kevinmarete: Good catch — fixed. `reindexAllDatasets` catches `Exception`, so an `Error` escaping a pass would leave `reindexRunning` set with nothing draining it: the same failure mode as the rejected submission, silently disabling reindexing for the life of the process. Now released before the rethrow.

Added a regression test and confirmed it fails with the release removed (`Wanted but not invoked: indexDatasets`) rather than just passing alongside it.

fboulnois (approved):
👍

src/main/java/org/broadinstitute/consent/http/service/DACAutomationRuleService.java
```diff
@@ -89,26 +102,104 @@ public List<DACAutomationRule> findAllByDacId(Integer dacId) {
...
+        logWarn("Dataset reindex after DAC rule toggle terminated unexpectedly");
+      }
+    }
+  }
+
+  /**
+   * Failures are logged rather than raised: the toggle that triggered this is already committed and
+   * audited, so nothing here may fail it, and the next reindex corrects the documents either way.
+   */
+  private void reindexAllDatasets() {
```
otchet-broad: Having looked at our production database again recently, I'm reminded about the number of external entries we now have that never need to be reindexed.  I'm going to backtrack partially on the statement about reindexing all when a rule changes.  We only need to reindex just the datasets/studies that are assigned to the DAC with the rule change.  DatasetDAO supports a lookup to find the datasets associated with a DAC: findDatasetsAssociatedWithDac.  It's probably best to use that for the list of dataset IDs needed to index instead of the findAllDatasetIds because in prod, the vast majority are now not affiliated with DACs.

Rule toggles should probably be FIFO ordered so order of change is preserved.  A simple queue should suffice.

kevinmarete: Both done in 0ac8ad2f — good call on the external entries, that reasoning had gone stale under me.

Scoped to the toggled DAC via `findDatasetsAssociatedWithDac` as you suggested. Worth noting it is broader than the `findDatasetIdsByDacIds` I had used originally: it also matches datasets carrying the DAC as a `dataAccessCommitteeId` property, so property-assigned datasets are not missed. It returns full rows and its property joins can repeat a dataset, so I map to ids and dedupe. The IN list is now bounded by DAC size rather than corpus size, which also retires the objection Copilot kept raising.

FIFO queue: the single pending flag is now a `LinkedHashSet<Integer>` of DAC ids under the same lock, drained from the head. A Set rather than a list so a DAC toggled twice while queued gets one pass instead of two — the pending pass has not started and reads current state when it does. The meaningful change is that coalescing is per-DAC now rather than global: two toggles on different DACs each get their own pass.

Added a test that three DACs toggled during an in-flight pass reindex in order; confirmed it fails against a reverse-ordered queue. PR description updated to match.

otchet-broad (approved):
Awesome!  Thanks for the changes!

PR #3015, opened 2026-08-12
Revert "[DT-3865] Use database-managed sequence for dataset aliases"

kevinmarete:
Reverts DataBiosphere/consent#3009

PR #3009, opened 2026-08-10
[DT-3865] Use database-managed sequence for dataset aliases

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3865

### Summary

Replaces race-prone dataset alias allocation with a PostgreSQL sequence and rolling-deployment compatibility trigger. New application inserts omit aliases, while the database enforces non-null and unique values. Adds concurrency, migration, and rollback tests.

----
Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

PR #3016, opened 2026-08-12
[DT-3865] Restore dataset alias sequence with migration fix

kevinmarete:
## Summary

- Reapplies the database-managed dataset alias sequence from [DT-3865](https://broadworkbench.atlassian.net/browse/DT-3865) after the emergency revert in #3015.
- Explicitly casts the initial sequence value to `bigint` before calling PostgreSQL `setval`.
- Exercises the migration against a legacy `numeric` alias column, matching the dev schema.

## Why the original deployment failed

PR #3009 assumed `dataset.alias` was a `bigint`. In dev, the legacy column is `numeric`, so `MAX(alias) + 1` also produced a `numeric` value. PostgreSQL 16 could not resolve `setval(unknown, numeric, boolean)`, because the three-argument function expects a `bigint` sequence value. Liquibase therefore failed during startup and blocked the dev deployment.

We reverted #3009 through #3015 to restore deployability while correcting and validating the migration.

## What this PR does

The migration now calls `setval` with:

```sql
(COALESCE((SELECT MAX(alias) FROM dataset), 0) + 1)::bigint
```

This preserves all existing aliases, positions the sequence above the current maximum, supports old and new application instances through the compatibility trigger, and keeps aliases positive, non-null, and unique. The regression fixture uses `numeric` so this dev-specific type mismatch is covered going forward.

The failed Liquibase changeset was not recorded as applied, so the corrected changeset will run again on redeployment.

## Validation

- `./mvnw -q spotless:check`
- `xmllint --noout` on the Liquibase changelog
- Test sources compile successfully
- The PostgreSQL integration suite could not run locally because Docker Desktop was unavailable; CI will execute it

## Merge order

Depends on #3015. After #3015 merges, retarget this PR from the revert branch to `develop`.


kevinmarete:
Superseded by #3017, which is based on the actual squash-merged revert commit and therefore has the correct ancestry and full reapplication diff.

PR #3017, opened 2026-08-12
[DT-3865] Restore dataset alias sequence with migration fix

kevinmarete:
## Summary

- Reapplies the database-managed dataset alias sequence from [DT-3865](https://broadworkbench.atlassian.net/browse/DT-3865) after the emergency revert in #3015.
- Explicitly casts the initial sequence value to `bigint` before calling PostgreSQL `setval`.
- Exercises the migration against a legacy `numeric` alias column, matching the dev schema.

## Why the original deployment failed

PR #3009 assumed `dataset.alias` was a `bigint`. In dev, the legacy column is `numeric`, so `MAX(alias) + 1` also produced a `numeric` value. PostgreSQL 16 could not resolve `setval(unknown, numeric, boolean)`, because the three-argument function expects a `bigint` sequence value. Liquibase therefore failed during startup and blocked the dev deployment.

We reverted #3009 through #3015 to restore deployability while correcting and validating the migration.

## What this PR does

The migration now calls `setval` with:

```sql
(COALESCE((SELECT MAX(alias) FROM dataset), 0) + 1)::bigint
```

This preserves all existing aliases, including the valid legacy alias `0`, positions the sequence above the current maximum, and supports old and new application instances through the compatibility trigger. Newly generated aliases remain positive, while all aliases remain non-null, unique, non-negative integers within Java's 32-bit `Integer` range. The migration rejects unsafe fractional or out-of-range legacy values before casting, reserves room for the next sequence allocation, and a database `CHECK` constraint preserves that invariant afterward. The regression fixture uses `numeric` so this dev-specific type mismatch is covered going forward.

The failed Liquibase changeset was not recorded as applied, so the corrected changeset will run again on redeployment.

## Validation

- `./mvnw -q spotless:check`
- `xmllint --noout` on the Liquibase changelog
- Test sources compile successfully
- The PostgreSQL integration suite could not run locally because Docker Desktop was unavailable; CI will execute it

## Revert history

#3015 has merged. This PR is based on that revert and cleanly reapplies DT-3865 with the corrected migration.

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DatasetDAO.java
@@ -132,9 +132,9 @@ public interface DatasetDAO extends Transactional<DatasetDAO> {
       """
           INSERT INTO dataset
               (name, create_date, create_user_id, update_date,
-              update_user_id, object_id, dac_id, alias, data_use)
-          (SELECT :name, :createDate, :createUserId, :createDate,
-              :createUserId, :objectId, :dacId, COALESCE(MAX(alias),0)+1, :dataUse FROM dataset)
+              update_user_id, object_id, dac_id, data_use)
+          VALUES (:name, :createDate, :createUserId, :createDate,
+              :createUserId, :objectId, :dacId, :dataUse)
           """)
   @GetGeneratedKeys
   Integer insertDataset(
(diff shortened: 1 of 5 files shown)


rushtong (approved):
Manually tested locally against a recent dev db export 👍🏽

otchet-broad:
👍

otchet-broad (approved):

fboulnois (approved):
👍

PR #3019, opened 2026-08-13
[DT-3989] Restrict election open and cancel actions for admins

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3989

Follows [duos-ui #3852](https://github.com/DataBiosphere/duos-ui/pull/3852) (merged), which removed the Action column from the Admin DAR table.

**Security risk: low** — this change only removes authorization; no endpoint gains access.

### Summary

- Remove Admin access from election open and cancellation APIs.
- Hide election open and cancel actions from Admin collection summaries.
- Preserve Chairperson and Researcher access.
- Imply only the Researcher role when `roleName` is omitted on cancel; a Chairperson must request the role explicitly, since cancelling as a chair cancels the elections for every dataset that chair governs.
- Require a Chairperson to govern at least one of the collection's datasets before creating elections (previously a chair of an unrelated DAC got a `500` instead of a `403`).
- Reject election creation for a DAR with no dataset ids with a `400` instead of skipping the governing-DAC check (which then failed as a `500` downstream).
- Note on both endpoints that Admins escalate to the governing DAC Chairperson.
- Add authorization and regression tests.

### Notes for reviewers

- **Breaking API contract:** `Admin` is removed from the `roleName` enum on `PUT /api/collections/{id}/cancel`, and Admin callers now get `403`. No duos-ui call site passes `Admin` — `AdminManageDarCollections` passes no cancel/open handlers after #3852.
- Enforcement lives in `DarCollectionService`, with `@RolesAllowed` at the boundary and a reflection test pinning the annotation. The cancel endpoint still resolves the acting role in the resource because `roleName` selects it.
- The escalation path for the [operational risk raised on duos-ui #3852](https://github.com/DataBiosphere/duos-ui/pull/3852#pullrequestreview-4931084810) is a note on the two endpoint descriptions: Admins escalate to the governing DAC Chairperson.

----

- Verify all tests go green


otchet-broad (approved):
👍

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -796,6 +773,35 @@ public DarCollection approveDarCollection(
...
+    }
+    List<Integer> darDatasetIds = dar.getDatasetIds();
+    if (!darDatasetIds.isEmpty()) {
+      Set<Integer> governedDatasetIds =
+          Set.copyOf(datasetDAO.findDatasetIdsByDACUserId(user.getUserId()));
+      if (darDatasetIds.stream().noneMatch(governedDatasetIds::contains)) {
+        throw new ForbiddenException(CREATE_ELECTION_DAC_ERROR);
+      }
+    }
+    return dar;
```
rushtong: If the dataset ids ARE empty, we still return the DAR and skip the `governedDatasetIds` check. Is this intentional? It seems like this might be an error case. We currently require DARs to have a dataset id, but in the future, that could change if we had DARs for assets.

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
```diff
@@ -36,6 +37,11 @@
 @Path("api/collections")
 public class DarCollectionResource extends Resource {
 
+  private static final String CANCEL_ROLE_ERROR =
+      "Only chairpersons and researchers can cancel a collection";
```
rushtong: This is copied in `DarCollectionService` - maybe they can be consolidated.

rushtong (approved):
Looks good, couple questions inline. 👍🏽

fboulnois (approved):
👍

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
kevinmarete: Good catch - not intentional, it was fail-open. A chair of an unrelated DAC would have skipped the governance check here and then hit `IllegalStateException` (500) downstream, once `createElectionsForDarByUser` created nothing.

Fixed in 04ae8165: an empty dataset id list is now a `400 At least one dataset is required` (same wording `DataAccessRequestService` already uses), and the `governedDatasetIds` check runs unconditionally. So if DARs for assets ever land, this rejects rather than quietly skipping authorization.

src/main/java/org/broadinstitute/consent/http/resources/DarCollectionResource.java
kevinmarete: Consolidated in 04ae8165: the service constant is now `public` and the resource references `DarCollectionService.CANCEL_ROLE_ERROR`, following the `UserResource` -> `AcknowledgementService.DAR_CLOSEOUT_CHAIR_REF` precedent.

I kept the resource's runtime role checks themselves, since the `roleName`-omitted branch has to resolve the acting role before it can call the service.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -796,6 +774,36 @@ public DarCollection approveDarCollection(
...
+      throw new BadRequestException(
+          "DAR Collection ID: [%s] does not have any associated DAR ids"
+              .formatted(collection.getDarCollectionId()));
+    }
+    List<Integer> darDatasetIds = dar.getDatasetIds();
+    if (darDatasetIds.isEmpty()) {
+      throw new BadRequestException(CREATE_ELECTION_DATASET_ERROR);
+    }
+    Set<Integer> governedDatasetIds =
+        Set.copyOf(datasetDAO.findDatasetIdsByDACUserId(user.getUserId()));
```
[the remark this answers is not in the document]
kevinmarete: Confirmed and fixed in 282f13a0. `findDatasetIdsByDACUserId` joins `user_role` on `dac_id` without filtering on `role_id`, so a user who chairs DAC A and is only a Member of DAC B would have been treated as governing DAC B's datasets here.

The check now uses `getDatasetIdsForUserAndRoleId(user, UserRoles.CHAIRPERSON.getRoleId())`, which is what the chair summary path already uses, so the authorization check and what the chair sees in the UI agree. A chair whose role rows carry no `dacId` resolves to an empty list and gets a `403`, so it fails closed.

Added `testCreateElectionsForDarCollectionAsMemberOfGoverningDacIsForbidden` to cover the chair-of-A / member-of-B case. `findDatasetIdsByDACUserId` is untouched elsewhere - it still backs the view-permission path.

PR #2660, opened 2025-09-12
[DT-2150] Require DuosUser when Resource annotated with @PermitAll

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-2150](https://broadworkbench.atlassian.net/browse/DT-2150)

### Summary

Modifies every use of @PermitAll so that a Duos User is required.
~~Updates DuosUser class to support returning the base AuthUser class to reduce scope of impact.~~
Updates tests for methods requiring refactor
Removes mocking of userService.findByEmail.... in many places.
Removes specific tests that test resource level user not found exceptions because those are handled by the Request filter and authorization framework.

----
Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

- Label PR with a Jira ticket number and include a link to the ticket
- Label PR with a security risk modifier [no, low, medium, high]
- PR describes scope of changes
- Get a minimum of one thumbs worth of review, preferably two if enough team members are available
- Get PO sign-off for all non-trivial UI or workflow changes
- Verify all tests go green
- Test this change deployed correctly and works on dev environment after deployment

PR #3021, opened 2026-08-17
[DT-3997] Return 404 when an authenticated user has no DUOS account

kevinmarete:
### Addresses

[DT-3997](https://broadworkbench.atlassian.net/browse/DT-3997) — security risk: **low** (one endpoint changes a refusal from 401 to 404; no change to who is admitted).

### Summary

`GET /api/user/me` answered **401** for a caller with a valid token but no DUOS account, instead of the **404** it gave before the endpoint moved to `@Auth DuosUser`. Since 401 also means "rejected token", callers can't tell "register this person" from "tear down this session" — duos-ui's sign-in reads the 401 as a session failure and logs the user out before its registration fallback runs.

`/me` now takes `@Auth AuthUser` and does its own `findUserByEmail`, so a missing account produces a 404 from the resource. This is the ticket's first suggested shape and restores the pre-DT-2150 behavior exactly.

The ticket attributes the regression to DT-3788 (#2979), but `/me` moved to `@Auth DuosUser` earlier, in **DT-2150 (#2660, `4dc39f8b`)**; #2979 migrated the other `UserResource` endpoints and left `/me` alone. Worth knowing if anyone bisects for it.

### Scope

Exactly one endpoint changes. An earlier revision of this PR made the fix in `DuosUserAuthenticator`, which put the 404 on all **169 `@Auth DuosUser` endpoints across 27 resource files** — reverted after review, since that conflates "no DUOS account" with 404's existing "that resource doesn't exist" meaning across the whole authenticated API. Whether those endpoints *should* distinguish the two is a real question, left open deliberately rather than decided here.

Narrowing is sufficient, not just smaller: `/api/user/me` is the only endpoint an unregistered caller actually reaches. duos-ui's new-user path is `getMe` fails → `handleRegistration` → `POST /api/user` (on the `AuthUser` filter, untouched here) → `/tos_acceptance` → `POST api/sam/register/self/tos`, and `ToS.ts` is the only place it touches `api/sam/*`. The user exists before any Sam endpoint is called.

### Tests

`UserResourceTest.testGetMeUnregisteredUserReturnsNotFound` covers the authenticated-but-unregistered case at the resource level — possible only because the lookup moved back into the resource. `UserTests` drives the real filter and mapper against a live app: unregistered → 404, plus both routes to a 401 (token the proxy couldn't resolve to an identity, and no token at all). Full suite 4372 green, `spotless:check` clean.

The 404 body deliberately does not carry `UserService`'s `"Unable to find user with email: …"`, per `docs/API_GUIDELINES.md:17`. `api-docs.yaml` already claimed 404 for this path; added the 401 and worded both so the `headers == null` branch doesn't contradict them.

### Follow-up

`SamResource.postSelfTos` lines 99-106 catch `NotFoundException` and create the DUOS user, but that block is unreachable — the auth filter does the same lookup first and rejects the caller. Dead rather than harmful, since registration completes via `POST /api/user` before ToS acceptance. Untouched here.

duos-ui #3861 tolerates both 401 and 404 on this path; that can be tightened once every environment runs a consent build with this change.

----

- Verify all tests go green

[DT-3997]: https://broadworkbench.atlassian.net/browse/DT-3997?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/authentication/DuosUserAuthenticator.java
+++ b/src/main/java/org/broadinstitute/consent/http/authentication/DuosUserAuthenticator.java
@@ -29,7 +29,11 @@ public class DuosUserAuthenticator implements Authenticator<String, DuosUser>, C
           User duosUser = authorizationHelper.userService.findUserByEmail(authUser.getEmail());
           return Optional.of(new DuosUser(authUser, duosUser));
         } catch (NotFoundException e) {
+          // The token itself is good; only the DUOS account is missing. Returning an empty
+          // Optional here would surface as a 401, indistinguishable from a rejected token, so
+          // let the 404 propagate and keep "not registered yet" a distinct answer.
           logWarn("User not found, authentication incomplete: %s".formatted(authUser.getEmail()));
+          throw e;
         }
       }
       return Optional.empty();
--- a/src/main/java/org/broadinstitute/consent/http/authentication/DuosUserAuthenticator.java
+++ b/src/main/java/org/broadinstitute/consent/http/authentication/DuosUserAuthenticator.java
@@ -29,10 +29,7 @@ public class DuosUserAuthenticator implements Authenticator<String, DuosUser>, C
           User duosUser = authorizationHelper.userService.findUserByEmail(authUser.getEmail());
           return Optional.of(new DuosUser(authUser, duosUser));
         } catch (NotFoundException e) {
-          // Rethrow rather than return empty: an empty Optional becomes a 401, which a caller
-          // cannot tell apart from a rejected token. The token here is good, only the account.
           logWarn("User not found, authentication incomplete: %s".formatted(authUser.getEmail()));
-          throw e;
         }
       }
       return Optional.empty();
--- a/src/main/java/org/broadinstitute/consent/http/authentication/DuosUserAuthenticator.java
+++ b/src/main/java/org/broadinstitute/consent/http/authentication/DuosUserAuthenticator.java
@@ -29,9 +29,8 @@ public class DuosUserAuthenticator implements Authenticator<String, DuosUser>, C
           User duosUser = authorizationHelper.userService.findUserByEmail(authUser.getEmail());
           return Optional.of(new DuosUser(authUser, duosUser));
         } catch (NotFoundException e) {
-          // The token itself is good; only the DUOS account is missing. Returning an empty
-          // Optional here would surface as a 401, indistinguishable from a rejected token, so
-          // let the 404 propagate and keep "not registered yet" a distinct answer.
+          // Rethrow rather than return empty: an empty Optional becomes a 401, which a caller
+          // cannot tell apart from a rejected token. The token here is good, only the account.
           logWarn("User not found, authentication incomplete: %s".formatted(authUser.getEmail()));
           throw e;
         }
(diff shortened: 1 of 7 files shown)


otchet-broad:
Notes from claude:

The headline concern is altitude. The `throw e` lands in the shared `DuosUserAuthenticator`, which backs all 28 files using `@Auth DuosUser` — so the 404 isn't scoped to `/api/user/me`, it replaces the 401 on the entire authenticated API surface, where 404 already means "that resource doesn't exist." Two follow-ons from that: the api-docs change and all three new tests cover only `/me`, and the Sam registration endpoints (`POST api/sam/register/self` and friends) are themselves `@Auth DuosUser`, so the registration flow the PR is meant to unblock still dead-ends there. It works only via `POST /api/user`, which runs on the `AuthUser` filter this diff didn't touch — worth stating explicitly somewhere, since the goal depends on it.

Two other real ones: the `headers == null` path four lines down still returns 401 for what it logs as a 500, which the new api-docs 401 wording now contradicts; and `UserService`'s internal message (`Unable to find user with email: …`) becomes the public 404 body, against `docs/API_GUIDELINES.md:17`.

Notes from me:

To me this seems confusing that we're returning a 404 on all resources that use DuosUser when the user isn't in the system.  On the surface, that seems strange.  Isn't it true that whenever we can extract a token from the Oauth token, we'll create an account?  Why not redirect to the account creation endpoint?

kevinmarete:
Both of you are right about the altitude, and my PR body undersold it — I wrote "45 endpoints", which was a bad count. The real surface is **169 `@Auth DuosUser` endpoints across 27 resource files**, since the authenticator runs ahead of any role check. Reverted in `217ce18d`; `/me` now takes `@Auth AuthUser` and does its own `findUserByEmail`, which is the ticket's other suggested shape and the pre-DT-2150 behavior. One endpoint changes; the other 168 keep today's 401.

**@otchet-broad on auto-creation** — no. `AuthorizationHelper.getUserStatusInfo` auto-registers the caller with **Sam**, but nothing on the auth path inserts the DUOS row. That happens in `POST /api/user` (`createResearcher`), which runs on the `AuthUser` filter and is untouched here.

On the Sam endpoints still dead-ending registration — I checked duos-ui and that isn't the case. The new-user path is `getMe` fails → `handleRegistration` → `POST /api/user`, and only then `/tos_acceptance` → `POST api/sam/register/self/tos`. `ToS.ts` is the only place duos-ui touches `api/sam/*`. So the user exists before any Sam endpoint is called, and **`/api/user/me` is the only endpoint an unregistered caller reaches** — which is why fixing it alone is sufficient. One leftover: `postSelfTos` lines 99-106 catch `NotFoundException` and create the user, and that block can't run (the filter rejects first) and wouldn't need to anyway. Harmless dead code, flagging rather than touching it here.

A redirect wouldn't work either — the creation endpoint is a `POST` and the BFF probes with `GET`.

Other two fixed: the 404 body no longer echoes `UserService`'s `"Unable to find user with email: …"` (`docs/API_GUIDELINES.md:17`), and the 401 description now reads "missing or unusable credentials", which the `headers == null` branch no longer contradicts. The narrowing also made your first point actionable — with the lookup back in the resource, the unregistered case is now covered by a `UserResourceTest` case, which I'd argued was impossible when it lived in the filter.

Leaving the cross-cutting 401-vs-404 question for the other 168 endpoints to you rather than deciding it here.

otchet-broad (approved):

fboulnois (approved):
👍

PR #3022, opened 2026-08-17
[DT-3870] Parse and validate study-template v1 in Consent

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-3870

**Security risk: low** — this adds parsing of untrusted uploaded text, but nothing can reach it yet: the endpoint arrives in DT-3871. The 5 MiB cap is enforced against the stream before the file is buffered, decoding is strict UTF-8, no type is deserialized from the input, no database call is made, and no uploaded content reaches a log.

### Summary

- Adds `StudyTemplateValidationService`, a service below the resource layer that enforces the 5 MiB upload limit, parses the v1 CSV template, and maps it into a validated `StudyRegistrationRequest`. No endpoint, no draft persistence, no database access, and no logging of uploaded content.
- Validation runs in three stages, each stopping the next when it reports anything: file and header problems, then record-model problems, then cell conversion together with the ordinary registration validator. The second stage stops before the DTO exists because business violations derived from an untrustworthy record model are misleading rather than actionable.
- Parses into an intermediate template model (`TemplateRow` / `TemplateAssignment` / `TemplateRecord` / `ParsedStudyTemplate`) before any registration DTO is constructed, so row and column context survives into conversion.
- `StudyTemplateV1Fields` is the declarative catalogue of every supported CSV field, the wire property it maps to, and how its cells convert. Wire properties the contract keeps out of v1 are listed separately, so a producer who names `alternativeDataSharingPlanFileName` is told it belongs on the draft form rather than that it does not exist.
- Version dispatch is explicit through the `StudyTemplateParser` interface, keyed on the `templateVersion` cell, so a v2 parser can coexist with v1 rather than replace it.
- `StudyRegistrationRequestValidator` is reused unmodified. `RegistrationViolationAttributor` gives its violations the row that caused them by substitution rather than by reading its messages: one field at a time is set to a value satisfying its own rule, and whichever violations disappear belong to that field's row. The same mechanism drops a violation that is only derivative of a cell that failed conversion, so `publicVisibility,yes` reports the conversion error rather than that and "Public Visibility is required". A violation no row accounts for stays message-only.
- Promotes Apache Commons CSV from test to compile scope; parsing uses it rather than splitting lines or delimiters by hand.

Worth a reviewer's eye:

- The contract's ordering sentence reads literally as all conversion errors before all DTO violations, but `field-values.errors.json` interleaves them by row (2, 5, 6, 8, 9, then the message-only DAC violation). I matched the committed fixture: everything with a row in row order, then unlocated violations in validator order. Happy to change either side if the fixture was the accident rather than the contract.
- `accessManagement` is "required by this template contract" but has no rule in the registration validator, so the mapper enforces it. This is one place where a template submission is deliberately stricter than manual registration, where a missing `accessManagement` is treated as `controlled`.

### Tests

136 tests across two classes, covering 99.89% of instructions and 99.43% of branches in the new packages, with every method and class covered.

- `StudyTemplateValidationServiceTest` compares against all nine committed `.errors.json` manifests exactly, so error ordering and the truncation cap are pinned rather than merely present. Also parameterized dialect cases (LF, CRLF, BOM, quoted delimiters, escaped quotes, significant whitespace, blank lines), parameterized scalar conversion, the file, header, record-model and version-dispatch failure paths, and regression checks in both directions between template and direct registration validation.
- `StudyTemplateV1FieldsTest` maps every supported field through one valid template and asserts each lands on its wire property, with a completeness assertion against the catalogue so a new field cannot be added without a mapping expectation.

The single uncovered branch guards `validate` against returning an empty error list as a valid result. It is unreachable, because every path that produces no CSV records also reports why; I kept it rather than delete a guard whose failure mode would be a silent false "valid".

----

- Verify all tests go green


fboulnois (approved):
lgtm, some suggestions from Claude:

### Findings

**1. Per-consent-group violations collapse into one row-less error (medium)**
`src/main/java/org/broadinstitute/consent/http/service/studytemplate/RegistrationViolationAttributor.java:24`

`StudyRegistrationRequestValidator.collectViolations` dedupes into a `LinkedHashSet`, and attribution only fires when a probe substitution makes a violation disappear from the *whole request*. While a sibling consent group still triggers the same message, no probe ever resolves it.

Confirmed: two consent groups each with an empty `numberOfParticipants` cell produce exactly one error — `row=null, column=null, "Number of Participants is required"`. Same for `Dataset Name is required`.

For the contract's own sizing example (100 consent groups) a producer gets one message and no way to tell which datasets are broken, which cuts against "Validation collects independently actionable errors" in `docs/study-template-v1.md`.

Fix direction: validate each consent group in isolation (a single-group request per group, or index violations by group position) instead of relying on whole-request set difference.

**2. Data-use and access-management violations never carry a row (medium)**
`src/main/java/org/broadinstitute/consent/http/service/studytemplate/StudyTemplateV1Fields.java:319-356`

`generalResearchUse`, `hmb`, `poa`, `diseaseSpecificUse`, `otherPrimary`, and `accessManagement` all have `probeValue = null`, so no probe exists for them. The most likely consent-group error — the "must have exactly one primary data use" message — is always unlocated, even with a single consent group. Confirmed.

kevinmarete:
Thanks — both findings reproduced, and both are fixed in 11100fe1.

I wrote them up as failing tests first to confirm the mechanism. The root cause is shared: `collectViolations` dedupes into a `LinkedHashSet`, and the attributor located a violation by set difference over the *whole request*. A message two consent groups both produce can never disappear under any single probe substitution, because the sibling keeps producing it — so it stayed `row=null, column=null`. Finding 2 is the same failure reached from the other side: a rule spanning `generalResearchUse`/`hmb`/`poa`/`diseaseSpecificUse`/`otherPrimary`/`accessManagement` has no single field whose substitution satisfies it, so no probe could ever resolve it.

Fix, per your suggested direction: `StudyRegistrationRequestValidator` now exposes `collectStudyViolations` and `collectConsentGroupViolations(cg)`, and each scope is attributed against only its own violations. `collectViolations` is composed from the two, so its order and dedupe are byte-identical and `DatasetResource` is unaffected.

On finding 2 I did not give the data-use fields probe values, and I think the record-row fallback is the better answer. A consent-group violation that no field accounts for now carries the record's first row: with zero primary uses set there are no rows for a probe to substitute, and with two set no single-field substitution resolves the conflict — picking one of the two rows would be arbitrary. The row names the dataset without claiming one guilty cell. Happy to revisit if you'd rather have clearing probes that name one of the conflicting rows.

Now, for your 100-consent-group example, 100 broken datasets produce 100 located errors instead of one:

- two groups with an empty `numberOfParticipants` → rows 10 and 13, each on the offending cell
- controlled access with no primary data use → located on the group's record row

Two existing expectations moved because the DAC violation is now located: `field-values.errors.json` gains `"row": 10`, and `testValidate_ordersLocatedErrorsByRowThenUnlocatedViolations` was rebuilt on a template that still has a genuinely unlocated study violation (it had been relying on that fixture's last error being row-less), asserting the full error list rather than a rows-only projection. `docs/study-template-v1.md` records the per-group collection and the row semantics; study-scoped violations may still carry no row, so the error contract's "may contain only a message" stays accurate.

`Tests run: 387, Failures: 0` across the validator, update-validator, template, data-use, and `DatasetResource` suites; `spotless:check` clean.

otchet-broad:
More feedback from claude:

**Worth fixing before merge**
- `TemplateErrors:20` — the `MAX_ERRORS` cap trims the *reported* list but nothing bounds collection. A 5 MiB template (inside your own documented cap) built 476k error objects; it OOMs at `-Xmx256m` and adds 275 MB at `-Xmx1g`. `getRecords()` materializing everything compounds it. This is the one that bites once the upload endpoint is wired.
- `StudyTemplateV1Parser:384` / `CellConverter:73` — duplicate-item and unknown-value messages interpolate the raw cell, so emails and free text land in HTTP errors and logs. `StudyTemplateValidationService:139` already suppresses the commons-csv message for exactly this reason, so the rule is known, just unevenly applied.

**Error-quality cluster** (all point the producer at the wrong cell)
- Lowest-row tie-break in `RegistrationViolationAttributor` blames a *valid* `nihAnvilUse` row for a missing `dbGaPPhsID`; reordering rows changes which cell gets blamed. A test pins the current behavior, so it's a deliberate tie-break — worth revisiting rather than a slip.
- Reported `row` is the record ordinal, not the file line, so one blank line shifts every later number.
- `,,,,,` (what Excel emits for a blank row inside the used range) is treated as data and rejected, against the doc's "trailing blank lines ignored."
- A reused `fileType` recordId reports "Conflicting parentRecordId" — wrong rule, wrong column.

**Contract drift**: uppercase `HTTPS://` rejected; `INTEGER` silently accepts `+10` and `007`; and the doc's error-ordering paragraph is now stale (everything sorts by row, so DTO violations interleave with conversion errors).

kevinmarete:
Thanks — all of these reproduced, and they're fixed in e4db8ecc.

**Memory.** Confirmed before changing anything: a 5 MiB file of the shortest possible rows is 655,351 rows, and validating one throws `OutOfMemoryError` at `-Xmx192m`. The same file now completes at `-Xmx64m` in about 100 ms. Three parts to it. `TemplateErrors` bounds collection at ten times the reported cap while counting every error it was handed, so the "N further errors were omitted" number stays exact rather than being derived from the list that survived; a test drives 5,000 errors through it to pin that. `getRecords()` is gone — the file is streamed in one pass, so a large template no longer costs a parsed copy of itself alongside the row model built from it. And there is now a 50,000-row limit, because bounding bytes alone does not bound the row model: 50,000 is twenty times the largest study the contract sizes for, and a worst-case file at that limit holds in 66 MB. **That limit is a contract addition** — it's in the doc under File contract and in the invalid-input table, but flag it if you'd rather the endpoint carried it instead.

**Echoing cells.** No message quotes a `value` cell now: the duplicate-item message drops the item, and the unknown-enum message drops the value. I kept the structural columns (`templateVersion`, `recordType`, `recordId`, `parentRecordId`, `field`) in messages, since a producer who typed `studyColour` needs to see which token was rejected, and wrote that boundary into the doc so the next field added follows it. One thing I deliberately did not touch: `StudyRegistrationRequestValidator`'s own messages still echo, e.g. `Data Custodian Email is not a valid email address: <cell>`. Those are shared with the manual submission path, so changing them is a separate decision — happy to take it on if you want it.

**The nihAnvilUse tie-break.** You were right that the lowest-row rule was the wrong rule rather than a slip. Probes now carry what the field held when they were built, and attribution reads that: an empty cell outranks a filled one, so `dbGaP phs ID is required` lands on the empty `dbGaPPhsID` cell rather than on the valid choice that required it, whatever order the rows are in. For a field with no cell at all there is nothing honest to point at, so I added probes for the fields a record never mentions — they carry no row and never report one, they only establish that the violation belongs to the absent field, which stops a valid cell being blamed. In your example, `Principal Investigator Institution is required` and `NIH Grant or Contract Number is required` now carry no row instead of pointing at row 6. The test that pinned the old behaviour is rewritten around the new intent rather than deleted.

**Rows and blank lines.** A reported row is now the physical line the record starts on, so a blank line no longer shifts every number after it, and a multi-line quoted value reports the line it begins on. That is also what let `,,,,,` be ignored the same way a blank line is — the parser now keeps blank lines so line numbers stay true, and blank records are dropped explicitly. A file of nothing but empty cells reads as empty.

**Reused recordId.** It reports the uniqueness rule on the `recordId` column now: `Duplicate fileType recordId 'ft-one': each recordId identifies one record and must name one parentRecordId`.

**Contract drift.** `HTTPS://` is accepted — the scheme is matched case-insensitively, as URI syntax defines it. `+10` is rejected. I kept `007`: the contract says "base-10 digits with an optional leading minus sign", which `007` satisfies and `+10` does not, and rejecting a leading zero costs a producer something without protecting anything. Say so and I'll tighten both the converter and the doc. The error-ordering paragraph now describes what the service actually does: everything carrying a row sorts by row, so violations interleave with conversion errors rather than following them.

**Verification.** `Tests run: 344, Failures: 0` across the template, fields, fixtures, registration-validator, and `DatasetResource` suites; `spotless:check` clean; JaCoCo shows no missed lines or branches in `service/studytemplate` (the split of the CSV read path into a scan and a report step was partly to keep the cognitive complexity of one method well under Sonar's threshold).

One thing I could not act on: the comment mentions 13 findings ranked above it, but only these came through in the text. If the rest exist somewhere, send them and I'll work through them.

otchet-broad (approved):
Thanks for the updates!

PR #3024, opened 2026-08-18
[DT-3982] Serve data use groups on the DAR collection summary

kevinmarete:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3982](https://broadworkbench.atlassian.net/browse/DT-3982)

Security risk: no

### Summary

The Data Access Requests table renders one row per data-use group. Until now each console derived those groups in the browser: per visible collection a `getCollectionById`, a match lookup and a dataset-terms search, so roughly thirty requests to fill a page of ten. Everything needed is already in Postgres, so `getSummariesForRole` now attaches the groups and the console renders what it is given.

Work is batched across the whole response rather than per collection: one `findDatasetsByIdList`, one member-vote query, and one data use translation per **distinct** data use. That last part is the point — translation reads the ontology store for disease restrictions, so translating per dataset would have moved the N+1 into the service rather than removed it. A DAC's datasets share very few distinct data uses (19 distinct across 275 groups in a local run), so this collapses to a handful of reads.

Groups carry the translated `DataUseSummary` rather than a flat label, so the UI renders them with the same helper the Data Library grid uses and the two cannot drift apart. `DataUseSummary` and `DataUseTerm` needed `@Expose` to survive the list-context Gson the collection resources serialize with.

Member votes are attached only for Chairperson and Member. Researchers and Signing Officials get the same groups with an empty vote list, and the vote query is not issued at all for those roles — DAC voting is not theirs to see. `VoteDAO` gains `findDacVotesWithNamesByElectionIds`, with a mapper that adds the voting member's name, because the interface-level `VoteMapper` builds a `Vote` without one.

Additive only: no existing response field or behaviour changes, and `binCollectionToBuckets` is untouched, so the DAR review page and DUC addendum are unaffected.

### API change

`GET /api/collections/role/{roleName}/summary` gains one field on each `DarCollectionSummary`: `dataUseGroups`. Every other field is unchanged, and no other endpoint is affected — `/summary/{collectionId}` and `GET /api/collections/{id}` do not carry groups.

Each entry is one data-use group within the collection: the datasets that share a data use, that data use translated, and (Chair/Member only) the DAC's votes on it. `key` is stable within a collection and is what the table uses for its row id.

```jsonc
// GET /api/collections/role/Chairperson/summary   (values below are illustrative)
[
  {
    "darCollectionId": 855,
    "darCode": "DAR-855",
    "status": "Submitted",
    "datasetCount": 3,
    // ... all existing fields unchanged ...
    "dataUseGroups": [
      {
        "key": "bucket-2",
        "dataUse": {
          "primary": [
            { "code": "DS",   "description": "Data use is limited for studying: cancerophobia, diabetes mellitus [DS]" },
            { "code": "NPOA", "description": "Future use for population origins or ancestry research is prohibited. [POA]" }
          ],
          "secondary": [
            { "code": "POP-F",  "description": "Data use is limited to research on females. [RS-FM]" },
            { "code": "POP-PD", "description": "Data use is limited to pediatric research. [RS-PD]" }
          ]
        },
        "datasets": [
          { "datasetId": 2, "name": "Example Cohort Study", "datasetIdentifier": "DUOS-000002" }
        ],
        "votes": [
          { "userId": 101, "vote": true,  "displayName": "Member One" },
          { "userId": 102, "vote": false, "displayName": "Member Two" },
          { "userId": 103,                "displayName": "Member Three" }   // no vote yet: still pending
        ]
      },
      {
        "key": "bucket-7-9",
        "dataUse": { "primary": [ { "code": "HMB", "description": "Data is limited for health/medical/biomedical research. [HMB]" } ] },
        "datasets": [
          { "datasetId": 7, "name": "Example Registry",  "datasetIdentifier": "DUOS-000007" },
          { "datasetId": 9, "name": "Example Biobank",   "datasetIdentifier": "DUOS-000009" }
        ],
        "votes": []
      }
    ]
  }
]
```

Notes on the shape:

- `vote` is absent when the member has not voted yet, so a pending vote is never read as a denial. Present and `false` means denied.
- `votes` is always empty for Admin, Signing Official and Researcher, and the vote query is not issued for those roles at all.
- `dataUse` is the translated `DataUseSummary`, the same shape the dataset search index stores, so the UI renders it with the helper the Data Library grid already uses. It is null when a group's datasets carry no data use.
- A collection with no datasets visible to the caller returns an empty `dataUseGroups` array. The field is omitted entirely when no collection in the response has any visible datasets.

### Notes for review

- **The duos-ui side depends on this being deployed first.** [duos-ui#3858](https://github.com/DataBiosphere/duos-ui/pull/3858) reads `dataUseGroups`; if it ships first the Data Use, Datasets and Votes columns render empty across all six consoles.
- The endpoint returns every collection for the role and groups all of them, while the table shows ten at a time. Measured locally on 227 collections: the endpoint goes from ~0.91s to ~1.18s and the response from 123KB to 303KB, which gzips to 17.7KB → 27.0KB on the wire. The pre-existing lack of server-side pagination on this endpoint dominates that cost; narrowing it would mean pushing search and sort server-side for all six consoles, which is out of scope here.
- Verified end to end against a local stack with the seeded database: 227 collections all carrying groups, correct disease-restriction labels, member names on votes, and zero votes returned for the Researcher role.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
@@ -504,11 +504,7 @@ public class DarCollectionService implements ConsentLogger {
         .map(
             entry -> {
               List<Dataset> datasets = entry.getValue();
-              String key =
-                  "bucket-"
-                      + datasets.stream()
-                          .map(d -> String.valueOf(d.getDatasetId()))
-                          .collect(Collectors.joining("-"));
+              List<Integer> key = datasets.stream().map(Dataset::getDatasetId).toList();
               DataUseSummary dataUse =
                   summariesByDataUse.computeIfAbsent(
                       entry.getKey(),
--- a/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
@@ -17,9 +17,11 @@ import java.sql.SQLException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -56,15 +58,19 @@ import org.broadinstitute.consent.http.models.DarCollection;
 import org.broadinstitute.consent.http.models.DarCollectionSummary;
 import org.broadinstitute.consent.http.models.DataAccessRequest;
 import org.broadinstitute.consent.http.models.DataAccessRequestData;
+import org.broadinstitute.consent.http.models.DataUseGroup;
 import org.broadinstitute.consent.http.models.Dataset;
 import org.broadinstitute.consent.http.models.Election;
 import org.broadinstitute.consent.http.models.User;
 import org.broadinstitute.consent.http.models.UserRole;
 import org.broadinstitute.consent.http.models.Vote;
+import org.broadinstitute.consent.http.models.ontology.DataUseSummary;
+import org.broadinstitute.consent.http.models.ontology.DataUseTerm;
 import org.broadinstitute.consent.http.rules.DACAutomationRule;
 import org.broadinstitute.consent.http.rules.DACAutomationRuleType;
 import org.broadinstitute.consent.http.service.dao.DarCollectionServiceDAO;
 import org.broadinstitute.consent.http.util.ConsentLogger;
+import org.broadinstitute.consent.http.util.gson.GsonUtil;
 import org.glassfish.jersey.server.ContainerRequest;
 import org.jdbi.v3.core.Jdbi;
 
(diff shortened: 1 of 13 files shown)


src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
```diff
@@ -424,9 +432,131 @@ public List<DarCollectionSummary> getSummariesForRole(User user, UserRoles role)
...
+            dataset ->
+                datasetsByDataUse
+                    .computeIfAbsent(dataUseKey(dataset), k -> new ArrayList<>())
+                    .add(dataset));
+
+    return datasetsByDataUse.entrySet().stream()
+        .map(
+            entry -> {
+              List<Dataset> datasets = entry.getValue();
+              String key =
```
otchet-broad: Why not make the key a just an array of the dataset IDs that the DataUseSummary applies to?  This way the front end doesn't need to do any fancy parsing, it's just a list of dataset Ids returned?

otchet-broad (approved):
Looks good.  One minor suggestion to change the key.

src/main/java/org/broadinstitute/consent/http/service/DarCollectionService.java
kevinmarete: Done — the key is now the group's dataset IDs, so it serializes as `[1, 2]` with no parsing on the front end.

It is exactly `datasets[*].datasetId`, so I did consider dropping it and deriving it in the UI, but the grid needs a stable row identity per (collection, group) and that is what the key is for — better to keep it server-side than have each consumer re-implement the join. duos-ui #3858 is updated to match: typed as `number[]`, and the row ID joins the IDs explicitly rather than relying on array-to-string coercion.

fboulnois (approved):
👍

PR #3026, opened 2026-08-19
Adds a plan to address remove gating DAR/access from library cards.

otchet-broad:
### Addresses

[DT-3308](https://broadworkbench.atlassian.net/browse/DT-3308)

### Summary

Builds a plan to address removing the gating of DAR and Access from Library Cards by relying on the newly created researcher activation.

----

- Verify all tests go green

[DT-3308]: https://broadworkbench.atlassian.net/browse/DT-3308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/plans/README.md
+++ b/docs/plans/README.md
@@ -13,3 +13,4 @@ Use this directory for documents that describe a proposed path forward, migratio
 | `dataset-registration-schema-migration-plan.md` | Plan for migrating dataset/study registration away from `dataset-registration-schema_v1.json` backend validation. |
 | `elasticsearch-service-duos-ui-usage.md` | Plan for adding access control to the information in consent's Elasticsearch infrastructure. |
 | `es-security-capability-record.md` | Ticket A-1 record: measured Elasticsearch security capabilities (DLS/FLS/API keys/`run_as`) per environment, and the resulting Epic D / Epic E decision. |
+| `researcher-status-decoupling-plan.md` | Plan for introducing a persisted `users.researcher_status` flag so researcher eligibility no longer keys off Library Card presence, leaving the Library Card as the DAA pre-authorization container. |
--- a/docs/plans/researcher-status-decoupling-plan.md
+++ b/docs/plans/researcher-status-decoupling-plan.md
@@ -7,7 +7,12 @@ that found four unclassified Library Card gates, a scope boundary error, and sev
 mistakes in the draft.
 
 Revised a second time after an adversarial review of this plan found fifteen defects in it, all of
-which were confirmed against source and are now resolved in the body below. See
+which were confirmed against source and are now resolved in the body below.
+
+Revised a third time after a review found two further defects in the rollout mechanism: the
+single-flag-row cutover was not actually atomic across backend and already-loaded browser clients, and
+the Phase-3 reconciliation could not honestly reconstruct audit provenance for a hard-deleted Library
+Card. Both are resolved in [Rollout and Compatibility](#rollout-and-compatibility). See
 [Review-Driven Revisions](#review-driven-revisions) for the finding-to-section map.
 
 ## Changes from the Ticket
@@ -59,12 +64,13 @@ re-listed by line because the line numbers drift and the sweep is the guarantee 
 eligibility pool, "no renames in this change", and leaving the contractual Library Card Agreement
 text and PDF filenames alone all carry over exactly as the ticket specified.
 
-**Added after review — four items in neither the ticket nor the first revision.** Activation when
-registration links a pre-issued Library Card; a bidirectional status/card reconciliation run in the
-Phase-3 flip window; an admin researcher-status control on `AdminEditUser`, because after Phase 3
-deleting a Library Card no longer deactivates anyone; and alignment of the status endpoint's Signing
-Official scoping and failure code with the sibling SO-scoped endpoint on the same resource. Each is
-marked as a decision in its own section so a reviewer can veto it.
+**Added after review — five items in neither the ticket nor the first revision.** Activation when
+registration links a pre-issued Library Card; flag-gated dual-write of researcher status on every
+Library Card mutation path through Phases 1-2, plus a flag echo that makes stale browser clients fail
+loudly instead of skewing silently; an admin researcher-status control on `AdminEditUser`, because
+after Phase 3 deleting a Library Card no longer deactivates anyone; and alignment of the status
+endpoint's Signing Official scoping and failure code with the sibling SO-scoped endpoint on the same
+resource. Each is marked as a decision in its own section so a reviewer can veto it.
 
 ## Review-Driven Revisions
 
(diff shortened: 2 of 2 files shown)


kevinmarete (commented):
Excellent work, here are some findings:

1. **P1 — The feature-flag flip is not atomic across backend and browser clients.**  
   The plan says both sides reading one database flag makes the transition atomic, but an already-loaded UI can retain the disabled/old behavior while the backend begins enforcing researcher status. That UI could delete a Library Card intending to deactivate someone, while the backend leaves their status active. There is also an unprotected interval between reconciliation and flag enablement where Library Card writes can introduce fresh skew. See [lines 442–501](https://github.com/DataBiosphere/consent/blob/46846d57f6483f2d5fd270252eb0019489e56103/docs/plans/researcher-status-decoupling-plan.md#L442-L501).  
   The plan needs a concrete synchronization mechanism: dual-writing during compatibility phases, blocking/versioning legacy mutations, or a write freeze/transactional cutover with a client-drain strategy.

2. **P1 — Reconciliation cannot truthfully reconstruct audit provenance after card deletion.**  
   The plan assigns `SIGNING_OFFICIAL` to reconciliation changes, claiming they represent an SO’s earlier action. But Library Cards are hard-deleted, so the deactivation query has neither the actor nor original action time. Deletion can also originate from admin or institution-enforcement paths, making `SIGNING_OFFICIAL` factually wrong. See [lines 476–501](https://github.com/DataBiosphere/consent/blob/46846d57f6483f2d5fd270252eb0019489e56103/docs/plans/researcher-status-decoupling-plan.md#L476-L501).  
   Prefer dual-writing status/audit during Phases 1–2. Otherwise introduce an honest `MIGRATION`/`RECONCILIATION` source and define actor/timestamp semantics.

rushtong:
Some codex feedback:

[P1] Collateral Library Card deletions break the zero-skew guarantee. The plan deliberately deactivates only the named user even though deleteAllLibraryCardsByUser also deletes cards they created or updated for other researchers. While the flag is off, those researchers remain researcher_status=true without a card, contradicting both legacy semantics and the required empty pre-flip skew query. Capture deleted cards with DELETE … RETURNING user_id and deactivate every affected researcher transactionally while the flag is off, or narrow the delete. [Plan lines 545–554](https://github.com/DataBiosphere/consent/blob/021a7806ff233a968da33bbe531156ed7ac8c318/docs/plans/researcher-status-decoupling-plan.md#L545-L554)

[P1] The status endpoint is operational before status is authoritative. During Phase 1, an authorized API caller can deactivate a researcher and receive a successful response, but the Library Card gates remain authoritative, so the researcher retains access. “No UI exposes it” does not protect a documented endpoint from Swagger, scripts, or direct calls. Reject status changes while the flag is off, or deploy the endpoint only at cutover. [Plan lines 594–597](https://github.com/DataBiosphere/consent/blob/021a7806ff233a968da33bbe531156ed7ac8c318/docs/plans/researcher-status-decoupling-plan.md#L594-L597), [637–640](https://github.com/DataBiosphere/consent/blob/021a7806ff233a968da33bbe531156ed7ac8c318/docs/plans/researcher-status-decoupling-plan.md#L637-L640)

[P1] Removing the flag echo reopens the stale-client vulnerability. Browser tabs can retain the pre-flip bundle indefinitely. Removing the backend header check during Phase-3 cleanup lets such a tab delete a card while leaving the researcher active—the exact failure the echo was introduced to prevent. Removal needs an enforceable client-drain condition, such as forced reload/version rejection or bounded session expiration; otherwise retain a compatibility/version guard. [Plan lines 573–577](https://github.com/DataBiosphere/consent/blob/021a7806ff233a968da33bbe531156ed7ac8c318/docs/plans/researcher-status-decoupling-plan.md#L573-L577)

kevinmarete (commented):
I agree with the three issues raised by Greg. One additional detail: the flag-echo design adds `X-Researcher-Status-Gating` and a `409` response to `POST /api/libraryCards` and `DELETE /api/libraryCards/{id}`. Please add updating `src/main/resources/assets/api-docs.yaml` to ticket 5a, the test matrix, and the Definition of Done.

kevinmarete (commented):
Thanks for the new changes, I am seeing these new findings kindly review them:

1. **P1 — “One Phase-1 release” does not close the general rolling-deployment gap.**  
   The migration/backfill must complete before new dual-write code can run, while old application instances may continue serving ordinary Library Card creates and deletes without dual-writing. The plan recognizes this gap for redaction, but not for normal card mutations. Those operations can still create both skew populations after backfill and before old instances drain. See [lines 1465–1485](https://github.com/DataBiosphere/consent/blob/3a8b121e6ca61ef451939baf2b16ea862b0914f3/docs/plans/researcher-status-decoupling-plan.md#L1465-L1485).

   The rollout should require a post-drain reconciliation/remediation for ordinary card mutations—not merely an optional “if proceeding” repair—or separate schema creation, dual-write deployment, and backfill into safe stages.

2. **P1/P2 — The permanent header is a breaking API change without the repository-required consumer migration process.**  
   After the flip, every existing script or API client that omits the new header receives 409. Calling this “documentation, not breakage” is inaccurate. See [lines 1299–1309](https://github.com/DataBiosphere/consent/blob/3a8b121e6ca61ef451939baf2b16ea862b0914f3/docs/plans/researcher-status-decoupling-plan.md#L1299-L1309). The repository requires Comms review and advance notification to API consumers for breaking changes ([CONTRIBUTING.md](https://github.com/DataBiosphere/consent/blob/develop/CONTRIBUTING.md#L54-L65)).

   Add consumer inventory/telemetry, migration instructions, Comms sign-off, and advance notification as explicit rollout prerequisites.

3. **P2 — The test matrix contradicts the implementation section.**  
   The implementation correctly says the already-linked-card branch cannot be tested by redacting and re-registering because redaction now deletes the card ([lines 674–676](https://github.com/DataBiosphere/consent/blob/3a8b121e6ca61ef451939baf2b16ea862b0914f3/docs/plans/researcher-status-decoupling-plan.md#L674-L676)). The test matrix nevertheless requires that exact sequence ([lines 1593–1597](https://github.com/DataBiosphere/consent/blob/3a8b121e6ca61ef451939baf2b16ea862b0914f3/docs/plans/researcher-status-decoupling-plan.md#L1593-L1597)). It should instead construct the legacy already-linked state directly.

rushtong (approved):
TBH, these plan details are too complex for me to fully grok. The overall concept is clear, and I am guessing that the implementation steps will be more comprehensible. I'm approving despite this Codex feedback:

> **[P1] Serialize every status transition on the user row**
> 
> The new reconciliation endpoint locks its targets, but the shared transition helper is also called by the normal status endpoint, redaction, and Library Card dual-write paths, where no user-row serialization is required.
> 
> Concurrent transitions can therefore make decisions using the same stale status. This can produce duplicate or incorrectly ordered audit rows. More seriously, activation could race with redaction and leave a redacted user active.
> 
> Require the transition helper itself to serialize changes by either:
>  - locking the user row with SELECT ... FOR UPDATE before checking the current status; or
>  - using an atomic conditional update that changes the status only when it differs and reports whether a transition occurred.
> 
> Add concurrency tests covering:
> - competing activation and deactivation requests;
> - concurrent duplicate activations;
> - activation racing with redaction;
> - activation or deactivation racing with Library Card dual-write.

kevinmarete (approved):
Approving the overall plan. I agree with Greg’s concurrency concern: implementation should serialize every researcher-status transition on the target user and re-check eligibility/redaction state after acquiring the lock. Please also include the suggested concurrency tests so duplicate, competing, and redaction-racing transitions cannot produce incorrect status or audit history.

PR #3027, opened 2026-08-19
[DT-3960] Update error text on SO and IT director submission to match new FE language.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-3960](https://broadworkbench.atlassian.net/browse/DT-3960)

### Summary

The ticket updates the frontend to include specific, helpful text to communicate that a signing official can not be both the submitter (PI) and the SO on the same DAR.  Consent also provides an error for this that is displayed on DAR submission when the case arrises.  This ticket updates the consent error (and adjacent IT director error message) to match the language now displayed in the UI.

No tests required modification.  Coverage exists, but only confirms the exception was thrown, not the error text.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
+++ b/src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
@@ -678,12 +678,12 @@ public class DataAccessRequestService implements ConsentLogger {
 
     if (piEmail.equalsIgnoreCase(soEmail)) {
       throw new IllegalArgumentException(
-          "Principal Investigator email cannot be the same as Signing Official email");
+          "You cannot be both the requester and Principal Investigator and the Signing Official on the same Data Access Request.");
     }
 
     if (piEmail.equalsIgnoreCase(itEmail)) {
       throw new IllegalArgumentException(
-          "Principal Investigator email cannot be the same as IT Director email");
+          "You cannot be both the requester and Principal Investigator and the IT Director on the same Data Access Request.");
     }
   }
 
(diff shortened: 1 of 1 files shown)


kevinmarete (approved):
Looks good, just handle copilot feedback.

otchet-broad:
> Looks good, just handle copilot feedback.

That's a copy/paste error.  Good catch.  I'm inclined to not add tests for the specific test because it's not the control, it's just a side effect of it.  Make sense?

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -678,12 +678,12 @@ public void validateNoKeyPersonnelDuplicates(DataAccessRequestData darData) {
 
     if (piEmail.equalsIgnoreCase(soEmail)) {
       throw new IllegalArgumentException(
-          "Principal Investigator email cannot be the same as Signing Official email");
+          "You cannot be both the requester and Principal Investigator and the Signing Official on the same Data Access Request.");
```
otchet-broad:
```suggestion
          "You cannot be both the requester and the Signing Official on the same Data Access Request.");
```

src/main/java/org/broadinstitute/consent/http/service/DataAccessRequestService.java
```diff
@@ -678,12 +678,12 @@ public void validateNoKeyPersonnelDuplicates(DataAccessRequestData darData) {
...
     }
 
     if (piEmail.equalsIgnoreCase(itEmail)) {
       throw new IllegalArgumentException(
-          "Principal Investigator email cannot be the same as IT Director email");
+          "You cannot be both the requester and Principal Investigator and the IT Director on the same Data Access Request.");
```
otchet-broad:
```suggestion
          "You cannot be both the requester and the IT Director on the same Data Access Request.");
```

fboulnois (approved):
👍

PR #3030, opened 2026-08-20
[DT-4006] Query fix, tests, and future item.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-4006](https://broadworkbench.atlassian.net/browse/DT-4006)

### Summary

Fixes a cross-DAR data leak in `DataAccessRequestDAO#findDatasetApprovalsByDar` and adds
regression coverage, plus a documentation warning on a related query in `ElectionDAO`.

#### The bug

The window function that picks the last final vote for a dar-dataset pair was partitioned by
`e.dataset_id` alone:

```sql
PARTITION BY e.dataset_id
```

Because `reference_id` was missing from the partition key, all final votes on a dataset — across
*every* DAR in the system — landed in one partition, and `LAST_VALUE` returned whichever vote was
cast most recently anywhere. The result: one researcher's denial silently revoked another
researcher's correctly granted approval, and conversely an unrelated approval could make a denied
DAR look approved. The impact grows with dataset popularity, since every new election on a shared
dataset overwrites the outcome for all prior DARs on it.

#### The fix

```sql
PARTITION BY e.reference_id, e.dataset_id
```

Each dar-dataset pair now gets its own partition, so multi-election histories on a single request
still resolve correctly (the original intent of the `LAST_VALUE` logic) while votes on other DARs
are excluded. The Javadoc on the query was updated to explain why the reference id must be part
of the key.

#### Tests

Three new integration tests in `DataAccessRequestDAOTest`, each failing before the change:

- `..._OtherDarDenialDoesNotRevokeApproval` — a later denial on a different DAR must not strip an
  existing approval.
- `..._OtherDarApprovalDoesNotGrantApproval` — the inverse; a later approval elsewhere must not
  confer access on a denied DAR.
- `..._MultipleDatasetsSurviveOtherDarDenial` — reproduces the production shape, where a DAR
  approved on two datasets returned only one id after an unrelated denial on the shared dataset.

### Follow-up item (documentation only, no behavior change)

`ElectionDAO#findLastElectionsByReferenceIdsAndType` has the same partition-key shape
(`PARTITION BY e.reference_id, e.election_type`, no `dataset_id`), which collapses a multi-dataset
DAR to a single election. That is currently harmless: its only caller,
`DarCollectionService#cancelDarCollectionAsResearcher`, uses the result as an existence check, and
dropping rows cannot change whether the list is empty. Added Javadoc and an inline SQL comment so
the next caller that counts or iterates these results knows to add `dataset_id` to the partition
key first.

----

- Verify all tests go green

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
@@ -193,7 +193,9 @@ public interface DataAccessRequestDAO extends Transactional<DataAccessRequestDAO
    * LAST vote in the group of final votes for all data access elections. We need to group them due
    * to the case of multiple elections on a dar-dataset request. Election 1 may have been denied.
    * Election 2 may have been approved. Election 3 may have been denied again. When we partition
-   * over the election reference id, we'll get all final votes. The `LAST_VALUE` function selects
+   * over the election reference id and dataset id, we'll get all final votes for that one
+   * dar-dataset pair; the reference id must be part of the partition key so votes cast on other
+   * DARs for the same dataset cannot influence this DAR's result. The `LAST_VALUE` function selects
    * the last result in the partition, which would be `FALSE` in the example above. Outside the
    * JOIN, we filter on groupings where the final vote value is `TRUE` so the denied election in the
    * example would be filtered out.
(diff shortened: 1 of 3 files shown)


kevinmarete (approved):
👍🏾

fboulnois (approved):
👍

PR #3031, opened 2026-08-24
[DT-4013] Fix native thread leak in HttpClientUtil

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-4013

## Summary

Fixes the burst of `java.lang.OutOfMemoryError: unable to create native thread` errors seen in production ([DT-4013](https://broadworkbench.atlassian.net/browse/DT-4013)).

`HttpClientUtil.getHttpResponse` created a new `ScheduledThreadPoolExecutor` on every call and never shut it down. Core threads in these pools never terminate, so each call leaked at least one native thread. Status probes run this method through the health checks, so the leak grew steadily. The process eventually hit its thread limit, every `Thread.start()` failed, and the errors continued until the pod restarted.

## Changes

- Create one shared timeout executor in the constructor. Its daemon threads are named `http-client-timeout` for thread dumps.
- Cancel each scheduled timeout task once its request completes, so stale tasks do not sit in the queue.
- Add a regression test that runs 25 requests and asserts the timeout thread count does not grow past the configured pool size.

The fix keeps the existing cancel-based timeout design. It caps total request time, and callers depend on the `RequestFailedException` that cancellation produces.

## Testing

- `HttpClientUtilTest`: 5 tests pass, including the new leak regression test.
- To confirm in production after deploy: pod thread-count metrics must stay flat instead of growing steadily.

[DT-4013]: https://broadworkbench.atlassian.net/browse/DT-4013?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/util/HttpClientUtil.java
+++ b/src/main/java/org/broadinstitute/consent/http/util/HttpClientUtil.java
@@ -23,10 +23,11 @@ import jakarta.ws.rs.core.MediaType;
 import java.io.IOException;
 import java.net.URI;
 import java.nio.charset.Charset;
-import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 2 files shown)


otchet-broad (approved):
👍

fboulnois (approved):
👍

PR #3032, opened 2026-08-24
[DT-4011] Map SamAzureB2CException to 409 on GET /api/user/me

rushtong:
## Addresses 
https://broadworkbench.atlassian.net/browse/DT-4011

## Summary

Sam rejects users whose `azureB2cId` cannot be updated (they previously signed in with a different provider). `GET /api/user/me` responded with 500 for these users, or 404 when they had no entry in Consent. Both answers send the client down a dead end: the 500 aborts sign-in with a generic message, and the 404 sends the user into a registration that cannot succeed.

## Changes

- Register `SamAzureB2CException` in `Resource.DISPATCH` as a 409 Conflict instead of the fallback 500.
- Reuse the actionable "different authentication provider" message with the Terra support link on the combinedState path. Both Sam paths now build the message from one method, `SamDAO.getAzureB2CErrorMessage`.
- Skip the async registration post when Sam reports the B2C conflict. That post always answered 409 and only added log noise.
- Probe Sam before answering 404 for an unregistered caller when the auth filter got no Sam status. The conflict then answers 409 instead of 404. A genuinely new user costs no extra Sam call: the auth filter registers them and supplies a status.
- Sonar cleanups in touched code: two S3457 format-string warnings in `SamDAO.getErrorMessage`, and the S1141 nested try in `UserResource.getUser`.

`/me` now has three outcomes: 200 registered, 404 start registration, 409 provider conflict.

## Testing

- New tests: unregistered caller with a B2C conflict answers 409 with no registration post; unregistered caller with a healthy auth-filter status answers 404 with no second Sam call.
- Updated tests: the two B2C `/me` tests now assert 409; the SamDAO B2C test asserts the actionable message and support link.
- Verified locally against dsde-dev Sam with a conflicted user.

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/SamDAO.java
@@ -124,10 +124,7 @@ public class SamDAO implements ConsentLogger {
       String body = response.parseAsString();
       if (!response.isSuccessStatusCode()) {
         if (body.toLowerCase().contains("cannot update azureb2cid for user")) {
-          throw new SamAzureB2CException(
-              String.format(
-                  "AzureB2C authentication Error for user %s. Please contact support for help with this error.",
-                  authUser.getEmail()));
+          throw new SamAzureB2CException(getAzureB2CErrorMessage(authUser.getEmail()));
         }
         String errorMsg =
             String.format(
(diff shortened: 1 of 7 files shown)


kevinmarete (approved):

fboulnois (approved):
👍

PR #3036, opened 2026-08-26
[DT-4030] Stop double-unescaping DAR data on read

kevinmarete:
### Addresses

[DT-4030](https://broadworkbench.atlassian.net/browse/DT-4030). Surfaced by the [DT-3863](https://broadworkbench.atlassian.net/browse/DT-3863) recompute, not caused by it. Supersedes #3035, which GitHub never dispatched any Actions run for.

Security risk: low.

### Summary

`RowMapperHelper.translate` unescaped the JSON string Postgres returns before parsing it. `data_access_request.data` is `jsonb`, so that string is already valid JSON — unescaping it turned an escaped quote in free text into a bare one, ending the string early and leaving the object unterminated.

This parses the column as Postgres returns it. The write path needed no change: jdbi's `JsonArgumentFactory` bound to `::jsonb` already stores clean JSON, so the unescape could only ever damage data.

### Confirmed against prod

- **One DAR is unreadable.** `5b3b585fc3da7a0001352d53` fails with `MalformedJsonException: Unterminated object at line 3 column 292 path $.rus`. That single record also makes `GET /api/dar/v2` return 400 for every admin, and is why dataset 1 was the one failure when DT-3863's recompute ran.
- **One DAR is silently altered.** `c02b1671-84c6-43bb-aae7-29f25125df8f` stores a literal `\t` in three collaborator fields, which `unescapeJava` decoded to a tab on every read. The backslash is still in the database, so it has never been written back — but a save following a corrupted read would have persisted the tab. Read-only damage today, data loss on the next edit.
- **The stored data was never at fault.** The unreadable row's `rus` is valid `jsonb` holding a quote and a newline, and no row in the table carries the double-escape fingerprint (`\"` surviving into a decoded value).

### Review notes

- The legacy repair is removed rather than kept behind a fallback. No double-escaped rows survive, so no compensating migration is owed, and `jsonb` guarantees the raw value parses — a fallback would be dead code.
- The carried-over `NullPointerException` catch is dropped as unreachable: Gson returns null for a JSON null, and both `validateOntologyEntries` and `getOntologies` null-guard.

No API contract change, so `api-docs.yaml` is untouched.

### After deployment

Merging deploys to dev only; staging and prod are separate Sherlock promotions.

**dev and staging — one call each:** `POST /api/datause/legacy/recomputeMatches`. It re-reads every abstaining dataset reachable by a DAR, so `run.failed` measures the fix directly: 10 → 0 in dev, 18 → 0 in staging.

The dev figure is measured rather than predicted. Run against a dev database snapshot, same snapshot and same call with only the read path differing: `develop` returned `failed: 10` on the same ten dataset ids live dev reports, the fix returned `failed: 0` and rebuilt 510 DARs against 435. A remaining failure has a different cause and is worth reporting rather than retrying — this class never recovers on a rerun.

**prod — DT-3863 step 4:** thirteen of dataset 1's fourteen DARs have already been reprocessed under the corrected `SINGLE(HMB)`. The fourteenth, `5b3b585fc3da7a0001352d53`, returns 400 until this ships — afterwards a single `POST /api/match/reprocess/purpose/5b3b585fc3da7a0001352d53` completes step 4. The rebuilt matches should come back non-abstaining; V5 abstained on the old `MULTIPLE(HMB,OTHER)` shape, so a row still abstaining would mean something other than the Data Use is responsible.

`recomputeMatches` is not re-run in prod: the correction made dataset 1 canonical, so it is no longer a candidate.

----

- Verify all tests go green

[DT-4030]: https://broadworkbench.atlassian.net/browse/DT-4030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DT-3863]: https://broadworkbench.atlassian.net/browse/DT-3863?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/mapper/RowMapperHelper.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/mapper/RowMapperHelper.java
@@ -7,7 +7,6 @@ import java.sql.SQLException;
 import java.util.Date;
 import java.util.Objects;
 import java.util.Optional;
-import org.apache.commons.text.StringEscapeUtils;
 import org.broadinstitute.consent.http.models.DataAccessRequestData;
 import org.broadinstitute.consent.http.util.ConsentLogger;
 import org.jdbi.v3.core.result.RowView;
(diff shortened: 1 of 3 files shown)


otchet-broad (approved):
👍

rushtong (approved):
Nice simplification 👍🏽

PR #3039, opened 2026-08-27
[DT-3873] Assert data submitters cannot read another user's draft

kevinmarete:
### Addresses

[DT-3873](https://broadworkbench.atlassian.net/browse/DT-3873).

Security risk: no. Test-only.

### Summary

Draft ownership was asserted for chairpersons only. The template-validation endpoint admits data submitters as well, so a template draft is reachable by that role and the case is worth asserting rather than inferring from the chairperson pair.

Adds the matching pair to `DraftServiceDAOTest`: a data submitter cannot read another user's draft, and can read their own.

### Also verified, no change

Nothing on the template path logs uploaded content — the `studytemplate` package, `DraftService`, and `DraftServiceDAO` have no log statements, and the one exception it throws carries a fixed constant.

`@Timed` gives latency and throughput but does not separate success from failure. `@ResponseMetered` would, but no resource in consent uses it, so adding it to one endpoint alone was left out of this PR. Recorded on the ticket as a repo-wide question.

----

- Verify all tests go green

[DT-3873]: https://broadworkbench.atlassian.net/browse/DT-3873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/test/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAOTest.java
+++ b/src/test/java/org/broadinstitute/consent/http/service/dao/DraftServiceDAOTest.java
@@ -198,6 +198,28 @@ class DraftServiceDAOTest extends DAOTestHelper {
   }
 
   @Test
+  void testDataSubmitterCannotReadAnotherUsersDraft() throws SQLException {
+    // The validation endpoint admits data submitters too, so this is asserted, not assumed.
+    User owner = createUser();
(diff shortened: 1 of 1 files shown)


fboulnois (approved):
👍

rushtong (approved):
👍🏽

PR #3047, opened 2026-09-04
[DT-4061] Consent modifications to support redesigned study pages.

otchet-broad:
### Addresses

### Summary

Adds the backend for the redesigned study page: PI contact details, researcher ratings/comments, first-class registration assets, and study-level metrics and recommendations.

PI details on Study

Four new study columns (pi_institution_id, pi_orcid, pi_linkedin_url, pi_website_url), plumbed through Study, StudyPatch, StudyDAO, and DatasetServiceDAO. Study reads join institution so the PI institution name comes back with the study. Registration already collected the PI institution as the numeric piInstitution study property; the create path now writes the column as well, and a backfill changeset copies the property into the column for existing studies (skipping non-numeric values and ids with no institution row).

Because these are columns rather than study properties, they follow JSON semantics instead of the blank-string-deletes convention used elsewhere in StudyPatch: absent is a no-op, explicit null clears, a blank string normalizes to a clear, a value sets. StudyPatch now records which fields arrived as explicit JSON nulls to make that distinction possible.

Behavior change: PATCH authorization                                                                                                                 
StudyResource.patchStudyById now requires creator/custodian/admin on this study, not merely a study-editing role somewhere in DUOS. A role-only callewho could PATCH a publicly visible study on develop now getsstration PUT path.

Comments & ratings

New study_comment table (rating 1–5 enforced by check construdy, FK cascade on study and user) with GET (list + average), POST (upsert), and DELETE (own comment only) under api/dataset/study/{studyId}/comments. Posting requires the Researcher role and a library card;    reads are @PermitAll subject to study visibility.

Asset promotion

Eight asset lists (models, workspaces, presentations, publiclectualProperties, biospecimens, funding) move out of theclient-managed, unvalidated assets blob into their own study_property rows and become first-class registration fields. Reads fall back to the legacy object for studies not re-registered since; a present promoteven when empty, so removing the last asset of a type isn'tundone by a stale copy. Elasticsearch reassembles the original assets shape, so search consumers are unaffected.

Seven new GET /api/dataset/study/{studyId}/assets/* endpoints. Two naming mismatches worth knowing: /assets/intellectualProperty reads intellectualProperties, /assets/fundingResources reads fundindpoint — it's reachable only through the registration GET.

Metrics & recommendations

- Study-scoped DAR summaries in a single query rather than o summary from the most recently submitted qualifying DAR so a pending progress report can't overwrite the grant's title, RUS, and date.
- Research outputs (presentations, publications, IP) aggregaeports.
- findSimilar — public studies sharing the PI or a data type, ranked by overlap; a blank PI name never matches another blank one.
- findFrequentlyRequestedWith — public studies co-requested ived DARs, and progress reports are excluded from scoring.

Cross-cutting

All twelve new @PermitAll GETs go through a shared DatasetSecess, so a non-public study is readable only by creator,custodians, and admins. New services are registered as Guice singletons (docs/ai/CLAUDE.md updated with the provider pattern). AuthorizationHelper now
null-guards a user with no role rows, which previously turne

Reviewer notes

- Four Liquibase changesets; two are data migrations with noangeset is destructive by design — it removes promoted keysfrom the legacy object after copying them. The registration PUT is a full replace: a client omitting
- Unit/integration tests added for each new DAO, service, resource, and both migrations. The five scripts/verify-study-*.sh runbooks drive a live instance manually and are not part of CI.
----

- Verify all tests go green


rushtong:
Some unfiltered AI findings while I go through this manually.

---
**Claude**

**Findings**
- StudyAssets.java:93 — assemble removes legacy asset keys with equalsIgnoreCase, but findAssetList reads the legacy object with an exact-case get(key). A key that differs only in case is deleted and never put back.
- MetricsService.java:54 — findProgressReportsByStudyId applies no approval filter, so research outputs from progress reports on ungranted requests reach the public study page.
Detail

**1. Case-variant legacy asset keys disappear.**

Take a study that carries assets = {"Publications": [...]} and was never re-registered. Both migration changesets use exact-case JSON operators, so they leave the key alone. Then on every read:

1. assemble copies the legacy object.
2. Line 93 runs removeIf("publications"::equalsIgnoreCase) and deletes "Publications".
3. Line 78 looks up get("publications") — exact case — and gets null.
4. findAssetList returns an empty list, so line 95 puts nothing back.

The key then vanishes from the registration GET and from the search index. Before this change, findMapPropValue returned the legacy map as it stood and kept it. Pick one convention for both sides: either read the legacy map case-insensitively, or remove from it case-sensitively.

**2. Research outputs skip the approval gate.**

The sibling query, findSummaryMetricApprovedDARsByStudyIdIncludesExpired, works hard to require a final access vote of TRUE. findProgressReportsByStudyId at DataAccessRequestDAO.java:294 checks only parent_id IS NOT NULL, a submission date, and non-archived status. The new DAO test at DataAccessRequestDAOTest.java:1397 builds a parent DAR with no election or vote and still expects both reports back. So on a public study, any signed-in caller can see outputs from a request nobody granted. If a progress report can only exist after a grant in practice, say so in the Javadoc; otherwise add the vote filter.

One nit outside the findings: findProgressReportsByDatasetId at DataAccessRequestDAO.java:318 has no production caller — only the DAO test uses it.

---
**Codex**

**PI institution edits can be reverted by registration saves.** PATCH updates the institution column but leaves the piInstitution study property unchanged. Registration GET reads that stale property, and a subsequent PUT writes it back into the column. Changing institution A to B—or clearing it—can therefore be undone by an unrelated registration edit. Keep both representations synchronized or make registration reads use the column.

**Legacy asset edits can be silently discarded.** Registration GET returns assets both top-level and inside assets. A legacy client that preserves the response and edits only assets.publications sends the unchanged top-level copy too. promotedValue() always selects that copy, so additions, edits, and deletions are ignored despite a successful save. The compatibility path needs to handle conflicting representations.

fboulnois (commented):
Can we split this out into smaller PRs? 8k lines is not easily human-reviewable.

Aside from that, at first glance I see that tests have been added using bash scripts. I am concerned about the maintainability of this approach. Additionally, I think given their size, each could be a separate reviewable component.

src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
```diff
@@ -105,6 +105,12 @@ protected boolean authorize(AuthUser authUser, String role) {
     boolean authorize = false;
     try {
       User user = userService.findUserByEmail(authUser.getEmail());
+      // A user with no user_role rows has a null role list, not an empty one. Without this guard
+      // the authorizer throws a NullPointerException and Jersey turns what should be a plain
+      // denial into a 500 on every @RolesAllowed endpoint. Mirrors User#hasAnyUserRole.
+      if (user == null || user.getRoles() == null) {
+        return false;
+      }
```
rushtong, line 113: This might also be in `User.getRoles()` if `roles` were instantiated with an empty list by default. That would also protect other callers from hitting this edge case.

otchet-broad:
I'm going to split this up into multiple PRs and remove the bash scripts.  Those scripts have been helpful testing against the variants in dev and prod databases, but aren't needed moving forward.

PR #3049, opened 2026-09-08
[DT-4061] Centralize study read access in DatasetService

otchet-broad:
**Base:** `develop` · 7 files, +238/-16 · **Migration:** no · **UI impact:** none

Extracts the study read-visibility rule out of `StudyResource` into
`DatasetService#verifyStudyVisibilityAccess`, so the asset, comment, and
metrics endpoints later in the stack all share one gate rather than each
re-deriving it. A study whose `public_visibility` is NULL now reads as "not
public" — 404 for an unapproved caller — instead of throwing on the unboxing
and returning 500.

Also guards `AuthorizationHelper` against a user with no `user_role` rows,
whose role list is null rather than empty. That NPE surfaced as a 500 on
every `@RolesAllowed` endpoint instead of a plain denial. Corrects the
`ConsentModule` singleton guidance in `docs/ai/CLAUDE.md` to describe the
provider-parameter pattern the module actually uses.

No behavior change for any caller who could already read a study, and no
migration. The PATCH ownership gate that originally shared this refactor is
now branch 1b.

**Depends on:** nothing. Every other branch depends on this one.

---

(95 files, +8168), so it was split into eight PRs. **This PR targets `develop`,

| # | Branch | Base |
|---|---|---|
| 1 | `otchet-dt-4061-study-visibility-authz` ← **this PR** | `develop` |

the descendants.

independently.


kevinmarete (commented):
`verifyStudyVisibilityAccess` treats null visibility as private, but `canReadStudy` still treats every value except `false` as public. A null-visibility study can therefore return 404 through the new study gate while remaining readable through dataset routes. Please define this rule once, ideally by changing `canReadStudy` to require `Boolean.TRUE.equals(...)` and delegating the new gate to it. Please also cover the dataset path with a null-visibility test.

fboulnois:
Some issues I filtered that Claude and Codex identified:

### Duplicate authorization gates return different statuses for the same predicate

DatasetService.java:375 defines findStudyByIdForRead, which wraps canReadStudy and throws ForbiddenException. FileStorageObjectService.java:442 uses this method for study files. The new verifyStudyVisibilityAccess wraps the same predicate but throws NotFoundException.

In the downstream stack, otchet-dt-4061-study-asset-endpoints adds a StudyAssetService that uses the 404 gate, while FileStorageObjectService continues using the 403 gate. As a result, two study-asset routes will disagree about whether a hidden study is forbidden or absent. That distinction matters because a 403 confirms that the study exists, while a 404 does not.

### Redundant initService() calls

DatasetServiceTest.java contains redundant initService() calls at lines 1197, 1211, 1228, 1249, and 1270. The method already runs under @BeforeEach, and the other tests in this class do not call it explicitly.

fboulnois (approved):
Some issues I filtered that Claude and Codex identified:

### Duplicate authorization gates return different statuses for the same predicate

DatasetService.java:375 defines findStudyByIdForRead, which wraps canReadStudy and throws ForbiddenException. FileStorageObjectService.java:442 uses this method for study files. The new verifyStudyVisibilityAccess wraps the same predicate but throws NotFoundException.

In the downstream stack, otchet-dt-4061-study-asset-endpoints adds a StudyAssetService that uses the 404 gate, while FileStorageObjectService continues using the 403 gate. As a result, two study-asset routes will disagree about whether a hidden study is forbidden or absent. That distinction matters because a 403 confirms that the study exists, while a 404 does not.

### Redundant initService() calls

DatasetServiceTest.java contains redundant initService() calls at lines 1197, 1211, 1228, 1249, and 1270. The method already runs under @BeforeEach, and the other tests in this class do not call it explicitly.

kevinmarete (approved):
👍🏾

kevinmarete (approved):
👍🏾

fboulnois (approved):
👍

rushtong:
I'm not 100% clear on the actual scope of this PR - it is substantially larger than the description. Claude and Codex both flagged this. Can it be broken up in any way? For example changes to `AuthorizationHelper` look like a simple bug fix that could stand on its own.

> [**P1] Restore the PR to its intended scope.** PR #3050 and #3051 were merged into this PR’s head branch, so #3049 now contains about 90 files and 9,271 diff lines—including new endpoints, schema migrations, comments, metrics, assets, and PI details—despite describing a 7-file visibility/auth refactor. For example, [ConsentApplication.java now registers unrelated asset/comment resources](https://github.com/DataBiosphere/consent/pull/3049/files). Merging #3049 would deploy later stack branches prematurely and invalidate the stated independent review/landing order. Reset the head to the original visibility commit or rebase/recreate the PR with only its intended changes.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -186,6 +194,128 @@ SELECT DISTINCT ON (dar.collection_id) dar.*
...
+          )
+          SELECT
+              c.dar_code,
+              latest_dar.submission_date,
+              latest_dar.reference_id,
+              latest_dar.update_date,
+              latest_dar.data ->> 'projectTitle' AS project_title,
+              latest_dar.data ->> 'nonTechRus' AS non_tech_rus,
+              COALESCE(latest_dar.data ->> 'piName', u.display_name) AS pi_name,
+              i.institution_name
```
rushtong: In the dataset-scoped counterpart to this query, these two values are coerced to `NULL::text`. Should we do that here too?

src/main/java/org/broadinstitute/consent/http/resources/StudyResource.java
```diff
@@ -155,16 +155,24 @@ public Response getStudyById(@Auth DuosUser duosUser, @PathParam("studyId") Inte
...
+      // registration PUT path does.
+      if (!datasetService.isCreatorCustodianOrAdmin(user, study)) {
+        throw new ForbiddenException("Study with ID " + studyId + " is not updatable");
+      }
       StudyPatch studyPatch = StudyPatch.fromJson(json);
       if (!studyPatch.isPatchable(study)) {
         return Response.status(Status.NOT_MODIFIED).entity(study).build();
       }
-      Study patchedStudy = datasetService.patchStudy(studyId, duosUser.getUser(), studyPatch);
+      Study patchedStudy = datasetService.patchStudy(studyId, user, studyPatch);
```
rushtong, line 177: This line is short-circuited if the study is not patchable (which is correct). The new code to `retireLegacyPiInstitutionProp` is never hit here if the user is only patching `piInstitutionId` since that field isn't considered in `studyPatch.isPatchable()`

src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
```diff
@@ -1,30 +1,91 @@
...
+   */
+  public List<DarMetricsSummary> generateDarSummaries(Integer datasetId, User user) {
+    datasetService.findDatasetByIdForRead(user, datasetId);
     return darDAO.findSummaryMetricApprovedDARsByDatasetIdIncludesExpired(datasetId);
   }
+
+  public List<DarMetricsSummary> generateStudyDarSummaries(Integer studyId, User user) {
+    requireStudy(studyId, user);
+    return darDAO.findSummaryMetricApprovedDARsByStudyIdIncludesExpired(studyId);
+  }
```
rushtong, line 61: This study version call (via `/dar-summaries/study/{studyId}` -> `generateStudyDarSummaries()`), has slightly different behavior compared to the dataset version (`/dar-summaries/{datasetId}` -> `generateDarSummaries()`). The dataset version can return a 403 for hidden datasets, but the study version doesn't go through that same logic so I think there is a potential for inconsistent responses here.

src/main/java/org/broadinstitute/consent/http/authentication/AuthorizationHelper.java
```diff
@@ -105,6 +105,12 @@ protected boolean authorize(AuthUser authUser, String role) {
     boolean authorize = false;
     try {
       User user = userService.findUserByEmail(authUser.getEmail());
+      // A user with no user_role rows has a null role list, not an empty one. Without this guard
+      // the authorizer throws a NullPointerException and Jersey turns what should be a plain
+      // denial into a 500 on every @RolesAllowed endpoint. Mirrors User#hasAnyUserRole.
+      if (user == null || user.getRoles() == null) {
+        return false;
+      }
```
rushtong, line 113: There is another instance of this problem in [Resource.java](https://github.com/DataBiosphere/consent/blob/49bad9e1fadabd17eb4bba29745cc36ec991fb1d/src/main/java/org/broadinstitute/consent/http/resources/Resource.java#L326). Might it be better to instantiate an empty roles array on `User` instead?

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
```diff
@@ -186,6 +194,128 @@ SELECT DISTINCT ON (dar.collection_id) dar.*
...
+              FROM data_access_request dar
+              INNER JOIN dar_dataset dd ON dd.reference_id = dar.reference_id
+              INNER JOIN study_datasets sd ON sd.dataset_id = dd.dataset_id
+              INNER JOIN (
+                  SELECT DISTINCT e.reference_id, e.dataset_id,
+                      LAST_VALUE(v.vote) OVER(
+                          PARTITION BY e.reference_id, e.dataset_id
+                          ORDER BY v.create_date
+                          RANGE BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING
+                      ) last_vote
```
rushtong, line 237: Claude finding:

> **Study DAR summary windows over every dataaccess vote**
>
> The final_access_vote derived table runs `DISTINCT` plus `LAST_VALUE` over all `dataaccess` elections and votes before joining to the study's datasets, and because the bound is a join to the `study_datasets` CTE rather than a constant, Postgres cannot push it into the subquery as it can for the dataset-scoped query.
>
> Every `GET /api/metrics/dar-summaries/study/{id}` sorts and partitions the full election x vote set (tens of thousands of rows in prod) to answer for a study with a few datasets. 
> 
> Fix: join `study_datasets` on `e.dataset_id` inside the derived table so the election (`reference_id`, `dataset_id`) index prunes before the window; semantics are unchanged since the partition is already per dataset.

otchet-broad:
@rushtong  - ultimately this branch morphed into a consolidation branch of multiple consent related PRs before merging to develop.  I've been tracking all of the expanding scope and am trying to consolidate down to these two branches (this one in consent, and a similar one for DT-4062 in duos-ui) to do a full run through all of the change this work has necessitated before landing on develop.

src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java
otchet-broad: This changes as a result of work you've now seen.

src/main/java/org/broadinstitute/consent/http/service/MetricsService.java
otchet-broad: I've made [https://broadworkbench.atlassian.net/browse/DT-4181](https://broadworkbench.atlassian.net/browse/DT-4181) to resolve the differences.  There are inconsistencies that predate this code and should be resolved holistically.

PR #3051, opened 2026-09-08
[DT-4061] Add PI institution, ORCID, and profile links to studies

otchet-broad:
**Base:** branch 1 · 20 files, +1165/-22 · **Migration: yes (2 changesets)** · **UI impact:** additive only

Adds `pi_institution_id`, `pi_orcid`, `pi_linkedin_url`, `pi_website_url` to
the `study` table and surfaces them on `Study`, `StudyPatch`, and the study
registration read path.

Registration already collected the PI institution as the numeric
`piInstitution` study property, so `createStudy` now records the column too,
and a backfill changeset copies the property into the column for studies
registered earlier. Without the backfill every existing study shows a blank
PI Institution until someone re-PATCHes it. The backfill skips values that
are not plain integers and ids with no matching institution row, which the
foreign key would otherwise reject.

Review focus: these are columns, not study properties, so they deliberately
do **not** follow the "blank string deletes" convention the patchable
properties use. They follow the JSON convention instead — absent is a no-op,
explicit `null` clears, blank normalizes to a clear. Jackson cannot tell an
absent field from an explicit null, so `StudyPatch.fromJson` re-reads the
body to record which field names arrived as literal nulls.

A registration PUT carries the PI institution but not the profile links,
which are PATCH-only, so `DatasetRegistrationService` carries the stored
links forward rather than dropping them; a study conversion does the same.
`executeUpdateStudy` now raises `NotFoundException` for a missing study
instead of dereferencing null.

Additive on the wire: four new fields on study reads, accepted on PATCH, no
existing field changes shape.

**Depends on:** branch 1 — shares hunks in `DatasetService.java` and
`DatasetServiceTest`. That file overlap is the only reason it is stacked here.

---

(95 files, +8168), so it was split into eight PRs. **This PR targets `otchet-dt-4061-study-visibility-authz`,

| # | Branch | Base |
|---|---|---|
| 2 | `otchet-dt-4061-study-pi-details` ← **this PR** | `otchet-dt-4061-study-visibility-authz` |

the descendants.

independently.


kevinmarete (commented):
Here are some findings:

1. **P1 — PI institution can be silently reverted.**  
   PATCH updates the new column, but the registration response builder still reads the legacy study property at [SchemaFromStudy.java#L89](https://github.com/DataBiosphere/consent/blob/8f0be37b3366e44508f684291c33309b887ec5fb/src/main/java/org/broadinstitute/consent/http/models/dataset_registration_v1/builder/SchemaFromStudy.java#L89).

   Example: institution `1` is backfilled, PATCH changes it to `2`, registration GET still returns `1`, and a subsequent registration PUT writes `1` back into the column. The builder should treat `study.piInstitution` as authoritative, with a PATCH → registration GET → PUT regression test.

2. **P2 — Study responses fabricate an institution creation date.**  
   [StudyReducer.java#L26](https://github.com/DataBiosphere/consent/blob/8f0be37b3366e44508f684291c33309b887ec5fb/src/main/java/org/broadinstitute/consent/http/db/mapper/StudyReducer.java#L26) constructs a partial `Institution`, whose constructor initializes `createDate` to the current time at [Institution.java#L44](https://github.com/DataBiosphere/consent/blob/8f0be37b3366e44508f684291c33309b887ec5fb/src/main/java/org/broadinstitute/consent/http/models/Institution.java#L44). Every study read therefore returns a false, changing `piInstitution.createDate`. The query should map the actual institution metadata, or the API should use a dedicated ID/name summary model.

fboulnois:
Some issues I identified with Claude and Codex:

### 1. An out-of-range value aborts the backfill

I ran the shipped SQL against `postgres:16.10-alpine` with `piInstitution = 99999999999`. It passes `^[0-9]+$`, then fails the integer cast:

```
ERROR:  value "99999999999" is out of range for type integer
```

This fails the entire changeset, including the backfill of valid rows. Limiting the digit count avoids the error:

```sql
AND btrim(sp.value) ~ '^[0-9]{1,9}$'
```

I confirmed this backfills the valid row and skips the out-of-range value. It also excludes valid 10-digit integer IDs, so a full range check would be needed to support those.

### 2. Both write paths need to validate `piInstitutionId`

**PATCH:** `DatasetService.java:826-830` catches every exception and throws `InternalServerErrorException`, so `{"piInstitutionId": 999999}` returns 500 instead of 400 when the institution doesn't exist.

**Registration:** `checkPiInstitutionRequired` only checks for null. Previously, a nonexistent ID was stored as a text property and registration succeeded. Now the foreign key constraint fails the entire registration transaction. This changes behavior for existing clients and needs explicit validation.

### 3. PATCH leaves conflicting institution values in study responses

After a PATCH, the study response contains the new `piInstitution.id` and the old `properties[piInstitution]` value. `SchemaFromStudy` handles this for registration payloads, but raw study responses still expose both. Update the legacy property in `patchStudy` or stop returning it.

### 4. The backfill test doesn't cover the `EXISTS` guard

`testSkipsValuesThatCannotBeAppliedToTheColumn` uses `-1` as an orphan ID, but that fails the regex before reaching `EXISTS`. Use a positive ID with no institution row, such as `2000000000`, and add a separate out-of-range case for item 1.

### 5. The timestamp fix needs a test

The commit fixes the constructor's default `createDate` by always overwriting it in the reducer, but no test asserts `getPiInstitution().getCreateDate()`.

Since `hasOptionalColumn` catches every exception and returns empty, a broken alias or mapping could silently produce null dates. Test both populated and null timestamps.

### 6. Registration updates fetch the study twice

`updateStudyFromRegistration` calls `findStudyById` just to read three strings. That opens a `REPEATABLE_READ` transaction and loads properties, dataset IDs and files. `executeUpdateStudy` then fetches the study again in the write transaction.

Read the links inside `executeUpdateStudy`, where the study is already loaded, to avoid the extra fetch and the gap between reading and writing.

### 7. Smaller issues

- The 18-argument constructor and 13-argument `updateStudy` are easy to misuse, especially with several adjacent `String` arguments and all-null call sites. A builder or test fixture would help, as would assertions more specific than `any()` for every argument.
- `piInstitutionId` accepts `"123"` as an integer, unlike the DTO's strict string and boolean deserializers. Add equivalent integer validation for consistency.
- Add a test showing that `{"piName": null}` remains a no-op while `{"piOrcid": null}` clears the field.
- `Dataset.yaml` reuses `Study.yaml` and documents `piInstitution`, but `DatasetDAO.findDatasetStudyById` doesn't select `pi_institution_id`, so the field is always null on that route.

I also checked that `stringPatchProps()` excludes the PI columns: a blank `piOrcid` correctly avoids deleting a study property of that name.

fboulnois (approved):
Some issues I identified with Claude and Codex:

### 1. An out-of-range value aborts the backfill

I ran the shipped SQL against `postgres:16.10-alpine` with `piInstitution = 99999999999`. It passes `^[0-9]+$`, then fails the integer cast:

```
ERROR:  value "99999999999" is out of range for type integer
```

This fails the entire changeset, including the backfill of valid rows. Limiting the digit count avoids the error:

```sql
AND btrim(sp.value) ~ '^[0-9]{1,9}$'
```

I confirmed this backfills the valid row and skips the out-of-range value. It also excludes valid 10-digit integer IDs, so a full range check would be needed to support those.

### 2. Both write paths need to validate `piInstitutionId`

**PATCH:** `DatasetService.java:826-830` catches every exception and throws `InternalServerErrorException`, so `{"piInstitutionId": 999999}` returns 500 instead of 400 when the institution doesn't exist.

**Registration:** `checkPiInstitutionRequired` only checks for null. Previously, a nonexistent ID was stored as a text property and registration succeeded. Now the foreign key constraint fails the entire registration transaction. This changes behavior for existing clients and needs explicit validation.

### 3. PATCH leaves conflicting institution values in study responses

After a PATCH, the study response contains the new `piInstitution.id` and the old `properties[piInstitution]` value. `SchemaFromStudy` handles this for registration payloads, but raw study responses still expose both. Update the legacy property in `patchStudy` or stop returning it.

### 4. The backfill test doesn't cover the `EXISTS` guard

`testSkipsValuesThatCannotBeAppliedToTheColumn` uses `-1` as an orphan ID, but that fails the regex before reaching `EXISTS`. Use a positive ID with no institution row, such as `2000000000`, and add a separate out-of-range case for item 1.

### 5. The timestamp fix needs a test

The commit fixes the constructor's default `createDate` by always overwriting it in the reducer, but no test asserts `getPiInstitution().getCreateDate()`.

Since `hasOptionalColumn` catches every exception and returns empty, a broken alias or mapping could silently produce null dates. Test both populated and null timestamps.

### 6. Registration updates fetch the study twice

`updateStudyFromRegistration` calls `findStudyById` just to read three strings. That opens a `REPEATABLE_READ` transaction and loads properties, dataset IDs and files. `executeUpdateStudy` then fetches the study again in the write transaction.

Read the links inside `executeUpdateStudy`, where the study is already loaded, to avoid the extra fetch and the gap between reading and writing.

### 7. Smaller issues

- The 18-argument constructor and 13-argument `updateStudy` are easy to misuse, especially with several adjacent `String` arguments and all-null call sites. A builder or test fixture would help, as would assertions more specific than `any()` for every argument.
- `piInstitutionId` accepts `"123"` as an integer, unlike the DTO's strict string and boolean deserializers. Add equivalent integer validation for consistency.
- Add a test showing that `{"piName": null}` remains a no-op while `{"piOrcid": null}` clears the field.
- `Dataset.yaml` reuses `Study.yaml` and documents `piInstitution`, but `DatasetDAO.findDatasetStudyById` doesn't select `pi_institution_id`, so the field is always null on that route.

I also checked that `stringPatchProps()` excludes the PI columns: a blank `piOrcid` correctly avoids deleting a study property of that name.

kevinmarete (commented):
Here are two findings:

1. The PI institution backfill accepts any digit-only string and then casts it to PostgreSQL `integer`. A value such as `99999999999` passes `^[0-9]+$` but raises `integer out of range`, aborting the entire Liquibase changeset. Please use a range-safe conversion or explicitly verify that the value fits an integer, and add an out-of-range migration test.

2. The registration and PATCH paths do not validate that `piInstitutionId` references an existing institution before writing it. An invalid ID reaches the foreign-key constraint, and PATCH converts the resulting exception into a 500. Please validate the institution explicitly and return an appropriate 400 response, with coverage for both write paths.

kevinmarete (commented):
Still seeing one more finding:

**P1 — Clearing `piInstitutionId` is undone by the legacy fallback**

For a migrated study, both `study.pi_institution_id` and the legacy `piInstitution` property contain the institution ID. PATCHing with `{"piInstitutionId": null}` clears only the column, leaving the legacy property unchanged. `SchemaFromStudy` then falls back to that property when the column is null, so registration GET returns the old institution and a subsequent registration PUT restores it in the column.

This violates the documented explicit-null-clears behavior. Please remove or update the legacy property when PATCH clears the institution, or otherwise distinguish an intentional clear from an unbackfilled legacy value. Please also add a regression test covering PATCH null → registration GET → registration PUT.

kevinmarete (commented):
Thanks for addressing the earlier feedback. I found one remaining issue before I can approve: `DatasetDAO.findDatasetStudyById` does not select `pi_orcid`, `pi_linkedin_url`, or `pi_website_url`. As a result, dataset responses return these fields as `null` in the embedded study even when values exist. Please add the three columns with the appropriate `s_` aliases and cover this dataset lookup path with a test. Once that is fixed, I am happy to approve.

kevinmarete (commented):
One blocker remains before I can approve: `DatasetDAO.findDatasetStudyById` does not select `pi_orcid`, `pi_linkedin_url`, or `pi_website_url`. Because the embedded `Study` uses the `s_` bean-mapping prefix, dataset responses return these fields as null even when values are stored.

Please add:
```sql
s.pi_orcid AS s_pi_orcid,
s.pi_linkedin_url AS s_pi_linkedin_url,
s.pi_website_url AS s_pi_website_url
```

Please also add a `DatasetDAO` regression test that stores all three values and verifies they are populated through the dataset lookup path. Once that is addressed, I’m comfortable approving.

kevinmarete (approved):
Excellent 👍🏾

PR #3052, opened 2026-09-08
[DT-4061] Add study ratings and comments

otchet-broad:
**Base:** branch 2 · 17 files, +944 · **Migration: yes (1 changeset)** · **UI impact:** none

Adds a `study_comment` table and the endpoints behind it: `GET`/`POST
/api/dataset/study/{studyId}/comments` and `DELETE
/api/dataset/study/{studyId}/comments/{commentId}`. The list response carries
the comments plus their average rating.

One comment per user per study, enforced by a unique constraint on
`(study_id, user_id)`, so POST upserts rather than accumulating rows. Rating
is constrained to 1–5 in the database as well as the service. Posting
requires an active researcher — the Researcher role plus a library card.
Deletes only ever remove the caller's own comment: the user id is part of the
DELETE predicate rather than a check performed before it.

Reads go through `verifyStudyVisibilityAccess`, so comments on a study that
is not publicly visible are readable only by its creator, its custodians, and
admins — the same rule the study itself uses.

Entirely additive: new table, new endpoints, no change to any existing
response, and nothing in the running UI calls these yet.

**Depends on:** branch 1 (visibility gate).

---

(95 files, +8168), so it was split into eight PRs. **This PR targets `otchet-dt-4061-study-pi-details`,

| # | Branch | Base |
|---|---|---|
| 3 | `otchet-dt-4061-study-comments` ← **this PR** | `otchet-dt-4061-study-pi-details` |

the descendants.

independently.


fboulnois (approved):
Some minor issues identified:

### 1. Creators and custodians can rate their own studies

`requireStudy` allows creators and custodians to read a non-public study. `post` then checks only the Researcher role and a library card, so they can submit ratings that affect their own study's average.

Decide whether to allow this, prevent them from posting, or exclude their ratings from the average.

### 2. There is no moderation path

Only authors with the Researcher role can delete comments. Admins cannot remove someone else's comment, and authors who lose the role cannot delete their own.

Since comments on public studies are readable by every authenticated user, decide how moderation and deletion after a role change should work before launch.

### 3. `commentText` accepts numbers and booleans

`rating` is parsed explicitly to avoid Gson coercion, but `commentText` accepts any JSON primitive. `{"commentText": 123}` stores `"123"`, and `{"commentText": true}` stores `"true"`.

The non-string test covers only an object. Add an `isString()` check and tests for numbers and booleans.

### 4. Comments have no length limit

The column is `TEXT`, the spec has no `maxLength`, and the service applies no limit. Large comments are stored and returned on every list request. Add a limit to both the spec and the service.

### 5. The list endpoint has no pagination

`findByStudyId` loads every comment, and Java computes the average over the full list. That may be fine initially, but responses can grow without limit. Decide on pagination now to avoid changing the response shape later.

### 6. Posting reloads every comment to find the saved one

`StudyCommentService.java:53-57` calls `findByStudyId` and filters for the ID returned by `upsert`. Use `findById` to fetch one row, and handle a missing result explicitly. The current `.orElseThrow()` produces a 500 if the comment is deleted between the write and read.

### 7. Visibility checks load more study data than they need

`requireStudy` calls `studyDAO.findStudyById`, which opens a `REPEATABLE_READ` transaction and runs three queries, including data the visibility check doesn't use.

Use `findStudyDetailsById`, as `verifyPublicVisibilityAccess` does for datasets. The same pattern appears elsewhere in this stack and could be fixed consistently.

### 8. Missing tests

- Verify that the DAO rejects ratings of 0 and 6 through `ck_study_comment_rating`.
- Verify that an upsert preserves `create_date` and advances `update_date`.
- Verify list ordering by `create_date DESC, study_comment_id DESC`.

### 9. API documentation

- Document DELETE's 403 response from the Researcher role requirement.
- Clarify that “newest first” means creation order. Editing a comment does not change its position.

kevinmarete (commented):
Here is a finding:

`commentText` currently accepts any JSON primitive rather than only strings. `isJsonPrimitive()` is also true for numbers and booleans, and `getAsString()` consequently turns values such as `123` and `false` into stored strings.

Please check `element.getAsJsonPrimitive().isString()` and add tests showing that numeric and boolean comment values return 400.

kevinmarete (approved):
👍🏾

PR #3055, opened 2026-09-08
[DT-4061] Promote study registration asset lists to first-class fields

otchet-broad:
**Base:** branch 5 · 20 files, +1227/-16 · **Migration: yes (2 changesets, data rewrite)** · **UI impact:** designed to be none — verify

The riskiest PR in the stack, and the only one that rewrites existing data.

`DatasetRegistrationSchemaV1.yaml` declared `assets` as client-managed
metadata "preserved and returned as-is by the backend; not
backend-validated". Backend endpoints and dashboard counts came to depend on
individual keys inside it, so that contract no longer held. This promotes the
eight lists the backend reads — `models`, `workspaces`, `presentations`,
`publications`, `clinicalTrials`, `intellectualProperties`, `biospecimens`,
`funding` — to top-level registration fields, each stored in its own
`study_property` row, with schemas for the item types.

Compatibility is the point of the `StudyAssets` helper. Reads fall back to
the legacy object for a study not yet rewritten, and writes are still
accepted through it, so a client that has not moved to the top-level fields
keeps working. Registration reads and the search index both still return the
full `assets` object — every promoted list plus whatever unpromoted keys
remain — so no consumer sees a shape change.

Two details deserve review attention. A promoted property that parses is
authoritative even when empty, so a submitter who removed the last asset of a
type does not get it restored from a stale legacy copy. And "provided" means
present, not non-empty: registration reads return each list both top-level
and inside `assets`, so an edit clearing the top-level list arrives beside
the pre-edit legacy copy, and treating `[]` as "not provided" would resurrect
what was just removed.

Migration promotes each array-valued key out of `assets` into its own row,
then strips the promoted keys and drops any `assets` row left empty. Guarded
to skip non-object and non-array values and to skip a study that already has
the target row.

**Rollback hazard:** once the strip changeset runs, code from before this
branch can no longer see these lists — they no longer live in `assets`.
Rolling back the deploy requires restoring the data.

Also corrects `IntellectualProperty.filingDate` from boolean to string.

**Depends on:** nothing in branches 1-5 — no code dependency on any of them.
Placed last deliberately, because it is the only data rewrite in the stack and
the hardest to roll back. It can be reordered earlier if that suits review.

---

(95 files, +8168), so it was split into eight PRs. **This PR targets `otchet-dt-4061-study-recommendations`,

| # | Branch | Base |
|---|---|---|
| 6 | `otchet-dt-4061-study-asset-fields` ← **this PR** | `otchet-dt-4061-study-recommendations` |

the descendants.

independently.


kevinmarete (commented):
The promotion skips a key when its value is not an array, but the strip changeset removes that key regardless. For example, an existing unvalidated `assets` value containing `"models": "custom-value"` will lose that data. Please retain wrong-shaped values unless a value was successfully promoted or an authoritative first-class property already exists, and add a migration test for a JSON object containing a promoted-name key with a non-array value.

kevinmarete (commented):
Here is a finding:

The runtime compatibility handling can discard legacy values that the migration deliberately preserves.

For example, the migration retains an unvalidated value such as `"models": "custom-value"` because it cannot promote it to an array. However, `StudyAssets.assemble()` removes the legacy `models` key and restores it only when it resolves to a nonempty list, so the value disappears from registration and search responses. On the next registration write, `promotedValue()` ignores the non-collection value while `stripPromoted()` removes it, deleting the only stored copy.

Please preserve wrong-shaped promoted-name values through both reads and writes, or explicitly migrate/reject them without silently losing data. Add a round-trip test covering a legacy `assets` object with a non-array value under a promoted key.

kevinmarete (commented):
One finding:

`promotedValue()` performs a case-sensitive lookup with `assets.get(key)`, while `stripPromoted()` removes list-valued keys using `equalsIgnoreCase`. Consequently, a legacy value such as `"Models": [...]` is not promoted but is removed from `assets`, losing the data during a registration update. Please use the same case-insensitive lookup in both paths and add a round-trip test for a differently cased promoted key.

fboulnois:
Some minor suggestions identified with Claude:

1. **`src/main/java/.../db/DataAccessRequestDAO.java:171`**: `pi_name` and `institution_name` were added to the dataset-scoped DAR-summary query, whose only caller `GET /api/metrics/dar-summaries/{datasetId}` is `@PermitAll` with no visibility/DAC check (`generateDarSummaries` only checks the dataset exists). Any authenticated user can enumerate dataset ids and harvest granted-PI identity + affiliation. The new study-scoped sibling in the same PR does gate via `requireReadableStudy`.
2. **`src/main/java/.../resources/StudyResource.java:300`**: `patchStudyById` authorizes a write with a read predicate (`verifyStudyVisibilityAccess` → true for any `public_visibility = true` study). The PR adds `piInstitutionId`/`piOrcid`/`piLinkedinUrl`/`piWebsiteUrl` to that path, so any DATASUBMITTER anywhere can rewrite PI identity/affiliation on any public study. The gate is pre-existing; the widened field set makes it worth fixing now.
3. **`src/main/resources/changesets/changelog-consent-2026-09-02-study-assets.xml:24`**: The migration casts `sp.value::jsonb` in the same `WHERE` as `sp.key = 'assets'`; Postgres doesn't guarantee qual ordering, and non-JSON `study_property.value` rows (bare strings) would raise `invalid input syntax for type json` and abort the Liquibase run. `idx_study_prop_key` makes the safe plan likely, so this is fragility rather than a certain break; a `CASE` wrapper or a key-filtered CTE makes it deterministic.

kevinmarete (approved):
👍🏾

fboulnois (commented):
Two security suggestions, identified with Claude and Codex:

1. **Data exposure:** The branch adds `COALESCE(latest_dar.data ->> 'piName', u.display_name) AS pi_name`
  and `i.institution_name` (via new `LEFT JOIN users` / `LEFT JOIN institution`) to
  `findSummaryMetricApprovedDARsByDatasetIdIncludesExpired`, surfaced through new
  `DarMetricsSummary` fields. That query backs `GET /api/metrics/dar-summaries/{datasetId}`, which is
  `@PermitAll`, and `MetricsService.generateDarSummaries` does an **existence check only** with no user
  parameter at all. Confirmed: the resource's `@Auth DuosUser user` is unused
  (`@SuppressWarnings("unused")`), and every *new* study-scoped sibling in the same file
  (`generateStudyDarSummaries`, `generateStudyResearchOutputs`, `getSimilarStudies`,
  `getFrequentlyRequestedWith`) routes through `requireStudy` → `DatasetService.requireReadableStudy`.
  The legacy dataset-scoped route was given the new PII columns without the gate.
  **Recommendation:** Pass the already-available `User` into `generateDarSummaries` and gate it via
  `requireReadableStudy(dataset.getStudyId(), user)`. If the route must stay ungated for
  compatibility, drop `pi_name`/`institution_name` from it and expose them only on the gated
  study-scoped query.
2. **Broken access control:** `patchStudyById` authorizes a write with a read predicate.
  `checkPublicVisibilityForUser` now delegates to `datasetService.verifyStudyVisibilityAccess` →
  `canReadStudy`, which is
  `Boolean.TRUE.equals(study.getPublicVisibility()) || isCreatorCustodianOrAdmin(...)`. On any public
  study the first disjunct short-circuits, leaving only the class-level
  `@RolesAllowed({ADMIN, CHAIRPERSON, DATASUBMITTER})`. Contrast `deleteStudyById`, which additionally
  requires creator-or-admin. **The weak gate is pre-existing**: diffed against `90901c6a`, where the
  inline version was semantically identical for `true`. What this branch changes is the surface behind
  it: `StudyPatch` gains `piInstitutionId`, `piOrcid`, `piLinkedinUrl`, `piWebsiteUrl`.
  **Recommendation:** Use the write predicate the sibling endpoints already use
  (`isCreatorCustodianOrAdmin`), and rename the helper so a read check can't be mistaken for a write
  check again.

fboulnois (approved):
👍

PR #3058, opened 2026-09-09
[DT-3942] Add the match dataset_id migration surface and snapshot

kevinmarete:
## Addresses

**Stacked on #3057** and targeting `km-dt-3942-match-dataset-id` so the diff shows only this phase. GitHub will retarget it to `develop` when #3057 merges. This needs #3057 deployed everywhere before the run is executed anywhere.

## Summary

Phase 2a: the snapshot tables and the admin surface that reprocesses the match rows still identifying their dataset by legacy text, so they rebuild carrying a real `dataset_id` at the current algorithm version.

Reprocesses rather than backfills. The legacy `consent` values are consent UUIDs whose dataset mapping was dropped in 2024, so no mapping can be derived without guessing — endorsed on review, and it is what the application already does whenever a DAR is edited.

Modelled on the Ticket 4 migration surface (`LegacyDataUseResource`/`LegacyDataUseService`) that DT-4009 retired, so retiring this one is the same clean delete.

### Release sequencing

The constraints are deliberately **not** in this PR. Liquibase applies changesets on deploy, so shipping them here would run them before any environment had cleared the rows they reject, and the precondition would halt the deploy. Three releases:

| Release | Contents | Gate to the next |
| --- | --- | --- |
| A (#3057) | Nullable `dataset_id`, dual write, dataset-correlated election join | Deployed everywhere |
| B (this PR) | Snapshot tables and the admin migration surface | `run` reports `readyForConstraints` |
| C | Non-null and `(purpose, dataset_id)` uniqueness constraints; retire the surface | Confirmed good in prod |

### What it does

**Snapshot** — `match_migration_snapshot` and `match_migration_rationale_snapshot`, created empty and populated when the run executes, because the affected set drains on its own as DARs are re-matched. No foreign key to `match_entity` by design: the run deletes the very rows captured, so a reference would either block the delete or cascade the evidence away. Inserts are `ON CONFLICT DO NOTHING`, so a re-run after a partial failure keeps the first, pre-migration capture instead of overwriting it with post-run state.

**Affected set** — defined by what the constraints will reject (no `dataset_id`, or a `v1`/`v2`/absent stamp) rather than by algorithm version alone. Versions are reported separately because the acceptance criteria count those populations, but they are not what selects the work.

**Skip guard** — `findByReferenceId` filters archived DARs, so a purpose whose DAR is archived or missing would rebuild to nothing and lose its rows. Those purposes are skipped and reported by id, never reprocessed.

**Restartable** — each purpose is rebuilt in its own transaction, one retry apiece, continuing past failures and reporting `failedPurposeIds` so a rerun can be scoped. Failures log the exception class, never the message, which can quote a DAR's free text.

**Endpoints**, all `@RolesAllowed(ADMIN)` and documented in OpenAPI:

- `GET /api/match/migration/dataset-id/population` — read-only, recounts the affected set
- `GET /api/match/migration/dataset-id/reconciliation` — read-only, for a later confirmation pass
- `POST /api/match/migration/dataset-id/run` — snapshots, then rebuilds; returns the population either side plus the reconciliation

Pre-deployment, post-deployment, and rollback validation is recorded under Ticket 5 of `docs/plans/data-use-primary-consistency-plan.md`, alongside the rest of the epic's design.

### Testing

28 new tests, all passing. 14 DAO tests run against real Postgres and cover the population counts, the archived/missing-DAR partitioning, snapshot idempotency, and each reconciliation invariant separately — a row still present, a row replaced under a new id, a purpose rebuilt to nothing, and a skipped row that is still affected. 9 service tests cover snapshot-before-rebuild ordering, the skip guard, retry-once, continuing past a failure, and both ways `readyForConstraints` can be refused. 5 resource tests cover the endpoints and their failure path. The 31 Phase 1 match tests still pass; spotless clean.

### Reviewer notes

- `readyForConstraints` also refuses on a pre-existing `(purpose, dataset_id)` collision, which this migration does not create and cannot fix. That is step 6 of the ticket, but it does mean the run can report "not ready" for something needing a separate decision.
- The skipped set has no owner yet. Production has zero archived-DAR purposes today, so it may stay empty — but any that appear keep a null `dataset_id` and would halt release C's precondition. Worth deciding with the run results rather than now: delete them, or exclude them with a partial constraint.
- `SnapshotReconciliation.purposesHoldingMatches` is named for the state, not the cause: a skipped purpose still holds its original rows and counts there without having been rebuilt. Called out because an operator reads these numbers.
- `findResolvablePurposes` returns an unbounded list. It is bounded in practice by the affected population — a few hundred ids, shrinking — so it is not paginated.
- `MatchMigrationDAO` repeats the archived-status condition from `DataAccessRequestDAO.findByReferenceId` rather than sharing it. The run has to predict that lookup exactly, so if it changes these queries should be revisited deliberately rather than silently following it.

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/MatchMigrationDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/MatchMigrationDAO.java
@@ -16,7 +16,7 @@ import org.jdbi.v3.sqlobject.statement.SqlUpdate;
  * {@code v1} or {@code v2} stamp, or has no stamp at all. Versions are reported separately because
  * the acceptance criteria count those populations, but they are not what selects the work.
  *
- * <p>{@code ARCHIVED_TEST} repeats the archived-status condition from {@code
+ * <p>{@code NOT_ARCHIVED} repeats the not-archived condition from {@code
  * DataAccessRequestDAO.findByReferenceId} rather than sharing it. That lookup is what a reprocess
  * resolves its DAR through, so the run has to predict it exactly; if it changes, these queries have
  * to be revisited deliberately, not silently follow it.
@@ -30,8 +30,7 @@ public interface MatchMigrationDAO {
         OR LOWER(m.algorithm_version) IN ('v1', 'v2'))
       """;
 
-  String ARCHIVED_TEST =
-      "(LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)";
+  String NOT_ARCHIVED = "(LOWER(dar.data->>'status') != 'archived' OR dar.data->>'status' IS NULL)";
 
   /**
    * Counted in one statement so the totals cannot disagree with each other, and recounted on every
@@ -52,7 +51,7 @@ public interface MatchMigrationDAO {
         FROM affected a
         JOIN data_access_request dar ON dar.reference_id = a.purpose
         WHERE """
-          + ARCHIVED_TEST
+          + NOT_ARCHIVED
           + """
       ),
       counts AS (
@@ -95,7 +94,7 @@ public interface MatchMigrationDAO {
       WHERE """
           + AFFECTED_PREDICATE
           + " AND "
-          + ARCHIVED_TEST
+          + NOT_ARCHIVED
           + """
       ORDER BY m.purpose
       """)
@@ -113,7 +112,7 @@ public interface MatchMigrationDAO {
           SELECT 1 FROM data_access_request dar
           WHERE dar.reference_id = m.purpose
             AND """
-          + ARCHIVED_TEST
+          + NOT_ARCHIVED
           + """
         )
       ORDER BY m.purpose
@@ -143,20 +142,21 @@ public interface MatchMigrationDAO {
   int snapshotAffectedMatches();
 
   /**
-   * Keyed on the snapshot rather than on {@code match_entity}, so it can never capture rationales
-   * for a match the other half did not capture. Re-runnable for the same reason as that half.
+   * Joined against the snapshot rather than {@code match_entity}, so it can never capture
+   * rationales for a match the other half did not capture.
+   *
+   * <p>Carries each row's own {@code match_rationale.id}, which makes the capture exact where
+   * comparing text would not: two rationales identical in text are still two rows, and restoring
+   * has to produce two. That id is also what makes the insert re-runnable, by the same {@code ON
+   * CONFLICT} idiom as the matches half.
    */
   @SqlUpdate(
       """
-      INSERT INTO match_migration_rationale_snapshot (match_id, rationale)
-      SELECT r.match_entity_id, r.rationale
+      INSERT INTO match_migration_rationale_snapshot (rationale_id, match_id, rationale)
+      SELECT r.id, r.match_entity_id, r.rationale
       FROM match_rationale r
       JOIN match_migration_snapshot s ON s.match_id = r.match_entity_id
-      WHERE NOT EXISTS (
-        SELECT 1 FROM match_migration_rationale_snapshot existing
-        WHERE existing.match_id = r.match_entity_id
-          AND existing.rationale = r.rationale
-      )
+      ON CONFLICT (rationale_id) DO NOTHING
       """)
   int snapshotAffectedRationales();
 
@@ -177,7 +177,7 @@ public interface MatchMigrationDAO {
             SELECT 1 FROM data_access_request dar
             WHERE dar.reference_id = m.purpose
               AND """
-          + ARCHIVED_TEST
+          + NOT_ARCHIVED
           + """
           )
       ),
--- a/src/main/java/org/broadinstitute/consent/http/db/MatchMigrationDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/MatchMigrationDAO.java
@@ -1,0 +1,202 @@
+package org.broadinstitute.consent.http.db;
+
+import java.util.List;
+import org.broadinstitute.consent.http.models.matchmigration.MatchMigrationPopulation;
+import org.broadinstitute.consent.http.models.matchmigration.SnapshotReconciliation;
+import org.jdbi.v3.sqlobject.config.RegisterConstructorMapper;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+
+/**
+ * Backs the one-off migration that replaces {@code match_entity.consent} with a real {@code
+ * dataset_id}. Retired with the rest of the migration surface once the constraints are on.
+ *
+ * <p>The affected set is defined by what the constraints will reject rather than by algorithm
+ * version alone: a row needs reprocessing if it has no {@code dataset_id}, or still carries a
+ * {@code v1} or {@code v2} stamp, or has no stamp at all. Versions are reported separately because
+ * the acceptance criteria count those populations, but they are not what selects the work.
+ *
+ * <p>{@code ARCHIVED_TEST} repeats the archived-status condition from {@code
+ * DataAccessRequestDAO.findByReferenceId} rather than sharing it. That lookup is what a reprocess
+ * resolves its DAR through, so the run has to predict it exactly; if it changes, these queries have
+ * to be revisited deliberately, not silently follow it.
+ */
+public interface MatchMigrationDAO {
+
+  String AFFECTED_PREDICATE =
+      """
+      (m.dataset_id IS NULL
+        OR m.algorithm_version IS NULL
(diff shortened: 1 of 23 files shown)


src/main/java/org/broadinstitute/consent/http/resources/MatchMigrationResource.java
```diff
@@ -0,0 +1,67 @@
...
+  @GET
+  @Path("/population")
+  @Produces(MediaType.APPLICATION_JSON)
+  @RolesAllowed({Resource.ADMIN})
+  public Response getPopulation(@Auth DuosUser duosUser) {
+    try {
+      return Response.ok().entity(service.findPopulation()).build();
+    } catch (Exception e) {
+      return createExceptionResponse(e);
+    }
```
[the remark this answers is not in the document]
kevinmarete: Traced this and I don't think it applies here, so leaving the three catches as they are.

**The `e.getMessage()` echo is the fallback path, not the one these endpoints take.** `Resource` dispatches `UnableToExecuteStatementException` and `PSQLException` to `unableToExecuteExceptionHandler` (Resource.java:195-196), which returns a canned message from `vendorCodeStatusMap` rather than the exception's own. The echo at Resource.java:265-269 is only reached by exception types with no registered handler.

**DAR free text cannot reach these catches.** Every exception that could carry it arises inside `reprocessMatchesForPurpose`, which `MatchMigrationService` catches at lines 111 and 119 and never rethrows — deliberately, for exactly this reason ("The class, never the message: a failure raised while matching can quote a DAR's free text"). What reaches the resource is DAO failures on SQL that takes no bind parameters at all: `findPopulation`, `reconcile`, `snapshotAffectedMatches`, `snapshotAffectedRationales`, `findResolvablePurposes` and `findUnresolvablePurposes` are all parameterless.

**Returning a generic 500 instead would lose the PSQL-state to status mapping**, collapsing mapped statuses into a flat 500, and would depart from `.github/copilot-instructions.md`, which asks to reuse `Resource.createExceptionResponse`.

One residual I did check: `createExceptionResponse` logs the message before dispatch, so a Postgres constraint message carrying key values would land in the logs. Those are DAR reference ids, which this service already logs by id in its own failure lines and which appear in logs app-wide, so it is not new exposure.

For the record, the earlier finding on this PR — `readyForConstraints` not reaching the JSON — was correct and in fact wider than reported: `reconciles` and `blocksConstraints` had the same defect. Responses here serialize with Gson (`JerseyGsonProvider`), which reads fields, so the suggested `isX()` rename would not have worked either; all three are now record components, with a serialization test guarding the contract.

src/main/java/org/broadinstitute/consent/http/db/MatchMigrationDAO.java
```diff
@@ -0,0 +1,227 @@
...
+          WHERE dataset_id IS NOT NULL
+          GROUP BY purpose, dataset_id HAVING COUNT(*) > 1
+        ) dupes) AS duplicate_purpose_dataset_pairs
+      )
+      SELECT counts.*,
+             (counts.affected_matches > 0 OR counts.duplicate_purpose_dataset_pairs > 0)
+               AS blocks_constraints
+      FROM counts
+      """)
+  MatchMigrationPopulation findPopulation();
```
otchet-broad, line 96: Two quick points:
Would this benefit from some indexes being added?
We used to have strict rules about using String variables in SQL statements and refactored those because of Sonar complaints.  I see Sonar isn't complaining, but this may come back to bite us.  The concern stems from changing the string can result in SQL injection.

kevinmarete: **Indexes** — the snapshot tables carry their own (`idx_match_migration_snapshot_purpose`, `idx_match_migration_rationale_snapshot_match_id`), and the join side is already covered: `data_access_request.reference_id` has a unique constraint from changelog-consent-81.0.

On `match_entity` I don't think one helps. `AFFECTED_PREDICATE` is a disjunction of two `IS NULL`s and a `LOWER()` call, so no btree index can serve it — Postgres scans the table either way, and these queries are single-pass over a small table run a handful of times by an admin. The one index that would matter is `match_entity (purpose)`, which doesn't exist, but that's the application's own hot path (`findMatchesByPurposeId`, `deleteMatchesByPurposeId`) rather than anything specific to the migration, and #3059 retires this surface — so an index added here would be created and dropped for a few executions. Happy to raise it as its own ticket if you want it for the app paths.

**SQL built from String constants** — the injection failure mode is closed by the compiler rather than by convention, which is why I was comfortable with it. Annotation element values must be constant expressions, so a runtime value cannot be spliced into `@SqlQuery` at all; the moment someone swaps a constant for something dynamic it stops compiling. I checked rather than assuming:

```
@Q("SELECT 1 " + mutable) void nonFinalField() {}
                ^
error: element value must be a constant expression
```

So the SQL is fixed in the class file at compile time, which is also why S2077/S3649 don't fire — there is no dynamic construction to flag. On top of that, none of these queries take a bind parameter: `findPopulation`, `reconcile`, `snapshotAffectedMatches`, `snapshotAffectedRationales`, `findResolvablePurposes` and `findUnresolvablePurposes` are all parameterless, so no external input reaches them.

The constants are there so the definition of the affected set cannot drift between the population count, the snapshot and the reconciliation — if those three ever disagreed, the run would snapshot one set and reprocess another. That's the property I most wanted to protect in a one-shot migration. If you'd still rather see the SQL inlined per method, I'll do it, but it trades that guarantee for six copies of the same predicate.

fboulnois (commented):
Can we split this into smaller PRs? Proposed split:

1. Phase 2 execution, validation, and rollback documentation changes (~60 lines)
2. Read-only population report with (~350 lines):
  - `MatchMigrationDAO` with `AFFECTED_PREDICATE`, `NOT_ARCHIVED`, and `findPopulation`
  - `MatchMigrationPopulation`, including the derived `blocksConstraints` field
  - `MatchMigrationResource` with `GET /population` only
  - `ConsentApplication` and `ConsentModule` wiring, and `DAOTestHelper`
  - OpenAPI: the population path and schema, and their `api-docs.yaml` entries
  - Five `findPopulation` DAO tests and two resource tests
3. Snapshot tables and reconciliation (~490 lines):
  - The Liquibase changeset for `match_migration_snapshot` and `match_migration_rationale_snapshot`, and its `changelog-master.xml` include
  - `RESOLVABLE_PURPOSES`, `findResolvablePurposes`, `findUnresolvablePurposes`
  - `snapshotAffectedMatches`, `snapshotAffectedRationales`, `reconcile`
  - `SnapshotReconciliation`, including the derived `reconciles` field
  - `GET /reconciliation`, and its OpenAPI path and schema
  - Ten snapshot and reconciliation DAO tests, and a reconciliation resource test
4. Executing `run` (~560 lines):
  - `MatchMigrationService` with `run()` and `reprocessWithOneRetry`
  - `MatchMigrationRunReport`, and `MatchMigrationRunResult` with the derived `readyForConstraints`
  - `POST /run`, and its OpenAPI path and schema
  - Nine service tests and three resource tests

If this is too much, then it could make sense to combine 1 and 2.

kevinmarete:
@fboulnois split as you proposed, four PRs plus the constraints release, each stacked on the previous:

| | PR | Contents | Size |
| --- | --- | --- | --- |
| 1 | #3061 | Phase 2 execution, validation, and rollback documentation | +55/-1 |
| 2 | #3062 | Read-only population report | +479 |
| 3 | #3063 | Snapshot tables and reconciliation | +516 |
| 4 | #3064 | Executing `run` | +586/-8 |
| 5 | #3065 | Release C, unchanged from #3059 | +103/-1541 |

Two departures from your list. I kept the docs separate rather than taking the combine-1-and-2 fallback, since the plan describes endpoints across all three code PRs and reads as the context for reviewing them. And `findResolvablePurposes`/`findUnresolvablePurposes` moved from 3 to 4 with their two DAO tests — they select work for the run and nothing in 3 calls them; `RESOLVABLE_PURPOSES` stays in 3 because `snapshotAffectedMatches` embeds it.

Worth knowing before you review 3: its snapshot writers have no production caller until 4. That is inherent to splitting a migration surface — DAO tests exercise them — and the alternative is merging 3 and 4 back into ~1100 lines.

Verified the split changes nothing: #3064's tree was byte-identical to this branch and #3065's to #3059, before one review fix on top — the three OpenAPI path specs were missing a `summary`, had `operationId` dangling after `responses`, and their ids omitted the `dataset-id` path segment. 33 tests green.

This PR and #3059 closed on their own when I renamed their branches into the numbered sequence; GitHub retargeted #3059's base but dropped both head refs rather than following them. Nothing was lost, but the threads here can't be reopened, so please pick the review up on the new stack.

PR #3061, opened 2026-09-11
[DT-3942][1/5] Document the Phase 2 release sequence and validation

kevinmarete:
## Addresses

https://broadworkbench.atlassian.net/browse/DT-3942 — security risk: no

First of four, splitting #3058 on review feedback. Documentation only; nothing here changes behaviour.

## Summary

Records the Phase 2 release sequence for the match `dataset_id` migration, and the checks an operator makes either side of the run.

Three releases, because the constraints must not deploy until the run that clears them has happened. Their changeset halts on the same conditions the run reports, so it never applies a constraint the data cannot satisfy — but it does not stop the deploy: `ConsentApplication` logs a Liquibase failure and starts Jersey anyway. Releasing C early leaves an environment running C's code with the constraints silently unapplied, and C's election join no longer tolerates a null `dataset_id`, so unmigrated rows drop out of match reads while looking healthy. Quiet rather than loud, which is why the ordering is written down rather than assumed.

| Release | Contents | Gate to the next |
| --- | --- | --- |
| A (#3057) | Nullable `dataset_id`, dual write, dataset-correlated election join | Deployed everywhere |
| B (#3062, #3063, #3064) | Snapshot tables and the admin migration surface | `run` reports `readyForConstraints` |
| C (#3059) | Non-null and `(purpose, dataset_id)` uniqueness constraints; retire the surface | Confirmed good in prod |

Also records what to do about the figures drifting: the affected set drains on its own as DARs are re-matched, so the counts in this plan and in the ticket are illustrative and every decision is made from a fresh `population` call.

### Reviewer notes

- The endpoints this section names do not exist until #3063 and #3064. Landing the plan first is deliberate — it is the context for reviewing the three code PRs behind it.
- Rollback is an insert from the snapshot tables, which carry no foreign key to `match_entity` and so survive the deletion of the rows they describe. Restored rows return under new `match_id`s, which is why reconciliation keys on presence rather than identity.

### DIFF
--- a/docs/plans/data-use-primary-consistency-plan.md
+++ b/docs/plans/data-use-primary-consistency-plan.md
@@ -10,7 +10,7 @@ In progress. Ticket-by-ticket state:
 | 2. Canonical primary classification on Data Use writes | Done. `DataUsePrimaryClassifier`/`DataUsePrimaryValidator` back registration, admin Data Use replacement, and dataset-to-study conversion. |
 | 3. Explicit legacy and unsupported matcher behavior | Done. `DataUseMatcherV5` classifies before matching and abstains on Other-only, NONE/null, and MULTIPLE. |
 | 4. Normalize legacy records and reprocess affected matches | Done. |
-| 5. Replace alias-derived internal dataset references | In progress. Alias allocation moved to a database sequence (DT-3865). For matches (DT-3942), Phase 1 has landed: nullable `match_entity.dataset_id`, dual write, and a dataset-correlated election join. The reprocess and the non-null and uniqueness constraints follow once Phase 1 is deployed everywhere. |
+| 5. Replace alias-derived internal dataset references | In progress. Alias allocation moved to a database sequence (DT-3865). For matches (DT-3942), Phase 1 has landed: nullable `match_entity.dataset_id`, dual write, and a dataset-correlated election join. Phase 2a adds the snapshot tables and the admin migration surface that runs the reprocess. The constraints are release C, gated on that run reporting clean. |
 | 6. Align duos-ui with the canonical classification | Done in duos-ui (DT-3866). The Data Use translation collapse is an owned follow-up (DT-4008). |
 
 ## Objective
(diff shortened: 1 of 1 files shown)


kevinmarete:
Not taken: **resolve or exclude `archivedOrMissingPurposes` before C.** `readyForConstraints` already refuses while any skipped purpose remains — it is `reconciles() && !blocksConstraints()`, and a skipped purpose keeps a null `dataset_id`, so `blocksConstraints` is true. The finding assumes readiness tolerates the skipped set; it does not.

kevinmarete:
Second Copilot round addressed; all five threads resolved.

- **The precondition does not fail the deploy.** This was the important one and the claim was mine, not Copilot's. `ConsentApplication.run` catches `LiquibaseException`, logs it, and starts Jersey anyway, so releasing C early leaves an environment running C's code with the constraints silently unapplied — quieter and worse than the failed deploy this section promised. Rewritten to say the gate protects the data, not the rollout, and that the ordering is therefore a release-plan requirement.
- Skipped purposes now stated to block C outright, not merely "left for a decision" — they keep a null `dataset_id`, so they hold `readyForConstraints` false.
- Rollback before C is no longer described as two inserts: restored matches take new `match_id`s while the rationale snapshot references the pre-run ones, so the procedure now says to restore matches first and remap each rationale onto its new parent.
- Rollback after C now sequences the application back to B before restoring, since C's code drops the match-side `IS NULL` election-join tolerance and would not read restored null-`dataset_id` rows.
- `duplicatePurposeDatasetPairs` separated from the counters that drain on their own; it counts collisions between current rows the run never reprocesses.
- Rerun wording qualified: the snapshot never overwrites pre-run state, but it can grow as rows become affected between attempts.
- Cleanup marked as post-migration rather than a gated release, and widened to drop the legacy `consent` column and its `purpose_consent` constraint alongside the snapshot tables.

fboulnois (approved):
👍

otchet-broad (approved):
👍

PR #3062, opened 2026-09-11
[DT-3942][2/5] Report the match rows blocking the dataset_id constraints

kevinmarete:
## Addresses

**Stacked on #3061** and targeting `km-dt-3942-1-migration-plan-docs` so the diff shows only this step. Second of four, splitting #3058 on review feedback. Read-only: nothing here writes to a table.

## Summary

Counts the match rows that would make the `dataset_id` constraints fail, so a migration run is always scoped from a fresh figure rather than from an earlier report.

The affected set is defined by what the constraints will reject — no `dataset_id`, or a `v1`/`v2`/absent algorithm stamp — rather than by algorithm version alone. Versions are reported separately because the acceptance criteria count those populations, but they are not what selects the work.

Everything is counted in one statement, so the totals cannot disagree with each other, and recounted on every call because the affected set drains on its own as DARs are re-matched.

`GET /api/match/migration/dataset-id/population`, `@RolesAllowed(ADMIN)`, documented in OpenAPI.

### Reviewer notes

- `blocksConstraints` is derived in the query into a record component rather than exposed as a method. Responses serialize with Gson, which reads fields and drops computed accessors, so a method would not reach the operator reading the response.
- It also refuses on a pre-existing `(purpose, dataset_id)` collision, which this migration neither creates nor fixes. That is step 6 of the ticket, but it means the report can say "blocked" for something needing a separate decision.
- `MatchMigrationDAO` repeats the archived-status condition from `DataAccessRequestDAO.findByReferenceId` rather than sharing it. A reprocess resolves its DAR through that lookup, so these queries have to predict it exactly; if it changes they should be revisited deliberately rather than silently following it.
- `MatchMigrationService` takes only `Jdbi` here. It gains `MatchService` in #3064, where the first code that reprocesses anything lands.

### Testing

7 new tests, all passing. 5 DAO tests run against real Postgres and cover each counter separately — what the constraints reject, v2 and null stamps counted apart, archived purposes partitioned from resolvable ones, purposes holding more than one affected row, and pre-existing pair collisions. 2 resource tests cover the endpoint and its failure path. Spotless clean.

### DIFF
--- a/src/main/java/org/broadinstitute/consent/http/db/MatchMigrationDAO.java
+++ b/src/main/java/org/broadinstitute/consent/http/db/MatchMigrationDAO.java
@@ -1,0 +1,80 @@
+package org.broadinstitute.consent.http.db;
+
+import org.broadinstitute.consent.http.models.matchmigration.MatchMigrationPopulation;
+import org.jdbi.v3.sqlobject.config.RegisterConstructorMapper;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+
+/**
+ * Backs the one-off migration that replaces {@code match_entity.consent} with a real {@code
+ * dataset_id}. Retired with the rest of the migration surface once the constraints are on.
+ *
+ * <p>The affected set is defined by what the constraints will reject rather than by algorithm
+ * version alone: a row needs reprocessing if it has no {@code dataset_id}, or still carries a
+ * {@code v1} or {@code v2} stamp, or has no stamp at all. Versions are reported separately because
+ * the acceptance criteria count those populations, but they are not what selects the work.
+ *
+ * <p>{@code NOT_ARCHIVED} repeats the not-archived condition from {@code
+ * DataAccessRequestDAO.findByReferenceId} rather than sharing it. That lookup is what a reprocess
+ * resolves its DAR through, so the run has to predict it exactly; if it changes, these queries have
+ * to be revisited deliberately, not silently follow it.
+ */
+public interface MatchMigrationDAO {
+
+  String AFFECTED_PREDICATE =
+      """
+      (m.dataset_id IS NULL
+        OR m.algorithm_version IS NULL
+        OR LOWER(m.algorithm_version) IN ('v1', 'v2'))
+      """;
+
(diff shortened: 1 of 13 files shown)


kevinmarete:
Not taken: **`blocksConstraints` should ignore rows that already carry a `dataset_id`.** The premise is that a `v2`-stamped row with a `dataset_id` already satisfies release C. It does not: C's precondition rejects `dataset_id IS NULL OR algorithm_version IS NULL OR LOWER(algorithm_version) IN ('v1','v2')`, so a stamped-but-current row halts that deploy exactly like a null one. This flag is deliberately the same set as the gate it predicts.

kevinmarete:
Second Copilot round addressed. Both were fair coverage gaps:

- `testFindPopulationDoesNotBlockConstraintsWhenNothingIsAffected` — the clear state was never asserted, so a regression that always returned `true` would have passed. It is the release gate, so it should be provable in both directions.
- `testPopulationResponseCarriesTheDerivedGate` — asserts `blocksConstraints` survives Gson serialization. The equivalent assertion existed for the run endpoint in #3064 but not for this one, and Gson dropping the field is exactly the failure the record component exists to prevent.

fboulnois (approved):
👍

src/main/java/org/broadinstitute/consent/http/db/MatchMigrationDAO.java
```diff
@@ -0,0 +1,80 @@
...
+   * call because the affected set drains as DARs are re-matched.
+   */
+  @RegisterConstructorMapper(MatchMigrationPopulation.class)
+  @SqlQuery(
+      """
+      WITH affected AS (
+        SELECT m.match_id, m.purpose, m.dataset_id, m.algorithm_version
+        FROM match_entity m
+        WHERE """
+          + AFFECTED_PREDICATE
```
otchet-broad: I'm not sure about others, but in a review, it would be easier to review with the entire SQL statement instead of having to compute the substitution in my head.

otchet-broad (approved):
Nice work.  Thanks for the test coverage!

PR #3063, opened 2026-09-11
[DT-3942][3/5] Snapshot the match rows a migration run can destroy

kevinmarete:
## Addresses

**Stacked on #3062** and targeting `km-dt-3942-2-match-population` so the diff shows only this step. Third of four, splitting #3058 on review feedback.

## Summary

The safety net for the migration run: tables that capture what a run can destroy, the SQL that fills them, and the query that proves the capture balances against what the run left behind.

`match_migration_snapshot` and `match_migration_rationale_snapshot` are created empty and populated when the run executes, because the affected set drains on its own as DARs are re-matched. Neither carries a foreign key to `match_entity` by design: the run deletes the very rows captured, so a reference would either block the delete or cascade the evidence away.

**Wider than the affected set, on purpose.** A reprocess deletes every match row for a purpose and rebuilds from the DAR, so a purpose holding both a legacy row and an already-current one would lose the current row with no way back if only the legacy one were captured. The snapshot takes every row of a purpose a run will touch, plus any affected row elsewhere — the latter covering skipped purposes, which are never deleted.

**Re-runnable.** `ON CONFLICT DO NOTHING` keeps the first capture authoritative, so a re-run after a partial failure cannot overwrite pre-migration state with post-migration state, and does not fail on the rows it already holds.

**Reconciliation reads presence, not identity.** A reprocess deletes and re-inserts, so replacements arrive under new ids: a snapshotted row that is gone was handled, one still present was not. `reconciles` asserts two equalities rather than one, so a run that skipped more than it meant to fails here instead of looking complete.

`GET /api/match/migration/dataset-id/reconciliation`, `@RolesAllowed(ADMIN)`, documented in OpenAPI.

### Reviewer notes

- The snapshot writers have no production caller until #3064. They are exercised by DAO tests here; splitting the migration surface at all means the schema and the code that fills it land before the code that calls it.
- Rationales carry their own `match_rationale.id` rather than being keyed on text. Two rationales identical in text are still two rows, and a restore has to produce two — which is also what makes the insert re-runnable.
- `SnapshotReconciliation.purposesHoldingMatches` is named for the state, not the cause: a skipped purpose still holds its original rows and counts there without having been rebuilt. Called out because an operator reads these numbers.
- Keep both tables until the migration is confirmed good in production; #3059 drops them with the rest of the surface.

### Testing

12 new tests, all passing. 10 DAO tests run against real Postgres: six on the snapshot — affected rows and their rationales, a current row sharing a reprocessed purpose, a current row on a purpose that is not reprocessed, two rationales sharing their text, re-runnability, and rationales of uncaptured matches — and four on reconciliation, covering a row still present, a row replaced under a new id, a purpose rebuilt to nothing, and a skipped row that is still affected. 2 resource tests cover the endpoint and its failure path. Spotless clean.


kevinmarete:
Copilot review addressed. Fixed:

- The changeset's lifecycle comment said the snapshot tables are "dropped with the migration surface". They are not — #3065 keeps them deliberately and a later cleanup drops them. Corrected.

Not taken as code:

- **Snapshot writes need a transaction or advisory lock.** The two writes are not atomic with each other, which is true, but making them so would not close the window the finding is really about. `snapshotAffectedRationales` joins against `match_migration_snapshot`, so a match inserted between the two calls is captured by neither and its rationales are consistently absent — there is no torn capture. The real exposure is wider than these two statements: a row arriving on a to-be-reprocessed purpose any time between the snapshot and that purpose's rebuild is deleted uncaptured, and only a lock held across the whole run would prevent it.

  That lock would be a new pattern — there are no advisory locks anywhere in this codebase — for an operator-triggered one-off whose window is minutes. Recorded the constraint instead, in #3064's service javadoc and the `run` OpenAPI description: the run is not serialized against DAR activity and should be executed in a quiet window. Happy to add a lock if you would rather have one, but it should be a deliberate choice rather than a side effect of this PR.

kevinmarete:
Second Copilot round. Both remaining threads are the same concurrency argument from the first round, restated with new mechanisms, so I have left them open rather than resolving them — they are a standing design question, not something this PR silently ignores.

The new specifics are fair: a reprocess deletes `match_rationale` before the match row, so a rationale can be missed between the two capture statements; and the purpose selector is evaluated before the runner fetches its work, so a DAR unarchived in that gap gets only its affected rows captured while the rebuild deletes all of them.

Both are windows in the same larger one — the interval between capture and rebuild, which no arrangement of these two statements closes. Only a lock held across the whole run does, and that would be this codebase's first advisory lock, introduced by a migration PR for an operator-triggered one-off. The constraint is recorded in #3064's service javadoc and the `run` OpenAPI description: not serialized against DAR activity, run it in a quiet window.

If the team wants the lock, I would rather do it deliberately as its own change against `MatchService` than fold it in here. Happy to be overruled.

fboulnois (approved):
👍

src/main/java/org/broadinstitute/consent/http/db/MatchMigrationDAO.java
```diff
@@ -77,4 +93,107 @@ GROUP BY purpose, dataset_id HAVING COUNT(*) > 1
...
+   * match row for a purpose and rebuilds from the DAR, so on a purpose holding both a legacy row
+   * and an already-current one, capturing only the legacy row would let the run delete the other
+   * with no way back. So: every row of a purpose that will be reprocessed, plus any affected row
+   * elsewhere - the latter covering unresolvable purposes, which are skipped and never deleted.
+   *
+   * <p>{@code ON CONFLICT DO NOTHING} keeps the first capture authoritative: a re-run after a
+   * partial failure must not overwrite pre-migration state with post-migration state, and must not
+   * fail on the rows it already holds.
+   */
+  @SqlUpdate(
```
otchet-broad, line 110: This inspired me to look at how matches are preserved as the system around them can change.  A gap I see is that we can be changing the code for matches and have no way to get back to answer the question: what value was shown to the DAC when the election was created and the user voted.  This is probably a question for product.  The risk we have is that we're not memorializing the history here.  Could the value have changed between when the user first viewed the information and when they actually voted?  Do we need to alert them?

This does not block this PR, but it does leave the question for product.

otchet-broad (approved):
👍

src/main/java/org/broadinstitute/consent/http/db/MatchMigrationDAO.java
kevinmarete: Agreed this is a product question rather than something this PR should decide. The snapshot here is deliberately narrow — it exists so a one-off migration can be rolled back, and it is dropped once the migration is done, so it is not an audit trail and does not record what a DAC member saw at election creation versus at vote time.

Raised as [DT-4098](https://broadworkbench.atlassian.net/browse/DT-4098) so it does not live only in this thread. Filed outside DT-3860, since vote provenance and DAC notification are a different concern from the data use classification work.

PR #3064, opened 2026-09-11
[DT-3942][4/5] Run the match dataset_id migration and report on it

kevinmarete:
## Addresses

**Stacked on #3063** and targeting `km-dt-3942-3-match-snapshot` so the diff shows only this step. Last of four, splitting #3058 on review feedback; #3058 carries the original review threads and closed when its branch was renamed into this stack.

## Summary

Executes the migration: rebuilds every affected purpose so its matches carry a real `dataset_id` at the current algorithm version, and reports enough for an operator to see that the run accounted for everything it touched.

Reprocesses rather than backfills. The legacy `consent` values are consent UUIDs whose dataset mapping was dropped in 2024, so no mapping can be derived without guessing — endorsed on review, and it is what the application already does whenever a DAR is edited. A purpose whose DAR carries no dataset associations rebuilds to nothing, which deletes its rows; that is the intended outcome for it.

**Snapshot first.** The capture from #3063 runs before anything is rebuilt, which is what makes the run reversible.

**Skip guard.** `findByReferenceId` filters archived DARs, so a purpose whose DAR is archived or missing would rebuild to nothing and lose its rows for no reason. Those purposes are skipped and reported by id, never reprocessed.

**Restartable.** Each purpose is rebuilt in its own transaction, so a failure leaves that purpose's existing matches in place. One retry apiece — a transient database error is worth a second attempt, a programming error is not worth more than that — then the run continues, reporting `failedPurposeIds` so a rerun can be scoped.

**Redacted.** Failures carry a purpose id and an exception class, never a message, which can quote a DAR's free text.

`POST /api/match/migration/dataset-id/run`, `@RolesAllowed(ADMIN)`, documented in OpenAPI. Returns the population either side of the run plus the reconciliation.

### Reviewer notes

- `readyForConstraints` is the single gate, derived into a record component for the same Gson reason as the other two: it requires both that the reconciliation balances and that nothing affected remains beyond what was skipped.
- It also refuses on a pre-existing `(purpose, dataset_id)` collision, which this migration neither creates nor fixes. That is step 6 of the ticket, but it means a run can report "not ready" for something needing a separate decision.
- The skipped set has no owner yet. Production has zero archived-DAR purposes today, so it may stay empty — but any that appear keep a null `dataset_id` and would halt release C's precondition. Worth deciding with the run results rather than now: delete them, or exclude them with a partial constraint.
- `findResolvablePurposes` returns an unbounded list. It is bounded in practice by the affected population — a few hundred ids, shrinking — so it is not paginated.
- The endpoints keep `Resource.createExceptionResponse`, which `.github/copilot-instructions.md` names as the pattern to reuse and which returns `e.getMessage()`. Called out because the run's own logging deliberately redacts: these are admin-only routes, so the audience for a message is an operator who can already read DAR content, but say so if you would rather they returned a bare 500.

### Testing

14 new tests, all passing; 33 across the stack. 9 service tests cover snapshot-before-rebuild ordering, the skip guard, retry-once, continuing past a failure, the snapshot totals, and both ways `readyForConstraints` can be refused. 2 DAO tests against real Postgres cover the resolvable/unresolvable partition, including a match whose purpose has no DAR at all. 3 resource tests cover the endpoint, its failure path, and that all three derived gates survive Gson. Spotless clean.


kevinmarete:
Copilot review addressed.

- **Rebuild is not serialized with DAR updates.** Correct, and worth stating plainly: `MatchService.reprocessMatchesForPurpose` reads its DAR and computes matches *outside* the transaction it then deletes and inserts in. A DAR edited between those two points loses its own rebuild to whichever writer commits last.

  That race predates this PR and belongs to every caller of that method, including the ordinary DAR-edit path — this migration does not introduce it, though running it in bulk does raise the exposure. Fixing it properly means either making resolvability and replacement atomic inside `MatchService`, which changes behaviour for every existing caller, or holding a lock across the whole run, which would be the first advisory lock in this codebase.

  Neither belongs in a migration PR. Recorded the constraint where an operator and the next reader will see it — the service javadoc and the `run` OpenAPI description now say the run is not serialized against DAR activity and should be executed in a quiet window, with the reports re-checked afterwards. If the underlying race is worth closing, it is its own ticket against `MatchService`.

kevinmarete:
Second Copilot round addressed. Fixed — and these were real documentation defects, not just wording:

- `MatchMigrationRunReport.failedPurposeIds` and its OpenAPI description both said the ids let a rerun be scoped. They do not; `run` takes no arguments. Now described as investigation only. Same correction already made in #3061.
- `MatchMigrationRunResult.readyForConstraints` and its schema said "no affected row is left except the ones deliberately skipped", which reads as though skipped purposes are tolerated. They are not: they keep a null `dataset_id`, count as affected, and hold the gate false. The javadoc now says so explicitly. Worth flagging that this is the same behaviour I relied on to push back on a first-round finding against #3061 — the behaviour was right, the documentation of it was wrong.
- Added `testRunIsReadyForConstraintsWhenNothingIsLeftBehind`. The gate was only asserted on its two false paths.

Left open: the serialization thread on `reprocessMatchesForPurpose`, which is the concurrency question discussed on #3063.

fboulnois (approved):
👍

otchet-broad (approved):
👍

PR #3065, opened 2026-09-11
[DT-3942][5/5] Apply the match dataset_id constraints and retire the migration surface

kevinmarete:
## Addresses

**Stacked on the run PR** and targeting `km-dt-3942-4-match-migration-run` so the diff shows only this release. GitHub retargets it as the stack merges. Reopened from #3059, which closed when its branch was renamed into the split stack; the content is unchanged.

**Do not merge or deploy this until the migration run has cleared every environment.** The precondition stops the constraints being applied to data that cannot satisfy them, but it does not stop the deploy — see below.

## Summary

Release C closes the migration. Every match row identifies its dataset by foreign key, and a purpose holds at most one match per dataset.

### Gated, not trusted

The changeset's precondition halts on exactly the conditions `POST /api/match/migration/dataset-id/run` reports: any row without a `dataset_id`, any row still stamped `v1`/`v2` or unstamped, and any duplicate `(purpose, dataset_id)` pair. Releasing this early therefore does not apply a constraint the data cannot satisfy. It does not fail the deploy either: `ConsentApplication.run` catches the `LiquibaseException` and starts Jersey regardless, so the environment comes up running this release with the constraints unapplied and the migration endpoint already retired. Worse, `onFail="HALT"` halts the whole changelog rather than the one changeset, so every changeset added after this one stops applying in that environment until the data is fixed. The gate protects the data, not the rollout; the ordering is a release-plan requirement. Making Liquibase failure abort startup would change deploy behaviour for every changeset in the service and is not something this migration needs; deploy ordering covers the risk here. Flagging it rather than fixing it.

Its failure message says to deploy the *previous* release first, because this one removes the endpoint that clears the precondition — a message telling an operator to call an endpoint absent from the build they are deploying would be worse than no message.

This follows the gating convention DT-3865 established in this epic for the alias sequence.

### Keeps the election join's null tolerance

The match-side `IS NULL` branch looks dead once the constraints are on, and an earlier revision of this PR removed it. It stays: a halted precondition still starts the app, so an unmigrated row would silently drop out of every election that does carry a dataset id. The comment above the clause now records why, so it is not deleted as unreachable next time. `election.dataset_id` stays nullable and keeps its own branch.

### Retires the migration surface

Resource, service, DAO, four models, three paths, three schemas, three test classes, and the Guice/Jersey registrations — the same clean delete DT-4009 performed on the Ticket 4 surface it was modelled on.

**The snapshot tables deliberately stay.** They are the rollback path for the run, and this is the release most likely to need one; dropping a backup in the same release that applies an irreversible constraint removes the recovery exactly when it is wanted. A later cleanup drops them once production has settled. The plan doc's release table now lists that cleanup explicitly.

### Testing

`MatchDAOTest` 14/14, `MatchServiceTest` 15/15, `MatchResourceTest` 5/5. The legacy null-`dataset_id` read test is replaced by two that assert the database now refuses that shape: a match with no dataset, and a second match for the same `(purpose, dataset_id)`. A third covers the election join's surviving null tolerance, which those constraints otherwise make unreachable — it drops the non-null constraint inside a transaction to reach the halted-precondition state, then rolls back so the DDL does not outlive the test. Removing the `IS NULL` branch fails that test and only that test. Because the DAO suite runs liquibase against a real Postgres, a green run is also evidence the precondition passes and both constraints apply cleanly.

### Reviewer notes

- `addNotNullConstraint` and `addUniqueConstraint` each take an ACCESS EXCLUSIVE lock and scan the table. `match_entity` is small enough for that to be brief; the changeset comment notes to revisit with `NOT VALID` plus `VALIDATE` if it grows by an order of magnitude.
- Both statements are in one changeset, so a failure of either rolls both back rather than leaving the column non-null without its uniqueness rule.
- The two still-open questions from #3064 both land here: a pre-existing duplicate pair halts this precondition, and so does any skipped archived-DAR purpose. Neither exists in production today, but both need a decision before this deploys if they appear in the run.

### DIFF
--- a/docs/plans/data-use-primary-consistency-plan.md
+++ b/docs/plans/data-use-primary-consistency-plan.md
@@ -10,7 +10,7 @@ In progress. Ticket-by-ticket state:
 | 2. Canonical primary classification on Data Use writes | Done. `DataUsePrimaryClassifier`/`DataUsePrimaryValidator` back registration, admin Data Use replacement, and dataset-to-study conversion. |
 | 3. Explicit legacy and unsupported matcher behavior | Done. `DataUseMatcherV5` classifies before matching and abstains on Other-only, NONE/null, and MULTIPLE. |
 | 4. Normalize legacy records and reprocess affected matches | Done. |
-| 5. Replace alias-derived internal dataset references | In progress. Alias allocation moved to a database sequence (DT-3865). For matches (DT-3942), Phase 1 has landed: nullable `match_entity.dataset_id`, dual write, and a dataset-correlated election join. The reprocess and the non-null and uniqueness constraints follow once Phase 1 is deployed everywhere. |
+| 5. Replace alias-derived internal dataset references | In progress. Alias allocation moved to a database sequence (DT-3865). For matches (DT-3942), Phase 1 has landed: nullable `match_entity.dataset_id`, dual write, and a dataset-correlated election join. Phase 2a adds the snapshot tables and the admin migration surface that runs the reprocess. The constraints are release C, gated on that run reporting clean. |
 | 6. Align duos-ui with the canonical classification | Done in duos-ui (DT-3866). The Data Use translation collapse is an owned follow-up (DT-4008). |
 
 ## Objective
@@ -226,8 +226,8 @@ failure. The audit utility must not log the raw JSON. Classification rules:
 The `match_usage` CTE deliberately isolates the legacy alias-derived join needed to reconcile the
 current schema. It is not the target design. Before relying on match counts, separately report null,
 noncanonical, duplicate-mapping, and unmatched `match_entity.consent` values; do not silently omit
-them. Ticket 5 backfills and validates a real `dataset_id` relationship before application reads or
-writes stop using the legacy text column.
+them. Ticket 5 reprocesses affected matches and validates a real `dataset_id` relationship before
+application reads or writes stop using the legacy text column.
 
 **Acceptance criteria**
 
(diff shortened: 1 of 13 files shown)


kevinmarete:
Not taken: **remove the null `election.dataset_id` branch.** The finding correctly spotted that the code and the PR description disagreed — the description was the wrong half and is now corrected. An election that never recorded a `dataset_id` is legitimate historical data, and dropping the branch would silently remove matches from those elections.

Not taken here: **plan doc instructs calling retired endpoints.** Those instructions are release B's. #3061 now labels the sections by release rather than deleting them.

kevinmarete:
Second Copilot round. Both findings are correct and both threads stay open, because neither is fully closed by this PR.

**`onFail="HALT"` does not stop the deploy.** Verified: `ConsentApplication.run` catches the `LiquibaseException` from `initializeLiquibase`, logs it, and starts Jersey. So an environment given release C early runs C's code with the constraints never applied — and, as the finding notes, without the endpoint that would clear the gate. My changeset comment and the plan both claimed this "fails the deploy". That was wrong and is now corrected in both places: the precondition protects the data, not the rollout, and the ordering is a release-plan requirement this file cannot enforce.

Making a Liquibase failure abort startup or fail readiness would change deploy behaviour for every changeset in the service, so it is not something to slip into a migration PR. It needs its own ticket and a deliberate decision. Thread left open so that decision is visible rather than buried.

**`purpose_consent` is still active.** Also correct. `Match` sets `consent` to `dataset.getDatasetIdentifier()`, so post-migration the legacy unique constraint on `(purpose, consent)` enforces the same rule as the new `(purpose, dataset_id)` — redundant rather than contradictory, since identifier and id are 1:1. It is dead weight the migration should eventually remove, but removing it properly means dropping the `consent` column too, and that column is still in `MatchDAO.insertMatch`. That is the cleanup release, which #3061 now lists explicitly: drop the snapshot tables, the `consent` column, and `purpose_consent` together. Doing it here would widen C from "apply constraints" to "remove the legacy column", which is the change most likely to need its own rollback.

fboulnois (approved):
👍

otchet-broad (commented):
Note from Claude about something there's a good bit of discussion already on in this PR: 

 Removing the match-side IS NULL tolerance from the election join creates a silent-data-loss mode, and it buys nothing this release.

MatchDAO.java:48-54 now reads AND (e.dataset_id IS NULL OR e.dataset_id = match_entity.dataset_id). The justification — "match_entity.dataset_id is NOT NULL from this release" — is only true if the precondition passed, and the precondition exists precisely because it might not. When it halts: Liquibase throws, ConsentApplication.java:141-145 logs and starts Jersey anyway, dataset_id stays nullable, and for every surviving legacy row NULL = e.dataset_id is NULL → the row drops out of findMatchesForLatestDataAccessElectionsByPurposeIds against any election that does carry a dataset id. Matches disappear from match reads with no error logged anywhere.

I think the important point to consider is that last sentence: Matches disappear from match reads with no error logged anywhere.

kevinmarete:
Taken — the match-side null tolerance is back in `findMatchesForLatestDataAccessElectionsByPurposeIds` as of e676cfaa. You were right that it buys nothing to remove: the branch is unreachable if the precondition passed, and a guard against silent row loss if it halted, which `ConsentApplication` does not treat as fatal.

I also rewrote the comment above the clause. It previously asserted `match_entity.dataset_id` is NOT NULL from this release, which is the premise your note disproves — it now records why the branch has to survive looking dead, so it does not get deleted as unreachable next time.

One gap worth flagging rather than burying: this ships without direct test coverage. The case is only reachable when the constraints failed to apply, and `DAOTestHelper` runs the full changelog, so a match row with a null `dataset_id` cannot be inserted to exercise it — `testAMatchCannotBeStoredWithoutADataset` in this PR asserts exactly that. `MatchDatasetIdConstraintsMigrationTest` builds a pre-migration schema by hand and could represent the halted state, but it is raw JDBC with no JDBI wiring, so covering the join there is a larger change than the fix. Happy to do it if you think the clause should not go in uncovered.

otchet-broad (approved):
👍

PR #3069, opened 2026-09-15
[DT-3800] Activate researcher status from an institutional email domain

kevinmarete:
## Addresses

https://broadworkbench.atlassian.net/browse/DT-3800 — security risk: low

## Summary

A user whose email domain belongs to an institution is now an active researcher without waiting for a signing official. Active status today means holding a library card — that is how `SigningOfficialDashboardDAO` counts active researchers — so registration issues one. `UserService.createUser` resolves an eligible signing official and hands it to `UserServiceDAO.createUser`, which inserts the card in the same transaction as the user row.

Registration is the only issuance point. `InstitutionAndLibraryCardEnforcement` still runs on every authenticated request and over every user when an institution's domains change, and still only removes cards, so a card revoked through `DELETE /api/libraryCards/{id}` stays revoked. The flip side is that no user already in the system is activated: the mechanism reaches registrations from here onward.

That is also why issuance is inside the create transaction rather than best-effort after it. Nothing retries a card that fails to appear, and a user who committed without one could not re-register past the duplicate-email check, so a failure has to roll the registration back and let the retry re-register cleanly.

The issuer is constrained: `needsLibraryCardRemovedForUser` deletes any card whose issuing SO's own email domain does not resolve back to the institution, so `UserDAO.findLibraryCardIssuerByInstitution` applies that same rule when picking one. Among eligible SOs it takes whoever has issued the most cards, so the card names the institution's most active signing official rather than an arbitrary one, and no eligible SO leaves the researcher inactive rather than issuing a card the next pass deletes.

### Pre-existing bug fixed here

`deleteAllLibraryCardsByUser` deleted `library_card` rows without clearing `lc_daa`, whose foreign key is `NO ACTION`, so removing a card that had any DAA raised a constraint violation — verified against Postgres. Enforcement runs that removal on the `AuthorizationHelper` path and only `NotFoundException` is caught, so an affected researcher was served a 500 on every API call. Now clears the relations in one statement, as `deleteLibraryCardById` already did, and targets `user_id` alone: the `create_user_id`/`update_user_id` predicate dated to #1984's FK removal and had no live caller, so clearing one SO's card also deleted every card they had issued. Independent of this ticket, but this ticket widens who holds a card, so it is fixed here rather than deferred to decoupling ticket 5a.

### Decided with the PO

Auto-creating a Library Card is confirmed as the interim mechanism, since persisted researcher status does not exist yet and the intent is to replace or rename the card as researcher status. On the reservation that a card "is currently an object that may have content": an auto-issued card has none — `lc_daa` is empty until an SO attaches an agreement, and every authorization read inner-joins it, so the card authorizes no dataset.

Four further gates key on card *presence* rather than `lc_daa` — `DatasetDAO.getApprovedDatasets`, `ResearcherDashboardDAO`, `ElectionDAO.findElectionsWithCardHoldingUsersByElectionIds`, and `TDRService`'s approved-users list. All four filter views of *already-approved* DARs, which a user being registered has none of, so confining issuance to registration keeps them empty for this population in practice. They still matter to decoupling ticket 4, which has to preserve presence semantics when these gates move to persisted status.

Two later calls from the PO shaped the current design:

- **First login, not every sign-in.** Issuance was originally in `handleUserWithInstitutionInMap`, which runs on every authenticated request and would have undone an explicit revocation on the user's next call.
- **Attribution to the most likely voucher.** An auto-issued card records a real SO in `create_user_id`, the "who vouched" field in the SO and admin views, so ranking by cards already issued picks the one most plausibly behind it. A sentinel issuer was rejected because attributing to a real SO of the institution is what keeps `needsLibraryCardRemovedForUser` from deleting the card on the next pass.

**Consequence for the decoupling plan:** "issuing a card no longer activates anyone" is wrong as written for the registration path — at Phase 3 it would silently retire this activation. Decoupling ticket 9 must set researcher status on a domain match at registration, not only clear it on a mismatch; recorded under *Institution and domain enforcement*.

Two smaller calls, each reversible in a line:

- **No issuance email** — the template asserts an SO acted.
- **Not role-filtered** — SOs and admins at a mapped domain also move to active in the SO dashboard counts.

Redacted users are unaffected: `redacted_<base64>` has no `@`, so it matches no domain.

### Testing

Full suite green — 4681 tests, 0 failures. The DAO tests run against real Postgres, which is what confirms the domain-match SQL and the issuer ranking. Also stabilised `testAsyncEnforceInstitutionAndLibraryCardRulesForAllUsers`, a pre-existing flake unchanged on `develop`: `verify(spy, timeout(...))` returns when the method is entered, so Mockito's strict-stub check raced the body's last call.


fboulnois (approved):
A few suggestions:

**1. Auto-issuance silently undoes card revocation**
`src/main/java/org/broadinstitute/consent/http/service/feature/InstitutionAndLibraryCardEnforcement.java:162`

Auto-issuance runs on every authenticated request, so it re-creates a card that an SO or admin just revoked via `DELETE /api/libraryCards/{id}`, and it nullifies the institution-mismatch removal rule.

**2. `deleteAllLibraryCardsByUser` now deletes other users' cards and their DAA pre-authorizations**
`src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java:211`

The statement targets `create_user_id`/`update_user_id` in addition to `user_id`, so removing one SO's card also deletes every card that SO issued and the associated `lc_daa` rows. Previously the FK aborted the statement; now it is silent loss of other users' DAA pre-authorizations. The auto-issue path then re-creates empty cards that still pass the presence-only gates.

**3. Auto-issued cards attribute an SO who took no action**
`src/main/java/org/broadinstitute/consent/http/service/feature/InstitutionAndLibraryCardEnforcement.java:183`

The auto-issued card records an arbitrary SO in `create_user_id`, which is the field used as "who vouched" in both the API/SO views and `needsLibraryCardRemovedForUser`. That SO performed no action and is never notified (no email is sent).

**4. Unbounded repeated query on every request for users who can never be issued a card**
`src/main/java/org/broadinstitute/consent/http/service/feature/InstitutionAndLibraryCardEnforcement.java:176`

For a card-less user at an institution with no domain-matching SO, `findLibraryCardIssuerByInstitution` runs on every authenticated request and can never succeed. Same for any user permanently conflicting on `uk_lc_user_email`.

kevinmarete:
Thanks @fboulnois. Taking (2) and (4) as defects, fixed in 6d2983b3 and e5caa315: `deleteAllLibraryCardsByUser` now targets `user_id` only, so clearing one SO's card no longer deletes the cards they issued or those cards' `lc_daa` rows (the `create_user_id`/`update_user_id` predicate dates to #1984's FK removal and has no live caller — both callers mean "clear this user's card"); and short-TTL negative caches stop both permanently-failing retries you flagged — the issuer lookup for an institution with no eligible SO, and issuance for a user whose email is held by another user's card. Expiry bounds how long the fix for either goes unnoticed.

(1) and (3) are the feature's design rather than defects, so they need a product call. @jlaw-codes, two questions:

1. **Revocation.** Auto-issuance runs on every authenticated request, so deleting a card via `DELETE /api/libraryCards/{id}` is undone on that user's next request whenever their email domain maps to an institution with an eligible SO. Should a domain match make a user unconditionally active, or should an explicit revocation stick and suppress re-issuance?

2. **Attribution.** An auto-issued card records a domain-matching SO in `create_user_id` — the "who vouched" field surfaced in the SO and admin views — for someone who took no action and gets no email. Is that acceptable for now, or should auto-issued cards be distinguishable from SO-issued ones? Note that attributing them to a real SO of the institution is what keeps `needsLibraryCardRemovedForUser` from deleting them on the next pass, so a sentinel issuer means changing that removal rule too.

kevinmarete:
Both answered, implemented in 08364ddb.

**1. Revocation** — first login only, not every sign-in. Issuance moved out of `handleUserWithInstitutionInMap` into registration, so enforcement is back to only removing cards and a revoked card stays revoked without needing a suppression flag. The trade-off is that no existing user is activated, and mapping a new domain to an institution no longer activates its current researchers — the mechanism reaches registrations from here onward.

That removed the retry that used to cover a failed insert, so issuance now happens inside `UserServiceDAO.createUser`'s transaction: a failure rolls back the user row and the retry re-registers cleanly, rather than committing a user who is permanently inactive and cannot re-register past the duplicate-email check. `insertLibraryCardIfAbsent` and the two negative caches went with it.

**2. Attribution** — stays a real SO, now the institution's most active one. `findLibraryCardIssuerByInstitution` ranks eligible SOs by cards already issued rather than taking the lowest user id, so the `create_user_id` shown in the SO and admin views names whoever most plausibly vouched. A sentinel issuer would have meant changing the removal rule, so it was not taken.

otchet-broad:
I remain concerned that the library card issuance is attributed to a user **who didn't perform this action.**. If an audit of our system were to occur, how would we be able to state if a user took this action or the system did?  I think we need a way to do that so that we don't misattribute this.  Alternatively, we stage the removal of the library card concept ahead of this work. @jlaw-codes - it would be good to get your thoughts on this.

otchet-broad:
Codex's thoughts on the matter:

otchet-broad:
The current implementation creates a real attribution risk.

  Automatic issuance writes the selected SO’s user ID into library_card.create_user_id, exactly as manual issuance does. There is no flag or audit event distinguishing “system-selected SO”
  from “SO actually issued this card.”

  Downstream consequences include:

  - The API describes createUserId as the “user who created the Library Card record” (schema). Admin or UI consumers can therefore present that SO as the issuer.
  - Enforcement treats that SO as the actual issuer and uses their email domain to decide whether the card remains valid (source).
  - GA4GH passport claims describe card-backed status as asserted by = so, with the automatic card’s creation timestamp (AffiliationAndRole, ResearcherStatus). That can communicate externally
    that an SO vouched for the researcher.

  - The issuer-selection query counts automatic cards as cards “issued” by that SO. Because the most prolific issuer is selected, automatic attribution reinforces itself and may concentrate
    many system actions under one SO’s identity (query).

  There is one limitation: the automatic path does not add DAA assignments, and later DAA ADD/REMOVE actions have their own actor-specific audit records. Thus the SO should not be blamed for
  a particular DAA authorization solely because they appear on the card.

  But for card issuance and researcher-status provenance, yes—the data says the selected SO created or vouched for the card even though the system did it. The PR records this as an
  intentional product decision, but it is not truthful audit attribution. A safer design would record a system actor or an explicit issuance_source=AUTO_DOMAIN_MATCH, while separately
  retaining the eligible SO association needed by enforcement.

kevinmarete:
@otchet-broad Confirmed — there is no discriminator. An auto-issued card writes the selected SO into `create_user_id`, the same field manual issuance writes and the one `LibraryCard.yaml` documents as "ID of user who created the Library Card record", so nothing in the data or the API separates the two. On the downstream consequences: the Passport flip is real (`AffiliationAndRole.by()` returns `system` without a card and `so` with one), though `ResearcherStatus.by()` already returns `so` unconditionally and does not change here; enforcement does resolve `create_user_id`'s domain to keep the card valid; and the self-reinforcement is the sharpest of them — the ranking subquery at `UserDAO.java:547` counts auto-issued cards too, so the chosen SO's count grows with each registration and attribution concentrates on one person by construction. Your DAA caveat holds: `lc_daa` stays empty on these cards and DAA actions keep their own actor records.

@jlaw-codes this is Monday's attribution question with a sharper edge on it — "most plausibly vouched" is still not "vouched". Two options:

1. **Distinguish auto-issued cards now** — an `issuance_source` column, and `by()` returning `system` for them. Cost: `needsLibraryCardRemovedForUser` has to stop validating a card through `create_user_id`'s domain and instead ask whether the institution has *an* eligible SO, which changes the removal rule for every existing card, not just auto-issued ones.
2. **Stage DT-3800 behind the decoupling**, as @otchet-broad suggests. `researcher_status_audit` is where provenance belongs — a system actor is a row there rather than an overloaded issuer field — and ticket 9 already has to set status on a domain match at registration. Cost: auto-activation waits for Phase 3.

I lean to (2); (1) builds a discriminator that Phase 3 discards and pays for it by loosening the removal rule in the meantime. If we ship as-is instead, the one change I would make first is keeping auto-issued cards out of the issuer ranking so attribution stops concentrating — `lc_daa` presence is the only signal available to separate the two, so that is a mitigation rather than a fix.

jlaw-codes:
Agree, seems best to wait for the fuller implementation

kevinmarete:
Closing this. @jlaw-codes agreed with @otchet-broad that activation should wait for the fuller researcher-status implementation, so this mechanism is not shipping.

The blocker is attribution. An auto-issued card writes a real SO into `create_user_id` — the same field manual issuance writes, and the one `LibraryCard.yaml` documents as the card's creator — so nothing in the data or the API separates a card the system issued from one an SO issued. The issuer ranking compounds it: auto-issued cards count toward the tally that picks the issuer, so attribution concentrates on one person by construction. Distinguishing them inside this design means a discriminator column the fuller implementation would discard, paid for by loosening `needsLibraryCardRemovedForUser` for every existing card in the meantime.

DT-3800 stays open and blocked, carrying the requirement forward: whatever ends up representing researcher status has to activate a user on an institutional domain match at registration, and has to be able to record that the system rather than a signing official did it.

One fix here is independent of the feature and will ship on its own ticket. `deleteAllLibraryCardsByUser` deletes `library_card` rows without first clearing `lc_daa`, whose foreign key is `NO ACTION`, so enforcement's removal path throws on the `AuthorizationHelper` path — where only `NotFoundException` is caught — and an affected researcher is served a 500 on every API call. The same statement also targets `create_user_id`/`update_user_id`, so clearing one SO's card deletes every card that SO issued; that predicate has had no live caller since #1984.

Thanks @fboulnois and @otchet-broad for the reviews. Branch kept so the discussion stays readable.

PR #3070, opened 2026-09-17
[DT-4061] Report the requesting institution on DAR summaries

otchet-broad:
### Addresses

### Summary

Both DAR summary routes now report where a grant went without naming who holds
it. The dataset route gains the requester's institution it never carried; the
study route stops selecting the PI name it had been returning, and
`DarMetric.yaml` drops the property to match — Gson omits nulls here, so the
field is simply absent from responses.

This also fixes which DAR each summary describes. The dataset query sourced its
display record from anything submitted against the dataset, so a pending
progress report could stand in for the grant — overwriting the title and RUS and
resetting an expired grant to current. It now keeps the qualifying reference ids
and sources from those, as the study query already did.

Institution is withheld for a dataset belonging to no study.
`findDatasetByIdForRead` has no study visibility to test there, so it returns
those to every authenticated caller, and the affiliation would otherwise be
harvestable by walking ids. The request itself still comes through.

----

- Verify all tests go green


kevinmarete (approved):
Looks good 👍🏾

fboulnois (approved):
A few suggestions identified with Claude and Codex:

**1. Institution redaction uses study presence instead of authorization outcome:** `src/main/java/org/broadinstitute/consent/http/service/MetricsService.java:47`

Admins and other independently authorized users lose `institutionName` for study-less datasets. Redaction should depend on why access was granted, not whether `studyId` is null.

**2. Removing `piName` is an uncoordinated breaking API change:** `src/main/resources/assets/schemas/DarMetric.yaml:26`

The field is now always null and omitted by Gson, potentially breaking consumers that display requester names. Coordinate the change with consumers or retain and deprecate the field for one release.

**3. An unreviewed closeout can become the latest DAR:** `src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java:182`

The closeout UNION branch does not require an election. A newly submitted closeout can replace the approved DAR and expose the closeout submitter’s institution and other incorrect summary data. Exclude closeouts or document and test this behavior.

**4. `piName` remains as dead plumbing across application layers:** `src/main/java/org/broadinstitute/consent/http/models/DarMetricsSummary.java:12`

The DAO always returns null, but the mapper, model, service, and tests still carry `piName`. Remove the record component, SQL columns, mapper access, redaction logic, and vacuous null assertions.

**5. Positional record copying risks silently swapping fields:** `src/main/java/org/broadinstitute/consent/http/service/MetricsService.java:57`

The nine-argument copy can still compile after same-typed fields are reordered or inserted, while placing values in incorrect slots. Move redaction into a named model method and test preserved fields.

**6. The study-less path fetches requester identity only to discard it:** `src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java:191`

The query joins users and institutions and maps their data before the service removes it. Once redaction follows the authorization outcome, avoid these joins when requester identity will be withheld.

**7. DAO documentation describes the replaced selection behavior:** `src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java:120`

The javadoc says the newest linked DAR drives the summary, but the query now selects the newest qualifying DAR. Update it to describe the approved-DAR and closeout rules after resolving closeout behavior.

**8. A renamed test documents behavior it no longer tests:** `src/test/java/org/broadinstitute/consent/http/db/DataAccessRequestDAOTest.java:1196`

The javadoc and `piName` fixture describe PI-name precedence, while the test asserts that `piName` is null. Rewrite the documentation, remove irrelevant setup, and add dataset-route coverage for institution identity.

**9. OpenAPI does not document withheld `institutionName` values:** `src/main/resources/assets/paths/darSummariesByDatasetId.yaml:21`

For some requests, `institutionName` is null and therefore omitted from JSON, although the schema presents it as a normal string. Mark it nullable and document when requester identity is withheld.

PR #3071, opened 2026-09-17
[DT-4061] Clear the Sonar findings on the study work

otchet-broad:
### Addresses

### Summary

Clears the nine Sonar issues open on #3049, all in code merged from the DT-4061
branches.

`SchemaFromStudy.build` was flagged CRITICAL at cognitive complexity 16. Nothing
in it is complicated: the whole body sat inside one `if (nonNull(study))`, which
costs a point itself and puts six enum conversions, a ternary and its `&&` a
level deeper, where each costs double. Inverting it to an early return removes
the wrapper and de-nests the rest. Most of that file's diff is re-indentation; a
null study still yields an empty schema, which now has a test.

`MetricsService` held a `DatasetDAO` it never read. Removing it made a stub and
a mock unnecessary, and made the accompanying "never fans out per dataset"
verification tautological — the service has no such DAO to call now.

`StudyCommentDAOTest` slept to separate timestamps. A sleep buys that only
probabilistically, and every run pays the wait regardless; the rows are
backdated instead.

The rest are removals: an unused migration helper, an unused test factory,
redundant `eq()` matchers, and a list built inside an `assertThrows` lambda.

----

- Verify all tests go green


kevinmarete (approved):
👍🏾

fboulnois (approved):

PR #3072, opened 2026-09-18
[DT-4101] Clear lc_daa and scope the bulk library card delete

kevinmarete:
### Addresses

https://broadworkbench.atlassian.net/browse/DT-4101 — security risk: low

### Summary

`deleteAllLibraryCardsByUser` deleted `library_card` rows without clearing `lc_daa`. That foreign key is `NO ACTION`, so removing a card that holds any DAA raised a constraint violation — and enforcement runs that removal on the `AuthorizationHelper` path, where only `NotFoundException` is caught, so the affected researcher was served a 500 on every API call until the data was fixed by hand. The statement now clears the relations and the cards together, as `deleteLibraryCardById` already did.

It also targeted `create_user_id`/`update_user_id`, so clearing one signing official's card deleted every card that SO issued. Both callers mean "clear this user's own card", and that predicate has had no live caller since #1984 removed the foreign keys it was written for. Today the `lc_daa` violation aborts the statement before the wider delete can commit, so fixing only the first half would have turned an error into silent data loss — hence both in one change.

Review then found that neither delete path wrote an `lc_daa_audit` REMOVE row, so a card's pre-authorizations disappeared while `findAuditsByLcUserId` still reported them as authorized on the trailing ADD. Both statements now return the relations they clear and record a REMOVE against them. `deleteLibraryCardById` takes the acting user — the resource already holds it — and reads `lc_user_id` from the card in the same statement, skipping cards that exist against an email with no user, since that column is `NOT NULL`. The bulk delete records the card owner as the initiator, as both of its callers run during that user's own login.

Draft until product/security confirm the narrowed predicate ([DT-4101 comment](https://broadworkbench.atlassian.net/browse/DT-4101)). If the wide revocation was intended, it should become an explicit cascade with audit rows rather than a side effect of a delete predicate.

### Testing

`LibraryCardDAOTest` 40/40, `LibraryCardServiceTest` 36/36 and `LibraryCardResourceTest` 18/18 against real Postgres. New DAO coverage: a card holding a DAA deletes cleanly and records a REMOVE, deleting one user's card leaves the cards they issued and those cards' `daaIds` intact, and the admin delete path records a REMOVE attributed to the acting user. Each assertion fails against the previous statements — on `fk_lc_id`, on the issued card having been deleted, and on the missing REMOVE row.

Found while reviewing DT-3800. #3069 closed without shipping, so the fix moves here on its own ticket.


fboulnois (approved):
Suggestions:

**1. Bulk delete drops lc_daa rows without writing an lc_daa_audit REMOVE row:** `src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java:194`

The `daa_deletes` CTE removes pre-authorizations, but unlike `deleteLibraryCardDaaRelation` and `DaaServiceDAO.bulkRemoveUsersFromDaa`, no REMOVE audit row is inserted. `findAuditsByLcUserId` will still show the DAA as authorized (last action ADD). Fix: add `RETURNING lc_id, daa_id` to the CTE and an `INSERT INTO lc_daa_audit ... SELECT ... 'REMOVE', NOW() FROM daa_deletes`.

**2. Dropping the `create_user_id`/`update_user_id` predicate narrows authorization scope inside an FK crash fix:** `src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java:192`

Previously, an SO changing institution revoked every card they issued. Now only the SO's own card is removed; researchers keep cards and DAA access until they individually log in. This may be intended, but it is outside the FK bug scope (see `docs/ai/prompts/bugfix.md`). Confirm with product/security or split into its own PR.

**3. Comment misstates why the CTE works:** `src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java:189`

The lc_daa rows do not "go first". A CTE imposes no ordering; both deletes run in one statement and the `NO ACTION` FK is checked at end of statement. A maintainer may later split this into two `@SqlUpdate` calls in the wrong order. Reword to: "fk_lc_id is NO ACTION (checked at end of statement), so deleting lc_daa in the same statement satisfies it."

**4. The lc_daa cascade is hand-rolled in two DAO methods:** `src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java:194`

Both `deleteLibraryCardById` and `deleteAllLibraryCardsByUser` now duplicate the cascade. Any future path deleting a `library_card` row must remember to delete `lc_daa` first or it throws `fk_lc_id`. Consider a Liquibase changeset making `fk_lc_id` `ON DELETE CASCADE`, or routing this method through the same SQL as `deleteLibraryCardById`.

**5. The `targets` CTE is redundant:** `src/main/java/org/broadinstitute/consent/http/db/LibraryCardDAO.java:191`

`uk_lc_user_id` makes `user_id` unique, so `targets` resolves to at most one row. Both `IN (SELECT id FROM targets)` subqueries can be replaced with the direct predicate: `DELETE FROM lc_daa WHERE lc_id IN (SELECT id FROM library_card WHERE user_id = :userId)` followed by `DELETE FROM library_card WHERE user_id = :userId`.

**6. No test covers a user who both owns a card and issued another user's card with lc_daa relations:** `src/test/java/org/broadinstitute/consent/http/db/LibraryCardDAOTest.java:312`

The two new tests cover own-card-with-DAA and issued-card-without-DAA separately. Neither asserts the issued card's `lc_daa` rows survive `deleteAllLibraryCardsByUser(signingOfficial)`. A regression widening the `daa_deletes` predicate back to `create_user_id` would pass both. Add a case asserting the researcher's card and its `daaIds` remain intact.

kevinmarete:
Thanks — 1, 3, 5 and 6 applied in 0e4c7177.

**1.** REMOVE rows now written on both paths. `deleteLibraryCardById` takes the acting user (the resource already holds it) and reads `lc_user_id` from the card in the same statement, guarded on `user_id IS NOT NULL` since cards can exist against an email with no user and the column is NOT NULL. The bulk path records the card owner as initiator — both of its callers run during that user's own login.

**3.** Reworded to the mechanism you gave.

**5.** `targets` gone; `uk_lc_user_id` made it a one-row subquery.

**6.** The issued-card test now gives that card a DAA and asserts its `daaIds` survive.

**4.** Declining this one. Now that both paths write audit rows they still need the explicit `DELETE ... RETURNING` to know what to audit, so a cascade would not remove the duplication — it would only change the failure mode for a future caller that forgets `lc_daa`, from a loud `fk_lc_id` violation to silently dropping the relations *and* their audit trail. The loud version is the better default, and the schema change sits outside the bugfix scope either way.

**2.** Agreed it needs a call. Asking product/security on DT-4101 before this merges.

kevinmarete:
Closing — team and PO agreed the library card replacement under [DT-3215](https://broadworkbench.atlassian.net/browse/DT-3215) retires `deleteAllLibraryCardsByUser` and the `lc_daa` relation this fixes, so the change and the open question about the narrowed predicate go with it.

Branch `km-dt-4101-library-card-delete-lc-daa` stays, so the diff and its DAO tests are here if the fix is needed before the replacement ships.

[DT-3215]: https://broadworkbench.atlassian.net/browse/DT-3215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3075, opened 2026-09-18
[DT-3940][3/4] Stop writing the legacy match consent identifier

kevinmarete:
### Addresses

[DT-3940](https://broadworkbench.atlassian.net/browse/DT-3940) — match release 2 of 3. Stacked on #3074.

**Security risk: low** — a nullable column and one fewer bound parameter on an insert; no endpoint, authorization or returned value changes.

### Summary

Match inserts drop `consent` and the column becomes nullable to accept what they leave behind. This is safe only because #3074 stopped reading it, so the changeset halts until that release has applied here — it checks for the snapshot tables being gone, which is that release's signature.

`purpose_consent` stays for now. Postgres unique constraints ignore nulls, so it cannot reject a row written after this, and `match_entity_purpose_dataset_unique` has enforced the real rule since the `dataset_id` constraints were applied. 4/4 drops both the constraint and the column, and needs this release deployed everywhere first.


src/main/java/org/broadinstitute/consent/http/db/MatchDAO.java
```diff
@@ -66,10 +66,10 @@ List<Match> findMatchesForLatestDataAccessElectionsByPurposeIds(
   @SqlUpdate(
       """
         INSERT INTO match_entity
-          (consent, dataset_id, purpose, match_entity, failed, create_date,
+          (dataset_id, purpose, match_entity, failed, create_date,
            algorithm_version, abstain)
         VALUES
-          (:consent, :datasetId, :purpose, :match, :failed, :createDate,
+          (:datasetId, :purpose, :match, :failed, :createDate,
```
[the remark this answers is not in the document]
kevinmarete: Same disposition as the equivalent thread on #3074, and left unresolved for the same reason.

Correct that a halted precondition does not stop the service starting, so deploying this before its predecessor would fail every match insert. That is loud rather than silent, and the precondition names which release is missing. Startup-on-migration-failure is DT-4099; the ordering is a release-plan requirement, as it was for DT-3942's A, B and C.

fboulnois (approved):
A few suggestions:

**1. Runtime correctness now depends on a Liquibase HALT the application swallows:** `src/main/java/org/broadinstitute/consent/http/db/MatchDAO.java:68`

If the snapshot tables still exist, the precondition HALTs, `ConsentApplication` logs the exception and starts anyway, and `match_entity.consent` stays NOT NULL. Every `insertMatch` then fails, silently breaking all DAR matching. Earlier releases in the stack tolerated their own gate halting; this one does not. Consider making `initializeLiquibase` failure fatal.

**2. The rollback UPDATE ignores the still-present `purpose_consent` unique constraint:** `src/main/resources/changesets/changelog-consent-2026-09-18-match-consent-nullable.xml:41`

A legacy row whose stored consent text differs from its dataset's derived identifier can collide with a re-derived NULL row for the same purpose. The rollback aborts midway and the column stays nullable while the changelog believes it rolled back. Either drop `purpose_consent` here or guard the UPDATE against collisions.

**3. The precondition does not assert `match_entity_purpose_dataset_unique` exists:** `src/main/resources/changesets/changelog-consent-2026-09-18-match-consent-nullable.xml:22`

The changeset's safety argument (that `purpose_consent` becoming toothless is fine) depends on that constraint, but only snapshot-table absence is checked. If an operator dropped the constraint, NULL-consent rows would fall under no uniqueness rule at all. Add a `pg_constraint` check, as the drop-snapshots changeset already does.

**4. Fifth verbatim copy of the Testcontainers/Liquibase migration-test harness:** `src/test/java/org/broadinstitute/consent/http/db/MatchConsentNullableMigrationTest.java:30`

`startPostgres`, `stopPostgres`, `connection`, `update`, `rollback`, `execute` and `queryObject` duplicate four sibling migration tests, each booting its own Postgres container. Extract a shared abstract `MigrationTestHelper` with a single static container and have each test supply only its changelog and pre-migration schema.

**5. The overlap test never exercises the mixed-writer collision the changeset header describes:** `src/test/java/org/broadinstitute/consent/http/db/MatchConsentNullableMigrationTest.java:74`

An old-release row with a consent value and a new-release NULL row for the same `(purpose, dataset_id)` is never asserted to be rejected, and the rollback test does not assert the pre-existing `DUOS-000042` row is left untouched. If the unique constraint were missing this suite would still pass. Add both assertions.

kevinmarete:
Thanks — four of five applied.

**1. Correctness rests on a HALT the app swallows.** Same disposition as the equivalent thread on #3074: real, but making `initializeLiquibase` fatal changes startup behaviour for every environment and every changeset, which is too broad for this stack. Tracked in DT-4099, which weighs aborting startup against failing the readiness probe and also covers narrowing `onFail="HALT"` to the failing changeset.

**2. Rollback UPDATE ignores `purpose_consent`.** Fixed. The UPDATE now re-derives every row rather than only the NULL ones, which makes `consent` a function of `dataset_id` — so `purpose_consent` can only be violated where `match_entity_purpose_dataset_unique` already was, and that constraint is now asserted in the precondition. It also matches #3076's rollback, which never had the `IS NULL` filter. `rollbackSurvivesALegacyValueThatDisagreesWithItsDataset` covers exactly the collision you described.

**3. Precondition does not assert `match_entity_purpose_dataset_unique`.** Fixed — added the `pg_constraint` check, scoped by `conrelid`. While there I also moved the snapshot-table half off `relnamespace = 'public'::regnamespace` onto `to_regclass`, so it resolves through `search_path` like everything else; same latent issue you raised on #3074.

**4. Fifth copy of the harness.** Fixed in #3073 via `MigrationTestHelper`; this test extends it.

**5. Overlap test never exercises the mixed-writer collision.** Fixed. `migrationRejectsTheMixedWriterCollisionOnTheReplacementConstraint` asserts both orderings of an old-release row with a value and a new-release NULL row on one `(purpose, dataset_id)`, `migrationHaltsWhenTheReplacementConstraintIsMissing` covers the constraint being absent, and the rollback test now asserts the pre-existing `DUOS-000042` row re-derives to the value it already carried.

fboulnois (approved):
👍

src/main/resources/changesets/changelog-consent-2026-09-18-match-consent-nullable.xml
```diff
@@ -0,0 +1,63 @@
...
+    <!--
+      The snapshot tables are gone only if the release that stopped reading the column applied its
+      changeset here; without it, this release's nulls reach a reader that still wants a value.
+      match_entity_purpose_dataset_unique is the other half: the argument for letting purpose_consent
+      go toothless below is that it is not the last uniqueness rule over a purpose's matches.
+    -->
+    <preConditions onFail="HALT" onError="HALT"
+      onFailMessage="The release that stops reading match_entity.consent must have applied here first; its changeset drops match_migration_snapshot, and match_entity_purpose_dataset_unique must be present because purpose_consent stops constraining the rows written after this. Deploy that release before this one.">
+      <dbms type="postgresql"/>
+      <sqlCheck expectedResult="1">
```
otchet-broad, line 34: The precondition sums three predicates and expects 1.  That unintentionally accepts any state where exactly one predicate is true.

These should probably be separate preconditions.

otchet-broad (changes requested):

src/main/resources/changesets/changelog-consent-2026-09-18-match-consent-nullable.xml
kevinmarete: Right, and it was letting a real state through. `0 + 0 + 1` is the state I meant, but `1 + 0 + 0` and `0 + 1 + 0` also sum to 1 — a surviving snapshot table with `match_entity_purpose_dataset_unique` already dropped satisfied the gate, which is exactly what it exists to block.

Each predicate is now its own `sqlCheck`. Kept `to_regclass` over `tableExists` since the schema is not `public`.

The existing halt tests each broke one predicate, summing to 2 and 0, so both passed and the suite stayed green; only the combinations expose it. Added those two, plus `match_migration_rationale_snapshot`, which no test named. Against the old precondition they fail with "Expected LiquibaseException to be thrown, but nothing was thrown".

Worth noting `match-dataset-id-constraints` uses the same summing idiom but expects 0, which is sound — a sum of non-negative counts is 0 only when every term is.

otchet-broad (approved):
👍

PR #3076, opened 2026-09-18
[DT-3940][4/4] Drop the match consent column and its legacy constraint

kevinmarete:
### Addresses

[DT-3940](https://broadworkbench.atlassian.net/browse/DT-3940) — match release 3 of 3, and the end of the dataset identity contract migration. Stacked on #3075.

**Security risk: low** — drops a column and constraint nothing reads or writes by this point; no endpoint or authorization changes.

### Summary

`match_entity` now identifies its dataset only by foreign key. `purpose_consent` goes with the column it constrains; `match_entity_purpose_dataset_unique` has covered the same rule since the `dataset_id` constraints were applied.

**Do not merge and deploy this together with #3075.** The changeset halts unless the column is already nullable and the replacement constraint is present, which catches an environment that skipped a release — but not one that deploys both in the same run, where an instance still writing a value would fail every match insert. The ordering is a release-plan requirement, as it was for A, B and C.


fboulnois (approved):
A few suggestions identified with Claude:

**1. The irreversible drop never verifies that surviving consent values match the identifier derived from `dataset_id`:** `src/main/resources/changesets/changelog-consent-2026-09-19-drop-match-consent.xml:27`

The 09-09 gate checked NULLs, `algorithm_version` and duplicates, but never `consent = 'DUOS-' || LPAD(alias)`. A hand-fixed row with `consent='DUOS-000042'` but `dataset_id` pointing at alias 57 passes, `dropColumn` discards the value for good, and rollback fabricates `'DUOS-000057'`. Add a third `sqlCheck` term counting disagreeing rows (expected 0) so the drop is provably lossless.

**2. Dropping a column breaks server-side prepared `SELECT m.*` statements on release-3 pods during the rolling deploy:** `src/main/java/org/broadinstitute/consent/http/db/MatchDAO.java:24`

Once pgjdbc has named-prepared a match query (after `prepareThreshold` executions), the next call on that pooled connection fails with `0A000 'cached plan must not change result type'`. pgjdbc only re-prepares outside a transaction, and `MatchDAO` is `Transactional`, so reads surface as 500s until the connection recycles. List columns explicitly instead of `m.*`, or call out the error window in release notes.

**3. Sixth verbatim copy of the Liquibase migration-test harness:** `src/test/java/org/broadinstitute/consent/http/db/MatchConsentColumnDropMigrationTest.java:38`

Container lifecycle, `connection()`, `update()`, `rollback()`, `execute()`, `queryObject()` and `columnExists()` duplicate the five sibling `*MigrationTest` classes, and CI pays a sixth container start-up. Extract an abstract `MigrationTestHelper` with a static shared container, leaving each test only its `CHANGELOG` constant and schema setup.

**4. `constraintExists()` matches by `conname` alone, not scoped to `match_entity`:** `src/test/java/org/broadinstitute/consent/http/db/MatchConsentColumnDropMigrationTest.java:133`

The changeset precondition scopes to `conrelid = 'match_entity'::regclass`, but the test helper passes if a same-named constraint exists on any table. It also concatenates the constraint name into SQL, a pattern that becomes injectable if the helper ever takes a non-constant. Add the `conrelid` filter and bind the name as a parameter.

**5. The plan's status table omits DT-3940's authoring state:** `docs/plans/data-use-primary-consistency-plan.md:13`

The row records that all three DT-3942 releases are written but says nothing about DT-3940, even though this branch completes the fourth and final release described in the Phase 3 table. The DT-3942 stack updated this row when its last release landed. One sentence (all four DT-3940 releases written; deploy order as Phase 3) keeps the plan authoritative.

kevinmarete:
Thanks — all five applied.

**1. The drop never verifies surviving values match the derived identifier.** Fixed. Added a second `sqlCheck` (expected 0) counting rows whose `consent` disagrees with `'DUOS-' || LPAD(alias)`, so a hand-fixed row halts the deploy instead of being discarded and rebuilt as a different value. The `::BIGINT` cast is load-bearing there too — without it a `numeric` `42.0` would render `DUOS-0042.0` and every row would look like it disagreed. `migrationHaltsWhenASurvivingValueDisagreesWithItsDataset` covers it. I also moved the nullable half of the first check onto `pg_attribute`, matching the `search_path` point from #3074.

**2. `SELECT m.*` breaks cached prepared statements mid-deploy.** Fixed by listing the columns explicitly in all three queries, with a note on the interface saying why. Better than a release note given the failure lasts the life of the pooled connection.

**3. Sixth copy of the harness.** Fixed in #3073 via `MigrationTestHelper`; this test extends it. One container now serves all six migration classes.

**4. `constraintExists()` unscoped and concatenated.** Fixed — scoped by `conrelid = 'match_entity'::regclass` and the name bound as a parameter. `MigrationTestHelper.queryObject` grew a varargs overload for this, and the other helpers that concatenated a varying name were moved onto it too.

**5. Plan status omits DT-3940.** Fixed — row 5 now records that all four DT-3940 releases are written and deploy in the Phase 3 order, each gated on the previous.

fboulnois (approved):

otchet-broad (approved):
👍

PR #3078, opened 2026-09-21
[DT-4128] Withhold identity on the authorization outcome, not on study presence

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-4128](https://broadworkbench.atlassian.net/browse/DT-4128)

### Summary

The rule keyed on the dataset having no study, which is a proxy for "nothing about this caller was checked" - and wrong for anyone allowed in on their own merits. An admin, the dataset's creator, or a reader of its study lost the requester's institution on a study-less dataset for no reason.

findDatasetByIdForRead now has a sibling that reports which rule let the caller through, and the metrics service withholds only for NO_STUDY. The set of callers allowed in is unchanged: the creator check moves ahead of the study lookup, where the study branch already admitted them, so it grants nothing new - it only names the reason. A creator with a hidden study no longer reads that study at all, which its test now asserts rather than stubbing.

Redaction is a named method on DarMetricsSummary. The nine-argument positional copy it replaces still compiled after two same-typed components were reordered, and would have moved values into the wrong slots silently.

Not done here: the review also suggests skipping the users and institution joins when identity will be withheld. That needs a second copy of a sixty-line query, and two copies of this one drifting apart is the bug this branch already fixed once. Left for a measurement that says the join costs anything.

----

- Verify all tests go green


src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
```diff
@@ -196,10 +212,25 @@ protected Dataset verifyPublicVisibilityAccess(Dataset dataset, User user) {
...
+
+  /** The dataset, and the rule that let this caller read it. */
+  public record DatasetRead(Dataset dataset, DatasetReadBasis basis) {}
+
+  /** Why a caller was allowed to read a dataset. */
+  public enum DatasetReadBasis {
+    /** An admin, who may read anything. */
+    ADMIN,
+    /** The user who created the dataset. */
+    DATASET_CREATOR,
```
rushtong, line 226: Do we care about data custodians as well?

rushtong (approved):
One question inline, but looks good overall 👍🏽

src/main/java/org/broadinstitute/consent/http/service/DatasetService.java
otchet-broad: Good catch!  I'll add a test to demonstrate why we don't need it for completeness.

fboulnois (approved):
A few suggestions identified with Claude:

**1. Redaction gate is weaker than its comment claims:** [MetricsService.java:53](src/main/java/org/broadinstitute/consent/http/service/MetricsService.java#L53)

Redaction keys on `NO_STUDY`, but `STUDY_READABLE` is also returned when a study is merely publicly visible, so any authenticated caller still gets `institutionName` for every dataset in a published study. The comment's claim that such callers were "allowed in on their own merits" is false in that case, since nothing about them was checked. Fix the gate or the comment.

**2. Javadoc overstates the safety it argues for:** [DarMetricsSummary.java:19](src/main/java/org/broadinstitute/consent/http/models/DarMetricsSummary.java#L19)

The Javadoc says "the compiler names every field", but the body is a plain positional 9-arg canonical constructor call. Reordering two same-typed components would silently move values into the wrong slots, which is exactly the failure the comment claims is prevented. Use a named or builder-style copy, or drop the claim.

**3. Archived-filter asymmetry drops collections from dataset metrics:** [DataAccessRequestDAO.java:186](src/main/java/org/broadinstitute/consent/http/db/DataAccessRequestDAO.java#L186)

The closeout branch of `qualifying_dars` has no archived filter, but the new `latest_dar` subquery does. A collection qualifying only via an archived closeout enters `approved_collections` yet produces no `latest_dar` row, so the INNER JOIN drops it; it previously appeared, sourced from a non-archived DAR. Same asymmetry at line 272.

**4. `piName` is dead but still on the wire:** [DarMetric.yaml:24](src/main/resources/assets/schemas/DarMetric.yaml#L24)

The schema drops `piName`, but both queries still select `NULL::text AS pi_name`, the record and `DarMetricsSummaryMapper` still carry it, and responses still serialize `"piName": null`. Remove it end to end, or keep it documented as always-null.

otchet-broad:
@fboulnois Items 1 and 2 are now addressed on this branch.  3 and 4 are addressed outside in DT-4130 and DT-4129 and are ready to merge down.

PR #3079, opened 2026-09-21
[DT-4030] Fix Metrics Queries.

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-4130](https://broadworkbench.atlassian.net/browse/DT-4130)

### Summary

One DAR can be granted some of the datasets it asks for and denied the rest. The qualifying CTE admitted a closeout on the strength of nothing — no election, no approval, just a reference to the dataset — so a denied dataset still listed the collection among its approved DARs, sourced from the closeout and showing that submission's title and dates as though access had been granted.

Requiring an approval on the queried dataset is the whole fix. Once the CTE holds only approved DARs, a closeout can neither qualify a collection nor win the display record, so the closeout branch and its ranking flag are both gone.

The study query keeps its own rule: a grant denied on one dataset but approved on another in the same study still appears there, because the study did grant access. Only the denied dataset's page omits it.

Dates now span the whole grant. The submission date is the earliest approved submission — the day access began, not the day it was last renewed. Expiry is measured from the newest approved submission, since a renewal restarts the term, and a closeout supersedes that term because it revokes access the day it is filed. The study list is ordered by the date it displays; ordering by the sourced DAR's own date floated a renewed old grant to the top and made the cards read out of order.

A closed-out grant stays in the history — these queries report how a dataset has been used and do not decide who may reach it.
----

- Verify all tests go green


rushtong (approved):
Looks good. The PR title says DT-4030 but I think it should be DT-4130 👍🏽

kevinmarete (approved):

PR #3080, opened 2026-09-21
[DT-4129] Stop carrying a PI name the metrics never report

otchet-broad:
### Addresses

[https://broadworkbench.atlassian.net/browse/DT-4129](https://broadworkbench.atlassian.net/browse/DT-4129)

### Summary

 Raised by @fboulnois on #3070, items 4 and 9. Follows the decision that neither the study nor the dataset page names the requester.

Both metrics queries selected NULL::text AS pi_name, and every layer below dutifully plumbed it: the record component, the mapper read, the service's redaction of an always-null value, and four assertions whose only claim was that a constant was constant. None of them could fail.

No client loses anything. The field never reached develop — DarMetricsSummary there has six components. piName, submissionDate and institutionName were all added by this stack, so this removes it before it ships rather than after.

One test survives, sited where it can still fail. Rather than restate the shape of a record that no longer has the component, it serializes a summary through the response path and compares the whole served field set, so a name reappearing under any spelling breaks it. Verified live: dropping one expected key fails the test against the real key list.

DarMetric.yaml now marks institutionName nullable and states when it is withheld — the rule DT-4128 establishes: only where nothing about the caller was established, meaning a dataset with no study. Gson omits nulls, so the field is absent rather than null, and the description says so.

----

- Verify all tests go green


src/main/resources/assets/schemas/DarMetric.yaml
```diff
@@ -25,4 +25,11 @@ properties:
     description: The reference identifier for the DAR
   institutionName:
     type: string
-    description: The institution of the requester the DAR was granted to
+    nullable: true
```
rushtong, line 28: Claude caught this:

> One small note, not a bug
> The schema at `DarMetric.yaml` marks institutionName with `nullable: true`. The API file declares OpenAPI 3.1.2, and that version does not define nullable. The field is also omitted rather than sent as null, so the flag does not describe the real behavior. Other schemas in the repo already use nullable: true, so this follows an existing convention. You can leave it, or remove it and rely on the description text.

I had to dig into it - it looks like [nullable exists in 3.0.4](https://spec.openapis.org/oas/v3.0.4.html#schema-nullable), but [no longer does in 3.1.2](https://spec.openapis.org/oas/v3.1.2.html#handling-null-values)

rushtong (approved):
Looks good 👍🏽

kevinmarete (approved):

src/main/resources/assets/schemas/DarMetric.yaml
otchet-broad: Made DT-4177 to address this across the API.

